PR #26, opened 2018-01-30
feature(db migration) Create a script for migrate old DB

hmoreno-fun:
This script take 2 .csv files with the old DB tables:
newsfeed_article and newsfeed_articlecategory

The script can be launch with:
sudo docker-compose exec app python manage.py old_db_migrate

It will create articles and categories with the Django ORM.


jmaupetit:
This work is obsolete right? If yes, can we close this PR.

sampaccoud:
Not really obsolete but we never intended to merge it, just pull the branch
manually on the server at the moment where we will want to load history in
the new install.

jmaupetit:
I am not comfortable with the idea to let a PR open for a temporary work that will never be merged. So here is what propose:

1. Close this PR :wink: 
2. Move the django command and related csv to a public gist
3. Open a new PR to:
    * add the required dev dependency (_e.g._ `Beautifulsoup4`)
    * document how to fetch and use the public gist

jmaupetit:
If we need Beautifulsoup to run this script, how will you launch the script in a container context? Installing it as a dev dependency is not a real pain.

PR #31, opened 2018-02-02
Add cache invalidation mechanism for js/css files

sampaccoud:
## Purpose

The js/css files are the result of a build done by webpack. Each time they are updated, browsers will get an outdated version from cache on the webserver or CDN which are difficult to invalidate.
By appending as querystring a unique identifier calculated on the content of the file we make sure browsers always get the updated version.

## Proposal

Use Django's `ManifestStaticFilesStorage` to collect static files (only in production settings).

### DIFF
--- a/richie/settings.py
+++ b/richie/settings.py
@@ -377,6 +377,13 @@ class Production(Base):
 
     ALLOWED_HOSTS = values.ListValue(None)
 
+    # For static files in production, we want to use a backend that includes a hash in
+    # the filename, that is calculated from the file content, so that browsers always
+    # get the updated version of each file.
+    STATICFILES_STORAGE = (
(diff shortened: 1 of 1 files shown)


ormsbee:
Have you looked at https://github.com/owais/django-webpack-loader ? We use it in edx-platform to bridge webpack and Django, and it supports optimized builds.

sampaccoud:
We can now merge this PR as the disturbing file references in CSS have been removed.

@ormsbee It's not clear to me what django-webpack-loader would improve to the current Webpack build we have? As for cache invalidation, the ManifestStaticFilesStorage included in Django seems to do a good job. It caches the mapping between each file and its renamed version in a json file manifest.

PR #78, opened 2018-04-09
Get courses from api to display on the frontend

mbenadda:
Drop course fixtures \o/


richie/js/data/course/reducer.ts
```diff
@@ -0,0 +1,64 @@
...
+    // NB: we still use string as the index type as keys of an objects are always converted to strings
+    items: { [index: string]: Course['id'] };
+    queryKey: string;
+    total_count: number;
+  };
+}
+
+export const course: Reducer<CourseState> = (
+  state: CourseState = initialState,
+  action?: CourseAdd | CourseListGetSuccess,
```
jmaupetit: :joy:

richie/js/data/course/reducer.ts
```diff
@@ -0,0 +1,64 @@
...
+) => {
+  if (!action) { return state; } // Compiler needs help
```
jmaupetit: :joy:

richie/js/data/course/sideEffects/getList.ts
```diff
@@ -0,0 +1,62 @@
...
+    offset: number;
+    total_count: number;
+  };
+  objects?: Course[];
+}
+
+// Wrap fetch to handle params, headers, parsing & sane response handling
+// NB: some of this logic should be move in a separate module when we reuse it elsewhere
+export function fetchCourses(params?: CourseListGet['params']): Promise<Response> {
+  return fetch('/api/v1.0/course' + formatQueryString(params), {
```
jmaupetit: Open question: instead of hard-coding API endpoints, what about listing them somewhere as application settings?

mbenadda:
@jmaupetit a new commit that handles your comment

webpack.config.js
```diff
@@ -1,4 +1,4 @@
-const SETTINGS = require('./richie/settings.json');
+const SETTINGS = require('./richie/js/settings.json');
```
jmaupetit: :clap:

richie/js/settings.json
```diff
@@ -1,7 +1,10 @@
 {
+  "API_ENDPOINTS": {
+    "COURSE": "/api/v1.0/course"
```
jmaupetit: I would have preferred to define an `API_ROOT_URL` setting and don't include this root url in endpoints to avoid repetition, but we are limited by the JSON format here. Do you think a plain JS/TS file would better suite our needs for the application's settings.

mbenadda: Less repetition in the file itself, but we still need a type definition file along the JS/JSON as we can't import TS into the webpack config

On the other hand JSON is more generally interoperable/reusable later on, and I don't think repetition is necessarily a bad thing in a JSON settings file

jmaupetit: Okay. Let's try this then.

PR #79, opened 2018-04-09
Bump Django CMS to version 3.5.1

sampaccoud:
All dependencies and plugins had to be updated as well because they were listed explicitly in requirements.


mbenadda:
It does not work on my machine, and I did a docker hard reset

PR #80, opened 2018-04-10
Add missing Django-parler setting

rmoch:
The lack of this setting constant and the default french language of our project
probably results in a edge case conflict where ironicaly `safe_translation_getter`
fails to choose language to translate

### DIFF
--- a/richie/settings.py
+++ b/richie/settings.py
@@ -78,15 +78,6 @@ class Base(DRFMixin, ElasticSearchMixin, Configuration):
     USE_I18N = True
     USE_L10N = True
     USE_TZ = True
-    LANGUAGE_CODE = 'fr'
-    # Careful! Languages should be ordered by priority, as this tuple is used to get
-    # fallback/default languages throughout the app.
-    # Use "en" as default as it is the language that is most likely to be spoken by any visitor
-    # when their preferred language, whatever it is, is unavailable
-    LANGUAGES = (
-        ('en', _('en')),
-        ('fr', _('fr')),
-    )
 
     # Templates
     TEMPLATES = [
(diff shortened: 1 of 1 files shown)


richie/settings.py
```diff
@@ -185,6 +176,29 @@ class Base(DRFMixin, ElasticSearchMixin, Configuration):
...
+    # Use "en" as default as it is the language that is most likely to be spoken by any visitor
+    # when their preferred language, whatever it is, is unavailable
+    LANGUAGES = (
+        ('en', _('en')),
+        ('fr', _('fr')),
+    )
+
+    PARLER_LANGUAGES = {
+        None: (
+            {'code': 'fr',},
```
sampaccoud: you should list all languages here with their fallbacks
```
   # app-specific
    'PARLER_LANGUAGES': {
        1: [
            {
                'code': 'en',
                'fallbacks': ['de', ],
            },
            {
                'code': 'de',
                'fallbacks': ['en', ],
            },
        ],
        'default': {
            'code': 'en',
            'fallbacks': ['en'],
            'hide_untranslated': False
        }
},
```

richie/settings.py
```diff
@@ -185,6 +176,47 @@ class Base(DRFMixin, ElasticSearchMixin, Configuration):
...
+    LANGUAGES = (
+        ('en', _('en')),
+        ('fr', _('fr')),
+    )
+
+    PARLER_LANGUAGES = {
+        None: (
+            {
+                'code': 'fr',
+                'fallbacks': [],
```
sampaccoud: fallback should be "en" ? or you're sure it will fallback to the default below?

richie/settings.py
```diff
@@ -185,6 +176,47 @@ class Base(DRFMixin, ElasticSearchMixin, Configuration):
...
+    #PARLER_LANGUAGES = {
+    #    None: (
+    #        {'code': 'en',},
+    #        {'code': 'fr',},
+    #    ),
+    #    'default': {
+    #        'fallbacks': ['en'],          # defaults to PARLER_DEFAULT_LANGUAGE_CODE
+    #        'hide_untranslated': False,   # the default; let .active_translations() return fallbacks too.
+    #    }
+    #}
```
sampaccoud: deadcode to be removed ?

PR #88, opened 2018-04-18
Display facet results in filter pane

mbenadda:
Simply take the orgs & subjects from facets we received from the server and display them along with the proper count (if available) in the left sidebar.

NB: this required a slight API change to avoid overcomplicating things on the client side.

### DIFF
--- a/richie/js/components/searchFilterGroupContainer/searchFilterGroupContainer.tsx
+++ b/richie/js/components/searchFilterGroupContainer/searchFilterGroupContainer.tsx
@@ -2,7 +2,7 @@ import values from 'lodash-es/values';
 import { connect } from 'react-redux';
 
 import { RootState } from '../../data/rootReducer';
-import FilterDefinition from '../../types/FilterDefinition';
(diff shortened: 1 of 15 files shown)


richie/js/components/searchFilterGroupContainer/searchFilterGroupContainer.tsx
```diff
@@ -15,16 +15,16 @@ const hardcodedFilters: { [machineName: string]: FilterDefinition; } = {
     humanName: 'Language',
```
sampaccoud: have you thought of i18n?

mbenadda: There's a Trello card for it (https://trello.com/c/ax6JbVfZ/74-manage-i18n-in-ts-tsx-files) but I think now that the project is public it's better to create issues so here it is: https://github.com/openfun/richie/issues/89

Right now I think it's better to move the features forward as fast as we can but it will soon be time to take care of this. I have already done some rough work on this.

Issue #90, opened 2018-04-18
Use TypeScript's `strict` mode


mbenadda:
TypeScript has a `strict` mode (nothing to do with the browser's) that enables stricter compilation errors, and will evolve over time as new, more stringent compiler options are added in. 

Right now it already enables a bunch of options we did not manually enable (https://www.typescriptlang.org/docs/handbook/compiler-options.html), the most significant of which is `strictNullChecks`.

We'll have to do some cleaning up after we enable it but it should help avoid errors.

jmaupetit:
Fixed in #91

PR #91, opened 2018-04-21
Use typescript strict mode

mbenadda:
As we talked about @sampaccoud I updated `tsconfig.json` in order to enable strict mode. This of course required some changes to the code to still compile (and did catch some potential errors).

I could not find a good online resource about this so I wrote a blog post if anyone's interested: http://mbenadda.com/typescripts-strict-compiler-option/

Related to issue #90

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -3,7 +3,7 @@
     "jsx": "react",
     "lib": [ "dom", "es5", "scripthost", "es2015.promise", "es2015.iterable" ],
     "module": "commonjs",
-    "noImplicitAny": true,
+    "strict": true,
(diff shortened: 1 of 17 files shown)


tsconfig.json
```diff
@@ -3,7 +3,7 @@
     "jsx": "react",
```
sampaccoud: WAT ! is this a gitmoji ?

mbenadda: Alright I put back the wrench 🚫💪

sampaccoud: I was just joking to say "I saw it" and didn't dislike it #litote

PR #96, opened 2018-04-26
Update configurations to run Richie on OpenShift

jmaupetit:
## Purpose

To run Richie in various environments for various clients on OpenShift, we need to be able the add flexibility in some settings, particularly the `ALLOWED_HOSTS` setting.

## Proposal

We decided to get the `ALLOWED_HOST` value from the environment, because it's easily manageable thanks to Ansible & OpenShift.

### DIFF
--- a/richie/settings.py
+++ b/richie/settings.py
@@ -261,8 +261,15 @@ class ContinuousIntegration(Test):
 
 
 class Production(Base):
-    """Production environment settings"""
-    pass
+    """Production environment settings
+
+    You must define the DJANGO_ALLOWED_HOSTS environment variable in Production
+    configuration (and derived configurations):
+
+    DJANGO_ALLOWED_HOSTS="foo.com,foo.fr"
+    """
+
+    ALLOWED_HOSTS = values.ListValue(None)
 
 
 class Staging(Production):
--- a/richie/settings.py
+++ b/richie/settings.py
@@ -272,6 +272,15 @@ class Production(Base):
     ALLOWED_HOSTS = values.ListValue(None)
 
 
+class Feature(Production):
+    """
+    Feature environment settings
+
+    nota bene: it should inherit from the Production environment.
+    """
+    pass
+
+
 class Staging(Production):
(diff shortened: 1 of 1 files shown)


sampaccoud:
My first thoughts were: we should do configuration the same way as for edxapp ie one yaml setting file and one yaml credential file.

Then I remembered that we are using django-configuration and maybe it is only working with environment variables? How do you plan to set environment variables? I think we can still have a setting file in configMap that we mount as environment variables in the container. Is it what you have in mind?

jmaupetit:
For this particular setting, this env var is set in the DC and initialized with: `{{ project_name }}-richie.{{ domain_name }}`. Sensible settings will be set in a ConfigMap loaded as a Docker `env_file` (not tested yet).

jmaupetit:
I think I was unclear/mistaken in my explanations. Let me try to clarify what I have in mind:

1. This PR makes the `ALLOWED_HOSTS` setting **configurable** via an environment variable for `Production` instances (& derived). More PRs are about to come to make things easily configurable _via_ environment variables.
2. This environment variable value (and others) will be **set** in a vaulted file treated as an OpenShift Configuration Map, hence managed by Arnold and version tracked in Matsuo.
3. In Richie's OpenShift Deployment Configuration, we will use this file as an env_file injected in Richie's pod (container).

To me, there is no need to make things over-complicated as they are `edxapp`. Using environment variables for configuration is a standard, and by doing so, Richie will not be bound to Matsuo, Arnold, or any deployment tool. WDYT?

sampaccoud:
Yes! I'm in line with this.

Just one detail: this particular setting could be in the "not vaulted" setting file as it is not secret.

The idea is to only vault the settings that are credentials because it wil be easier to consult the functional settings if they are not vaulted.

jmaupetit:
> Just one detail: this particular setting could be in the "not vaulted" setting file as it is not secret.

True. I made a mistake, particularly since it will be set via an Ansible template.

PR #107, opened 2018-05-01
Pluralize resource names across back & front code

### LINKED ISSUE #95: Resource names should be consistent throughout the project
## Expected behavior

Resources have a canonical name that is reused throughout all the places where their name is expressed as a string, such as URL routes, keys in objects or serializable elements, etc.

## Actual behavior

Resources are named by their plural or singular forms alternatively, eg.
- `organization` in API routes, client state keys, facet names...
- `organizations` in foreign keys on eg. courses, in API filters...

## Proposed solution

Standardize on the plural form as suggested by https://swaggerhub.com/blog/api-design/api-design-best-practices/.

This will prevent us from running into issues due to inconsistent naming and having to transform singulars into plurals and plurals into singulars when reaching for a resource using its name in different places.

mbenadda:
## Purpose

As discussed in #95, we need to make resource names consistent across the project so we don't have to convert between singular/plural form.

## Proposal

Pluralize resource names across the board so everything fits together nicely. We've only left out case when it makes no sense to pluralize, for exemple a single record type names or a single record passed through a prop

- [x] Pluralize resource names throughout backend code...
- [x] ... and throughout frontend code

Fixes #95

### DIFF
--- a/apps/search/description.apib
+++ b/apps/search/description.apib
@@ -44,7 +44,7 @@ A single subject as saved in our ElasticSearch index. Includes a small subset of
 # Group Course
 Course-related resources for the richie search API.
 
-## GET /course/{course_id}
+## GET /courses/{course_id}
 A single course object. Access through this API is read-only (and thus GET only) as we're interacting with ElasticSearch and not the underlying model.
 
 See the Course data structure for more information.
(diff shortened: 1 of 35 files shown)


jmaupetit:
One small thing: it appears that `apps/search/tests/test_viewsets_courses.py` should be reformatted with `black`. Once done, ship it!

PR #109, opened 2018-05-01
Clean up resource list params in reducer

mbenadda:
## Purpose

As we're not caching list responses anymore, we don't need the query key. On the other hand, we do need the current params to modify them as users interact with the filters.

## Proposal

Set defaults and clean up APIListCommonRequestParams.

It makes more sense to have them specified and then use Partial<> when we accept incomplete subsets.

### DIFF
--- a/richie/js/data/genericReducers/resourceList/resourceList.ts
+++ b/richie/js/data/genericReducers/resourceList/resourceList.ts
@@ -1,10 +1,14 @@
 import get from 'lodash-es/get';
 
-import { APIResponseListFacets } from '../../../types/api';
+import { API_LIST_DEFAULT_PARAMS as defaultParams } from '../../../settings.json';
+import { APIListCommonRequestParams, APIResponseListFacets } from '../../../types/api';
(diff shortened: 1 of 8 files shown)


richie/js/data/genericReducers/resourceList/resourceList.ts
```diff
@@ -1,19 +1,24 @@
...
 
-import { APIResponseListFacets } from '../../../types/api';
+import { APIListCommonRequestParams, APIResponseListFacets } from '../../../types/api';
 import Resource from '../../../types/Resource';
-import { ResourceListGetSuccess } from '../../genericSideEffects/getResourceList/actions';
+import { ResourceListGet, ResourceListGetSuccess } from '../../genericSideEffects/getResourceList/actions';
 
 export const initialState = {};
 
+export const defaultParams = { limit: 20, offset: 0 };
```
jmaupetit: Shouldn't it be configurable in the settings?

mbenadda:
@jmaupetit done

PR #112, opened 2018-05-02
Add & configure sentry

jmaupetit:
## Purpose

Sentry is a service used to aggregate application failure reports. It is a central tool to monitor and debug application's behavior. A must-have for every Django application.

### `version.json`

Following Mozilla's Dockerflow, we now use CircleCI to generate a `version.json` file that will be copied at the root of the `/app` directory during Docker image built. This file describes the version of the containerized application [1].

### Sentry configuration

- [x] Add raven python dependency (sentry python client)
- [x] Configure raven to use a DSN from the DJANGO_SENTRY_DSN environment variable
- [x] Configure logging to use sentry as the root logs handler
- [x] Add a get_release utility that fetches current release for Sentry from a version.json file or the latest git commit sha1 (if available)
- [x] Wrap wsgi server execution so that we catch low level errors from the wsgi server
- [x] Add richie directory as a testpath for pytest
- [x] Ignore `.pytest_cache`

## TODO

Integrate Raven.js to report front-end failures (this will be implemented in a new PR, see #113).

## References

[1] https://github.com/mozilla-services/Dockerflow/blob/master/docs/version_object.md


richie/settings.py
```diff
@@ -234,6 +272,49 @@ class Base(DRFMixin, ElasticSearchMixin, Configuration):
...
+        "root": {"level": "WARNING", "handlers": ["sentry"]},
+        "formatters": {
+            "verbose": {
+                "format": "%(levelname)s %(asctime)s %(module)s "
+                "%(process)d %(thread)d %(message)s"
+            }
+        },
+        "handlers": {
+            "sentry": {
+                "level": "ERROR",  # To capture more than ERROR, change to WARNING, INFO, etc.
```
mbenadda: Don't we actually want to be able to send more than just errors to sentry?

I'm thinking events like ES index automated management, synchronization with fun-apps/openedx etc. should be able to send INFO and WARNING level events so we have a history and can quickly check how the latest actions went when we need to debug something.

richie/settings.py
```diff
@@ -234,6 +272,49 @@ class Base(DRFMixin, ElasticSearchMixin, Configuration):
...
+                "formatter": "verbose",
+            },
+        },
+        "loggers": {
+            "django.db.backends": {
+                "level": "ERROR", "handlers": ["console"], "propagate": False
+            },
+            "raven": {"level": "DEBUG", "handlers": ["console"], "propagate": False},
+            "sentry.errors": {
+                "level": "DEBUG", "handlers": ["console"], "propagate": False
```
mbenadda: If we set `propagate: False` here, does that not mean that the `"sentry"` handler will not receive events from this logger?

richie/settings.py
```diff
@@ -13,6 +15,39 @@
...
+        pass
+
+    # Get latest commit sha1 from the local git repository
+    try:
+        return raven.fetch_git_sha(project_dir)
+    except raven.exceptions.InvalidGitRepository:
+        pass
+
+    # Default: not available
+    return "NA"
```
mbenadda: I know this is just a failsafe, but maybe we could add datetime info here, just in case we start seeing these it might help us to pinpoint their source?

(of course we could also try more complicated stuff like detect the OS etc. but I'm thinking datetime is trivial to add and could prove helpful)

richie/settings.py
jmaupetit: `WARNING` seems relevant to me, but not `INFO`: at this level, logs should be sent to syslog and further analyzed/mined with an ELK based dashboard.

richie/settings.py
jmaupetit: Contextual informations are already collected and sent by Sentry/raven (OS, user, server hostname, ip, etc.). I think those metadata would be sufficient to track who's sending sentry events from it's laptop :stuck_out_tongue:

richie/settings.py
mbenadda: Ideally yes, but as a first approximation, until we have an ELK instance running, Sentry could be helpful there.

I've seen @sampaccoud do that before, and I think it's better than dropping that data entirely.

richie/settings.py
mbenadda: Alright then, `"NA"` is good enough!!

richie/settings.py
jmaupetit: The `sentry` handler is the root logger, hence it will receive all logs who's level is `> WARNING`. This particular logger is using the `console` handler at a `DEBUG` level with propagate set to `False`, consequently the `sentry` handler will not receive logs `>= DEBUG, <WARNING` for this particular logger.

To make a long explanation short: yes.

:wink:

mbenadda: Ok then. I'm not familiar with this and I was trying to understand things by reading the code and referring to the docs.

Thank you

jmaupetit: That was a pretty interesting remark. Thank you for asking.

richie/settings.py
mbenadda: Maybe add an issue so we remember to take care of this later :)

jmaupetit: Done in #114.

richie/settings.py
```diff
@@ -13,6 +15,39 @@
...
+    [1] https://github.com/mozilla-services/Dockerflow/blob/master/docs/version_object.md
+    """
+    project_dir = os.path.abspath(os.path.join(BASE_DIR, os.pardir))
+
+    # Try to get the current release from the version.json file generated by the
+    # CI during the Docker image build
+    try:
+        with open(os.path.join(project_dir, "version.json")) as version:
+            return json.load(version)["commit"]
+    except FileNotFoundError:
```
rmoch: Shall we handle malformed json ?

richie/tests.py
```diff
@@ -0,0 +1,50 @@
...
+
+# pylint: disable=unused-argument
+class GetReleaseTestCase(TestCase):
+    """Battle test the settings.get_release function"""
+
+    @mock.patch("builtins.open", side_effect=FileNotFoundError)
+    @mock.patch("raven.fetch_git_sha", side_effect=InvalidGitRepository)
+    def test_returns_default_without_version_file_or_git_repository(self, *args):
+        """
+        When not version.json file exists nor a Git repository, we expect that
```
rmoch: s/not/no ?

richie/settings.py
jmaupetit: Good remark. I don't think so as it is auto-generated by the CI.

richie/tests.py
jmaupetit: Yes! Well spotted thanks. This is now fixed.

PR #115, opened 2018-05-03
🐛(search) fix subjects+organizations terms queries in the same query

mbenadda:
## Purpose

The "terms" query does not support multiple different keys on the same query. This is not documented but the issue arose when we attempted it on our live ES instance.

We received a 500 error on the API (400 error from ES) when attempting to filter on orgs & subjects at the same time.

## Proposal

Use a "bool" query that allows us to combine multiple queries (and therefore 2 or more "terms" queries) into one.

It's also structurally cleaner as it allows us to express our constraints as arrays of invididual constraints
with a filter/ranking/necessity marker.

Docs: https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-bool-query.html

### DIFF
--- a/apps/search/tests/test_viewsets_courses.py
+++ b/apps/search/tests/test_viewsets_courses.py
@@ -150,16 +150,22 @@ class CoursesViewsetTestCase(TestCase):
         # The ES connector was called with appropriate arguments for the client's request
         mock_search.assert_called_with(
             body={
-                "query": {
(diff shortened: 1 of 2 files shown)


jmaupetit:
Can you add a link to "bool queries" documentation in this PR description (and also in the code)?

mbenadda:
done

PR #116, opened 2018-05-03
Update the filters and trigger a request when the user clicks on a filter

mbenadda:
## Purpose

This is the most basic form of the expected functionality for our frontend: let users pick a filter in the left sidebar, add it to the current search and get new, relevant search results. 

## Proposal

Build handlers for filter adding and removal in `SearchFilterGroupContainer` as it is the place where all the actionable information is gathered.

We then pass down the handlers through to `SearchFilter` which just uses them as click handlers and passes its own value back upstream.

Note: removal functionality was not included in the UI as this PR was already complicated enough as it is — and it would required us first to manage to show the active filters so they can be deactivated.

The same goes for facet management, which is separate logic: adding/removing should work the same no matter the actual filters we show.

### DIFF
--- a/richie/js/components/searchFilterGroupContainer/computeNewFilterValue.ts
+++ b/richie/js/components/searchFilterGroupContainer/computeNewFilterValue.ts
@@ -7,34 +7,38 @@ export function computeNewFilterValue(
   existingValue: Maybe<Nullable<string | number | Array<string | number>>>,
   relevantValue: string | number,
 ) {
+  // There is no existing value for this filter
   if (!existingValue) {
-    // There is no existing value for this filter
-    return action === 'add' ?
+    return {
       // ADD: Make an array with the existing value
-      [ relevantValue ] :
+      add: () => [ relevantValue ],
       // REMOVE: There's nothing that could possibly removed, return null
-      null;
-  } else if (typeof existingValue === 'string' || typeof existingValue === 'number') {
-    // The existing value for this filter is a single primitive type value
-    return action === 'add' ?
-      // ADD: Make an array with the existing value and the new one
-      [ existingValue, relevantValue ] :
-      (existingValue === relevantValue ?
-        // REMOVE: Return nothing if we had to drop the existing value we had
-        null :
-        // REMOVE: Keep the existing value if it's not the one we needed to drop
-        existingValue);
-  } else {
-    // The existing value is an array of strings or numbers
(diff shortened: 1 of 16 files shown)


richie/js/components/searchFilterGroupContainer/computeNewFilterValue.ts
```diff
@@ -0,0 +1,40 @@
...
+    // The existing value for this filter is a single primitive type value
+    return action === 'add' ?
+      // ADD: Make an array with the existing value and the new one
+      [ existingValue, relevantValue ] :
+      (existingValue === relevantValue ?
+        // REMOVE: Return nothing if we had to drop the existing value we had
+        null :
+        // REMOVE: Keep the existing value if it's not the one we needed to drop
+        existingValue);
+  } else {
```
jmaupetit: Can you add a comment for this block (type of existing value)?

richie/js/components/searchFilterGroupContainer/computeNewFilterValue.ts
```diff
@@ -0,0 +1,40 @@
...
+      return array.length === 0 ? null : array;
+    };
+
+    return action === 'add' ?
+      // ADD: Just push the new value into our existing array of values
+      [ ...existingValue, relevantValue ] :
+      // REMOVE: Return the existing array of values without the one we needed to remove
+      nullEmptyArray(existingValue.filter((v) => v !== relevantValue));
+  }
+}
```
jmaupetit: This function is quite complex to read. Not sure how it can be improved... But one thing for sure: comments help a lot. Thx! :+1:

richie/js/components/searchFilterGroupContainer/getFilterFromState.ts
```diff
@@ -0,0 +1,66 @@
...
+        ...state.filterDefinitions[machineName],
+        values: getFacetedValues(state, facets, machineName),
+      };
+
+    // Values from state are usable as-is for hardcoded filters
+    default:
+      return state.filterDefinitions[machineName];
+  }
+
+    /* tslint:disable:variable-name */
```
jmaupetit: strange indentation

mbenadda:
@jmaupetit I took a quick stab at writing a simpler to understand version. Unfortunately the clearest (to me) did not pass type check due to a compiler weakness.

Tell me if you think it's better this way :)

mbenadda:
Looks have not changed much lately

PR #117, opened 2018-05-03
add `duplicate-code` exception to .pylintrc

rmoch:
## Purpose
Remove this warning, at least for now, because it not disablable on-line
and we do not want to factorize reported code this early
https://github.com/PyCQA/pylint/issues/214


rmoch:
👍

PR #122, opened 2018-05-07
Add prettier

mbenadda:
## Purpose

We already use code-formatting on the python side to avoid splitting hair over pointless style differences.
There's no reason not to use black's older brother on the JS side :)

## Proposal

- [x] add prettier to our dependencies
- [x] make a *minimal* config file
- [x] apply prettier to the whole codebase
- [x] add a prettier check step to the CI
- [ ] let each developer integrate prettier into their editor

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -437,7 +437,7 @@ jobs:
             - ./node_modules
           key: v3-front-dependencies-{{ checksum "package.json" }}
 
-  lint-front:
+  lint-front-tslint:
     docker:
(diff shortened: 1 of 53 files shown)


mbenadda:
@jmaupetit done! For some reason I still see the old `lint-front` task in the GitHub checks though

jmaupetit:
Thx! 

>For some reason I still see the old lint-front task (in addition to the 2 new names) in the GitHub checks though

I fixed this in the protected branch requirement settings.

PR #123, opened 2018-05-07
Use url for search params

mbenadda:
## Purpose

Search should interact with the URL: filter changes by the user should be reflected in query string params for the page & Search should use query string params to initialize itself when it first loads.

## Proposal

- [x] Add a clean way to use history pushState from connected components
- [x] Trigger this URL+history update whenever the user changes the filters
- [x] Init with query string params present in the URL
- [x] Use an established package to interact with query strings instead of homegrown functions


richie/js/data/genericSideEffects/getResourceList/getResourceList.ts
```diff
@@ -31,11 +31,11 @@ export interface GetListSagaSpecifics {
 // NB: some of this logic should be move in a separate module when we reuse it elsewhere
 export function fetchList(
   resourceName: keyof RootState['resources'],
-  params?: ResourceListGet['params'],
+  params: ResourceListGet['params'] = API_LIST_DEFAULT_PARAMS,
 ): Promise<Response> {
   const endpoint = API_ENDPOINTS[resourceName];
 
-  return fetch(endpoint + formatQueryString(params), {
+  return fetch(endpoint + '?' + stringify(params), {
```
jmaupetit: Why not using a template literal here? I love template literal.

PR #130, opened 2018-05-15
Add courses to the CMS

sampaccoud:
## Purpose

Add courses to the CMS.

## Proposal

We added a course model, a subject model and the relationships between the courses and subjects/organizations.

- [x] Add course and subject models
- [x] Add wizards to create a course and a subject page
- [X] Add templates for the detail view of courses and subjects
- [x] Add admin view for the course model
- [x] Add toolbar menu items to edit a course
- [x] Refactor organizations to use the improvements made while developing courses


apps/courses/cms_wizards.py
```diff
@@ -0,0 +1,198 @@
...
+        A page `path` is limited to 255 chars, therefore the course page slug should
+        always be shorter than (255 - length of parent page path - 1 character for the "/")
+        """
+        parent_page = Page.objects.filter(
+            reverse_id=self.model.ROOT_REVERSE_ID, publisher_is_draft=True
+        ).first()
+        if not parent_page:
+            raise forms.ValidationError(
+                _(
+                    "You must first create a parent page and set its `reverse_id` to "
```
mbenadda: This error pops up when attempting to create a new subject or organization.

I'm not really sure what's expected of me at that point...

- Either we should be able to create a Subject / Organization from the wizard
- Or the wizard should not even show the "Organization" & "Subject" options?

sampaccoud: This is because we create the course pages as children of a "all courses" page. So we need it to exist and identify it by its reverse_id set to "courses" (see page advanced settings).

You can create a basic site with all the necessary root pages by running the command "create_cms_data".

I need to add some documentation.

sampaccoud:
Fixups TODO:

- [ ] add documentation about root pages and "create_cms_data"

apps/courses/models.py
```diff
@@ -0,0 +1,142 @@
...
+        """
+        self.full_clean()
+        super().save(*args, **kwargs)
+
+        if self.pk:
+            # pylint: disable=no-member
+            self.organizations.add(self.main_organization)
+
+
+class Subject(PageExtension):
```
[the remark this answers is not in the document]
sampaccoud: The common practice in Django to put several models in the same file and split them in several files when the file get too long. 

142 lines is still short I think.

apps/courses/models.py
```diff
@@ -0,0 +1,142 @@
...
+        """
+        Enforce validation each time an instance is saved
+        Make sure the main organization is also included in `organizations` as a m2m relation
+        """
+        self.full_clean()
+        super().save(*args, **kwargs)
+
+        if self.pk:
+            # pylint: disable=no-member
+            self.organizations.add(self.main_organization)
```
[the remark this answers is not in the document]
sampaccoud: No ordering on the ManyToMany field. 
Adding ordering adds a lot of complexity and my thought was that it is not a good design to use ordering to designate what is the main organization.
So there was 2 reasons to prefer a separate field for that.

I think we also need this field in the index because we want to display the main organization on the list view...

apps/courses/cms_wizards.py
mbenadda: That was it!

apps/courses/templates/courses/cms/course_detail.html
```diff
@@ -0,0 +1,74 @@
+{% extends "richie/fullwidth.html" %}
+{% load cms_tags i18n %}
+
+{% block content %}
+<div class="course-template">
+
+  <h1 class="course-template__title">{{ current_page.get_title }}</h1>
```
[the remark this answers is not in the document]
mbenadda: You're right, `course-template` is not very expressive.
I had to do this with Search as I did not find a sensible name for an element whose sole purpose is to contain a react component (and `search-container` is already used inside react as container components are a pattern there)

For courses I was careful not to use up too broad names (I used `course-glimpse` and `course-glimpse-list`) so `course` is still available, but in that same spirit, +1 for `course-detail`

apps/courses/tests/test_admin_course.py
```diff
@@ -0,0 +1,157 @@
...
+        )
+
+    def test_course_change_view_post(self):
+        """
+        Validate that the course can be updated via the admin.
+
+        In particular, when a course is updated from the admin, the main organization should still
+        be added as a many-to-many relation in "organizations". This is being taken care of by
+        overriding the `save_related` method on the ModelAdmin and not by the override to the
+        `save` method on the model (see explanation in the docstring of the`save_related` method
```
jmaupetit: I think I get your point, but this phrase needs to be a bit rephrase.

apps/courses/cms_toolbars.py
```diff
@@ -0,0 +1,57 @@
...
+            page_extension, admin_url = self.get_page_extension_admin()
+            if page_extension:
+                edit_mode_inactive = not self.toolbar.edit_mode_active
+                page_menu = self.toolbar.get_or_create_menu("page")
+                # Create the new menu item as a modal
+                page_menu.add_modal_item(
+                    _("{:s} settings").format(self.model.__name__),
+                    url=admin_url,
+                    disabled=edit_mode_inactive,
+                    position=4,
```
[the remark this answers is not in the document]
sampaccoud: Glad you ask ;-)

I tried to get the position of the "Page settings..." item to put it below but it seems that the toolbar is built in now predefined order so sometimes it is not there yet. There's a way to order alphabetically but that's not what we want... etc. 

So. A better way than what I did is maybe to have a setting... Maybe we can have a shared setting for all page extensions since they can not coexist in our case. 

Should I call it PAGE_EXTENSION_TOOLBAR_ITEM_POSITION ?

apps/courses/cms_wizards.py
```diff
@@ -0,0 +1,198 @@
...
+        widget=forms.TextInput(),
+        label=_("Course key"),
+        help_text=_("Course key of the active session"),
+    )
+
+    model = Course
+
+    def clean_active_session(self):
+        """
+        Ensure that a course does not already exist with this active session key."""
```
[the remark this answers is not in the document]
sampaccoud: Let's try to  make it 3 lines...

apps/courses/cms_toolbars.py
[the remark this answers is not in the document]
jmaupetit: Yes! Should we prefix this kind of setting with `RICHIE_` to quickly identify that it's a custom settings?

apps/courses/templates/courses/cms/course_detail.html
```diff
@@ -0,0 +1,74 @@
...
+      {# If the current course page is the published version, show only the subjects that are published #}
+      {% elif subject.public_extension %}
+        <li class="course-template__subjects__item">
+          {{ subject.public_extension.extended_object.get_title }}
+        </li>
+      {% endif %}
+      {% if forloop.last %}
+      </ul>
+      {% endif %}
+    {% endfor %}
```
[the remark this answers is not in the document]
sampaccoud: Hmm in fact when the list is empty we:
-  don't want to have an empty <ul></ul>
- don't want to have the "no subjects" message styled in a <li></li>  similar to a the <li> to display a result...

So i propose to just add the empty clause like this: 
```python
    {% for ... %}
        .... (same as now)
    {% empty %}
      <div class="subject-detail__subjects--empty">
        {% trans "No associated courses" %}
      </div>
    {% endfor %}
```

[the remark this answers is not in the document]
jmaupetit: I usually have a `li.empty` class to deal with it but what you propose is perfectly fine!

apps/courses/cms_wizards.py
```diff
@@ -0,0 +1,198 @@
...
+        total_size = len(parent_page.get_slug()) + 1 + len(self.cleaned_data["slug"])
+        if total_size > 255:
+            raise forms.ValidationError(
+                _(
+                    "This slug is too long. The length of the path built by prepending the slug "
+                    "of the parent page would be {:d} characters long and should be less "
+                    "than 255".format(total_size)
+                )
+            )
+        return self.cleaned_data["slug"]
```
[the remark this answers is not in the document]
jmaupetit: Moreover will this validation error message be explicit for our end-users? Can we also ask them to decrease the (parent) page title size?

apps/courses/cms_toolbars.py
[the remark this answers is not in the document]
sampaccoud: Maybe we should have a `settings.py` file in the app and the RICHIE_ prefix is only required to override the setting in the project conf?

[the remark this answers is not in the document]
jmaupetit: :thinking: I'm not sure this is a good idea to dilute configuration in multiple files.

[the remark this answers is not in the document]
sampaccoud: I have in mind to later package "courses" and "search" as third party applications so that people can start their own CMS project and add these applications as functionalities to their CMS. 

That's why I think it's good that defaults for our settings are handled in the application itself.

```python

PAGE_EXTENSION_TOOLBAR_ITEM_POSITION = getattr(
    settings, "RICHIE_PAGE_EXTENSION_TOOLBAR_ITEM_POSITION", 4)
```

apps/courses/cms_wizards.py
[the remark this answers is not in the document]
jmaupetit: I'm not sure they fully understand what a slug is. But ok, forget about this comment: we'll see how it goes.

apps/courses/cms_toolbars.py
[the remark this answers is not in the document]
jmaupetit: Ok, I wasn't aware that we might package those apps separately. In this case, that makes sense to me.

Instead of naming this file `courses/settings.py` shouldn't we call it `courses/defaults.py`? I'm not sure if this is a convention/good practice, but this is how I usually proceed.

apps/courses/tests/test_admin_course.py
sampaccoud: I think the problem is better explained in the docstring of the "save_related" method. 
So in this docstring, I propose to just say:
```python
In particular, make sure that when a course is updated from the admin, the main organization is
automatically added to the many-to-many field "organizations".
See http://stackoverflow.com/a/1925784/469575 for details on the issue.
```

jmaupetit: Yes.

apps/courses/cms_toolbars.py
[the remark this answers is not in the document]
sampaccoud: Let's do that

apps/courses/templates/courses/cms/course_detail.html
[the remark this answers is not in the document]
rmoch: why not more classical loop structure 
```python
  {% if subject in course.subjects.exists ... %}
      <ul>
        {% for subject in course.subjects.all %}
               

         {% endfor %}
      </ul>
  {% else %}
    <div class="subject-detail__subjects--empty">
        {% trans "No associated courses" %}
    </div>
  {% endfif %}
```

?

README.md
```diff
@@ -1,185 +1,164 @@
-# Installing FUN CMS
+# Richie, A FUN CMS for Open edX
```
jmaupetit: Lowercase A?

README.md
```diff
@@ -1,185 +1,164 @@
...
 
+However, if you need to build a complete website with flexible content to aggregate your courses,
+in several languages and from different sources, **you will soon need a CMS**.
 
-## Installing a fresh server
+For example, the [edx.org](https://www.edx.org) website is built on `Drupal` for content pages and
+on [edX course discovery](https://github.com/edx/course-discovery) for the catalog.
 
-### Version
+At `France Université Numérique`, we wanted to build a similar portal but with `Python` and
```
jmaupetit: IMO "France Université Numérique" shouldn't be written has code, but rather emphasized with double quotes.

README.md
```diff
@@ -1,185 +1,164 @@
...
+Among the functionalities that `Richie` offers out of the box:
 
-If you are using another operating system or distribution, you can use [`Vagrant`](https://docs.vagrantup.com/v2/getting-started/index.html) to get a running Ubuntu 16.04 server in seconds.
+- multi-site and multi-lingual by default,
+- advanced access rights and moderation,
+- catalog of courses synchronized with one or more `Open edX` instances,
+- search engine based on `Elasticsearch` and pre-configured to offer full-text queries,
+  multi-facetting, auto-complete,...
+- flexible custom pages for courses, organizations, subjects, teachers (and their inter-relations),
+- Extensible with any third-party `DjangoCMS` plugin or any third-party `Django` application.
```
jmaupetit: This whole section should stand in a "motivations" section from the project's documentation, not in a README. IMO a README should go straight to the point: one phrase to describe what purpose serves the project, a short section explaining how to install & test it, contributing, license and voilà.

README.md
```diff
@@ -1,185 +1,164 @@
...
-post
-```sh
-sudo apt-get -y update
-sudo apt-get -y dist-upgrade
-```
-
-
-## Database part
+`Richie` is a **container-native application** but can also be installed on your machine the [old
+fashioned way](./docs/native_installation.md).
```
jmaupetit: The "old fashioned way" is a bit pejorative. What about "a more classical way"? Or "but it can also be installed on your system without virtualization tool".

README.md
```diff
@@ -1,185 +1,164 @@
...
-
-We need NPM to install the dependencies and run the build, which depends on a version of Nodejs specified in `.nvmrc`. See [the repo](https://github.com/creationix/nvm) for instructions on how to install NVM. To take advantage of `.nvmrc`, run this in the context of the repository:
-
-    nvm install
-    nvm use
-
-As a prerequisite to running the frontend build for either JS or CSS, you'll need to [install yarn](https://yarnpkg.com/lang/en/docs/install/) and download dependencies _via_:
-
-    yarn install
+Settings are defined using Django Configuration for different environments:
```
jmaupetit: Maybe add a link to Django-Configurations project: https://github.com/jazzband/django-configurations/

README.md
```diff
@@ -1,185 +1,164 @@
...
 
-- JS build
+- Development: settings for development on developpers' local environment,
+- Test: settings used to run our test suite,
+- ContinousIntegration: settings used on the continuous integration platform,
+- Feature: settings for deployment of each developers' feature branches,
+- Staging: settings for deployment to the staging environment,
+- PreProduction: settings for deployment to the pre-production environment,
+- Production: settings for deployment to the production environment.
```
jmaupetit: IMO this should go in a complete project doc, not in a README.

docs/docker_development.md
```diff
@@ -0,0 +1,75 @@
...
+
+You can easily see the latest logs for a container:
+
+    $ docker-compose logs [app|db|elasticsearch]
+
+Or follow the stream of logs:
+
+    $ docker logs --follow --until=2s
+
+If you need to debug inside a container, you can open a Linux shell with the `exec` command:
```
jmaupetit: "debug a running container" would be more explicit.

docs/docker_development.md
```diff
@@ -0,0 +1,75 @@
...
+
+    $ docker-compose exec [app|db|nginx] bash
+
+While developing on `Richie`, you will also need to run a `Django shell` and it has to be done in
+the `app` container:
+
+    $ docker-compose exec app python manage.py shell
```
jmaupetit: I would prefer running a new container:

```bash
$ docker-compose run --rm app python manage.py shell
```

docs/docker_development.md
```diff
@@ -0,0 +1,75 @@
...
+
+
+### Cleanup
+
+If you work on the Docker configuration and make repeated modifications, remember to periodically clean the unused docker images by running:
```
jmaupetit: images & containers

README.md
sampaccoud: Ok

README.md
sampaccoud: ok. then let's just write: "on your machine"

README.md
sampaccoud: I moved it to "docker_development.md"

README.md
jmaupetit: Indeed. The simpler, the better.

.circleci/config.yml
```diff
@@ -16,6 +16,22 @@ generate-version-file: &generate-version-file
...
+
+    working_directory: ~/fun
+
+    steps:
+      - checkout
+
+      - run:
+          name: Check absence of fixup commits
+          command: |
+            ! git log | grep fixup!
```
jmaupetit: I'd rather write:
```
! git log | grep "fixup!"
```
to protect the exclamation mark

PR #133, opened 2018-05-15
Move organizations to the courses application

funmoocbot:
This PR depends on https://github.com/openfun/richie/pull/130

## Purpose

The Organization model is closely related to the Course model so it does not make sense to have it in a separate application.

Furthermore, we want to publish "courses" as a third-party application so it should integrate the whole functional thing.

## Proposal

- [x] move organizations code to the courses application
- [x] reset the initial migration for all models in the courses application because the project still has no running instance


apps/courses/tests/test_cms_toolbars.py
```diff
@@ -1,59 +1,27 @@
 """
-Test suite of the toolbar extension for course pages
```
sampaccoud: This refactor factorizes the test so it can be run for organizations and courses without writing duplicate code.

apps/courses/factories.py
```diff
@@ -1,13 +1,48 @@
...
+        """
+        Automatically create a related page with a random title
+        """
+        return create_page(self.title, "courses/cms/organization_detail.html", "en")
 
-from apps.organizations.factories import OrganizationFactory
+    @factory.lazy_attribute
+    def code(self):
+        """
+        Since `name` is required, let's just slugify it to get a meaningful code (and keep it
```
mbenadda: s/name/title ?

apps/courses/factories.py
```diff
@@ -1,13 +1,48 @@
 """
 Courses factories
 """
-import factory
+from django.utils.text import slugify
 
 from cms.api import create_page
+import factory
+
+from .models import Course, Organization, Subject
```
jmaupetit: Surprising importation ordering. I think it's time to integrate [`isort`](https://github.com/timothycrosley/isort). If you agree, I can declare a new issue to integrate it later. WDYT?

sampaccoud: Why surprising this one? C < O < S ?

jmaupetit: No i'm targeting import statements order : from/import/from. `isort` would be useful here.

sampaccoud: Oh ok. I didn't know this rule and was applying alphabetical ordering.
Definitely +1 for isort.

PR #143, opened 2018-05-19
🐛(search) do not apply dimension D's filter when faceting on it

mbenadda:
## Purpose

We were having an issue where counts on facets were not what we expected, as counts for eg. the organizations facets were only including courses that matched the current organizations filter.

This made the facet counts pretty much useless for us.

## Proposal

We can instead facet on everything and use a specific filter for each aggregation, that reproduces the main query but omits the current dimension. See [a simple example](https://blog.madewithlove.be/post/faceted-search-using-elasticsearch/) for this pattern.


sampaccoud:
@mbenadda should be rebased and resolve conflicts but otherwise GTM.

PR #157, opened 2018-05-26
Pin deps so greenkeeper updates them

### LINKED ISSUE #142: Pin dependencies in package.json to get auto-updates from Greenkeeper
## Bug Report

**Expected behavior/code**
Greenkeeper should update every possible version for every dependency

**Actual Behavior**
It only updates major versions (semver)

**Steps to Reproduce**
N/A (see Greenkeeper bot Pull Requests)

**Environment**
N/A

**Possible Solution**
https://greenkeeper.io/faq.html#pinning-versions
Pinning the versions (removing ^ will cause greenkeeper to upgrade all dependencies)

mbenadda:
## Purpose

Greenkeeper only attemps to upgrade non-major versions of dependencies that have their versions pinned in package.json. 
We want it to update every possible version for every dependency, but it currently does not.

## Proposal

Pin our dependencies so it can actually do its job.

Fixes #142.


sampaccoud:
You could add the Fix ref in your commit

PR #159, opened 2018-05-27
Use "match" as the params name for basic searches

mbenadda:
## Purpose

Courses use "match" as the param name for searches, whereas Subjects & Organizations use "name". This is unintuitive in the client-side as it would force us to maintain a list matching those param names to model names.

## Proposal

Change those so everyone uses "match".


sampaccoud:
I would suggest a more common key for this to be "query" or even "q" (Google)

PR #160, opened 2018-05-27
Use Search as a template for courses in the demo-site

mbenadda:
## Purpose

Search is the intended template to display a list of courses. We need to update the demo site so that it uses it by default for Courses.

## Proposal

Update create_demo_site so it sets up that template for the courses page.
It is a template that does not have a user-editable maincontent area, which means we need to add a field in the page's definition and a flag on the `create_i18n_page` helper


apps/core/tests/test_helpers.py
```diff
@@ -63,6 +63,17 @@ def test_utils_create_i18n_page(self, mock_title, mock_plugin, mock_homepage):
 
         self.assertFalse(mock_homepage.called)
 
+    @mock.patch("apps.core.helpers.create_title")
+    @mock.patch("apps.core.helpers.add_plugin")
+    def test_utils_create_i18n_page_without_maincontent(self, mock_plugin, *_):
+        """
+        Make sure we can create pages without the maincontent placeholder. These are pages that
+        use specific templates to perform thheir function.
```
sampaccoud: s/thheir/their

PR #163, opened 2018-05-28
Clean up css

mbenadda:
## Purpose

When this project began it packed a lot of legacy CSS from FUN MOOC. We're not using most of it and due to the way it's written (basic CSS, one large file, etc.) it's not a base we can build off of.
We should start from a clean sheet and rebuild whatever we need over time.

## Proposal

- [x] refactor the large_banner plugin that was using styles from this legacy CSS
- [x] drop all unused styles


sampaccoud:
I would prefer starting from a clean sheet...

plugins/large_banner/templates/_large_banner.scss
```diff
@@ -0,0 +1,72 @@
...
+
+  &__logo {
+    margin: 0 0 0 0.5rem; // Leave some margin between title & logo
+
+    // Adjust logo size for smaller screens
+    max-width: 10rem;
+    max-height: 5rem;
+
+    @include media-breakpoint-up(sm) {
+      max-width: 11.25 rem;
```
jmaupetit: Extra space here.

PR #172, opened 2018-05-31
Person plugin

rmoch:
## Purpose
This a proposal of implementation of our plugins
As every entities (person, organisation...) are CMS pages, we have to store objects
specific information in CMS placeholder.
Here, a personne has a portrait, a short resume and, of course its page content
Then a person plugin, will show an except of a person page on other pages


apps/persons/migrations/0003_auto_20180528_1525.py
```diff
@@ -0,0 +1,16 @@
+# -*- coding: utf-8 -*-
```
sampaccoud: rerun makemigrations to have only one migration for 0002 and 0003?

apps/persons/tests/test_person_plugin.py
```diff
@@ -0,0 +1,88 @@
...
+        )
+
+        # Render the PersonPlugin
+        request = {}
+        renderer = ContentRenderer(request)
+        context = RequestContext(request)
+        context["request"] = request
+        html = renderer.render_plugin(person_plugin, context)
+
+        # Page title should be present
```
sampaccoud: the comment is not in line with what the code does right? You are not testing presence of title but of the url:
- modify comment
- add another test to check the presence of the title?

apps/persons/tests/test_person_plugin.py
```diff
@@ -0,0 +1,88 @@
...
+            '<a href="{url}">'.format(url=person_page.get_absolute_url()), html
+        )
+        # Person's portrait and its properties should be present
+        self.assertIn('<img src="" alt="{alt}">'.format(alt="portrait alternate"), html)
+        # Person's full name should be present
+        self.assertIn(
+            "{title} {first_name} {last_name}".format(
+                title=person.person_title.title,
+                first_name=person.first_name,
+                last_name=person.last_name,
```
sampaccoud: you are using this fullname in many places so maybe add a `get_fullname` method on the model or something like that?

apps/persons/templates/persons/plugins/person.html
```diff
@@ -0,0 +1,13 @@
+{% load thumbnail %}
+{% thumbnail portrait.picture 100x100 crop=opts.crop upscale=opts.upscale subject_location=opts.subject_location as thumbnail %}
+<p>
+    <a href="{{ instance.person.extended_object.get_absolute_url }}">
+        {{ instance.person.person_title.title }} {{ instance.person.first_name }} {{ instance.person.last_name }}
+    </a>
+</p>
+<p>
+    <img src="{{ thumbnail.url }}" {{ portrait.attributes_str }}>
```
sampaccoud: This is rendering blank in my manual tests.

apps/persons/tests/test_person_plugin.py
```diff
@@ -0,0 +1,88 @@
...
+        context = RequestContext(request)
+        context["request"] = request
+        html = renderer.render_plugin(person_plugin, context)
+
+        # Page title should be present
+        self.assertIn(
+            '<a href="{url}">'.format(url=person_page.get_absolute_url()), html
+        )
+        # Person's portrait and its properties should be present
+        self.assertIn('<img src="" alt="{alt}">'.format(alt="portrait alternate"), html)
```
sampaccoud: src should not be equal to a blanck string

apps/persons/tests/test_person_plugin.py
```diff
@@ -0,0 +1,88 @@
...
+        # Add portrait to related placeholder
+        placeholder = person_page.placeholders.get(slot="portrait")
+        add_plugin(
+            placeholder,
+            PicturePlugin,
+            "en",
+            **{"picture": image, "attributes": {"alt": "portrait alternate"}}
+        )
+        # A resume to related placeholder
+        placeholder = person_page.placeholders.get(slot="resume")
```
sampaccoud: As discussed, let's make it a raw text field without any editor

apps/persons/tests/test_person_plugin.py
```diff
@@ -0,0 +1,88 @@
...
+    """
+    Test that PersonPlugin correctly displays a Person's page placeholders content
+    """
+
+    def test_person_plugin_render(self):
+        """
+        Test that a PersonPlugin correctly renders person's page specific information
+        """
+        # Create a filer fake image
+        user = UserFactory(is_staff=True, is_superuser=True)
```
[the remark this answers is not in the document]
rmoch: done

apps/persons/tests/test_person_plugin.py
rmoch: yeah

apps/persons/tests/test_person_plugin.py
rmoch: updated

apps/persons/tests/test_person_plugin.py
rmoch: It's a fake image. How should we implement a test of this kind without storing any file ?

apps/persons/tests/test_person_plugin.py
rmoch: 👍

apps/persons/templates/persons/plugins/person.html
```diff
@@ -0,0 +1,18 @@
...
+     title="{{ page.get_title }} {% trans "avatar" %}">
+    {% with related_plugins.portrait.0 as picture_plugin %}
+    {% thumbnail picture_plugin.picture 200x200 crop=picture_plugin.use_crop upscale=picture_plugin.use_upscale as thumbnail %}
+      <img src="{{ thumbnail.url }}" {{ picture_plugin.attributes_str }}
+           alt="{{ page.get_title }} {% trans "avatar" %}">
+    {% endwith %}
+  </a>
+  <div class="person-plugin__body">
+    <a href="{{ page.get_absolute_url }}" title="{{ page.get_title }}">
+      <h4 class="person-plugin__body__title">{{ page.person.get_full_name }}</h4>
```
sveetch: This should be SEO nicer but should be semantically not right if template base does not open an <h3> before, do we stand on idea than user may have to be carefull about HTML semantic changes he do ?

sampaccoud: Good catch. How can we do it?

sveetch: There is no real solution except adding this level information as a plugin attribute than user could change from frontend edition.

I saw your solution about using `{% with ... %}` in upstream template that include plugin to define a title level variable but finally it ends with user needing to override this template. It almost the same than just letting user to override templates to accommodate them to its title tree. So i think we should stand on this way to keep it simple

PR #173, opened 2018-05-31
(cms) Replace obsolet image plugin cmsplugin-filer by djangocms-picture

rmoch:
`cmsplugin-filer` was a dependency used by Aldryn_newsblog to handle images, but its deprecated
Now we do not use newsblog anymore we can replace it by `djangocms-picture`


sampaccoud:
can you add the PR description as commit message as there is not enough information in the commit message now.

PR #182, opened 2018-06-02
Add bootstrap styles to nav

mbenadda:
## Purpose

Navigation has no styles and is displayed as a vertical list with the default list-styles.

## Proposal

It should use the bootstrap navbar classes so it is displayed horizontally.

- [x] Override the Django CMS menu to set proper classes on the elements
- [x] Use @extend to set bootstrap navbar styles on them
- [x] Drop some FUN-specific styles that do not belong in this project and affected the navbar look


jmaupetit:
It's so sad that not all bootstrap component styles are available as scss mixins.

richie/static/scss/_main.scss
```diff
@@ -11,16 +11,15 @@
...
+@import "../../../node_modules/bootstrap/scss/nav";
+@import "../../../node_modules/bootstrap/scss/navbar";
 @import "../../../node_modules/bootstrap/scss/type";
 @import "../../../node_modules/bootstrap/scss/utilities/display";
 @import "../../../node_modules/bootstrap/scss/utilities/flex";
 @import "../../../node_modules/bootstrap/scss/utilities/text";
 
 
 // FUNCMS
-// Global styles
```
sampaccoud: remove also the above "FUNCMS" line?

PR #187, opened 2018-06-04
Rework active filters without facets

mbenadda:
## Purpose

We used to handle both active filters and facets together, which caused us to display useless counts next to active filters, and more importantly to *not* display active filters that were not part of the facets.

Handling them separately helps us solve both these issues and also gets rid of some awkward code in `SearchFilterGroup`.

## Proposal

Pass active filter values separately from `SearchFilterGroupContainer`.

- [x] improve the filter integration test (which was broken by our other changes where it should not have been due to internal refreshing issues)
- [x] Add a function in `utils/filters` to create this new list of active filter values
- [x] Rewire `SearchFilterGroup` and `SearchFilterGroupContainer` to handle this new list

### DIFF
--- a/richie/js/components/SearchFilterGroup/SearchFilterGroup.spec.tsx
+++ b/richie/js/components/SearchFilterGroup/SearchFilterGroup.spec.tsx
@@ -19,8 +19,8 @@ describe('components/SearchFilterGroup', () => {
     } as FilterDefinitionWithValues;
     const wrapper = shallow(
       <SearchFilterGroup
+        activeFilterValues={[]}
         addFilter={addFilter}
-        currentValue={undefined}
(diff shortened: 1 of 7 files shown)


richie/js/components/SearchFilterGroup/SearchFilterGroup.spec.tsx
```diff
@@ -89,4 +91,48 @@ describe('components/SearchFilterGroup', () => {
         .text(),
     ).toContain('Value Three');
   });
+
+  it('deduplicates keys between the filter and thhe active filter values', () => {
```
jmaupetit: `s/thhe/the/`

PR #203, opened 2018-06-07
💚(gitlint) ignore bots git message lint

jmaupetit:
## Purpose

Pyup and Greenkeeper bots git commit title does not comply with our imposed pattern :cry: thus blocking each PR that the bot makes.

## Proposal

We choose to add a `git-lint` rule to ignore the linting of these commits to be able to reformat them using the "squash and merge" strategy while merging the PR.

### DIFF
--- a/.gitlint
+++ b/.gitlint
@@ -63,9 +63,10 @@ words=wip
 # regex=[^@]+@foo.com
 
 [ignore-by-title]
-# Allow empty body only when upgrading dependendies
-regex=^⬆️.*$
-ignore=B6
+# Allow empty body & wrong title pattern only when bots (pyup/greenkeeper)
+# upgrade dependencies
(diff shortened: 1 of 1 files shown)


.gitlint
```diff
@@ -63,10 +63,21 @@ words=wip
...
+# Allow empty body only when upgrading dependencies
 regex=^⬆️.*$
 ignore=B6
 
+[ignore-by-title]
+# Allow empty body & wrong title pattern only when pyup upgrades dependencies
+regex=^Update (.*) from (.*) to (.*)$
+ignore=B6,UC1
+
+[ignore-by-title]
```
sampaccoud: good if it works to duplicate the rule. Otherwise, do both in the same regex...

jmaupetit: That's what I finally did. :cry:

PR #212, opened 2018-06-11
Improve organization fixtures

sampaccoud:
## Purpose

We want our demo organization pages to look realistic.

## Proposal

First remove the existing fake content which does not look realistic.

Then improve the organization factory to generate complete organization pages with fake data (logo, banner image and description). 

For images we added a few fixture images that can be randomly used on each organization.

Example organization page (no css... yet):

screenshot-2018-6-12 ray ltd


apps/courses/tests/test_factories_organization.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+class OrganizationFactoryTestCase(TestCase):
+    """
+    Unit test suite to validate the behavior of the Organization factory
+    """
+
+    def test_organization_factory(self):
+        """
+        OrganizationFactory should be able generate plugins with realistic fake data: logo,
```
jmaupetit: should be able *to* generate

PR #227, opened 2018-06-18
⬆️(package) upgrade front-end dependencies

jmaupetit:
* update @types/react to version 16.3.18
* update karma to version 2.0.3
* update webpack-cli to version 3.0.8

Closes #222, #224, #225

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "@types/jasmine": "2.8.8",
     "@types/lodash-es": "4.17.0",
     "@types/query-string": "5.1.0",
-    "@types/react": "16.3.17",
+    "@types/react": "16.3.18",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -49,8 +49,8 @@
     "tslint": "5.9.1",
     "typescript": "2.8.3",
     "webpack": "4.12.0",
-    "webpack-cli": "3.0.6",
-    "xhr-mock": "2.4.0"
```
mbenadda: Do you really want to update this dep to a preview version? Not sure it's a good idea as we don't need this latest version?

jmaupetit: You are right. I've rolled back xhr-mock.

mbenadda:
GTM then, thanks!

Issue #228, opened 2018-06-18
Layout & style the organization view


mbenadda:
## Feature Request

**Is your feature request related to a problem or unsupported use case? Please describe.**
The organization view is implemented in Django but it's missing any layout or styling.

<img width="1435" alt="capture d ecran 2018-06-18 a 10 57 30" src="https://user-images.githubusercontent.com/1932937/41527233-be939714-72e6-11e8-9694-a0262ab67649.png">

**Describe the solution you'd like**
Use BEM classes and bootstrap defaults to make a proper layout for the organization view.

**Describe alternatives you've considered**
N/A

**Discovery, Documentation, Adoption, Migration Strategy**
This is roughly what I'm going for:
```
+------------------------------------------------------------------------------+
+                                                                              +
+                                  * image *                                   +
+                                                                              +
+                                                                              +
+                                                                              +
+                                                     +------------------+     +
+                                                     +                  +     +
+                                                     +                  +     +
+-----------------------------------------------------+                  +-----+
                                                      +       logo       +
                                                      +                  +
  The Incredible Course title                         +                  +
                                                      +------------------+

  Lorem ipsum dolor sit amet, consectetur adipiscing elit. Sed turpis arcu,
  vulputate a orci vitae, semper molestie quam. Vivamus quam enim, tincidunt
  eget lorem eget, consequat volutpat odio. Cras eget mi sed elit tincidunt
  fringilla a a sapien. Mauris at luctus diam. Suspendisse condimentum, arcu
  eu ullamcorper pretium, elit sapien iaculis lacus, quis auctor risus quam in.
  Praesent interdum metus turpis, quis mattis ligula imperdiet sed. Nam in
  tortor non magna blandit hendrerit in eu odio. Duis luctus rutrum nisi, in
  dignissim lectus. Maecenas vel ante fermentum, mollis massa eget, sollicitudin
  odio. Maecenas malesuada pulvinar dictum. Phasellus vel porttitor tellus. Ut
  varius orci sed dapibus aliquet.

  Duis in dapibus mi, non porttitor libero. Cras tristique purus sed mauris
  feugiat, vel egestas diam fringilla. Pellentesque sit amet mauris quis libero
  molestie interdum. Donec nec sodales dolor, lacinia interdum mi. Suspendisse
  ac est purus. Nullam vel sapien ante. Maecenas hendrerit leo vitae iaculis
  facilisis. Phasellus ornare pellentesque imperdiet. Nam cursus bibendum
[...]

PR #234, opened 2018-06-20
💄(courses) lay out & style the organization detail view

mbenadda:
## Purpose

The template for the organization detail view existed but did not include any CSS. 

## Proposal

Do some mobile-first layout work and add some basic styling so the default version of the website without any style override can be shown or used as an example directly.

Part 1/2 of work towards #228.

Part 2 will be to reuse `CourseGlimpse` component styles to have a proper list of related courses as shown in #228 — it just expanded the scope of this PR a little too much.


sveetch:
Nothing special to say, PR is pretty simple and efficient, we may think later to use flex behavior to align image and contain it properly to its container

PR #241, opened 2018-06-22
Make pyup less noisy

jmaupetit:
## Purpose

Greenkeeper and Pyup are useful services configured to ensure that front and back-end dependencies are always up-to-date. After a few months working with them, it appears that the update frequency is too high and finally counter-productive. So the decision has been made to increase the signal vs noise ratio.

## Proposal

- [x] back-end: switch to pyup weekly updates 
- [x] ~front-end: switch back to caret versions instead of pinned dependencies \*~

\* using this pattern, Greenkeeper will only submit a PR for major dependency updates or when the CI breaks for minor/patch updates (see [Greenkeeper's documentation](https://greenkeeper.io/docs.html#greenkeeper-step-by-step)).

Edit: we decided to keep pinned dependencies for the front-end and switch to Renovate to schedule monthly updates. This switch will be implemented in a separated PR.


mbenadda:
Hey Julien, thanks for taking some time to improve this 👍

I imagine there is no way to configure greenkeeper to do updates once a week like pyup? As I understand the docs, moving forward it will pretty much not do anything apart from major version changes (which don't occur all that often).

jmaupetit:
Unfortunately no: there is no way to set the update frequency with greenkeeper :disappointed: But I think we might have come with a reasonable solution by using caret versions: being notified when a minor or patch release breaks your CI or when a major release is out seems relevant.

mbenadda:
> when a minor or patch release breaks your CI

This is never going to happen as the lockfile precisely exists to prevent the CI (or anyone that installs using the lockfile) from pulling newer versions. Therefore I think we should just remove it entirely rather than add 30 seconds to all our builds for a negligible benefit.


I still think dependency updates need to be taken care of on a (reasonably) regular schedule; just not as a daily annoyance as greenkeeper has been.

I offer 2 solutions:
- accept it as a manual job (takes 10 minute developer time every week/whatever) — this is my favored solution
- try another tool, hopefully a better one than greenkeeper (Renovate? it's free for OS and available on GH marketplace)

Ping @rmoch @sampaccoud you might want to weigh in on this

jmaupetit:
> This is never going to happen as the lockfile precisely exists to prevent the CI (or anyone that installs using the lockfile) from pulling newer versions.

Totally true, I missed that point.

I don't think that it must be a manual job: to me it's the best way to never upgrade our dependencies.

Maybe other (better?) tools can be investigated, but in the meantime I propose to:

* remove the `--frozen-lockfile` flag from the yarn install step in the CI (since every new release is already tested by greenkeeper, there is no reason to keep this file locked)
* test greenkeeper with those settings during a few weeks and see how it goes.

WDYT?

sampaccoud:
@mbenadda I agree with @jmaupetit that an automatic PR has value. This will remind us to keep the project updated when we will be working on something else. Once a month for frontend  and once a week for backend seems enough though.

Renovate seems to allow scheduling so let's give it a try? 

> Receive automated Pull Requests whenever dependencies need updating. Or whenever you schedule it for.

mbenadda:
yarn's `--frozen-lockfile` (and the similar npm thing) are a godsend.

Without them the CI is unreliable as it installs random stuff and breaks on completely unrelated branches. Essentially builds from the same source don't have the same output anymore. This means:
- you lose a lot of time to track down problems and upgrade deps when you would rather be finishing your feature
- you also lose your ability to deploy in the meantime as your builds don't work and you can't revert as the breakage it implicitly contained in some `^` in your package.json

I've been through this in the past and it's pretty painful.

Let's leave Greenkeeper untouched until we try Renovate. I agree with both of you that this can be a valuable service if it is good enough.

jmaupetit:
> Without them the CI is unreliable as it installs random stuff and breaks on completely unrelated branches.

I might be wrong, but this cannot happen since every new dependency release will be tested by greenkeeper and a new PR will be submitted when something breaks. No?

> Essentially builds from the same source don't have the same output anymore. [...] I've been through this in the past and it's pretty painful.

True. But were you notified in the past when a minor or patch upgrade broke your CI?

No trolling. I'm trying to understand the pros and cons.

mbenadda:
> this cannot happen

I'd say _might not happen as often_ probably? We don't know that greenkeeper is instantaneous and we're running builds all the time.

> notified [...] when a minor or patch upgrade broke your CI?

This obviously can't cover new feature work done. Maybe not critical but this is still a hole in our caret mitigation strategy.

Other cons:
- we might be warned, but it still makes those upgrades a "drop everything and fix this *now*" situation, rather than a deliberate and organized thing
- if a bunch of people work on the project, they'll have different deps installed depending on the time they pulled and that can't be fixed by a `yarn install` as the versions are still matching 


In my opinion using carets means losing control over dependencies. I'd rather not go down that rabbit hole than do it and hope greenkeeper mitigates the issues it causes

jmaupetit:
We must not forget that the lockfile is here to ensure that everyone will install locked dependencies (and locked 3rd-party dependencies) as long as dependency versions listed in the `package.json` match locked dependencies.

Hence, if a dependency upgrade breaks the CI, Greenkeeper will tell us so, the lockfile will be updated, and a `yarn install` on a developer's machine **will fix** the issue.

That said, I totally agree with you that not pinning dependencies is a really bad idea and feel a bit ashamed to propose caret versions as a "solution" to our initial problem that was: increase signal _vs_ noise ratio of dependency updates.

Therefore, to make everyone happy, I think our best option would be to stay with pinned dependencies and configure monthly updates with Renovate (if it has support for lockfile upgrade). @mbenadda should I go for it?

mbenadda:
Yes definitely! 🙏

Maybe in another PR so you can go ahead with the pyup stuff which is OK. (and maybe I can do it myself since I'm the one who suggested dropping greenkeeper for renovate 😉)

jmaupetit:
Okay, I've dropped my greenkeeper-related commit. 

> I can do it myself since I'm the one who suggested dropping greenkeeper for renovate

Do as you pleased. In any case, feel free to ping me if you need some help on the CI.

Issue #243, opened 2018-06-22
Show *why* a course appears in search results


mbenadda:
## Feature Request

**Is your feature request related to a problem or unsupported use case? Please describe.**
Sometimes it is non-obvious why a course appears in the list of results, even when search is properly configured and the course *does belong* in the results.

The first example of such behavior we've encountered is filtering by organizations:
- a course can have several linked organizations and therefore appear in the search results for any of them;
- however we only display the main organization for each course in the search results (for lack of space to show more);
- users will probably see this as a bug (`"Why is this course from University X appearing in my search for University Y? This is broken!"`).

**Describe the solution you'd like**
There are two ways to go about this: either consider this a general problem — understanding search result relevance — and find a general solution, or solve it for the `organizations` case only as hopefully other cases don't suffer from this (because the reason is obvious or because the relevant information does not appear in the search results at all)

**Describe alternatives you've considered**
Note all of these but the last one involve showing the org the user searched for when there is a mismatch between an active org filter and a course's main org.

General solutions:
- Typography, like bold or colored org name when the filter is applied
- Visual cue next to the org name

Specific solutions:
- Show a helpful "in collaboration with X" before the name of matched secondary orgs
- Show a separate subtitle "Courses made in collaboration with Univ Z, Univ Z2, Univ Z3" without those courses below

**Discovery, Documentation, Adoption, Migration Strategy**
N/A

**Do you want to work on it through a Pull Request?**
✅

PR #244, opened 2018-06-22
🐛(search) show the proper organization for courses

### LINKED ISSUE #134: Use `main_organization` on courses in the search app & in the frontend
## Bug Report

**Expected behavior/code**
Any use of related organizations for a course should use the `main_organization` field to identify the main org (for search, display etc.)

**Actual Behavior**
The `search` app does not make use of the `main_organization` field on courses and instead relies on the ordering of the `organizations` field to pick out the first org as the main one.
This behavior is not consistent with the way it is implemented in the new `Course` model.

**Steps to Reproduce**
1. Open the frontend to check out API responses & displayed elements in the list of courses

**Environment**
N/A

**Possible Solution**
Add `main_organization` in the ES index (for higher-priority in search results later on) and in the API responses in `search`. Make use of it on the frontend to display a course's main org.

N/A

mbenadda:
## Purpose

Courses can have several linked organizations, listed in the API on the `organizations` field. We operated on the assumption that the "main" organization, the one that should be displayed in search
results, was the first in the list.

That assumption was wrong in our current use, and also encodes some behavior implicitly.

Closes #134 

## Proposal

Use a specific field `organization_main` to convey that information instead.

Note #243 is a related problem that we decided not to solve in this PR.


sampaccoud:
You called the field "OrganizationMain" and "organization_main" but it is called "main_organization" in Python. Two cases:

- you really think "main_organization" is a bad name (because let me guess... generic then specific) and you should rename the field in Python as well. I would argue that name "main_organization" is often seen and derived from english,
- if you think "main_organization" is ok then we should stick to it everywhere.

I don't think having a discrepancy is a good option.

mbenadda:
I don’t see main_organization in python. I just see main_university which happens to be the legacy FUN name.

As for the name itself, it’s not just generic to specific, it’s also that it should imo be right next to organizations on the objects to make its existence and function obvious

sampaccoud:
@mbenadda https://github.com/openfun/richie/blob/master/apps/courses/models.py#L112

apps/courses/migrations/0001_squashed_0002_auto_20180625_1547.py
```diff
@@ -0,0 +1,75 @@
+# -*- coding: utf-8 -*-
```
mbenadda: I think this is what you were looking for @sampaccoud ?

sampaccoud: I would have deleted and recreated 0001 since we have nothing in operation yet but what you did should be better.

PR #251, opened 2018-06-26
Define filters in settings

mbenadda:
## Purpose

In order to make it easier to extend/modify the filters available on the search courses view, define them in a JSON settings file that can be extended and/or replaced, instead of hardcoding them in out TypeScript code directly.

## Proposal

- [x] Upgrade to TypeScript > 2.9.2 to have a working `resolveJsonModule`
- [x] Fix some issues linked to synthetic imports with `resolveJsonModules` (because it needs `esModuleInterop`)
- [x] Move filter definitons & order to settings.json


package.json
```diff
@@ -45,10 +45,10 @@
     "karma-sourcemap-loader": "0.3.7",
     "karma-webpack": "3.0.0",
     "source-map-loader": "0.2.3",
-    "ts-loader": "4.2.0",
+    "ts-loader": "4.4.1",
     "tslint": "5.9.1",
-    "typescript": "2.8.3",
-    "webpack": "4.12.0",
+    "typescript": "^3.0.0-dev.20180626",
+    "webpack": "4.10.2",
```
jmaupetit: Was this webpack downgrade required?

mbenadda: Wow that was a rebase mistake

PR #260, opened 2018-06-29
Move project to app

jmaupetit:
## Purpose

At the time of writing, Richie is a Django project embedding FUN's opinionated vision of what such portal should be. Let's make Richie a distributable Django app that can be customized and integrated in an existing Django (CMS) project.

## Proposal

- [x] make `richie` a meta application that bundles sub-apps (core, courses, persons, search, etc.)
- [x] move the application to a `src/` directory
- [x] add a `sandbox/` directory (the minimal Django project that will be used for development/tests)
- [x] package the application (`setup.{py,cfg}`)
- [x] move tests in a `test/` directory
- [x] move front-end sources to a more appropriate directory than the sandbox


richie/settings.py
```diff
@@ -191,12 +191,12 @@ class Base(DRFMixin, ElasticSearchMixin, Configuration):
...
-        "apps.courses",
-        "apps.persons",
-        "apps.search",
-        "plugins.large_banner",
+        # Richie stuff
+        "richie.apps.core",
+        "richie.apps.courses",
+        "richie.apps.persons",
+        "richie.apps.search",
+        "richie.plugins.large_banner",
```
sampaccoud: This will be my only comment on this awesome PR.

I think this plugins directory is not worth one more level in our code tree.

Why don't we make it an app so we would have:

- `richie.core`
- `richie.plugins`
- `richie.courses`
- `richie.persons`
- `richie.search`

It is nicer to be able to call `richie.course` instead of `richie.apps.course`.

In `richie.plugins`, we can put all the plugins that come with Richie to help build a learning portal but are not directly related to one of the app (cards, heros, large banners,...). In the other apps we can put the plugins that are related to this app (like @rmoch has already done a person plugin in the `persons` app).

Another option is to put these general plugins in `richie.core`.

WDYT?

.circleci/config.yml
```diff
@@ -228,7 +228,7 @@ jobs:
               -f docker/compose/ci/docker-compose.yml \
```
sampaccoud: s/threated/treated in commit message

sandbox/settings.py
```diff
@@ -9,8 +9,8 @@
 import raven
 from configurations import Configuration, values
 
-from .configurations.elasticsearch import ElasticSearchMixin
-from .configurations.rest_framework import DRFMixin
+from configurations_mixins.elasticsearch import ElasticSearchMixin
```
sampaccoud: s/configurations_mixins/configuration_mixins

setup.cfg
```diff
@@ -1,3 +1,110 @@
...
+    djangocms-file==2.0.2
+    djangocms-googlemap==1.1.1
+    djangocms-link==2.1.2
+    django-parler==1.9.2
+    djangocms-picture==2.0.6
+    djangocms-snippet==2.0.0
+    djangocms-style==2.0.2
+    djangocms-text-ckeditor==3.6.0
+    djangocms-video==2.0.4
+    djangocms-picture==2.0.6
```
sampaccoud: this is repeated

richie/settings.py
jmaupetit: IMO it's more explicit to put apps in an _ad hoc_ directory. But I have no strong feelings about this design decision.

sandbox/settings.py
```diff
@@ -151,7 +151,8 @@ class Base(DRFMixin, ElasticSearchMixin, Configuration):
     MEDIA_ROOT = os.path.join(DATA_DIR, "media")
```
sampaccoud: weird because you are just undoing what you did in the previous commit. This is not really new and should be squashed with the previous commit?

setup.cfg
```diff
@@ -129,6 +129,4 @@ python_files =
     test_*.py
```
sampaccoud: Note about your commit message: The command to create a demo app could move to the sandbox... it is not really part of the app itself.

jmaupetit: Unless it is distributed with the application itself?

sandbox/settings.py
jmaupetit: In the first step I moved the file name. In this step I removed it. It's not really undoing what I've done.

setup.cfg
jmaupetit: If you think factories must be part of the tests only and that this management command should stand in the sandbox, let's do it! (in a new PR)

mbenadda:
I think the JS indirection level is not necessary, we should put the front files directly in `richie-front`

setup.cfg
jmaupetit: FTR it has been decided that we will distribute factories and the create demo app management command. That can be useful when working with Richie in a new Django-CMS project.

PR #272, opened 2018-07-02
Automate python package building in the CI

jmaupetit:
## Purpose

We need to add Richie's python package building & publication steps to the CI.

## Proposal

- [x] Make front-end builds distributable by making Richie's statics a target directory
- [x] Move Sass base sources to `src/richie-front/sass`
- [x] Automate the package build in the CI
- [x] Publish new releases (git tags) to PyPI


.circleci/config.yml
```diff
@@ -480,9 +562,6 @@ jobs:
           keys:
           - v3-front-dependencies-{{ checksum "package.json" }}
           - v3-front-dependencies-
-      # Copy an up-to-date yarn.lock file
```
mbenadda: Why are we removing this?

jmaupetit: Because nothing has been saved to the workspace, hence there is nothing that needs to be restored/attached.

mbenadda: I'm not sure what's generally going on, was just pointing it out because it seemed unrelated to the rest of the PR.

Anyway 👍

jmaupetit: You are right: it should have been cleaned when we removed the yarn lockfile generation step.

PR #277, opened 2018-07-03
♻️(settings) move from JSON settings to TS settings

### LINKED ISSUE #276: Define settings in a TS file instead of JSON
## Feature Request

We need to translate a lot of strings in the settings (for #89). This is somewhat painful to handle in a JSON file and would require some code.

If the settings are a TS file, the react-intl babel plugin will take care of string extraction for us.

Writing a script to pick up message descriptors in JSON settings.

Will work out of the box. Consumers who want to extend/modify the default settings will just have to replace the module and import+extend it from their own settings module.

In progress.

mbenadda:
## Purpose

Settings are generally easier to wield when they are written in the same language as the code that uses them. This is proving true here as we need to automatically extract MessageDescriptors from our settings.

This will be a lot easier to do with TS settings than with JSON. Also, it makes consuming those settings a breeze throughout the app.

## Proposal

Transliterate JSON settings to TS.

Closes #276.


webpack.config.js
```diff
@@ -47,7 +46,7 @@ module.exports = {
     // Only load the locales we need to support in moment.js
     new webpack.ContextReplacementPlugin(
       /moment[\/\\]locale$/,
-      new RegExp(SETTINGS.SUPPORTED_LANGUAGES.join('|')),
+      new RegExp(['de', 'en', 'fr'].join('|')),
```
mbenadda: Duplication for the sake of simplicity.

sveetch: But you are "hardcoding" available languages ?

mbenadda: Yes. I can't take them from the settings anymore because settings are a TS file and webpack conf can't import TS.

Issue #89, opened 2018-04-18
Manage i18n/l10n for TS & TSX files


mbenadda:
Right now all user-facing text in the frontend is written in English with no support for localization and no i18n infrastructure.

We should add some i18n management facilities and a way to tag, collect and load translatable/translated strings.

- [x] Add a babel transpilation step so we can use react-intl's babel plugin string extractor (prerequisite)
- [x] Define settings in a TS file so strings in there can be easily extracted
- [x] Add react-intl and MessageDescriptors for translatable strings
- [x] Automatically extract strings with the babel plugin
- [x] Fetch translations to be injected in the app at runtime

🚧 Manage translations => for now we will be pulling & committing them manually into the repository.

jmaupetit:
Do you have any third-party app in mind to achieve this?

mbenadda:
### Translation management

I think SaaS is the way to go.

Transifex seems like the most established choice if we're eligible for [Transifex for Open Source](https://docs.transifex.com/projects/open-source-project).
We can also consider [Crowdin](https://crowdin.com/page/open-source-project-setup-request). I'm familiar with [Smartling](https://www.smartling.com) but their pricing is opaque, they don't seem to have support open source, and their workflows are more geared towards marketing site translated by third-parties than actual product string management.

### Tooling

It seems there's nothing as simple as I would have hoped (and what we used to have in Angular 1) but there seems to be some adequate tools.

The most promising one is [React-intl](https://github.com/yahoo/react-intl) along with its very own [parser](https://github.com/yahoo/babel-plugin-react-intl).

🎉
- takes care of our translations needs
- handles dates & times (so we could get rid of moment?)
- uses [standards](http://www.unicode.org/cldr/charts/28/supplemental/language_plural_rules.html)

💩
- needs string IDs (a hassle)


It would also require us to use babel to get the benefit of its parser plugin. This does complicate the build process. However, string extraction is actually one of the painful parts of handling i18n so I think it could be worth it.
As a side benefit we could get to use [babel's polyfilling](http://babeljs.io/docs/en/babel-polyfill/).

mbenadda:
Thought on this topic @sampaccoud @jmaupetit ?

sveetch:
Hi, for online translation tool there is also:

- https://weblate.org/
- http://pootle.translatehouse.org/

sampaccoud:
I'm not qualified but React-intl seems to be the default choice for i18n with React. 
Maybe we have more time to discuss the translation tool.

[...]

PR #280, opened 2018-07-03
Manage translations through react-intl

mbenadda:
## Purpose

This is step 3 & 4 of #89. We need to manage strings inside the JS app and to collect them for translation.

## Proposal

- [x] Add a dependency (`react-intl`) to manage i18n for us
- [x] Mark all strings and replace them with full message descriptors
- [x] Automatically extract them during `yarn build` through a babel plugin

### DIFF
--- a/src/richie-front/js/components/CourseGlimpse/CourseGlimpse.spec.tsx
+++ b/src/richie-front/js/components/CourseGlimpse/CourseGlimpse.spec.tsx
@@ -2,6 +2,7 @@ import '../../testSetup.spec';
 
 import { render } from 'enzyme';
 import * as React from 'react';
+import { IntlProvider } from 'react-intl';
 
 import { Course } from '../../types/Course';
 import { Organization } from '../../types/Organization';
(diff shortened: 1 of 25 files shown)


src/richie-front/js/components/CourseGlimpse/CourseGlimpse.spec.tsx
```diff
@@ -18,12 +19,14 @@ describe('components/CourseGlimpse', () => {
...
-      <CourseGlimpse course={course} organizationMain={organization} />,
+      <IntlProvider>
+        <CourseGlimpse course={course} organizationMain={organization} />
+      </IntlProvider>,
     );
 
     expect(wrapper.html()).toContain('Course 42');
     expect(wrapper.find('img').attr('src')).toContain('/thumbs/small.png');
-    expect(wrapper.html()).toContain('Starts on Mar 12, 2018');
+    expect(wrapper.html()).toContain('Starts on', 'Mar 12, 2018');
```
sampaccoud: Curious: why this change?

tsconfig.json
```diff
@@ -2,7 +2,7 @@
   "compilerOptions": {
     "allowSyntheticDefaultImports": true,
     "esModuleInterop": true,
-    "jsx": "react",
+    "jsx": "preserve",
```
sampaccoud: curious why this change?

src/richie-front/js/components/CourseGlimpse/CourseGlimpse.spec.tsx
mbenadda: There is now a `<span></span>` around the date

tsconfig.json
mbenadda: See my note in the 2nd commit message:
> Note: we now need TypeScript to leave JSX in its output so it can be picked up by the babel plugin. [...]

src/richie-front/js/components/CourseGlimpse/CourseGlimpse.spec.tsx
sampaccoud: oh then maybe it's a good idea to include the span in the test?

PR #283, opened 2018-07-04
Clean unused imports & drop momentjs

mbenadda:
## Purpose

Just some routine maintenance, cleaning unused imports & removing an unused dependency.

## Proposal

What it says on the label


mbenadda:
It's supposed to ([this rule](https://palantir.github.io/tslint/rules/no-unused-variable/)), maybe we should investigate this sometime 🤷‍♂️

PR #285, opened 2018-07-05
🐛(person) fix Person plugin template escaped html

sveetch:
## Purpose

Due to Page extension relation, Person plugin template does not reach
commonly the Person resume attribute value so it is not marked as safe
content, leading to escaped HTML.

## Proposal

This little fix just add the `|safe` filter to the resume variable
usage.

### DIFF
--- a/src/richie/apps/persons/tests/test_person_plugin.py
+++ b/src/richie/apps/persons/tests/test_person_plugin.py
@@ -49,10 +49,16 @@ class PersonPluginTestCase(TestCase):
         # A resume to related placeholder
         resume_placeholder = person_page.placeholders.get(slot="resume")
         add_plugin(
-            resume_placeholder, PlaintextPlugin, "en", **{"body": "A short resume"}
+            resume_placeholder,
+            PlaintextPlugin,
+            "en",
+            **{"body": "<p>A short resume</p>"}
         )
         add_plugin(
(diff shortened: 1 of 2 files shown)


sampaccoud:
@sveetch I approved but actually I think all bug fixes should include a test that fails before the fix and succeeds after the fix. WDYT?

sveetch:
@sampaccoud i understand the need of a test for the fix. (we are talking about an unique test that fail if fix is broken, but in fact should allways succeed if everything is clean?)

I will see to improve this PR with the fix test before merging.

src/richie/apps/persons/tests/test_person_plugin.py
```diff
@@ -81,7 +81,10 @@ def test_person_plugin_render(self):
         # pylint: disable=no-member
         self.assertContains(response, image.file.name)
         # Short resume should be present
-        self.assertContains(response, "A short resume")
+        self.assertInHTML(
+            "<p>A short resume</p>",
+            str(response.content, encoding='utf8'),
+        )
```
sampaccoud: self.assertContains(..., html=true) looks better because you can pass it directly the response.

Is it because assertInHTML fails and not assertContains when the content is escaped?

sveetch: `assertInHTML` seems more relevant to me because it can not fail because of spaces or indentation around content or inside tag. By the way finally the assertContain may do the same thing internally with response.

PR #286, opened 2018-07-05
🐳(user) use an un-privileged user in tests

### LINKED ISSUE #282: Switch to a non-privileged USER for the development Docker image
## Purpose

As discussed in #279, running tests in the container using the root user may hide potential issues of the application (_e.g._ permission-related issues). We need to make our test suite more reliable and trustworthy by running the test suite using a non-privileged user.

## Proposal

- [ ] Switch to the user ID=10000 when new dependencies have been installed in the development image
- [ ] Configure pytest and the linters to run tests with the app in read-only mode

jmaupetit:
## Purpose

As discussed in #279, running tests in the container using the root user may hide potential issues of the application (e.g. permission-related issues). We need to make our test suite more reliable and trustworthy by running the test suite using a non-privileged user.

## Proposal

- [x] Switch back to an un-privileged user in the dev image
- [x] Use `tmpfs` for the `/data` directory that requires write permission (test + CI)

Fix #282

### DIFF
--- a/docker/compose/ci/docker-compose.yml
+++ b/docker/compose/ci/docker-compose.yml
@@ -21,8 +21,7 @@ services:
   app:
     image: "richie:${CIRCLE_SHA1}${IMAGE_SUFFIX}-dev"
     env_file: env.d/ci
-    volumes:
-      - ./dist:/app/dist
     depends_on:
       - "db"
       - "elasticsearch"
+    tmpfs: /data
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -417,15 +417,16 @@ jobs:
           command: |
             docker load < docker/images/dev/richie.tar
 
+      # Run commands using CircleCI's user and mount the app volume to have
+      # write permission on it, and thus, being able to create the package.
       - run:
-          name: Build python packages
+          name: Build python package
           command: |
-            docker-compose \
-              -p richie-test \
-              -f docker/compose/ci/docker-compose.yml \
-              --project-directory . \
-              run --rm --no-deps app \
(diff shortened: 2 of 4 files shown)


docker/compose/ci/docker-compose.yml
```diff
@@ -21,8 +21,16 @@ services:
...
-    volumes:
-      - ./dist:/app/dist
     depends_on:
       - "db"
       - "elasticsearch"
+    tmpfs: /data
+
+  # The "package" service defines volumes that SHOULD pre-exist before mounting
+  # them so that we can run commands using CircleCI's user and have write
+  # permission them (to create the package)
```
sampaccoud: have write permission ON them?

docker/compose/ci/docker-compose.yml
```diff
@@ -21,8 +21,16 @@ services:
...
     image: "richie:${CIRCLE_SHA1}${IMAGE_SUFFIX}-dev"
     env_file: env.d/ci
-    volumes:
-      - ./dist:/app/dist
     depends_on:
       - "db"
       - "elasticsearch"
+    tmpfs: /data
+
+  # The "package" service defines volumes that SHOULD pre-exist before mounting
```
sampaccoud: not clear this explanation :-(
I understand that you are starting the image with another user so that you can generate the dist. 
My first idea would be: do we really need docker-compose for this as it does not require the db... ? Just using docker run would be easier to understand and avoid an akward docker-compose conf?

jmaupetit:
This alpine-related bug leaves me perplex. I think I should wait and re-launch this job on Monday.

PR #288, opened 2018-07-08
✅(tests) Improve course test on html

sveetch:
## Purpose

Plugin section lacking of tests around rendering and Course details
test is not relevant enough.

## Proposal

Plugin section tests has been improved to correctly test nested
content and header level in its markup. Course detail tests use
'assertInHTML' and its template has been improved.


sveetch:
This is under work and WIP stuff since i put some pointers to try to understand issues with draft stuff. 

Some related organization items show twice when page is published, maybe because of draft mechanic with djangocms.

I have to find bug cause and after see if it would be better to have some Organization method from model or manager, or even template filter to get the right content loop instead of doing it from template.

sveetch:
Also, i see Person plugin test should be improved with the same stuff than for Section plugin, but it depend of #285 validation before.

tests/plugins/section/test_plugin.py
```diff
@@ -1,21 +1,20 @@
...
+from richie.apps.core.tests.utils import CMSPluginTestCase
 from richie.plugins.section.cms_plugins import SectionPlugin
 from richie.plugins.section.factories import SectionFactory
 
 
-class SectionTests(TestCase):
+class SectionTests(CMSPluginTestCase):
     """Section plugin tests case"""
 
+    @transaction.atomic
```
sveetch: Worth to mention i don't really know why i needed to turn this test in atomic version, else it was breaking every other test following it. Troubling since another similar test in Person plugin does not need it.

src/richie/apps/core/tests/utils.py
```diff
@@ -5,6 +5,47 @@
...
+
+    def get_practical_plugin_context(self, extra_context=None):
+        """
+        Build a template context with dummy request object and
+        instanciated content renderer suitable to perform full rendering of
+        any plugin.
+
+        Note:
+            CMSTestCase use a dummy AnonymousUser on default behavior you may
+            override with a custom user if you an ``user`` attribute to your
```
sampaccoud: you wrote: "if you ... an user" -> missing a word and "an" should be "a".

sampaccoud: Note: We don't want to use the self.user from TestCase. We should use UserFactory inside each test to be explicit and self referencing.

src/richie/apps/core/tests/utils.py
```diff
@@ -5,6 +5,47 @@
...
+            ``cms_content_renderer`` item.
+        """
+        context = self.get_context()
+        if extra_context:
+            context.update(extra_context)
+
+        renderer = self.get_content_renderer(request=context["request"])
+
+        # 'cms_content_renderer' is the attempted item name from CMS rendering
+        # machinary
```
sampaccoud: s/machinary/machinery

tests/apps/courses/test_templates_course_detail.py
```diff
@@ -41,62 +45,66 @@ def test_course_cms_published_content(self):
...
         # Publish and ensure content is correct
         page.publish("en")
         response = self.client.get(url)
-        self.assertContains(
-            response,
+        html = str(response.content, encoding="utf8")
+
+        self.assertEqual(response.status_code, 200)
+
+        self.assertInHTML(
```
sampaccoud: I still feel that assertContains(..., html=True) is better than assertInHtml because it can be passed a response.

tests/apps/courses/test_templates_course_detail.py
```diff
@@ -41,62 +45,66 @@ def test_course_cms_published_content(self):
         # The page should not be visible before it is published
         url = page.get_absolute_url()
         response = self.client.get(url)
-        self.assertEqual(response.status_code, 404)
```
sampaccoud: why remove this test?

tests/apps/courses/test_templates_course_detail.py
```diff
@@ -14,6 +14,10 @@
 class CourseCMSTestCase(TestCase):
     """
     End-to-end test suite to validate the content and Ux of the course detail view
+
+    It's worth to notice related draft items (Person, Organization) are only
+    displayed for draft page so admin can preview them. But draft items are
+    hided from published page so common users can not see them.
```
sampaccoud: s/hided/hidden

tests/apps/courses/test_templates_course_detail.py
```diff
@@ -14,6 +14,10 @@
 class CourseCMSTestCase(TestCase):
     """
     End-to-end test suite to validate the content and Ux of the course detail view
+
+    It's worth to notice related draft items (Person, Organization) are only
+    displayed for draft page so admin can preview them. But draft items are
```
sampaccoud: s/for draft page/on a draft course page/

tests/apps/courses/test_templates_course_detail.py
```diff
@@ -41,62 +45,66 @@ def test_course_cms_published_content(self):
...
+        self.assertInHTML(
             '<li class="course-detail__content__organizations__item">{:s}</li>'.format(
                 organization2.extended_object.get_title()
             ),
-            html=True,
+            html,
         )
+
         # Draft organization should not be in response content
-        # TODO: This is wrong, we show draft but marked with a class modifier,
```
sampaccoud: why do you remove this TODO?

tests/plugins/section/test_plugin.py
sampaccoud: then this is not a solution. When a test is breaking the following tests, it means it is doing something wrong. Like modifying the active language outside of a "with" clause or something like this?

tests/plugins/section/test_plugin.py
```diff
@@ -45,39 +44,115 @@ def test_section_context_and_html(self):
...
+
+        expected_title = title_format.format(section.title)
+
+        # Check expected title and nested content are there
+        self.assertInHTML(expected_title, html)
+        self.assertInHTML(nested_content, html)
+
+    def test_section_title_level(self):
+        """
+        Title level can be changed from context variable 'header_level'.
```
sampaccoud: s/Title level/Header level

tests/plugins/section/test_plugin.py
```diff
@@ -45,39 +44,115 @@ def test_section_context_and_html(self):
...
+            language='en',
+        )
+
+        expected_title = title_format.format(section.title)
+
+        # Check expected title and nested content are there
+        self.assertInHTML(expected_title, html)
+        self.assertInHTML(nested_content, html)
+
+    def test_section_title_level(self):
```
sampaccoud: s/title/header

tests/apps/courses/test_templates_course_detail.py
sveetch: That was an error from mine for unknow reason.

tests/plugins/section/test_plugin.py
sveetch: Stated than this is an unknow issue provoked by CMSTestCase mixin.

sveetch:
Paused this PR until we find a solution around CMSTestCase. Probably trying to make our own mixin to get a functional context/request/rendered which is needed to correctly build placeholder and plugin content.

src/richie/apps/courses/templates/courses/cms/_organization_list-item_snippet.html
```diff
@@ -0,0 +1,16 @@
+{% if current_page.publisher_is_draft %}
```
sveetch: This template is redundant with an almost identical one from `fix-check-if-page-extension-is-published` branch on the way to be promoted to a pull request from @sampaccoud. It will still until the new PR has been merged, since tests here depends from this HTML.

sveetch:
Getting ride of `assertInHtml` in profit of `assertContain` except from HTML fragment for plugins

src/richie/apps/core/tests/utils.py
```diff
@@ -5,6 +5,58 @@
...
+    rendering.
+
+    Note:
+        CMSTestCase mixin from DjangoCMS cause unknow issue with transactions
+        on some tests and lead to necessity to decorate involved tests with
+        ``django.db.transaction`` (or to use it as a context manager) which is
+        undesired.
+
+        We should change this mixin to not inherit anymore from CMSTestCase
+        and try to build ourself a correct context.
```
sampaccoud: Maybe change the whole note to:
```
Note:
    The CMSTestCase mixin from DjangoCMS is making queries in its "tearDown" method which
    causes database transaction errors with tests that have already generated a database error.

    A temporary workaround is to decorate such tests with ``django.db.transaction`` (or to use it as a context manager) so that the test has its own transaction that is rolled back before the TearDown method sends its queries to the database.
```

tests/plugins/section/test_plugin.py
```diff
@@ -1,21 +1,21 @@
...
 
-class SectionTests(TestCase):
+# pylint: disable=too-many-ancestors
+class SectionTests(CMSPluginTestCase):
     """Section plugin tests case"""
 
+    @transaction.atomic
     def test_section_title_required(self):
         """
         A "title" is required when instantiating a section.
```
sampaccoud: Add same comment as on the TestCase? or referred to it?

sampaccoud:
The issue around CMSTestCase encountered in this PR was submitted to Django CMS with a proposal for fixing it: https://github.com/divio/django-cms/issues/6451

sveetch:
Fixed docstring and rebase from last master

PR #293, opened 2018-07-11
Fix getting linked page extensions via many-to-many

sampaccoud:
## Purpose

Many to many relations between page extenions are copied when a page is published. We end up with links between draft and published versions of the page extension but the draft version should serve as the reference. I added tests to evidence the bug and then fixed it.

## Proposal

- [x] Improve tests to surface the issue
- [x] Add a custom "drafts" filter to easily access the relevant linked page extensions 

TODO: improving tests on the html generated by our templates will be done in https://github.com/openfun/richie/pull/288

### DIFF
--- a/tests/apps/persons/test_person_plugin.py
+++ b/tests/apps/persons/test_person_plugin.py
@@ -36,7 +36,6 @@ class PersonPluginTestCase(TestCase):
         other_page_title = "other page"
         create_page(other_page_title, "richie/fullwidth.html", settings.LANGUAGE_CODE)
         plugin_form = PersonPluginModelForm()
-        print(plugin_form)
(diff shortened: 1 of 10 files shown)


tests/apps/persons/test_person_plugin.py
```diff
@@ -36,7 +36,6 @@ class Meta:
         other_page_title = "other page"
         create_page(other_page_title, "richie/fullwidth.html", settings.LANGUAGE_CODE)
         plugin_form = PersonPluginModelForm()
-        print(plugin_form)
```
sveetch: Oops, nice catch

sampaccoud:
@sveetch 🎉 thanks. I don't like people to be stuck because of my bugs!

Issue #292, opened 2018-07-11
Implement default additional filters on the backend side


mbenadda:
## Feature Request

**Is your feature request related to a problem or unsupported use case? Please describe.**
The `'new'`, `'availability'` and `'language'` filters are described by the frontend's default settings; they are expected and displayed but not yet implemented in the API.

**Describe the solution you'd like**
We need to add the corresponding queries to the default backend settings and implement the logic that can handle settings-defined filters in the ViewSet itself.

**Describe alternatives you've considered**
N/A. Right now this is our best proposal for filters extensibility.

**Discovery, Documentation, Adoption, Migration Strategy**
We will need to document this feature, using the default filters as an example: users can extend/replace frontend and backend settings to customize filters and hand custom-made ES requests to our code.

**Do you want to work on it through a Pull Request?**
👍

PR #302, opened 2018-07-17
Prepare to implement default filters on the api

mbenadda:
## Purpose

As we were working on #292, we noticed some necessary improvements we should undertake before getting to the heart of the matter.

## Proposal

- [x] simplify indexers by using `@staticmethod` & `@classmethod` instead of instantiating them
- [x] make `ES_INDEXES` a named tuple instead of an array: we're going to need the names, it ensures validity for downstream devs, and it remains easy to consume
- [x] import indexers in viewsets through settings
- [x] break in our new viewset/list design with the simple orgs & subjects viewsets

### DIFF
--- a/src/richie/apps/search/indexers/subjects.py
+++ b/src/richie/apps/search/indexers/subjects.py
@@ -3,7 +3,8 @@ ElasticSearch subject document management utilities
 """
 from django.conf import settings
 
-from ..exceptions import IndexerDataException
+from ..exceptions import IndexerDataException, QueryFormatException
+from ..forms import SubjectListForm
 from ..partial_mappings import MULTILINGUAL_TEXT
 from ..utils.api_consumption import walk_api_json_list
 from ..utils.i18n import get_best_field_language
@@ -56,3 +57,40 @@ class SubjectsIndexer:
                 es_subject["_source"]["name"], best_language
             ),
         }
+
+    @staticmethod
+    def build_es_query(request):
+        """
+        Build an ElasticSearch query and its related aggregations, to be consumed by the ES client
+        in the Subjects ViewSet
+        """
+        # Instantiate a form with our query_params to check & sanitize them
+        params_form = SubjectListForm(request.query_params)
+
+        # Raise an exception with error information if the query params are not valid
+        if not params_form.is_valid():
+            raise QueryFormatException(params_form.errors)
+
+        # Build a query that matches on the name field if it was handed by the client
+        # Note: test_elasticsearch_feature.py needs to be updated whenever the search call
+        # is updated and makes use new features.
+        if params_form.cleaned_data.get("query"):
+            query = {
+                "query": {
+                    "match": {
+                        "name.fr": {
+                            "query": params_form.cleaned_data.get("query"),
+                            "analyzer": "french",
+                        }
+                    }
+                }
+            }
+        # Build a match_all query by default
+        else:
+            query = {"query": {"match_all": {}}}
+
+        return (
+            params_form.cleaned_data.get("limit"),
+            params_form.cleaned_data.get("offset") or 0,
+            query,
+        )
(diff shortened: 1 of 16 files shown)


src/richie/apps/search/viewsets/courses.py
```diff
@@ -2,14 +2,17 @@
...
 
 from elasticsearch.exceptions import NotFoundError
 from rest_framework.response import Response
 from rest_framework.viewsets import ViewSet
 
 from ..forms import CourseListForm
-from ..indexers.courses import CoursesIndexer
+
+# Get the courses indexer from the settings
+indexer = import_string(settings.ES_INDICES.courses)
```
jmaupetit: I think naming it `CoursesIndexer` was more explicit; why did you change it to a generic name?

mbenadda: I can name it `courses_indexer` but not `CoursesIndexer` :)

jmaupetit: :sweat: True. Let's do this.

src/richie/apps/search/indexers/subjects.py
```diff
@@ -55,3 +57,40 @@ def format_es_subject_for_api(es_subject, best_language):
...
+        # Build a query that matches on the name field if it was handed by the client
+        # Note: test_elasticsearch_feature.py needs to be updated whenever the search call
+        # is updated and makes use new features.
+        if params_form.cleaned_data.get("query"):
+            query = {
+                "query": {
+                    "match": {
+                        "name.fr": {
+                            "query": params_form.cleaned_data.get("query"),
+                            "analyzer": "french",
```
sampaccoud: it looks like this should be configurable but maybe not in the frame of this PR.

tests/apps/search/test_indexers_subjects.py
```diff
@@ -115,3 +121,42 @@ def test_format_es_subject_for_api(self):
             SubjectsIndexer.format_es_subject_for_api(es_subject, "en"),
             {"id": 89, "image": "example.com/image.png", "name": "Computer science"},
         )
+
+    def test_build_es_query(self):
```
sampaccoud: this was not tested yet or you forgot to remove the existing test code?

mbenadda: The return values for the viewsets were tested (the viewsets were previously building the queries).

Now it's tested more granularly on the indexer

sampaccoud: so you may want to change the test on the viewset to test what is now relevant to test? I'm thinking for example mocking the call to build_es_query and raising the exception to check that the response has the error?

sampaccoud: and stop testing the format of the query there... ?

mbenadda: Right now the tests on the viewset methods are perhaps "less obvious" (more integration than unit) but also definitely less coupled to the internal implementation than they would be with a mock. Those cases you mention are covered by the existing tests.

I do not think it is worth it to change these tests to move testing upstream and increase coupling when we have existing tests covering what we want to cover.

Also note it would be simple enough for subjects & orgs but the courses test is 600 line long. In my opinion it would not be a good investment to rewrite them on the indexers just for the sake of it.


I suggest we keep them as they are and move them to the indexers only if/when the necessity arises

sampaccoud: I follow you.

tests/apps/search/test_viewsets_organizations.py
```diff
@@ -225,4 +161,5 @@ def test_search_organizations_with_invalid_params(self):
 
         # The client received a BadRequest response with the relevant data
         self.assertEqual(response.status_code, 400)
+        print(response.data["errors"])
```
sampaccoud: print fail

Issue #220, opened 2018-06-14
Basic Standard template and look


2018D:
## Feature Request

**Is your feature request related to a problem or unsupported use case? Please describe.**
The basic template and static files in the app are barebone. I'd like to flesh it out so that right out the box we can have a nice portal website. Starting with the course-detail page.

**Describe the solution you'd like**
I'll edit the templates and the static files.
The templates will have limited placeholder to allow anyone to edit on the cms without breaking the webpage's structure. The limited placeholder will make up a form that people fill in.

**Describe alternatives you've considered**
Without the limited placeholder, you would have a lot of difficulty maintaining the same look and feel throughout the whole website, especially if you hand it over to non web designer.

**Discovery, Documentation, Adoption, Migration Strategy**
I'll use {% extends template %} on the plugins to have it as modular as possible. Basic look and feel will be maintained by the django-cms core.

**Do you want to work on it through a Pull Request?**
I'll make a pull request as soon as I'll have the course_detail page working how I want it to.

2018D:
Seeing that Mehdi is working on the organisation page, i went and focused this issue onto the course-detail page.

sveetch:
Hi @2018D just to let you know i'm working now on the course template to be able to structure and our default content for a Course page.

We will add about ten default content (we retained the ones you added in your branch) almost all with their own placeholder with a basic HTML structure directly using BEM convention (so without bootstrap components).

2018D:
About the «Course key of active course session» on the side, are you planning on making it into a link towards something? or is it there as information only?

PR #303, opened 2018-07-17
💄(courses) subject detail: template and style

2018D:
Copy organization detail style. Adapt copied template and style for subject detail. 
Adapt related tests.

## Purpose

To have a basic style for subject detail pages. #220 

## Proposal

Imitate what was done for organization detail. Adjust it to suit subject detail.

### DIFF
--- a/src/richie/apps/courses/templates/courses/cms/subject_detail.html
+++ b/src/richie/apps/courses/templates/courses/cms/subject_detail.html
@@ -4,50 +4,50 @@
 {% block content %}
 <div class="subject-detail">
 
-  <h1 class="subject-detail__title">{{ current_page.get_title }}</h1>
-
   {% with subject=current_page.subject header_level=2 %}
 
-    <div class="subject-detail__content">
-      <div class="subject-detail__content__banner">
-        {% placeholder "banner" %}
-      </div>
+    <div class="subject-detail__banner">
+      {% placeholder "banner" %}
+    </div>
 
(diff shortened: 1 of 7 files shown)


src/richie/apps/courses/templates/courses/cms/subject_detail.html
```diff
@@ -4,49 +4,50 @@
 {% block content %}
 <div class="subject-detail">
```
sveetch: You may add templatetag `{% spaceless%}...{% endspaceless %}` around content just under this div, so HTML is a bit lighter and unnecessary whitespaces won't never make tests fails.

2018D: I ended up not adding the spaceless tag after all.

sveetch: and because ... ?

2018D: I was trying to stay consistent with other templates.
I added them now.
Thanks for the review.

sampaccoud:
@2018D There is a problem on your PR as it is making unwanted changes on courses and organizations. I don't know what happened, maybe a rebase fail? Can you take a look at it?

2018D:
@sveetch recommanded that i add spaceless tags to the subject template. To be consistent, i also added them to course and organization's templates.

PR #304, opened 2018-07-18
Fix checking if page extension is published

sampaccoud:
## Purpose

When checking if a page is published, we were just checking the existence of a public version of the page. It turns out that pages can be published and later unpublished... so the public version of a page may exist but not be in a "published" status. 

Furthermore, the publication of a page is handled language by language.

## Proposal

- [x] Improve tests to surface the problem for all existing templates ;
- [x] Add a method "check_publication" on all page extension models to check the publication status of a page for a specific language or for the current language ;
- [x] Use the new "check_publication" method to check the publication status of a page extension in all our templates ;
- [x] Fix BEM structure of list items in templates (organizations, courses, subjects) ;
- [x] Related bug fix: reactivate linting on the "tests" directory.

### DIFF
--- a/src/richie/apps/courses/templates/courses/cms/organization_fragment.html
+++ b/src/richie/apps/courses/templates/courses/cms/organization_fragment.html
@@ -1,16 +1,16 @@
 {% if current_page.publisher_is_draft %}
     {% if organization.check_publication is True %}
-        <li class="course-detail__content__organizations__item{% if is_main %}--main{% endif %}">
+        <li class="course-detail__content__organizations__item{% if is_main %} course-detail__content__organizations__item--main{% endif %}">
             {{ organization.public_extension.extended_object.get_title }}
         </li>
     {% else %}
-        <li class="course-detail__content__organizations__item--draft{% if is_main %} course-detail__content__organizations__item--main{% endif %}">
+        <li class="course-detail__content__organizations__item course-detail__content__organizations__item--draft{% if is_main %} course-detail__content__organizations__item--main{% endif %}">
             {{ organization.extended_object.get_title }}
         </li>
     {% endif %}
     {# If the current page is the published version, show only the organizations that are published #}
 {% elif organization.check_publication is True %}
-    <li class="course-detail__content__organizations__item{% if is_main %}--main{% endif %}">
+    <li class="course-detail__content__organizations__item{% if is_main %} course-detail__content__organizations__item--main{% endif %}">
     {{ organization.public_extension.extended_object.get_title }}
     </li>
 {% endif %}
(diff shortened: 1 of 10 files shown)


src/richie/apps/courses/templates/courses/cms/organization_fragment.html
```diff
@@ -0,0 +1,16 @@
+{% if current_page.publisher_is_draft %}
+    {% if organization.check_publication is True %}
+        <li class="course-detail__content__organizations__item{% if is_main %}--main{% endif %}">
```
sveetch: BEM involve to add modifier as an additional class it should be something like:

```
course-detail__content__organizations__item{% if is_main %} course-detail__content__organizations__item--main{% endif %}
```

src/richie/apps/courses/templates/courses/cms/organization_fragment.html
```diff
@@ -0,0 +1,16 @@
...
+            {{ organization.public_extension.extended_object.get_title }}
+        </li>
+    {% else %}
+        <li class="course-detail__content__organizations__item--draft{% if is_main %} course-detail__content__organizations__item--main{% endif %}">
+            {{ organization.extended_object.get_title }}
+        </li>
+    {% endif %}
+    {# If the current page is the published version, show only the organizations that are published #}
+{% elif organization.check_publication is True %}
+    <li class="course-detail__content__organizations__item{% if is_main %}--main{% endif %}">
```
sveetch: Same feedback than the one from line 3

src/richie/apps/courses/templates/courses/cms/organization_fragment.html
```diff
@@ -0,0 +1,16 @@
+{% if current_page.publisher_is_draft %}
+    {% if organization.check_publication is True %}
+        <li class="course-detail__content__organizations__item{% if is_main %}--main{% endif %}">
+            {{ organization.public_extension.extended_object.get_title }}
+        </li>
+    {% else %}
+        <li class="course-detail__content__organizations__item--draft{% if is_main %} course-detail__content__organizations__item--main{% endif %}">
```
sampaccoud: @sveetch ok for your above comment but why don't we write this here then?
```
<li  class="course-detail__content__organizations__item course-detail__content__organizations__item--draft{% if is_main %} course-detail__content__organizations__item--main{% endif %}">
```

sveetch: Because i missed it :) You are correct this is the right way to do

sampaccoud: Ok so I need to change all templates as well: subject, organization, course...

PR #307, opened 2018-07-20
Build the courses query in CoursesIndexer

mbenadda:
## Purpose

Just like we did for the Organizations & Subjects, we're delegating query building to the Indexer from the ViewSet to enable extension & personalization of the query.

## Proposal

- [x] Move query building from `CoursesViewSet` to `CoursesIndexer`
- [x] Refactor the tests so the output from build_es_query is actually tested in the indexer tests.

PR 2/3 for #292


tests/apps/search/test_indexers_courses.py
```diff
@@ -193,3 +202,331 @@ def test_format_es_course_for_api(self):
                 "title": "Duis eu arcu erat",
             },
         )
+
+    def test_build_es_query_search_all_courses(self):
+        # Build a request stub
```
sampaccoud: All your methods in this file have no docstring. If there is no error it means that the linting is not done on these files!?

PR #309, opened 2018-07-20
✨(ckeditor) create a new simple text plugin with CKEditor widget

sveetch:
## Purpose

Common 'TextPlugin' plugin from djangocms_text_ckeditor has bad
behavior of creating orphan plugin required for its nesting plugin
feature. This behavior is breaking plugin limit constraint from
setting 'CMS_PLACEHOLDER_CONF'.

## Proposal

Creating our own plugin to a add text with CKEditor UI.
It does not accept nesting plugin, don't append orphan plugin and
respect constraint from setting 'CMS_PLACEHOLDER_CONF'. Plugins
readme has been updated also for some minor fixes.

Then after propagate usage of this plugin in every placeholder constraints and update test and factory to use our plugin instead of TextPlugin.

* [x] Create new simple text plugin with CKEditor widget;
* [x] Use our plugin instead of TextPlugin in configuration, factories and tests.


src/richie/plugins/simple_text_ckeditor/models.py
```diff
@@ -0,0 +1,40 @@
...
+    def __init__(self, *args, **kwargs):
+        super(SimpleText, self).__init__(*args, **kwargs)
+        self.body = force_text(self.body)
+
+    def save(self, *args, **kwargs):
+        super(SimpleText, self).save(*args, **kwargs)
+
+        # Clean HTML from potential XSS content
+        body = self.body
+        body = clean_html(body, full=False)
```
sampaccoud: Why not directly
```
self.body = clean_html(self.body, full=False)
```

src/richie/plugins/simple_text_ckeditor/cms_plugins.py
```diff
@@ -0,0 +1,36 @@
+"""
+Simple text CMS plugins
```
sampaccoud: s/plugins/plugin

src/richie/plugins/simple_text_ckeditor/models.py
```diff
@@ -0,0 +1,40 @@
...
+
+    def __str__(self):
+        return Truncator(strip_tags(self.body)).words(6, truncate="...")
+
+    def __init__(self, *args, **kwargs):
+        super(SimpleText, self).__init__(*args, **kwargs)
+        self.body = force_text(self.body)
+
+    def save(self, *args, **kwargs):
+        super(SimpleText, self).save(*args, **kwargs)
```
sampaccoud: Why do you need to save before sanitizing the body  ?  Actually you should sanitize in the model's clean method.

tests/plugins/simple_text_ckeditor/test_plugin.py
```diff
@@ -0,0 +1,52 @@
...
+
+class SimpleTextTests(TestCase):
+    """SimpleText plugin tests case"""
+
+    def test_simpletext_create_success(self):
+        """
+        Simple text plugin creation success
+        """
+        simpletext = SimpleTextFactory(body="Foo")
+        self.assertEqual("Foo", simpletext.body)
```
sampaccoud: You should test with a more complex string with accepted html and forbidden tags to test your sanitizer.

src/richie/plugins/simple_text_ckeditor/models.py
sveetch: Honestly i just copied a ligne from djangocms_text_ckeditor plugin.

src/richie/plugins/README.md
```diff
@@ -3,25 +3,28 @@
...
 
-They sould be added to Django project `INSTALLED_APPS` constant as follow:
+They should be added to Django project `INSTALLED_APPS` constant as follow:
 
     INSTALLED_APPS = (
         ...
-        'plugins.large_banner',
+        "richie.plugins.large_banner",
+        "richie.plugins.section",
+        "richie.plugins.simple_text_ckeditor",
```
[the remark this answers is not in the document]
sveetch: Look at the full file, the indentation is right.

src/richie/plugins/simple_text_ckeditor/models.py
sveetch: This is required because inherited CMSPlugin model is performing some job parent relation and plugin position: https://github.com/divio/django-cms/blob/develop/cms/models/pluginmodel.py#L259

sampaccoud: yes but you can sanitize you "body" field then call "super". But calling "super" twice is akward and inefficient IMO.

sveetch:
Fixed last review feedbacks and added some tests around html5lib cleaner and sanitizer

sandbox/settings.py
```diff
@@ -370,6 +371,54 @@ class Base(DRFMixin, ElasticSearchMixin, Configuration):
...
+                "CreateDiv",
+                "HorizontalRule",
+            ],
+            ["Source"],
+        ],
+    }
+    # Share the same config for djangocms_text_ckeditor field and derived
+    # CKEditor widgets/fields
+    CKEDITOR_SETTINGS["toolbar_HTMLField"] = CKEDITOR_SETTINGS["toolbar_CMS"]
+
```
sampaccoud: Let's not do like Open edX. I think we should have default settings declared directly in the plugin so that adding this plugin to a project just works.

I propose the following pattern: create a settings.py file in the plugin with somethinng like:
```
CKEDITOR_SETTINGS = getattr(settings, 'CKEDITOR_SETTINGS', {...your whole dict definition...})
```
In the plugin, you can use the setting value by doing:
```
from myplugin.settings import CKEDITOR_SETTINGS

def foo():
    ...
    print(CKEDITOR_SETTINGS)

File: src/richie/plugins/simple_text_ckeditor/cms_plugins.py [hunk 7]
```diff
@@ -0,0 +1,36 @@
...
+    name = _("Simple text")
+    render_template = "richie/simple_text_ckeditor/simple_text.html"
+    cache = True
+    module = settings.FUN_PLUGINS_GROUP
+    allow_children = False
+    disable_child_plugins = True
+
+    fieldsets = ((None, {"fields": ["body"]}),)
+
+    def render(self, context, instance, placeholder):
```
sampaccoud: missing method docstring. How come you don't get an error?

src/richie/plugins/simple_text_ckeditor/forms.py
```diff
@@ -0,0 +1,13 @@
+# -*- coding: utf-8 -*-
+from django import forms
+
+from djangocms_text_ckeditor.widgets import TextEditorWidget
+
+from .models import SimpleText
+
+
+class CKEditorPluginForm(forms.ModelForm):
+    class Meta:
```
sampaccoud: same missing module and class docstrings... ?

src/richie/plugins/simple_text_ckeditor/cms_plugins.py
sveetch: dont know, i'm making test-back and lint-back makefile action before each push

sandbox/settings.py
sveetch: As you may know, `CKEDITOR_SETTINGS` is common setting from djangocms_text_ckeditor itself that is used from its plugin. Furthermore we don't use it directly, this is the widget which get it from settings then pass it to the CKEditor UI configuration. Except rewriting our own widget, i don't think we can do what you are proposing.

The plugins with CKEditor UI can work when it's not setted, you may see this as a sample config from sandbox that we may eventually suggest from documentation.

src/richie/plugins/simple_text_ckeditor/cms_plugins.py
sveetch: By the way it seems the behavior since some time because older plugins like section and large_banner does not have docstring on this method too, also some methods like `__str__` or `save` method don't have one.

sveetch:
Added second part to replace TextPlugin explicit usage by CKEditorPlugin instead

PR #316, opened 2018-07-26
🐛(docker) use local user ID in the dev image

### LINKED ISSUE #305: App container is launched with wrong user ID
## Bug Report

**Expected behavior/code**
App container should be launched with user ID 1000 as configured with container FS permissions. This is required to create file like uploading image file in page contents.

**Actual Behavior**
Actually app container seems launched with user ID 10000 and so trying to upload a file lead to an error like `[Errno 13] Permission denied: '/data/media/organizations/logo/6.jpg'`.

**Steps to Reproduce**
1. Launch richie app container with `make run`;
2. Go to a page (logged as admin) where you can upload an image or any other file kind;
3. Upload a file;
4. You get a `Permission denied` error.

**Environment**
- Richie version: 0.1.0 (1696320)
- Platform: Ubuntu 18.04

**Possible Solution**
From @jmaupetit it seems related to our usage of `docker-compose up -d` in `make run` where we don't set an user.

jmaupetit:
# Purpose

When starting the development stack using the `docker-compose up` command, compose uses the image default `USER` to run defined command (_e.g._ `10000` in the "dev" Dockerfile). Hence the Django development server runs using this user ID. This is an issue when developing with local volumes mounted in this container, because, the server may want to create new files in these volumes and unfortunately the running user (with UID `10000`) has no permission to write in the aforementioned volumes.

## Proposal

- [x] Add a `UID` build `ARG` to be able to customize the default user ID in the "dev" docker image

Fix #305

### DIFF
--- a/docker/images/dev/Dockerfile
+++ b/docker/images/dev/Dockerfile
@@ -1,7 +1,12 @@
 # The base image we inherit from is richie:latest, but you can override this by
-# passing a build argument to your build command, e.g.:
+# passing the BASE_TAG build argument to your build command. You can also
+# override the default container running user ID (e.g. 1000) thanks to the UID
+# build argument (we recommend using yours). An example follows:
 #
-# docker build --build-arg BASE_TAG=${CIRCLE_SHA1} .
+# docker build \
+#   --build-arg BASE_TAG=${CIRCLE_SHA1}\
+#   --build-arg UID=$(id -u) \
+#   .
 #
 ARG BASE_TAG=latest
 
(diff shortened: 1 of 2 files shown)


docker/images/dev/Dockerfile
```diff
@@ -1,7 +1,12 @@
 # The base image we inherit from is richie:latest, but you can override this by
-# passing a build argument to your build command, e.g.:
+# passing the BASE_TAG build argument to your build command. You can also
+# override the default container running user ID (e.g. 1000) thanks to the UID
+# build argument (we recommend using yours). An example follows:
```
sampaccoud: Should be called USER (or USER_ID)? Not sure UID is explicit and the param in Dockerfile is called USER.

jmaupetit: It refers to the UID shell variable. I thought it was explicit...

sampaccoud: ok then

PR #317, opened 2018-07-27
Implement custom+default filters on the api

### LINKED ISSUE #292: Implement default additional filters on the backend side

mbenadda:
## Purpose

We need to implement some default filters for development in the sandbox and to enable providers to quickly have an up-and-running search experience when they start working with Richie.

We wanted these default filters to be easy to extend/replace without having to re-do or copy/pasta the whole query building logic which can be daunting for people unfamiliar with ES.

## Proposal

Provide a way to add filters through settings (and some useful defaults for them) by specifying key/value pairs along with form fields to validate them and actual ES query fragments to use them.

- [x] Set up default filters in our `defaults.py`, overridable through settings
- [x] Extend the courses list request params form with fields from defaults/settings
- [x] Use the custom filters as we build the main query & aggregations in the courses indexer
- [x] Structure consistent facets out of these filters for the API response

Note: we still have some work to complete this PR:
- [x] Update tests for the courses viewset
- [x] Enable multiple choices for custom filters (`ChoiceField` => `MultipleChoiceField` in defaults + support when we build the main query in the indexer)
- [ ] Support drilldown filters [this may be done in a separate PR] 

## Refs

Closes #292.
Follows #302 & #307.

### DIFF
--- a/src/richie/apps/search/defaults.py
+++ b/src/richie/apps/search/defaults.py
@@ -2,9 +2,52 @@
 Import custom settings and set up defaults for values the Search app needs
 """
 from django.conf import settings
+from django.forms import MultipleChoiceField
+
+import arrow
 
 # Define our aggregations names, for our ES query, which will match with the field
 # names on the objects & the facets we return on the API response
 RESOURCE_FACETS = getattr(
     settings, "RICHIE_SEARCH_RESOURCE_FACETS", ["organizations", "subjects"]
+)
+
+FILTERS_HARDCODED_DEFAULT = {
+    "availability": {
+        "field": MultipleChoiceField,
+        "choices": {
+            "coming_soon": [
+                {
+                    "range": {
+                        "start_date": {
+                            "gte": arrow.utcnow().datetime,
+                            "lte": arrow.utcnow().shift(weeks=+12).datetime,
+                        }
+                    }
+                }
+            ],
+            "current": [
+                {"range": {"start_date": {"lte": arrow.utcnow().datetime}}},
+                {"range": {"end_date": {"gte": arrow.utcnow().datetime}}},
+                {"range": {"enrollment_start_date": {"lte": arrow.utcnow().datetime}}},
+                {"range": {"enrollment_end_date": {"gte": arrow.utcnow().datetime}}},
+            ],
+            "open": [
+                {"range": {"start_date": {"lte": arrow.utcnow().datetime}}},
+                {"range": {"end_date": {"gte": arrow.utcnow().datetime}}},
+            ],
+        },
+    },
+    "language": {
+        "field": MultipleChoiceField,
+        "choices": {lang: [{"term": {"language": lang}}] for lang in ["en", "fr"]},
+    },
+    "new": {
+        "field": MultipleChoiceField,
+        "choices": {"new": [{"term": {"session_number": 1}}]},
+    },
+}
+
+FILTERS_HARDCODED = getattr(
+    settings, "RICHIE_SEARCH_FILTERS_HARDCODED", FILTERS_HARDCODED_DEFAULT
 )
--- a/src/richie/apps/search/defaults.py
+++ b/src/richie/apps/search/defaults.py
@@ -1,0 +1,10 @@
+"""
+Import custom settings and set up defaults for values the Search app needs
+"""
+from django.conf import settings
+
+# Define our aggregations names, for our ES query, which will match with the field
+# names on the objects & the facets we return on the API response
+RESOURCE_FACETS = getattr(
+    settings, "RICHIE_SEARCH_RESOURCE_FACETS", ["organizations", "subjects"]
+)
--- a/src/richie/apps/search/forms.py
+++ b/src/richie/apps/search/forms.py
@@ -3,6 +3,7 @@ Validate and clean request parameters for our endpoints using Django forms
 """
 from django import forms
 
+from .defaults import FILTERS_HARDCODED
 from .fields.array import ArrayField
 from .fields.datetimerange import DatetimeRangeField
 
@@ -11,6 +12,17 @@ class CourseListForm(forms.Form):
     """
     Validate the query string params in the course list request
     """
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        # Add in the fields for the custom filters from defaults/settings
+        for filter_key in FILTERS_HARDCODED:
+            self.fields[filter_key] = FILTERS_HARDCODED[filter_key]["field"](
+                choices=[
+                    (value, value) for value in FILTERS_HARDCODED[filter_key]["choices"]
+                ],
+                required=False,
+            )
 
     end_date = DatetimeRangeField(required=False)
     enrollment_end_date = DatetimeRangeField(required=False)
--- a/src/richie/apps/search/indexers/courses.py
+++ b/src/richie/apps/search/indexers/courses.py
@@ -3,6 +3,7 @@ ElasticSearch course document management utilities
 """
 from django.conf import settings
 
+from ..defaults import RESOURCE_FACETS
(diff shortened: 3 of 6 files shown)


src/richie/apps/search/defaults.py
```diff
@@ -0,0 +1,54 @@
...
+# Define our aggregations names, for our ES query, which will match with the field
+# names on the objects & the facets we return on the API response
+RESOURCE_FACETS = getattr(
+    settings, "RICHIE_SEARCH_RESOURCE_FACETS", ["organizations", "subjects"]
+)
+
+FILTERS_HARDCODED_DEFAULT = {
+    "availability": {
+        "field": ChoiceField(
+            choices=[("coming_soon", ""), ("current", ""), ("open", "")],
```
sampaccoud: Why empty human names ?

src/richie/apps/search/indexers/courses.py
```diff
@@ -146,59 +149,117 @@ def build_es_query(request, facets):
...
+                                filter(
+                                    lambda query: filter_key is not query[0], queries
+                                ),
+                                FILTERS_HARDCODED[filter_key]["choices"][choice],
+                            )
+                        }
+                    }
+                }
 
         # Build organizations and subjects terms aggregations for our query
```
sampaccoud: Outdated comment

src/richie/apps/search/defaults.py
```diff
@@ -0,0 +1,54 @@
...
+                {"range": {"enrollment_end_date": {"gte": arrow.utcnow().datetime}}},
+            ],
+            "open": [
+                {"range": {"start_date": {"lte": arrow.utcnow().datetime}}},
+                {"range": {"end_date": {"gte": arrow.utcnow().datetime}}},
+            ],
+        },
+    },
+    "language": {
+        "field": ChoiceField(choices=[("en", ""), ("fr", "")], required=False),
```
sampaccoud: I think you should just pass the class ChoiceField or MultipleChoiceField. The object you pass for choices is akward and a useless duplicate. A structure that would maybe be more self explanatory : 
```
FILTERS_HARDCODED_DEFAULT = {
    availability : [
        {
            name :  "coming_soon",
            human_name :  "Coming soon",
            query_fragment :  ...,
            field_class :  MultipleChoiceField
        },
        {...}
    ]
}
```

src/richie/apps/search/defaults.py
mbenadda: There is literally no use for these human names, they are not displayed anywhere not useful for anything as far as I can tell.

I think they're just required because the Django field classes were built with actual generated HTML forms in mind, which would need human names.

src/richie/apps/search/defaults.py
mbenadda: I don't think I understand your proposal.

The `ChoiceField` key is on `"language"` whose value can be either `"en"` or `"fr"`.

As far as I can tell there is no point in a ChoiceField if I don't specify the actual choices it allows. That would just be a `CharField` that does not check the actual content of the values...


Also in all the cases we want it's a `ChoiceField` but it might as well be something else entirely?

sampaccoud: my point was that in the __init__ of your form you could construct the choices and avoid this duplicate. Something like:
```python
self.choices = [(f["name"], f["human_name"]) for f in FILTERS_HARDCODED_DEFAULT["availability"]]
```

mbenadda: yes, that's what I understood after reading it the 2nd time and what I actually did this morning. Thanks :)

src/richie/apps/search/forms.py
```diff
@@ -12,6 +13,17 @@ class CourseListForm(forms.Form):
     Validate the query string params in the course list request
     """
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        # Add in the fields for the custom filters from defaults/settings
+        for filter_key in FILTERS_HARDCODED:
+            self.fields[filter_key] = FILTERS_HARDCODED[filter_key]["field"](
+                choices=[
+                    (value, "") for value in FILTERS_HARDCODED[filter_key]["choices"]
```
sampaccoud: Granted human names are useless, but maybe make it (value, value)?

src/richie/apps/search/indexers/courses.py
```diff
@@ -146,59 +157,126 @@ def build_es_query(request, facets):
...
+                        "bool": {
+                            # Concatenate all the lists of active query filters with
+                            # the relevant choice filter
+                            "must": reduce(
+                                lambda all_filters, query_list: all_filters
+                                + query_list[1],
+                                filter(
+                                    lambda query, fk=filter_key: fk is not query[0],
+                                    queries,
+                                ),
```
sampaccoud: [u for u in v[1] if v[0] is not query[0] for v in queries] if it works...

src/richie/apps/search/indexers/courses.py
```diff
@@ -146,59 +157,126 @@ def build_es_query(request, facets):
                 start, end = value
                 queries = [
                     *queries,
-                    {
-                        "range": {
-                            param: {
-                                "gte": start.datetime if start else None,
-                                "lte": end.datetime if end else None,
+                    (
```
sampaccoud: Using dictionary or even better a nametuple will avoid calling query[0] and query[1] in the following.

src/richie/apps/search/indexers/courses.py
```diff
@@ -146,59 +157,126 @@ def build_es_query(request, facets):
...
+            for choice in FILTERS_HARDCODED[filter_key]["choices"]:
+                # Create an aggregation for each filter/choice pair
+                filters_facets["{:s}@{:s}".format(filter_key, choice)] = {
+                    "filter": {
+                        "bool": {
+                            # Concatenate all the lists of active query filters with
+                            # the relevant choice filter
+                            "must": reduce(
+                                lambda all_filters, query_list: all_filters
+                                + query_list[1],
```
sampaccoud: bad name query_list is not really a list it's just a tuple

src/richie/apps/search/indexers/courses.py
```diff
@@ -130,6 +137,7 @@ def build_es_query(request, facets):
         # we pass them to ES. See the docs en bool queries.
         # https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-bool-query.html
         queries = []
+        key_fragment_pair = namedtuple("key_fragment_pair", ["key", "fragment"])
         for param, value in params_form.cleaned_data.items():
```
sampaccoud: call it key to be coherent with your namedtuple

src/richie/apps/search/indexers/courses.py
```diff
@@ -130,6 +137,7 @@ def build_es_query(request, facets):
         # we pass them to ES. See the docs en bool queries.
         # https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-bool-query.html
         queries = []
+        key_fragment_pair = namedtuple("key_fragment_pair", ["key", "fragment"])
```
sampaccoud: In fact the namedtuple should be declared on top of the file so its instanciated at code evaluation once for all.

mbenadda: But won't that require it to be in all caps like an export variable? Or does that not matter?

sampaccoud: IMO no problem. Hope pylint agrees !

Issue #273, opened 2018-07-02
Add a plugin to define the publishing licenses of a course


sampaccoud:
## Feature Request

**Is your feature request related to a problem or unsupported use case? Please describe.**
We need a plugin to define a licenses for:

- the content of a course
- the content created by participants in a course

**Describe the solution you'd like**
To add a license to the course content, the person editing the page can add a "license" plugin and choose from 8 possible licenses in a select box:

- One of the 7 variants of Creative Commons as defined here https://en.wikipedia.org/wiki/Creative_Commons_license#Seven_regularly_used_licenses
- Reserved Rights

A free text input is available for the person to give details (e.g. version of the cc license, adaptated version, etc.).

The license is displayed on the site with the free text and a set of logo+name+description defined by us (in the code) for each of the possible licenses.

**Discovery, Documentation, Adoption, Migration Strategy**
This is what we have today on the site:
![image](https://user-images.githubusercontent.com/1427165/42174775-af0a9b2e-7e23-11e8-9ff7-c634df11d33c.png)

sveetch:
Functionally it's done with PR #324 that has been merged but it lacks of some unitests, factory and to be included in demo-site.

sveetch:
Issue will continue in PR #341 with missing factory and tests

PR #324, opened 2018-08-01
✨(courses) create new plugin to manage course licences

sveetch:
## Purpose

Currently course licences are managed through text plugin where user
have to write in licence content again for each course. This is
related to issue #273.

## Proposal

We should have a dedicated plugin to manage licence. Site administrator
will write some licences once and user will be able to just select the
right one from the licence plugin with an additional and optional free
text for some specific details.


sveetch:
Work in progress, currently it lacks of factory, tests and correct plugin template.

src/richie/apps/courses/models.py
```diff
@@ -224,6 +228,61 @@ def copy_relations(self, oldinstance, language):
...
+
+
+@python_2_unicode_compatible
+class LicencePluginModel(CMSPlugin):
+    """
+    Licence plugin model.
+    """
+
+    licence = models.ForeignKey(Licence)
+    content = models.TextField(
```
sampaccoud: description? (when I hear content I expect more than text)

src/richie/apps/courses/models.py
```diff
@@ -224,6 +228,61 @@ def copy_relations(self, oldinstance, language):
...
+
+    Instances of this models should only be created by administrators.
+    """
+
+    name = models.CharField(_("name"), max_length=200)
+    logo = FilerImageField(
+        related_name="licence",
+        verbose_name=_("logo"),
+    )
+    link_url = models.CharField(
```
sampaccoud: link and url are synonym. Proposals:
"resource_link"? 
or just "url"?

src/richie/apps/courses/models.py
```diff
@@ -224,6 +228,61 @@ def copy_relations(self, oldinstance, language):
...
+
+    def __str__(self):
+        """Human representation of a person title"""
+        return "{model}: {name}".format(
+            model=self._meta.verbose_name.title(),
+            name=self.name,
+        )
+
+
+@python_2_unicode_compatible
```
rmoch: I think we do not expect this project to be python2 compatible.
There are other decorator alike in plugins

src/richie/apps/courses/models.py
```diff
@@ -224,6 +228,61 @@ def copy_relations(self, oldinstance, language):
...
+class Licence(models.Model):
+    """
+    Licence model.
+
+    Instances of this models should only be created by administrators.
+    """
+
+    name = models.CharField(_("name"), max_length=200)
+    logo = FilerImageField(
+        related_name="licence",
```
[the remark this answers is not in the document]
sveetch: You mean for the `name` content because you said pastly than this does not need to be translatable when i was starting it with Parler, no ?

src/richie/apps/courses/models.py
sveetch: I started with using ckeditor widget in mind but forgot to do so. It's should also be effective for the Licence.description field

src/richie/apps/courses/models.py
sveetch: Just "url" is perfect.

src/richie/apps/courses/models.py
sveetch: Nice catch, self remember to remove this from other plugins i did

src/richie/apps/courses/models.py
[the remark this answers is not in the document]
sampaccoud: ah yes sorry I meant for the verbose name of the field. But I didn't see that this is a related_name. The verbose name is actually missing :-)

src/richie/apps/courses/templates/courses/plugins/licence_plugin.html
```diff
@@ -0,0 +1,18 @@
+{% load i18n %}{% spaceless %}
+<div class="licence-plugin">
+  <h{{ header_level|default:2 }} class="licence-plugin__title">
+    {% if instance.licence.url %}<a href="{{ instance.licence.url }}">{% endif %}
+    {{ instance.licence.name }}
+    {% if instance.licence.url %}</a>{% endif %}
+  </h{{ header_level|default:2 }}>
+  <p class="licence-plugin__logo">
+    <img src="{{ instance.licence.logo.url }}" alt="">
```
sampaccoud: use easythumbnail to make sure we always have the same format of image?

src/richie/apps/courses/models.py
```diff
@@ -219,6 +220,28 @@ def copy_relations(self, oldinstance, language):
...
+    """
+    Licence model.
+
+    Instances of this models should only be created by administrators.
+    """
+
+    name = models.CharField(_("name"), max_length=200)
+    logo = FilerImageField(verbose_name=_("logo"), related_name="licence")
+    url = models.CharField(_("url"), blank=True, max_length=255)
+    content = models.TextField(_("Content"), blank=False, default="")
```
sampaccoud: s/Content/content

src/richie/apps/courses/forms.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+class LicenceFormAdmin(forms.ModelForm):
+    """
+    Licence model form used within Django admin
+    """
+
+    class Meta:
+        """
+        Form meta attributes
```
sampaccoud: same

src/richie/apps/courses/forms.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+class LicencePluginForm(forms.ModelForm):
+    """
+    Licence plugin form used to fill its content from frontend admin.
+    """
+
+    class Meta:
+        """
+        Form meta attributes
```
sampaccoud: is this necessary? we didn't put it anywhere else.

Issue #329, opened 2018-08-02
Add courseRun


2018D:
## Feature Request

**Is your feature request related to a problem or unsupported use case? Please describe.**
We need to add the courseRun to complete course session information(startDate, endDate, subscriptionDate,etc.) 

**Describe the solution you'd like**
Add a page model that we can attach by plugin to course page.

**Describe alternatives you've considered**
Have the courseRun created as plugin of course. What do you think @sveetch ?

**Discovery, Documentation, Adoption, Migration Strategy**
[schemaRichie.pdf](https://github.com/openfun/richie/files/2255014/schemaRichie.pdf)

**Do you want to work on it through a Pull Request?**
Proof of concept within the next 2 weeks.

2018D:
I had lengthy talks with Robert and we believe that having the courseRun only exist within the course is a mistake. One that openEdx has done. It now wants to seperate the runs from the courses according to the following doc: https://github.com/edx/open-edx-proposals/blob/mroytman/oep-0027/oeps/oep-0027-course-metadata-authoring.rst
We should avoid the initial mistake and separate our runs from the course right now. 

Associating the course run to a course in the same way a course is associated to it's main organization would allow us to integrate it to the course page and thus control which one are shown. The added benefit being that each and all course run page can be consulted by the pedagogical engineer's. Whereas having the complete history only in the course admin would not allow that role to consult them since it is likely we won't allow acces to the admin interface.

I'll rework Samuel's prototype on another branch.

2018D:
I'm proposing #338 as base for this.

PR #330, opened 2018-08-03
Add course runs

sampaccoud:
## Purpose

This is a POC for adding course runs.

The Ux can be greatly improved by doing a bit of customization and CSS on the admin template.
An alternative could be to declare a plugin...

## Proposal

- [x] Add a CourseRun model with a foreign key to the course
- [x] Add the course run as an inline on the course admin view
- [x] Remove the active_session key on the course


2018D:
I made a comment on issue #329

2018D:
the active_session key is not required but we do still need a course_key. We could rename the value or leave as is for now.

2018D:
did a rebase to prep the terrain

2018D:
How can we retrieve the course_run info after filling the inline form to show it on the html? I haven't been able to figure it out.

2018D:
work to be done:
[ ] fix publishing course does not copy or show draft course-runs on published course-detail
[ ] factories
[ ] unit testing: course admin and course-detail page to include course-run

Should we consider using the licence structure for the course-run?

sampaccoud:
@2018D I have fix the issues you raised. 
I propose to skip testing the course run custom admin widget/field as it is quite painful with little value and I don't have time now. If you think it's important, please do it or add an issue for someone to do it later?

src/richie/apps/courses/admin.py
```diff
@@ -2,17 +2,40 @@
...
 from cms.extensions import PageExtensionAdmin
 
+from .fields import CourseRunSplitDateTimeField
 from .forms import LicenceFormAdmin
-from .models import Course, Licence, Organization
+from .models import Course, CourseRun, Licence, Organization
+from .widgets import CourseRunSplitDateTimeWidget
+
+
+class CourseRunInline(admin.StackedInline):
```
sveetch: You may try `admin.TabularInline` it is commonly more clean and accessible

sampaccoud: Yes that's what I used at first but it requires some adaptation to fit in a line so as a first approach I switched to stacked inlines...

image

sveetch: Ah yeah, table makes troubleshooting with non text element like the date widtget.

2018D:
C'est bon pour moi 👍

sampaccoud:
Fixes https://github.com/openfun/richie/issues/329

PR #331, opened 2018-08-03
♻️(person) use new fill_* pattern on person factory and improve i18n

sampaccoud:
## Purpose

Update the person factory to use the same fill_* pattern as the other factories and improve its i18n support.

## Proposal

- [x] replace the "with_content" hook by "fill_portrait" and "fill_resume" hooks
- [x] loop on all languages when creating the plugins in the fill_* hooks
- [x] fix the tests and use the new scheme to build the demo site
- [x] manually check that the demo site now has multilingual persons


src/richie/apps/persons/factories.py
```diff
@@ -58,46 +59,35 @@ def title(self):
...
-            body = ["<p>{:s}</p>".format(p) for p in paragraphs]
-            add_plugin(
-                language=language,
-                placeholder=resume_placeholder,
-                plugin_type="CKEditorPlugin",
-                body="".join(body),
+
+    @factory.post_generation
+    # pylint: disable=unused-argument
+    def fill_resume(self, create, extracted, **kwargs):
```
sveetch: Isn't a good opportunity to put fill_texts() method from https://github.com/openfun/richie/blob/master/src/richie/apps/courses/factories.py#L210 in common Factory mixin so we can have the same logic to create text plugin share everywhere ?

sampaccoud: @sveetch that's enticing but maybe not in this PR? For the person bio, I have set "nb_paragraphs=1" so we need to rethink how to allow that if we factorize the code.

PR #335, opened 2018-08-09
Improve naming in tests

sampaccoud:
## Purpose

While coding, I often run pytest with the -k option on variable fractions of the name of a test method in order to run all tests about the same object or a specific subject. For example:

- pytest -k models_organization
- pytest -k factories_subject
- pytest -k templates_organization_detail
- ...

If we want this to work well, the names of our tests methods  must be **carefully crafted from the most generic to the most specific**. We have slowly derived from this best practice and it was becoming hazardous to target groups of tests in this manner.

In the process of changing the names of the methods, I also noticed that some test file names did not follow the recommandations in our handbook (https://openfun.gitbooks.io/handbook/content/python.html).

Let's take advantage of everyone's holiday to propose this refactoring.

## Proposal

- [x] improve tests file names
- [x] improve tests class names
- [x] improve tests method names


mbenadda:
You renamed indexers and courses tests by de-pluralizing resource names.

However, we agreed in the past to always pluralize these resource names to avoid having to deal with the overhead of singular/plural on them given they are used everywhere. Which is why those names were pluralized.

Do you think this was misguided? If we change the practice, at the very least we should also rename the indexers and viewsets themselves so their names are consistent with their tests (and the class names inside to match the class names?)

WDYT?

sampaccoud:
@mbenadda I had doubts on this but yes we can pluralize class names in the names of tests files and methods.

mbenadda:
So sorry, I missed your comment yesterday.

Your proposal makes sense, let's 🚢 it

PR #338, opened 2018-08-16
✨ (Courses) add course-run

2018D:
A course in the catalog can happen several times at different
dates. In the LMS, each MOOC corresponds to a course run so we
added a "resource_link" field in the course run to materialize
the relation.

## Purpose

Add course-run through a page instead of as a add-in to the course-page.

## Proposal

This is a prototype. The work is still in progress for #330 

- [x] add course-run model and page setup
- [ ] add factory
- [ ] add test


src/richie/apps/courses/models.py
```diff
@@ -112,6 +112,11 @@ class Course(BasePageExtension):
         related_name="main_courses",
         limit_choices_to={"extended_object__publisher_is_draft": True},
     )
+    course_run = models.ManyToManyField(
+        "CourseRun",
+        related_name="course_run",
```
rmoch: The related_name is the name of the relation from Course to CourseRun objects as seen from CourseRun object. So as a logical convention, it should be called `courses`, then when handling a CourseRun, you will access its Courses this way: course_run.courses (note that other relations like `organizations` use this convention)
Also as a M2M, this name should be plural: course_runs (again see `organisations` 😉  )

src/richie/apps/courses/models.py
```diff
@@ -190,6 +195,52 @@ def save(self, *args, **kwargs):
...
 
+class CourseRun(BasePageExtension):
+    """
+    The course run represents and records the occurence of a course between a start
+    and an end date.
+    """
+
+    course = models.ForeignKey(
+        "Course",
+        related_name="courserun",
```
rmoch: same convention, and snakecase, related_name should ne called `course_runs`

Also, not sure this relation is necessary, shall we not do just like for organizations: Course.course_runs and Course.main_course_run ?
If not, I still think this relational schema is probably wrong 🤔  WDYT ?

sandbox/settings.py
```diff
@@ -288,6 +288,7 @@ class Base(DRFMixin, ElasticSearchMixin, Configuration):
 
     CMS_TEMPLATES = (
         ("courses/cms/course_detail.html", _("Course page")),
+        ("courses/cms/course_run.html", _("Course run page")),
```
rmoch: Not sure how a course run page is usefull, but it should be called course_run_detail.html, 
(or did you mean course_runs.html ?)

src/richie/apps/courses/models.py
2018D: I'll make the required changes. Thanks :)

sandbox/settings.py
2018D: I think i'll go with course_run_detail.html

src/richie/apps/courses/models.py
2018D: I am aiming for the user to choose within the course only among the course runs that were created for it. Is it possible?

I like your idea of having a course.main_course_run as well as my existing course.course_runs. It would make the choice of course run to show on the course card much easier later on.

sampaccoud:
Closed in favor of https://github.com/openfun/richie/pull/330

PR #344, opened 2018-08-30
💄(layout) do basic integration for sandbox

sveetch:
## Purpose

Sandbox lacks of clean layout styles. It's not fully responsive,
accessible and does not have solid basic design identity.

## Proposal

This commit adds better header, footer and some minor content
adjustment. Everything done is responsive and global layout stand
on current 'fun-mooc.fr' design. Most of work lives in Sass sources
and base template. Some vanilla Javascript code has been added
for mobile menu toggle (with click on hamburger). Also some basic
layout has been added to person detail template to follow other
detail templates.

## TODO
- [x] Use variable to drive behavior for header and footer;
- [x] Put Header topbar Javascript to the right place;
- [x] Wait until demo-site command has been modified to adapt to new cms page tree for the right menus (one for header and another one for footer).


sandbox/templates/richie/base.html
```diff
@@ -1,8 +1,8 @@
-{% load cms_tags menu_tags sekizai_tags staticfiles %}
+{% load i18n cms_tags menu_tags static sekizai_tags staticfiles %}
 <!doctype html>
 <html>
     <head>
-        <title>{% block title %}This is my new project home page{% endblock title %}</title>
+        <title>{% block title %}Richie demonstration{% endblock title %}</title>
```
sampaccoud: add i18n?

sandbox/templates/richie/base.html
```diff
@@ -12,12 +12,95 @@
...
-        <div>
-            {% show_menu 0 100 100 100 %}
-            {% block content %}
+        {% block body_header %}
+        <nav class="topbar" id="main-menu">
+            <div class="topbar__brand">
+                <button class="topbar__hamburger"
+                        data-target="main-menu" aria-label="menu" aria-expanded="false">&#8801;</button>
+                <a href="/" class="topbar__brand__link">
+                    <img src="{% static "images/logo.png" %}" class="topbar__brand__logo" alt="">
```
sampaccoud: missing alt?

src/richie-front/scss/_colors.scss
```diff
@@ -0,0 +1,22 @@
...
+$gray97: #f7f7f7;
+$coral2: #f06d54;
+$cyan4: #009fa1;
+$deeppink4: #9c005d;
+$dodgerblue1: #08223c;
+$dodgerblue3: #335280;
+$dodgerblue5: #007aa6;
+$dodgerblue7: #168dc3;
+$firebrick3: #c7162b;
+$firebrick6: #f72c30;
```
sampaccoud: If these are color variables then maybe call them by a name that does not relate to colors? Maybe relate to role of the color? first/second? primary? main? ...

src/richie-front/scss/mixins/_generic.scss
```diff
@@ -0,0 +1,14 @@
+///
+/// Default base document font size
+/// Value stand on Bootstrap assertion of 16px default value from browsers
+/// But this is not a threat since rem will finally adapt relatively to any
+/// default value. However for sanity we use a variable.
+$global-base-font-size: 16px !default;
```
mbenadda: Why not just use `$font-size-base` from bootstrap?

src/richie-front/scss/_colors.scss
mbenadda: I'd even say: reuse the names from bootstrap & override bootstrap's defaults?

sandbox/templates/richie/base.html
```diff
@@ -12,12 +12,95 @@
...
+                </a>
+            </div>
+        </div>
+        {% endblock body_footer %}
+
         {% render_block "js" %}
+        <script type="text/javascript">
+        document.addEventListener('DOMContentLoaded', () => {
+            // Get all topbar burger elements
+            const $topbarBurgers = Array.prototype.slice.call(document.querySelectorAll('.topbar__hamburger'), 0);
```
mbenadda: What's the `$` prefix' meaning?

mbenadda:
I think we should run `prettier` formatting on `scss` files.

Right now it does not run as a linter on them (maybe it should) but that does not prevent us from running it on the file ourselves

src/richie-front/scss/mixins/_generic.scss
sveetch: For this function helper, you need a pixel value to computate the rem one and `$font-size-base` setting from bootstrap use `1rem`.

sandbox/templates/richie/base.html
sveetch: In fact this title should come from `django.contrib.sites` since site title should not be translatable, this is the site name not a punchline.

In my django project i have a template context processor that allways include in every template context for convenience but with richie there is nothing like this.

sandbox/templates/richie/base.html
sveetch: my bad, commonly it's me that shouts on webdesigners which forget it..

sandbox/templates/richie/base.html
sampaccoud: IMO site title should be translatable. It is an important info for SEO and should be in the current language! For other pages I suppose the block is overriden to display the page title in the current language.

sandbox/templates/richie/base.html
sveetch: It's a common practice when your object is a DOM element to quickly see it.

sveetch:
As promised to @2018D , some screenshots to show sample layout.

### Mobile
mobile

### Tablet 
tablet

### Desktop
xlarge

PR #345, opened 2018-09-03
WIP: ✅(demo) improve demo site page tree

sveetch:
## Purpose

Page tree is just a flat list of pages, all enabled in navigation.
But we don't want every pages to be in menus and need a more advanced
architecture to be able to create an additional menu for annexe page
like About, Help, Legal mentions, etc..

## Proposal

This commit hides some pages from menus and add a new page "Annexe"
where "About" page is moved in.


src/richie/apps/core/management/commands/create_demo_site.py
```diff
@@ -26,6 +26,7 @@
 
 logger = logging.getLogger("richie.commands.core.create_demo_site")
 
+DEMO_ANNEXE_PAGE_ID = "annexe"
```
sampaccoud: make it "annex" in english so:

ANNEX_PAGE_ID = "annex"

src/richie/apps/core/management/commands/create_demo_site.py
```diff
@@ -111,28 +130,59 @@ def clear_cms_data():
     Person.objects.all().delete()
 
 
-def create_demo_site():
+def recursive_page_creation(site, pages, parent=None):
```
sampaccoud: I think it is becoming too complicated for just one child page, and you have created empty pages so the final solution where you add content to each page will not look like this... 

For the moment, maybe just add this to the create_demo_site method:
```
languages=[l[0] for l in settings.LANGUAGES]
about_page = create_i18n_page(
    languages=languages,
    parent=pages_created["annex"],
    published=True,
)
create_text_plugin(
    about_page,
    "main_content",
    languages=languages,
)

File: src/richie/apps/core/factories.py [hunk 3]
```diff
@@ -79,6 +79,16 @@ def _after_postgeneration(cls, instance, create, results=None):
             for language in instance.extended_object.get_languages():
                 instance.extended_object.publish(language)
 
+    @factory.post_generation
+    # pylint: disable=no-self-use,unused-argument
+    def in_navigation(self, create, extracted, **kwargs):
```
sampaccoud: You can just modify the existing method "extended_object" like so:
```
@factory.lazy_attribute
def extended_object(self):
    return create_i18n_page(
        title=self.title,
        languages=self.languages,
        template=self.template,
        in_navigation=self.in_navigation,
        parent=self.parent,
    )
```

PR #348, opened 2018-09-06
Fix creating new course after related objects

sampaccoud:
## Purpose

The many-to-many relations between courses and their organizations/subjects were always pointing to the draft instance. This was causing a issue on the following edge case:

- create a subject and publish it,
- add a course and link the subject to it,
- publish the course.

Now the published course should be on the published subject page but it's absent :scream: 

## Proposal

Solving this edge case with the current relationships was convoluted because it required adding the missing link between the published subject and the draft course upon publishing the course. We ended up updating the relationships for both the draft and the published course. This was not necessarily a performance issue because the number of organizations and subjects for a course is expected to be small but this was hard to understand and lead to :hankey: code.

We decided to refactor the way relationships are made.

A draft course is still linked to its draft organizations and draft subjects. But when we publish the course, the published course is now linked to the published organizations corresponding to the draft organizations that the draft course had (same for subjects). 

When publishing an organization or a subject, we do not copy relations anymore. This is ok because:

1) a course does not normally come before its subjects or organizations
2) if a subject or organization is added later, it seems good that we have to re-publish the course before it is added to the published version.

:information_source: I started by splitting the models in several files as the models.py file was getting far too long!


mbenadda:
> When publishing an organization or a subject, we do not copy relations anymore. This is ok because:
> 
> 1. a course does not normally come before its subjects or organizations

I understand the rationale but this is worse than the existing implementation for users. It would need to be explained somewhere prominent in the docs too.

---

Imagine you created your subjects/orgs, then a bunch of courses. You publish all the subjects/orgs, then the courses.

Then you realize you did not publish one of the orgs. You have to go over all courses — because you might not remember exactly which ones are linked to what — to republish them one by one.

Hopefully nobody else had worked on any of those drafts in the meantime...

---

If you add eg. a new org with a logical publishing workflow it's also painful:

- make the new org draft
- make all the course drafts
- publish the org (empty)
- publish each of the org's courses

--- 

The one way this might make sense is if we don't allow users to add yet to be published orgs/subjects to courses. This way they'll notice in time they forgot to publish something.

The only issue is we'd have somewhat inconsistent state where you must publish an org/subject page with 0 courses and then have them appear there one by one. Not too big a deal if you add them all at once.

sampaccoud:
@mehdi actually what you need is that the public page exists, not that it is published.

The public page is created the first time the page is published and you can then depublish it.

src/richie/apps/courses/templates/courses/cms/organization_detail.html
```diff
@@ -26,7 +26,7 @@ <h1 class="organization-detail__title">{{ current_page.get_title }}</h1>
         </div>
 
         <ul class="organization-detail__content__courses">
-          {% for course in organization.courses.drafts %}
+          {% for course in organization.courses.all %}
```
2018D: Substituting "all" for "drafts" in the templates, list all courses including the one still in draft and the unpublished as well. The title won't appear because we aren't referencing the right "get.title" but they are in the list nonetheless.

src/richie/apps/courses/admin.py
```diff
@@ -44,18 +44,6 @@ def title(self, obj):
...
-        """
-        When a model is saved via the admin, the main object gets saved first (to make sure it has
-        a pk), then the many-to-many are cleared and the new values set to whatever came out of
-        the form (see http://stackoverflow.com/a/1925784/469575).
-
-        This means that we can not rely on our override of the `save` method on the model to modify
-        the `organizations` many-to-many field and make sure it includes the main organization.
-        """
-        super().save_related(request, form, formsets, change)
-        form.instance.organizations.add(form.instance.organization_main)
```
2018D: Removing this make it so that choosing a organization_main for a new course does not also add this main org to attached organizations.

sampaccoud: Yeah I removed it at some point but we should be able to keep this feature with the final solution I found. Let me revert this together with @mehdi suggestion so you can it.

PR #354, opened 2018-09-12
Add package.json/makefile commands to run prettier

mbenadda:
## Purpose

Right now we don't have easy-to-use commands to run prettier in write mode on the codebase (the CI just runs it in list-different mode).

## Proposal

Add a "yarn prettier" command that can be run directly by end-users with local installs of yarn & dependencies, and a corresponding lint-front-prettier in Makefile to enable users to run the same
from the container.


mbenadda:
@sampaccoud done

PR #355, opened 2018-09-14
💄(courses) upgrade organization_plugin

2018D:
## Purpose

The organization_plugin allows the creation of a custom organizations page by choosing which one appears and where.

## Proposal

The selector for organization_plugin showed all organizations.
Filter it to show only one version.
Renamed plugin template to indicate template.
Add style file to customize appearance.


src/richie/apps/courses/models/organization.py
```diff
@@ -98,9 +98,12 @@ class OrganizationPluginModel(CMSPlugin):
     Organization plugin model handles the relation from OrganizationPlugin
     to their Organization instance
     """
-
-    organization = models.ForeignKey(Organization)
-
+    organization = models.ForeignKey(
+            "Organization",
+            related_name="organization",
+            limit_choices_to={"extended_object__publisher_is_draft": True},
```
sampaccoud: @sveetch in the description of your PR you wrote "only publised" and her you are filtering "only draft". The description is wrong and this is good ;-)

src/richie/apps/courses/models/organization.py
```diff
@@ -98,9 +98,12 @@ class OrganizationPluginModel(CMSPlugin):
     Organization plugin model handles the relation from OrganizationPlugin
     to their Organization instance
     """
-
-    organization = models.ForeignKey(Organization)
-
+    organization = models.ForeignKey(
+            "Organization",
+            related_name="organization",
```
sampaccoud: I think this is wrong, it should be something like "plugin_model"?

src/richie/apps/courses/models/organization.py
2018D: Well, that was a mistake. I wanted to make sure i showed only one version.
Should it be "extended_object__publisher_is_draft": False
if i want to show published organization only ? Which one would be better ?

sampaccoud: No you should leave it like this. We only edit draft versions. The public version is set when you publish and DjangoCMS copies the draft to the published version.

2018D: allright 👍 Thanks a lot.

src/richie/apps/courses/models/organization.py
2018D: I've edited the related name according to:
http://docs.django-cms.org/en/latest/how_to/custom_plugins.html#for-foreign-key-relations-from-other-objects

src/richie/apps/courses/models/organization.py
```diff
@@ -98,9 +98,12 @@ class OrganizationPluginModel(CMSPlugin):
     Organization plugin model handles the relation from OrganizationPlugin
     to their Organization instance
     """
-
-    organization = models.ForeignKey(Organization)
-
+    organization = models.ForeignKey(
+            "Organization",
+            related_name="organization_plugin",
```
sampaccoud: Much better but you don't need to include "organization" in the name because this is a related name so it is going to be called from an organization object. You will have to write: "organization.organization_plugin" which is redundant.

You will agree that "organization.plugin_model" is much better and makes more sense to the reader.

sampaccoud:
@2018D you can add the migration but it will not do anything on the database. It's just that Django will find a difference in your model but it is just an application difference and the database table is the same.

2018D:
I saw what the migration was this morning after making it. You are right, it doesn't change much. The tests for it are however failing locally so i'll fix those before rebasing.

tests/apps/courses/test_cms_plugins_organization.py
```diff
@@ -50,18 +50,15 @@ def test_cms_plugins_organization_render(self):
             fill_banner=True,
             fill_logo=True,
             fill_description=True,
+            should_publish=True,
```
sampaccoud: then should we check the result when the organization is not published as well? on a draft page? on a published page? to cover all the case sof your template.

2018D: Good point... I'll see what i can do today.

2018D:
This is ready for review.

src/richie/apps/courses/templates/courses/plugins/organization_plugin.html
```diff
@@ -0,0 +1,52 @@
+{% load cms_tags i18n %}
+{% spaceless %}
+{% if current_page.publisher_is_draft %}
+  {% if instance.organization.check_publication is True %}
+    <div class="organization-plugin-container">
+      <div class="organization-plugin">
+        <a class="organization-plugin__body" href="{{ instance.organization.public_extension.extended_object.get_absolute_url }}"
```
2018D: I need to check the presence of this tag in 3 parts. Otherwise the test fails. I'm not sure why.

tests/apps/courses/test_cms_plugins_organization.py
2018D: It's done but i need to use 3 assetContains to check the complete presence of this tag:
<a class="organization-plugin__body" href="{{ instance.organization.public_extension.extended_object.get_absolute_url }}"
          title="{{ instance.organization.public_extension.extended_object.get_title }}"
          alt="{{ instance.organization.public_extension.extended_object.get_title }} {% trans 'logo' %}">

This doesn't seem good... Is it because it's on multiple lines?

2018D:
It doesn't work even if i put the html on a single line. Any ideas as to why it fails when i search for the complete line?

2018D:
Found my error, i had the tag on multiple lines in the html template. The test failed because of that. It's fixed now. The html lines are somewhat very long. Any problem there?

2018D:
will refactor this into a larger PR to show organization list in course detail

PR #356, opened 2018-09-17
💄(layout) improve course detail integration

sveetch:
## Purpose

So far the course detail render is very basic and course lacks of a
preview image in glimpse.

## Proposal

This commit add a new placeholder 'course_cover' to add a preview
image to use in glimpse instead of video teaser that does not fit
well in card format. Course factory, detail template and demo-site
command have been upgraded to include it.

Then we upgrade the detail template and CSS to fit to fun-mooc.fr
course detail page but with according to our content structure,
especially the course runs that are not managed in same way.


sampaccoud:
@sveetch I proposed a solution so that the cover images are in sync with the video teaser while still being picked randomly.

PR #361, opened 2018-09-21
Diversify course runs on the demo site

sampaccoud:
## Purpose

On the demo site, all course runs were created in the future. We want to have course runs in different states: some open, and some future, past, open or closed for enrollment, etc.

## Proposal

I chose the option to loosen the factory and had to fix a few tests as a consequence.

### DIFF
--- a/src/richie/apps/core/management/commands/create_demo_site.py
+++ b/src/richie/apps/core/management/commands/create_demo_site.py
@@ -3,11 +3,12 @@ create_demo_site management command
 """
 import logging
 import random
+from datetime import timedelta
 
 from django.conf import settings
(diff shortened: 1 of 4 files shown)


src/richie/apps/core/management/commands/create_demo_site.py
```diff
@@ -108,7 +109,7 @@ def get_number_of_course_runs():
     Returns a random integer between 0 and 5. We make it a convenience method so that it can
```
sveetch: now it's `between 1 and 5`

sampaccoud: Well spotted! Fixed.

PR #366, opened 2018-09-27
Migrate tests to Jest

mbenadda:
## Purpose

Per #339 we can't keep our existing technique to mock imports in Karma + Jasmine.

Our preferred solution was to move to Jest as it provides builtin mocking facilities and all around better tooling and a better developer experience for our use cases.

## Proposal

Update config files and some details in tests to run them with `jest` instead of `karma`/`jasmine`.


package.json
```diff
@@ -9,7 +9,6 @@
     "lint": "tslint -c tslint.json 'src/richie-front/js/**/*.ts?(x)'",
     "prettier": "prettier --write 'src/richie-front/js/**/*.+(ts|tsx|json|js|jsx)' '*.+(ts|tsx|json|js|jsx)'",
     "sass": "node-sass src/richie-front/scss/_main.scss src/richie/static/richie/css/main.css",
-    "test": "karma start ./karma.conf.js",
```
jmaupetit: Can we keep this test script `"test": "jest",`? And thus, use it in the `Makefile` and the CI as it was before this switch?

package.json
```diff
@@ -83,5 +78,8 @@
     "react-redux": "5.0.7",
     "redux": "4.0.0",
     "redux-saga": "0.16.0"
+  },
+  "resolutions": {
+    "babel-core": "7.0.0-bridge.0"
```
jmaupetit: This "resolution" field is yarn-specific, right? Is there a conflict that requires to force this `babel-core` release?

jest.config.js
```diff
@@ -0,0 +1,19 @@
+module.exports = {
```
jmaupetit: `s/precompiler by the library author/precompiled by the library author/`? in the commit message

src/richie-front/js/components/CourseGlimpse/CourseGlimpse.spec.tsx
```diff
@@ -1,4 +1,4 @@
-import '../../testSetup.spec';
```
jmaupetit: `s/a utility/an utility/`? in the commit message

package.json
mbenadda: I can see how it can be nice to keep the standard command throughout. Done.

src/richie-front/js/components/CourseGlimpse/CourseGlimpse.spec.tsx
mbenadda: I think it's `a utility` as `utility` starts with a consonant sound

package.json
mbenadda: It is yarn specific. One dependency (`babel-jest` AFAIK) loads babel 6 in our specific circumstances and therefore requires us to use this resolution + bridge to force babel 7.

I would have put in a comment but unfortunately JSON can't handle comments :(

mbenadda: I'll add something in a commit so the info can at least be found through a `git blame`

PR #368, opened 2018-10-01
🐛(licence) fill Licence with a sample image from fixtures

sveetch:
## Purpose

When Licence has been created, their logo image was left to the
'FilerImageFactory' which generate a simple plain color and dummy
image.

## Proposal

Licence factory has been changed to use a new Subfactory for image
file to randomly select a licence logo from fixtures.

### DIFF
--- a/src/richie/apps/courses/factories.py
+++ b/src/richie/apps/courses/factories.py
@@ -375,6 +375,25 @@ class SubjectFactory(BLDPageExtensionDjangoModelFactory):
             self.courses.set(extracted)
 
 
+class LicenceLogoImageFactory(FilerImageFactory):
+    """
+    Image field factory for Licence.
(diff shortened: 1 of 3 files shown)


src/richie/apps/courses/factories.py
```diff
@@ -375,6 +375,29 @@ def with_courses(self, create, extracted, **kwargs):
             self.courses.set(extracted)
 
 
+class LicenceLogoImageFactory(factory.django.DjangoModelFactory):
```
sampaccoud: I propose that you inherit from FilerImageFactory and just override the file method...

sveetch: @sampaccoud i've got a hardtime to do it but it's ok now

PR #373, opened 2018-10-02
💄(layout) style course subject to look as badges

sveetch:
## Purpose

Actually subjects in Course detail is just a simple list of subject
item.

## Proposal

This commit change course detail template and styles so the subject
now look like tiny badges under Course title. Tests has been update
to match against template changes.

course-detail

### DIFF
--- a/src/richie/apps/courses/templates/courses/cms/course_detail.html
+++ b/src/richie/apps/courses/templates/courses/cms/course_detail.html
@@ -13,6 +13,32 @@
         <h1 class="course-detail__title">{{ current_page.get_title }}</h1>
       </div>
 
+      <div class="course-detail__content__row course-detail__content__subjects">
+        {% for subject in course.subjects.all %}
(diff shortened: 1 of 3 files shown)


src/richie/apps/courses/templates/courses/cms/course_detail.html
```diff
@@ -13,6 +13,55 @@
...
+            <a class="course-detail__content__subjects__item" href="{{ subject.extended_object.get_absolute_url }}">
+              {{ subject.extended_object.get_title }}
+            </a>
+          {% endif %}
+        {% empty %}
+          <p class="course-detail__content__subjects__item course-detail__content__subjects__item--empty">
+            {% trans "No associated subjects" %}
+          </p>
+        {% endfor %}
+        {% for subject in course.subjects.all %}
```
sampaccoud: I think you pasted the code twice!? This is exactly the same code

PR #374, opened 2018-10-03
WIP:💄(course) move relevant course detail styles values to settings

sveetch:
## Purpose

Actually Course detail styles can not be piloted from settings.

## Proposal

This commit change Course detail stylesheet to use variable settings
for every relevant values (font size, margin, padding, colors,
etc..).


sveetch:
It misses some styles yet to be finished.

Issue #376, opened 2018-10-03
Improve filtering in the filters pane on the Search view


mbenadda:
## Feature Request

**Is your feature request related to a problem or unsupported use case? Please describe.**
Right now the search filters pane is pretty crude. The basic functionality is there (for organizations and subjects) but we're missing key parts and the experience is not great.

**Describe the solution you'd like**
- Implement the filters for other facets
  The filters are not implemented nor the facets displayed for other dimensions than organizations & subjects. They are available on the backend and just need to be plugged in here.

- Improve the readability of active filters
  Active filters are displayed at the top of their category and are made bolder. This is not enough to distinguish them. We need to add eg. a ticked box in the place where their facet count would be to make it obvious they are enabled.
  We can additionally list some active filters with an icon to remove them directly in the search box. This can be somewhat more complex (we need to scale it when users add a lot of filters) and is therefore a lower priority.

- Support more than 10 values for the organizations & subjects categories.
  The existing implementation is just picking up whatever the backend is returning as top facets for this category. This is fine as a default view, but we need to be able to filter by other values straight from the filters pane.
  We can do this by adding a small autosuggest field below the list of top filter values. Ideally it would display facet counts along with the suggestions. As a first version, we can use the generic endpoints and not display any counts.

**Describe alternatives you've considered**
Any other suggestions for improving the experience and exhaustiveness of our search filters is welcome.

**Discovery, Documentation, Adoption, Migration Strategy**
All of the suggestions above add visual elements and should be easy to discover this way. Of note: existing implementations might not be completely accessible: any refactors should look to improve on this. We need our search to be as accessible as possible.

**Do you want to work on it through a Pull Request?**
Yes. Also meant as proposals for @lunika to pick from.

- [x] Filters for Language, Availability & New courses are implemented
- [x] Active filters are more distinguishable
- [x] Non-top-10 organizations & subjects can be added to filters through a search box
- [x] They show facet counts
- [ ] Search box shows a (redundant) list of active filters

lunika:
> Implement the filters for other facets
>The filters are not implemented nor the facets displayed for other dimensions than organizations & subjects. They are available on the backend and just need to be plugged in here.

I would like to work on this point. Where can I found the api resources related to current harcoded filters ? I saw there is a documentation for the [search endpoints](https://github.com/openfun/richie/blob/master/src/richie/apps/search/description.apib) and I see nothing related to `availatbility`, `language` or `new` resources. 

[...]

PR #380, opened 2018-10-04
💄 (front) Active filters are more distinguishable

lunika:
When a filter is active, emphasizes it by adding a close button. Doing
this you know that this filter is active and you can only remove it.

Part of issue #376

A screenshot of what it looks like : 

screenshot from 2018-10-04 15-49-19

### DIFF
--- a/src/richie-front/js/components/SearchFilter/SearchFilter.tsx
+++ b/src/richie-front/js/components/SearchFilter/SearchFilter.tsx
@@ -21,6 +21,7 @@ export const SearchFilter = (props: SearchFilterProps) => {
           ? removeFilter(filter.primaryKey)
           : addFilter(filter.primaryKey)
       }
+      aria-label={isActive ? 'Close' : ''}
     >
       {typeof filter.humanName === 'string' ? (
         <span>{filter.humanName}</span>
@@ -29,6 +30,13 @@ export const SearchFilter = (props: SearchFilterProps) => {
       )}
       {filter.count || filter.count === 0 ? (
         <span className="search-filter__count">{filter.count}</span>
+      ) : (
+        ''
+      )}
+      {isActive ? (
+        <span className="search-filter__close" aria-hidden="true">
+          &times;
+        </span>
       ) : (
         ''
       )}
--- a/src/richie-front/js/components/SearchFilter/_SearchFilter.scss
+++ b/src/richie-front/js/components/SearchFilter/_SearchFilter.scss
@@ -19,6 +19,7 @@ $richie-search-filters-active-fontweight: null !default;
 $richie-search-filters-active-background: null !default;
 $richie-search-filters-active-divider-border: null !default;
 
+$richie-search-filters-icon-font-size: $font-size-base * 2 !default;
 
 .search-filter {
   @include m-o-list-group__item(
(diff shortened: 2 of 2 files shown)


src/richie-front/js/components/SearchFilter/SearchFilter.tsx
```diff
@@ -15,12 +15,13 @@ export const SearchFilter = (props: SearchFilterProps) => {
 
   return (
     <button
-      className={`search-filter ${isActive ? 'active' : ''}`}
+      className={`search-filter ${isActive ? 'active close' : ''}`}
```
mbenadda: What's the purpose of this `close` class?

lunika: It comes from the bootstrap documentation. It applies style on the button https://getbootstrap.com/docs/4.1/utilities/close-icon/

I see you use bootstrap on this project.

mbenadda: Ok. We're not using bootstrap directly, just `@import`ing some stuff we need. Also those are not buttons in the style of bootstrap as you can probably see.

Additionally following an impetus from @sveetch we have been dropping bootstrap from our styles.

mbenadda: We're mostly following BEM. There's no proper style guide outside of that. I think we're in the middle of a process to move from bootstrap to our own thing.

lunika:
The icon is bigger now and I rename my commit title.

screenshot from 2018-10-04 17-41-48

lunika:
I change the icon by a bigger one and the color is now red. Using a red color is more standard I think when you want to remove/delete an item.

WDYT about this improvement ?

screenshot from 2018-10-04 20-27-13

mbenadda:
I think it's still too small! IMO it's best if it's quite larger and in white.

Also on your screenshot it does not look perfectly centered relative to the facet counts

lunika:
I rollback to the white color and this time with a really bigger close icon.
I try but I fail to perfectly centered the facet. It's not the same font size, not the same thing displayed.

Here the result I have

screenshot from 2018-10-04 22-06-46

css is clearly my weakness...

src/richie-front/js/components/SearchFilter/_SearchFilter.scss
```diff
@@ -19,6 +19,7 @@ $richie-search-filters-active-fontweight: null !default;
 $richie-search-filters-active-background: null !default;
 $richie-search-filters-active-divider-border: null !default;
 
+$richie-search-filters-close-font-size: $font-size-base * 2 !default;
```
jmaupetit: I think we can make it more generic, _i.e._ `$richie-search-filters-icon-font-size`. WDYT?

lunika: why not but doing this, every new icon added in the search interface will have this size which is pretty big. Is it what you want ? I though the close icon is an exception but if it's not the case we can change this name yes.

jmaupetit: I'll let @mbenadda answer this question, but IMO, if the close icon is really bigger, we can do something like:

```scss
&__close {
  // [...]
  font-size: 1.5 * $richie-search-filters-icon-font-size;
}
```

mbenadda:
I was not completely satisfied with the result with `&times;` so I tried to improve on this (https://github.com/openfun/richie/commit/c2f2dce4eb0e262081b8f0d346d9ece178891caa)

- added i18n for the "Close" label
- move the `"Close"` label to the icon and added `aria-pressed` to the button
- used an svg icon for the close icon

What do you think?

lunika:
I'm ok with this. I only see difference on the font weight but I think your solution is better.

PR #381, opened 2018-10-04
🚸(front) enable usage of redux dev tools extension

lunika:
[redux devtools extension](https://github.com/zalmoxisus/redux-devtools-extension) is a nice browser extension allowing to follow state change in realtime.
It's a great tools when developing with react + redux.

this tool can be enable only in development if needed. The author wrote a
nice article about that : https://medium.com/@zalmoxis/using-redux-devtools-in-production-4c5b56c5600f

## Purpose

React development ecosystem has great tool to give a better development experience. If you probably know the [react devtool](https://github.com/facebook/react-devtools), it's brother the [redux devtools extension](https://github.com/zalmoxisus/redux-devtools-extension) is really great but need some configuration when you create the redux store.

## Proposal

Enable usage of redux devtools extension

You probably want to know if there are some overhead in production. You can read this article https://medium.com/@zalmoxis/using-redux-devtools-in-production-4c5b56c5600f and the project [FAQ](https://github.com/zalmoxisus/redux-devtools-extension/blob/master/docs/FAQ.md)

### DIFF
--- a/src/richie-front/js/data/configureStore.ts
+++ b/src/richie-front/js/data/configureStore.ts
@@ -1,4 +1,4 @@
-import { applyMiddleware, createStore } from 'redux';
+import { applyMiddleware, compose, createStore } from 'redux';
 import createSagaMiddleware from 'redux-saga';
 
 import { rootReducer, RootState } from './rootReducer';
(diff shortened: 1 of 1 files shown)


mbenadda:
Hey @lunika!

Thanks for the improvement suggestion. It's always great to get a better development experience. I actually used this extension while developing the Richie frontend.

My first reaction was that we should not ship this, but then the articles you linked convinced me it's mostly harmless.

I'm still reluctant to add development tool related stuff to our product code 😞. I think if we had a clear plan to use this to enable easier bug reporting (by having users send us their state) from staging/preprod/prod, I'd be a lot more open (enthusiastic even!) about shipping this.

What do you think?

mbenadda:
> How do you use it without the code I added ?

I add this code when I want to debug Richie, I just don't commit it.

> Writing a documenation explaining how to help us or by integrating the redux remotedev package ?

Yes.
Consider it just like another feature. We need the full story on the feature before merging it. In my opinion the feature that justifies merging redux-devtools utilities is: facilitate debugging by enabling one click state uploads from informed users when they encounter an error.

Now, we've not been as diligent as we'd like on documentation these last few months but in this case documenting the way our informed users will interact with the tool is part of the work IMO.

PR #384, opened 2018-10-05
Search missing facets

lunika:
## Purpose

deal with a point in #376 

> The filters are not implemented nor the facets displayed for other dimensions than organizations & subjects. They are available on the backend and just need to be plugged in here.

## Proposal

- [x] display a facet counter in front of each harcoded filters
- [x] display then as active once selected

### DIFF
--- a/src/richie-front/js/utils/filters/getActiveFilterValues.ts
+++ b/src/richie-front/js/utils/filters/getActiveFilterValues.ts
@@ -1,6 +1,11 @@
 import { ResourceListStateParams } from '../../data/genericReducers/resourceList/resourceList';
 import { RootState } from '../../data/rootReducer';
-import { filterGroupName, FilterValue } from '../../types/filters';
+import {
+  filterGroupName,
+  FilterValue,
+  hardcodedFilterGroupName,
+  resourceBasedFilterGroupName,
+} from '../../types/filters';
 
 // Return the list of *currenly active* filter values for a dimension built from the state & current params
 export const getActiveFilterValues = (
@@ -10,22 +15,18 @@ export const getActiveFilterValues = (
 ): FilterValue[] => {
   let currentValues = currentParams[machineName];
 
+  if (!currentValues) {
+    return [];
+  }
+
+  // Wrap base typed values in arrays to make manipulating them easier
+  if (typeof currentValues === 'string' || typeof currentValues === 'number') {
+    currentValues = [currentValues];
+  }
+
   switch (machineName) {
     case 'organizations':
     case 'subjects':
-      // There are no active values to return
-      if (!currentValues) {
-        return [];
-      }
-
-      // Wrap base typed values in arrays to make manipulating them easier
-      if (
-        typeof currentValues === 'string' ||
-        typeof currentValues === 'number'
-      ) {
-        currentValues = [currentValues];
-      }
-
       return (
         currentValues
           // Get the value for each our the keys we received
@@ -43,8 +44,17 @@ export const getActiveFilterValues = (
             primaryKey: String(value!.id),
           }))
       );
-
     default:
-      return [];
+      return currentValues
+        .map(key => {
+          return state.filterDefinitions[machineName].values.find(
+            value => value.primaryKey === key,
+          );
+        })
+        .filter(value => !!value)
+        .map(value => ({
+          humanName: value!.humanName,
+          primaryKey: value!.primaryKey,
+        }));
   }
 };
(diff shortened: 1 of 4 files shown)


src/richie-front/js/components/SearchFilterGroup/SearchFilterGroup.tsx
```diff
@@ -14,7 +14,6 @@ export interface SearchFilterGroupProps {
 
 export const SearchFilterGroup = (props: SearchFilterGroupProps) => {
   const { humanName, values } = props.filter;
-
```
lunika: no idea why there is this modification. Probably the linter who did that.

mbenadda: Then there's probably no point in keeping it in the commit :)

src/richie-front/js/utils/filters/getActiveFilterValues.ts
```diff
@@ -10,22 +15,18 @@ export const getActiveFilterValues = (
...
+  if (!currentValues) {
+    return [];
+  }
+
+  // Wrap base typed values in arrays to make manipulating them easier
+  if (typeof currentValues === 'string' || typeof currentValues === 'number') {
+    currentValues = [currentValues];
+  }
+
   switch (machineName) {
```
mbenadda: Now that we have implemented both sides, I think we could make this switch more generic with something like

```
switch (true) {
  case includes(Object.keys(FILTERS_RESOURCES), machineName):
    // ...

  case includes(Object.keys(FILTERS_HARDCODED), machineName):
    // ...
}
```

You did not change this code in your PR but I think it's in scope to stop hardcoding filter names here and use names from settings instead 😇

src/richie-front/js/utils/filters/getFilterFromState.ts
```diff
@@ -32,10 +33,33 @@ export function getFilterFromState(
...
+        values: getHarcodedFilterFacetedValues(facets, machineName),
+      };
   }
 
   /* tslint:disable:variable-name */
+  function getHarcodedFilterFacetedValues(
+    facets_: typeof facets,
+    resourceName: hardcodedFilterGroupName,
+  ) {
+    const courses = state.resources.courses;
```
mbenadda: I don't think this line does anything?

src/richie-front/js/utils/filters/getFilterFromState.ts
```diff
@@ -32,10 +33,33 @@ export function getFilterFromState(
 
     // Values from state are usable as-is for hardcoded filters
     default:
```
mbenadda: Maybe we can improve this switch the same way I suggested for the other one?

mbenadda:
Nice job! Juste some small improvements but it's almost ready to merge!

src/richie-front/js/utils/filters/getActiveFilterValues.ts
lunika: I tried to change it and it works.

But I have some trouble with typescript because it doesn't know anymore what can be the value of `machineName` and I have this error :

```
 Element implicitly has an 'any' type because type '{ courses?: Maybe<ResourceByIdState<Course> & ResourceListState<Course>>; organizations?: Maybe<ResourceByIdState<Organization> & ResourceListState<Organization>>; subjects?: Maybe<...>; }' has no index signature.
```

I looked on google but I see nothing allowing me to tell to typescript that here `machineName` is type `resourceBasedFilterGroupName` and in the other case of type `hardcodedFilterGroupName`. Do you know if there is an annotation or something else to do that ?

Thanks

mbenadda: Indeed it does not seem that typescript is able to either interpret `includes(Object.keys(FILTERS_RESOURCES), machineName)` as type discrimination, nor to apply type discrimination to an inverted switch like this.

What we can do is use a type guard and replace the switch with an `if/else`. Since we already need it in two separate places the type guard should probably be in `utils`.

Since the scope on this is increasing a bit it could probably be done in a separate PR though, so if you just fix the remaining unrelated comment this we should be ready to merge 👍

lunika: ok, I just push a commit resolving the other comments.

mbenadda:
Are you able to merge this one yourself (seeing as you're committing from a fork) or do you need to do it?

lunika:
No I can't merge my PR, I don't have wrtie access to this repo.

mbenadda:
done!

PR #386, opened 2018-10-08
♻️ (front) create a type guard to check machine name in search filters

lunika:
we don't want to hardcode filter name anymore. This PR creates a type
guard to  detect if the filter name is a resource or hardcoded one.

## Purpose

In PR #384 @mbenadda suggests to stop harcoding filter name in each switch statement. 

screenshot from 2018-10-08 11-21-46

## Proposal

screenshot from 2018-10-08 10-11-50

create a type guard allowing us to detect if the filter name is a `resource` or `hardcoded` one.

- [x] create the type guard
- [x] apply it in `getActiveFilterValues` and `getFilterFromState` functions

### DIFF
--- a/src/richie-front/js/utils/filters/getFilterFromState.ts
+++ b/src/richie-front/js/utils/filters/getFilterFromState.ts
@@ -7,6 +7,7 @@ import {
   hardcodedFilterGroupName,
   resourceBasedFilterGroupName,
 } from '../../types/filters';
+import { isResourceBasedFilterGroupName } from './isResourceBasedFilterGroupName';
 
 // Get (or build) a complete filter definition for `machineName` from the state, using:
 // - filterDefinitions for hardcoded filters and hardcoded props of resource based filters
@@ -22,22 +23,12 @@ export function getFilterFromState(
       state.resources.courses.currentQuery.facets) ||
     {};
 
-  switch (machineName) {
-    // Use the facets to build the values for resource based filters
-    case 'organizations':
-    case 'subjects':
-      return {
-        ...state.filterDefinitions[machineName],
-        values: getFacetedValues(state, facets, machineName),
-      };
-
-    // Values from state are usable as-is for hardcoded filters
-    default:
-      return {
-        ...state.filterDefinitions[machineName],
-        values: getHarcodedFilterFacetedValues(facets, machineName),
-      };
-  }
+  return {
+    ...state.filterDefinitions[machineName],
+    values: isResourceBasedFilterGroupName(machineName)
+      ? getFacetedValues(state, facets, machineName)
+      : getHarcodedFilterFacetedValues(facets, machineName),
+  };
 
   /* tslint:disable:variable-name */
   function getHarcodedFilterFacetedValues(
--- a/src/richie-front/js/utils/filters/isResourceBasedFilterGroupName.ts
+++ b/src/richie-front/js/utils/filters/isResourceBasedFilterGroupName.ts
@@ -1,0 +1,13 @@
+import includes from 'lodash-es/includes';
+
+import { FILTERS_RESOURCES } from '../../settings';
+import {
+  filterGroupName,
+  resourceBasedFilterGroupName,
+} from '../../types/filters';
+
+export function isResourceBasedFilterGroupName(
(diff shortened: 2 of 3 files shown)


src/richie-front/js/utils/filters/isResourceBasedFilterGroupName.ts
```diff
@@ -0,0 +1,14 @@
...
+import {
+  filterGroupName,
+  resourceBasedFilterGroupName,
+} from '../../types/filters';
+
+export function isResourceBasedFilterGroupName(
+  name: filterGroupName,
+): name is resourceBasedFilterGroupName {
+  return includes(
+    ['organizations', 'subjects'],
```
mbenadda: These should be taken from settings

src/richie-front/js/utils/filters/isResourceBasedFilterGroupName.ts
```diff
@@ -0,0 +1,14 @@
+import { includes } from 'lodash-es';
```
mbenadda: Alas, importing from lodash like this gets us the whole kitchen sink.

We can do `import includes from 'lodash-es/includes'` to only import the relevant function. It does not look as good but since importing the whole lodash kitchen sink would increase payload size by 50+% I think it's a reasonable tradeoff.

(That's AFAIK the best way. If you know a way to have clean imports and not get the kitchen sink please share it 🙌)

src/richie-front/js/utils/filters/getFilterFromState.ts
```diff
@@ -22,21 +23,16 @@ export function getFilterFromState(
...
-        values: getFacetedValues(state, facets, machineName),
-      };
-
-    // Values from state are usable as-is for hardcoded filters
-    default:
-      return {
-        ...state.filterDefinitions[machineName],
-        values: getHarcodedFilterFacetedValues(facets, machineName),
-      };
+  if (isResourceBasedFilterGroupName(machineName)) {
```
mbenadda: I know I'm the one who suggested an `if/else` but here a ternary would probably be more expressive and clearer

src/richie-front/js/utils/filters/isResourceBasedFilterGroupName.ts
lunika: Oh yes you are right, I forgot that we can import only the code wanted and all the lodash code base. :+1:

src/richie-front/js/utils/filters/isResourceBasedFilterGroupName.ts
```diff
@@ -0,0 +1,15 @@
...
+  filterGroupName,
+  resourceBasedFilterGroupName,
+} from '../../types/filters';
+
+export function isResourceBasedFilterGroupName(
+  name: filterGroupName,
+): name is resourceBasedFilterGroupName {
+  return includes(
+    Object.keys(FILTERS_RESOURCES),
+    name as resourceBasedFilterGroupName,
```
mbenadda: I don't think `as resourceBasedFilterGroupName` is needed here

src/richie-front/js/utils/filters/getFilterFromState.ts
```diff
@@ -22,22 +23,15 @@ export function getFilterFromState(
       state.resources.courses.currentQuery.facets) ||
     {};
 
-  switch (machineName) {
-    // Use the facets to build the values for resource based filters
-    case 'organizations':
-    case 'subjects':
-      return {
+  return isResourceBasedFilterGroupName(machineName)
+    ? {
```
mbenadda: I meant a ternary for the value of `values` since the rest is the same

src/richie-front/js/utils/filters/isResourceBasedFilterGroupName.ts
```diff
@@ -0,0 +1,15 @@
+import includes from 'lodash-es/includes';
```
mbenadda: This is a nitpick but we add a line break between dependencies imports and local imports (... which we probably need to document in the handbook!).

lunika:
The branch is rebase on top of master. Can you merge it once the CI green ?

PR #387, opened 2018-10-08
Improve SearchFiler active state

mbenadda:
## Purpose

Following #380, we have made some improvements to the active filter state. Part of #376.

Closes #380.


src/richie-front/js/components/SearchFilter/CloseIcon.tsx
```diff
@@ -0,0 +1,26 @@
+import * as React from 'react';
```
jmaupetit: Can we prepare the project for other icons to come by moving this component to something like `src/richie-front/js/components/icons/CloseIcon.tsx`?

mbenadda: This is a good idea. It will make it simpler to change things when we decide to standardize icons

PR #389, opened 2018-10-10
Refactor person plugin to fix draft vs published

sampaccoud:
## Purpose

The person plugin allows to display a person page+extension on a course page+extension.

The code as of today was not working properly and resulted in the draft version of the person being shown on the draft course as well as on the public course whatever its publication status.

## Proposal

- [x] use a foreign key to the person page instead of a foreign key to the person page extension in order to use page extensions as little as possible,
- [x] refactor the template to:
    * show the public person on the public course if available or nothing at all otherwise,
    * show the public person on the draft course if available or the draft person otherwise.
- [x] Refactor and complement the tests to secure this behavior.


src/richie/apps/persons/models.py
```diff
@@ -98,5 +102,36 @@ def __str__(self):
...
+    def check_publication(self, language=None):
+        """
+        Allow checking if the page is published without passing any language (unlike the
+        "is_published" method on the page object): if not explicitly passed as argument, the
+        language is retrieved from the context.
+
+        The actual check is subcontracted to the "is_published" method on the related Django CMS
+        Page object.
+
+        Note: We choose not to name our method "is_published" like Django CMS, because it is a
```
mbenadda: 👍

PR #390, opened 2018-10-11
✨(courses) add course plugin

2018D:
Add course plugin to enable publishing of
course glimpse for any course on any page of the CMS

## Purpose

Enables the production of custom courses cards page and courses gallery.


sampaccoud:
@2018D I think you missed this PR from earlier this week https://github.com/openfun/richie/pull/389 in which we discussed issues in our plugin implementation. The new person plugin+tests are fixing these issues so I think the course plugin should now follow this example.

I'm now moving many-to-manys between courses/subjects and courses/organizations to be also plugins.

2018D:
I sent this in right before the meeting. I did not remove afterward. Closing it now.

PR #391, opened 2018-10-11
Add subject plugin

sampaccoud:
## Purpose

We decided to replace the many-to-many fields on page extensions by page-to-page plugins. 

We found the plugin pattern (already used for persons) to be superior to using many-to-many relationships between pages:

- frontend editing is nicer and works the same as for other pieces of content,
- pages linked via plugins can be ordered,
- draft/public behavior is simpler and already works for persons. We did not manage to make it work as expected with many-to-many relationships as of today.

## Proposal

- [x] Add a page-to-page subject plugin,
- [x] Remove the many-to-many relationship between subjects and courses

While working of the subject plugins, I bumped into minor on-related issues that I fixed in separate commits:
 
- [x] typos/oversights on the recently refactored person plugin
- [x] randomloy failing test on course runs


mbenadda:
I think the styling for the subject tags on the course page is not finished, right?

capture d ecran 2018-10-12 a 18 05 22

PR #397, opened 2018-10-18
Replace organization many to many by plugin

sampaccoud:
## Purpose

Organization is the last model to still rely on many-to-many relationships with the Course model.

Other models (persons and subjects) have already moved to plugins so organizations had to follow. 

## Proposal

- [x] Fix the organization plugin to display as expected depending on its plublication status,
- [x] Replace the many-to-many relationship between organizations and courses by a page plugin placeholder,

:warning: Since plugins are sortable inside a placeholder, the main organization is now the first in the list of organizations linked to a course via plugins.


sampaccoud:
@2018D I pushed a fixup to correct the 2 bugs you surfaced. Thanks!

PR #404, opened 2018-10-25
Limit resource-based filters to 10 displayed suggestions

mbenadda:
## Purpose

When we're using facet counts to display a resource-based filter, we're only showing the top 10 available values in the list. (We'll be adding a UI element to filter using other values).

When there are no facet counts, we're displaying all possible values. This is not ideal as it might make the UI jitter on load & looks bad / makes it seem like there is a bug when we really are missing the facets.

## Proposal

Sample 10 random values from the full list in this case.

### DIFF
--- a/src/richie-front/js/utils/filters/getFilterFromState.spec.ts
+++ b/src/richie-front/js/utils/filters/getFilterFromState.spec.ts
@@ -1,3 +1,5 @@
+import { find } from 'lodash-es';
+
 import { CoursesState } from '../../data/courses/reducer';
 import { RootState } from '../../data/rootReducer';
 import {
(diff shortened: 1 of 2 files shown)


src/richie-front/js/utils/filters/getFilterFromState.spec.ts
```diff
@@ -140,14 +142,67 @@ describe('utils/filters/getFilterFromState', () => {
...
-      values: [
+      // Values might be mis-ordered as we're sampling them instead of just reusing the array
+      values: expect.arrayContaining([
         { humanName: 'Organization #Twenty-One', primaryKey: '21' },
         { humanName: 'Organization #Fourty-Two', primaryKey: '42' },
         { humanName: 'Organization #Eighty-Four', primaryKey: '84' },
-      ],
+      ]),
     });
+    // If the array contains only 3 values *and* contains our 3 values, we can reasonably thinl
```
jmaupetit: `s/thinl/think/`

src/richie-front/js/utils/filters/getFilterFromState.ts
```diff
@@ -61,7 +62,9 @@ export function getFilterFromState(
     // We don't have the facets yet or something broke upstream: provide some filtering
     // capabilities anyway (without counts, as we can't generate those)
     if (!facets_[resourceName] || !Object.keys(facets_[resourceName]).length) {
-      return values(state.resources[resourceName]!.byId)
+      // Return a maximum of 10 values so the page does not seem bugged when missing the facet
+      // counts, and the display is always consistent (as our facet counts cap at the top 10)
+      return sampleSize(values(state.resources[resourceName]!.byId), 10)
```
jmaupetit: Do you think the size of the sample should be configurable?

jmaupetit: Side remark: `sampleSize` is really a bad name for this function...

jmaupetit: Ok, let's do this!

PR #406, opened 2018-10-26
add course plugin

2018D:
## Purpose
Add course plugin to enable publishing of
course glimpse for any course on any page of the CMS
This allows the creation of custom course presentation pages or sections.

## Proposal
- [x] Add course plugin
- [x] Add style
- [x] Add tests


tests/apps/courses/test_cms_plugins_course.py
```diff
@@ -0,0 +1,158 @@
...
+                fields = ["page"]
+
+        course = CourseFactory()
+        other_page_title = "other page"
+        create_page(other_page_title, "richie/fullwidth.html", settings.LANGUAGE_CODE)
+        plugin_form = CoursePluginModelForm()
+        self.assertIn(course.extended_object.get_title(), plugin_form.as_table())
+        self.assertNotIn(other_page_title, plugin_form.as_table())
+
+    def test_cms_plugins_course_render(self):
```
sampaccoud: I think you need to have to separate tests: one for including a plugin on a draft page and one for including a plugin on a public page. See example here: https://github.com/openfun/richie/blob/master/tests/apps/courses/test_cms_plugins_subject.py

2018D: i had a few bad test, i had to fix. It should be ready for a final review.

2018D: i added the test for plugin on draft page, might as well rename the first one.. gimme a second

2018D: I've renamed it as well.

2018D:
the fails are from a test server being unavailable: "received unexpected HTTP status: 503 Service Unavailable"

src/richie/apps/courses/templates/courses/plugins/course_plugin.html
```diff
@@ -0,0 +1,12 @@
+{% load cms_tags i18n %}
+{% spaceless %}
+{% if current_page.publisher_is_draft or instance.check_publication is True %}
+  <div class="course-plugin-container {% if relevant_page.publisher_is_draft %} course-plugin-container--draft{% endif %}">
+    <div class="course-plugin">
+      <a class="course-plugin__body course-plugin__link" href="{{ relevant_page.get_absolute_url }}" title="{{ relevant_page.get_title }}" alt="{{ relevant_page.get_title }} {% trans 'logo' %}">
```
mbenadda: I don't think you should have both `__body` and `__link` on the same BEM element name.

If this is a `course-glimpse` and a `course-glimpse--link` just put both on the same class and use it here :)

2018D: should be better now 👍

src/richie/apps/courses/templates/courses/plugins/_course_plugin.scss
```diff
@@ -0,0 +1,15 @@
...
+  display: inline-block;
+  min-width: 250px;
+  margin: auto;
+}
+
+.course-plugin {
+  @extend .course-glimpse;
+  max-width: 100%;
+  
+  &__link {
```
mbenadda: Sorry about this, but I think a tenet of BEM is that a modifier class (such as `.course-glimpse--link`) is supposed to be used on the same element as the class it's modifying (`.course-glimpse`).

It may work in this instance but it feels wrong to use them as parent & child, even through an extend.

mbenadda: This is what I'm talking about:

capture d ecran 2018-11-13 a 17 35 52

`.course-glimpse` & `.course-glimpse--link` need to be **on the same element**, even if you're just applying them through an extend.

I guess what I'm thinking this should look like is more something like this:

```
.course-plugin {
  @extend .course-glimpse;
  @extend .course-glimpse--link;
  max-width: 100%;
}
```

I think this matches the way they were meant to be used, as can be seen eg. on the organization detail page https://github.com/openfun/richie/blob/master/src/richie/apps/courses/templates/courses/cms/organization_detail.html#L34.

Sorry if my previous comments were a little bit unclear 😅

2018D: Oh i clearly misunderstood the example then... Give me a moment, it's gonna be clean in a moment.

src/richie/apps/courses/models/course.py
```diff
@@ -213,6 +214,28 @@ def state(self):
...
+class CoursePluginModel(PagePluginMixin, CMSPlugin):
+    """
+    Course plugin model handles the relation from CoursePlugin
+    to their Course instance
+    """
+
+    page = models.ForeignKey(
+        Page,
+        related_name="course_plugins",
+        limit_choices_to={"publisher_is_draft": True, "course__isnull": False},
```
mbenadda: We need to filter out the snapshots here.

There's currently a bug on `master` where snapshots are included in the list of courses on organizations & subjects pages. @sampaccoud corrected this in #423 in [this commit](https://github.com/openfun/richie/pull/423/commits/6362e03a7b25c9f95fea0a3ac9dfc3f9d16ccab9).

We need to do something analogous here, otherwise the snapshots will appear in the dropdown to select a course.

sampaccoud: You should just need to add a clause:
```python
"node__parent__cms_pages__course__isnull": True
```
But before you do that, try writing a test that surfaces the issues ;-)

2018D: All done. This need review.

PR #410, opened 2018-10-31
Allow duplicating courses and keep course runs linked to their version of the course

sampaccoud:
## Purpose

This change is required so that when a course is being modified, the course runs associated to the course can still display the version of the course as it was before modification. 

## Proposal

A course can be snapshotted and the snapshot is placed as a child of the current course page. Course run pages are children of the course page in a specific version (either the current course page or one of the snapshots). We will use the parent page/child page relationship baked in the CMS instead of a course foreign key so that the current course can easily display all course runs by fetching all its descendants:

- [x] refactor course runs to become CMS pages, children to their related course,
- [x] refactor the course detail template to share markup with the new course run page,
- [x] refactor css to match the new markup on the course page and style the course run page,
- [x] add a wizard to easily create a new course run.

This was quite a big refactoring and I had the opportunity to make unrelated improvements:
- [x] make course run dates and page titles editable on the front end,
- [x] stop including language markers in strings for testing


sampaccoud:
@2018D yes I had the same idea but I didn't find an obvious place to put this link... so I concluded that we should make it a breadcrumb and left it for a future PR. If you have ideas on your side, we can open an issue and discuss how to do it?

sandbox/settings.py
```diff
@@ -288,6 +288,7 @@ class Base(DRFMixin, ElasticSearchMixin, Configuration):
```
mbenadda: that is placed* as a child

src/richie/apps/courses/cms_wizards.py
```diff
@@ -2,7 +2,8 @@
 CMS Wizard to add a course page
```
mbenadda: add a wizard to* easily...?

PR #411, opened 2018-11-01
Enable code reuse in viewsets

mbenadda:
## Purpose

We generally want to promote code reuse throughout Richie. Here, we're trying to apply this idea to consumers of our indexers, and ViewSets in particular.

As we were working on adding new endpoints for autocompletion to our ViewSets, we realized the functionality would be the same for all of our resources.

However, we encountered two issues that prevented us from just using a Mixin to share this code:
- indexers had different names for ES to API object formatted method. 
- there was no way for our Mixins to access the indexers.

## Proposal

- [x] Rename the method to `format_es_object_for_api` on all indexers. In our opinion, the indexers ought to have the same shape. All other common attributes and methods already share the same name.
- [x] Take a hint from Django,  add a `_meta` prop to our ViewSets. This way Mixins get to access the relevant indexer through `self._meta`.

### DIFF
--- a/src/richie/apps/search/utils/viewsets.py
+++ b/src/richie/apps/search/utils/viewsets.py
@@ -1,0 +1,17 @@
+"""
+Helpers to enable reuse for needs that are shared between viewsets.
+"""
+
+
+class ViewSetMetadata:
+    """
+    "Meta" class intended to be used as an attribute on ViewSets to provide a common set of
+    meta attributes, like the Django `class Meta` on eg. models.
+    """
+
+    def __init__(self, indexer):
+        """
+        Set the required attributes passed from the call site. Currently this only includes
+        the ViewSet's corresponding indexer.
+        """
+        self.indexer = indexer
--- a/src/richie/apps/search/indexers/courses.py
+++ b/src/richie/apps/search/indexers/courses.py
@@ -83,7 +83,7 @@ class CoursesIndexer:
         # This means there can be no overlap between the various buckets, but we can still
         # sort courses inside each bucket as we see fit by simply adding timestamps (ascending
         # order) or substracting them (descending order).
-        "sort_courses_list": {
+        "sort_list": {
             "script": {
                 "lang": "expression",
                 "source": (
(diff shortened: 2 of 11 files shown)


src/richie/apps/search/utils/viewsets.py
```diff
@@ -0,0 +1,19 @@
+"""
+Helpers to enable reuse for needs that are shared between viewsets.
+"""
+
+
+class ViewSetMeta:
```
sampaccoud: you can set it on self._meta on the viewset but don't call this calss Meta because Meta classes are something else...
Django is kind of misusing this word but it has become a common practice in other project to attach options to a set._meta attribute. But it is meta as in "metadata", not as in "metaclass".
Maybe you could call this class ViewSetOptions...

mbenadda: I just went for `ViewSetMetadata`, that way it can't be confused with a metaclass :)

src/richie/apps/search/utils/viewsets.py
```diff
@@ -0,0 +1,17 @@
...
+    "Meta" class intended to be used as an attribute on ViewSets to provide a common set of
+    meta attributes, like the Django `class Meta` on eg. models.
+    """
+
+    def __init__(self, indexer):
+        """
+        Set the required attributes passed from the call site. Currently this only includes
+        the ViewSet's corresponding indexer.
+        """
+        self.indexer = indexer
```
sampaccoud: Given how you use it, you could do:
```suggestion
for key, value in kwargs:
    setattr(self, key, value)
```

src/richie/apps/search/utils/viewsets.py
```diff
@@ -0,0 +1,17 @@
...
+"""
+
+
+class ViewSetMetadata:
+    """
+    "Meta" class intended to be used as an attribute on ViewSets to provide a common set of
+    meta attributes, like the Django `class Meta` on eg. models.
+    """
+
+    def __init__(self, indexer):
```
sampaccoud:
```suggestion
    def __init__(self, **kwargs):
```

PR #418, opened 2018-11-08
Use autocomplete endpoints in main search autosuggest

mbenadda:
## Purpose

We opened autocomplete endpoints specifically to have better suggestion results in search boxes with the autosuggest UI.

Autocomplete as a functionality needs to be very easy to understand for users. They need to see at a glance why so or so result is being suggested.

## Proposal

- [x] Add the autocomplete endpoints to the settings & call them instead of the search endpoints when generating suggestions.
- [x] ~Add a helper to find the query in the proposed suggestions and highlight it~


mbenadda:
@lunika the query highlighter is not sufficient as it does not work with diacritics. I'll update the backend to take care of this as well as diacritics-insensitive suggestions.

I think the 1st part of this PR (using the autocomplete endpoints) is still good but I'm going to forgo the second commit.

PR #421, opened 2018-11-12
enable redux devtools :tada:

### LINKED ISSUE #383: enable redux devtools by default
## Feature Request

Richie front application uses Redux. There is a lot of tools available to improve the developer experience when using redux. One of the most famous (and most used) is [redux devtools](https://github.com/zalmoxisus/redux-devtools-extension). 

To use it, you should configure it in your redux store, so you must, programmatically, add dedicated code in your codebase to allow its usage. In your codebase, it should be [here](https://github.com/openfun/richie/blob/master/src/richie-front/js/data/configureStore.ts). Without this modification it is not allow to use redux devtools.

For a better developer experience, it should be great to configure, by default, the usage of this tool. I've created a PR for this (#381) but this one will not be merged as is because openfun's team does not want to ship this code in production.

The configuration needed is explain [here](https://github.com/zalmoxisus/redux-devtools-extension#12-advanced-store-setup) and it can be use as is in production.

An alternative can be to configure the tool to be only enabled in development.
There is a dedicated chapter in the documentation about using it in production : https://github.com/zalmoxisus/redux-devtools-extension#14-using-in-production
You should also read this article, explaining why it is not bad to have this code in your production environment : https://medium.com/@zalmoxis/using-redux-devtools-in-production-4c5b56c5600f

There will be nothing to do. Richie will be configured to use redux devtools and if a new developer wants to contribute he has nothing to do.

A PR already exists #381

lunika:
## Purpose

redux devtools are not used and, as described in #383, this is useful in development (but also in production).

close #383 

## Proposal

enable redux devtools with a configuration parameter. This parameter can be change on each environement. The goal is to enable it in development and disable it otherwise.

This configuration can be enhanced later for other part of the project.


mbenadda:
Before adding a new system to manage configurations for your new redux devtools variable, I think you need to refactor existing settings to fit into this new system.

There is already a `settings` file in richie, located at `src/richie-front/js/settings.ts`.

In my opinion, settings are best expressed through typescript ([config supports this](https://github.com/lorenwest/node-config/wiki/Configuration-Files#typescript---ts)) as this lets us have proper typings & type checks for those settings, as opposed to "anything goes" JSON, which is also more painful to write.

The problem then is: I'm not sure if `config` is able to get us properly typed values in our code without typecasting at every use site. Have you looked into this?

src/richie-front/js/data/configureStore.ts
```diff
@@ -1,16 +1,25 @@
...
 
 const sagaMiddleware = createSagaMiddleware();
 
+declare const CONFIG: Config;
+
 export function configureStore(preloadedState: RootState) {
+  const composeEnhancers =
+    CONFIG.enableDevTools &&
+    (window as any).__REDUX_DEVTOOLS_EXTENSION_COMPOSE__
+      ? (window as any).__REDUX_DEVTOOLS_EXTENSION_COMPOSE__
```
mbenadda: I think `&&` and `||` would be clearer here. This relies on operator precedence which might be confusing for some

package.json
```diff
@@ -71,6 +72,7 @@
   },
   "dependencies": {
     "bootstrap": "4.1.3",
+    "config": "^2.0.1",
```
mbenadda: We always pin dependency versions (see other deps in this `package.json`)

src/richie-front/js/data/configureStore.ts
```diff
@@ -1,16 +1,25 @@
...
+import { applyMiddleware, compose, createStore } from 'redux';
 import createSagaMiddleware from 'redux-saga';
 
+import { Config } from '../types/config';
 import { rootReducer, RootState } from './rootReducer';
 import { rootSaga } from './rootSaga';
 
 const sagaMiddleware = createSagaMiddleware();
 
+declare const CONFIG: Config;
```
mbenadda: I think this would be better as an import. We could have a file that does EG:

```
declare const CONFIG: Config;
export CONFIG = CONFIG;
```

Also maybe we can keep the name settings as in the Django part of the project? What do you think?

lunika:
> In my opinion, settings are best expressed through typescript (config supports this) as this lets us have proper typings & type checks for those settings, as opposed to "anything goes" JSON, which is also more painful to write.

I tried to use config with typescript but I have several issues :
- config relies ont `ts-node` to execute typescript in node
- once `ts-node` installed I have this error : `Cannot parse config file: '/home/manu/dev/fun/richie/config/default.ts': SyntaxError: Unexpected token export`. I presume that the code generated is not e6 compliant and `export` is not recognized but even if I run webpack with node having `--experimental-modules` option enable I still have this error.

lunika:
> There is already a settings file in richie, located at src/richie-front/js/settings.ts.

These settings are not configurable and will not change depending the environment they will be used. These settings are here to not repeat code again and again so I'm not sure we should merge settings with new config.

The new config will be injected when the build is made by webpack, it replaces by the real value, I don't think this is the behaviour we want for settings.

src/richie-front/js/data/configureStore.ts
lunika: > Also maybe we can keep the name settings as in the Django part of the project? 

like this `ENABLE_DEV_TOOLS` ?

For the declare part, I created a `@types` folder is src and I use it in `tsconfig.json`

mbenadda: I meant naming the variable `settings` instead of `CONFIG` but I see this name is forced by the webpack plugin. Ignore this comment :)

lunika: No it's possible to change it https://www.npmjs.com/package/config-webpack#configuring

mbenadda: Now that you say it, existing settings are formatted like Django in all-caps snake-case. I think we can keep it that way.

mbenadda:
Here is a working shim:

```
import { Config } from './types/config';

// Declare global CONFIG as a Config object: it was injected this way by
// the `configWebpackPlugin` during the build.
declare const CONFIG: Config;

// Export a usable object we'll be able to cleanly import everywhere else.
export const config = CONFIG;

```

mbenadda:
As for `module.exports.default`, it would be nicer to be able to use `export` but I don't think we can easily override the `module` prop for `ts-node` as it is called through several levels of indirection.

Let's keep it like that 👍

PR #430, opened 2018-11-27
Pin npm deps in dockerfiles

mbenadda:
## Purpose

Builds were broken on `master` due to two things:
- `flatmap-stream` versions were depublished due to a malicious code injection
- `pylint-django` was not compatible with a new version of `astroid`

The `flatmap-stream` issue also revealed us our Dockerfiles were pulling incorrect versions of dependencies.

## Proposal

- [x] fix Dockerfiles so they use dependencies versions from the lock files;
- [x] update lock file to drop `flatmap-stream`;
- [x] update `pylint-django`;
- [x] pin `@types/react` in a resolution to ensure we don't waste time with incorrectly deduplicated lock files.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -599,9 +599,9 @@
     to-fast-properties "^2.0.0"
 
 "@types/cheerio@*":
-  version "0.22.9"
-  resolved "https://registry.yarnpkg.com/@types/cheerio/-/cheerio-0.22.9.tgz#b5990152604c2ada749b7f88cab3476f21f39d7b"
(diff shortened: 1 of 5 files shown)


yarn.lock
```diff
@@ -599,9 +599,9 @@
     to-fast-properties "^2.0.0"
 
 "@types/cheerio@*":
-  version "0.22.9"
```
sampaccoud: in commit message:
s/recenctly/recently

PR #443, opened 2019-01-17
Index courses from models

sampaccoud:
## Purpose

The Elasticsearch index should be created from the page models and their related plugins that form the content of a page.

Up to now it was created via API calls for demo purposes.

## Proposal

- [x] replace the API calls by models walkthroughs,
- [x] refactor the ES index to fit the information we have in models. In particular, I propose to index course runs and not courses.
- [x] fix tests


src/richie/apps/courses/models/course.py
```diff
@@ -132,6 +132,10 @@ def get_course_runs(self):
         """
         Returns a query yielding the course runs related to the course. They may be direct
         children of the course or children of a snapshot of the course.
+
+        For a draft course, the related draft course runs are retrieved.
+        For a public course, the related public course runs are retrieved, but only if they
```
mbenadda: The tests seem to cover all cases, but I just don't understand the explanation clearly here: what is a `public` course run if not a `published` course run?

src/richie-front/js/types/Course.ts
```diff
@@ -1,21 +1,14 @@
 import { Resource } from './Resource';
 
 export interface Course extends Resource {
-  end_date: string;
-  enrollment_end_date: string;
-  enrollment_start_date: string;
-  language: string;
-  organization_main: number;
+  start: string;
```
mbenadda: The props should be alphabetically ordered

src/richie/apps/search/description.apib
```diff
@@ -7,21 +7,15 @@ FORMAT: 1A
...
-+ short_description: In this course you'll learn to use Python to accomplish tasks and automate some of your workflows... (string, required) - Brief summary of the course, going into more detail than the title, but smaller than the whole syllabus
-+ start_date: `2018-02-01T06:00:00Z` (string, required) - ISO 8601 date, when this session of the course starts
++ start: `2018-02-01T06:00:00Z` (string, required) - ISO 8601 date, when this session of the course starts
 + subjects: [42, 44] (array[number], required) - IDs of the subjects this course has been tagged with
-+ thumbnails:
-    + about https://example.com/about.png (string) - URL to an image, a size of the thumbnail
-    + big https://example.com/big.png (string) - URL to an image, a size of the thumbnail
-    + facebook https://example.com/facebook.png (string) - URL to an image, a size of the thumbnail
-    + small https://example.com/small.png (string) - URL to an image, a size of the thumbnail
++ cover_image: `/about.png` (string) - URL to an image, a size of the thumbnail
```
mbenadda: this prop is out of order too

src/richie/apps/search/indexers/courses.py
```diff
@@ -161,20 +224,20 @@ def format_es_object_for_api(es_course, best_language):
...
-            "enrollment_end_date": es_course["_source"]["enrollment_end_date"],
-            "enrollment_start_date": es_course["_source"]["enrollment_start_date"],
             "id": es_course["_id"],
-            "language": es_course["_source"]["language"],
-            "organization_main": es_course["_source"]["organization_main"],
-            "organizations": es_course["_source"]["organizations"],
-            "session_number": es_course["_source"]["session_number"],
-            "short_description": get_best_field_language(
-                es_course["_source"]["short_description"], best_language
+            "start": es_course["_source"]["start"],
```
mbenadda: order props alphabetically?

tests/apps/search/test_indexers_courses.py
```diff
@@ -35,6 +37,66 @@ class CoursesIndexersTestCase(TestCase):
     and especially dynamic mapping shape in ES
     """
 
+    def test_indexers_courses_organizations_correspondance(self):
```
mbenadda: 👍

Maybe `test_indexers_related_objects_consistency`? (or correspond~a~**e**nce)

src/richie-front/js/settings.ts
```diff
@@ -76,10 +76,10 @@ export const FILTERS_HARDCODED: {
       {
         humanName: defineMessages({
           message: {
-            defaultMessage: 'Current session',
```
mbenadda: Nice catch!

tests/apps/search/test_indexers_subjects.py
```diff
@@ -18,38 +20,32 @@ class SubjectsIndexersTestCase(TestCase):
...
-            url=settings.SUBJECT_API_ENDPOINT + "?page=1&rpp=50",
-            match_querystring=True,
-            json={
-                "count": 51,
-                "results": [
-                    {"id": 62, "image": "example_cs.png", "name": "Computer Science"}
-                ],
-            },
+        subject1 = SubjectFactory(
+            page_title={"en": "my first subject", "fr": "ma première sujet"},
```
mbenadda: Using `première` to check it works with accents?

src/richie-front/js/utils/searchSuggest/suggestionHumanName.ts
```diff
@@ -2,17 +2,5 @@ import { modelName } from '../../types/models';
 import { ResourceSuggestion } from '../../types/searchSuggest';
 
 export const suggestionHumanName = (suggestion: ResourceSuggestion) => {
```
mbenadda: I was going to suggest we move this to the call site and drop the tests...

... but then I though we might add People to the suggestions and their human name will probably not be `title`.

mbenadda:
Also you could probably drop `api_consumption.py` now that there's nothing that uses `walk_api_json_list` anymore.

tests/apps/courses/test_models_course.py
```diff
@@ -24,6 +28,64 @@ def test_models_course_str(self):
...
+        course = CourseFactory(should_publish=True)
+        self.assertFalse(course.get_organizations().exists())
+        self.assertFalse(course.public_extension.get_organizations().exists())
+
+    def test_models_course_get_organizations(self):
+        """
+        The `get_organizations` method should return all organizations linked to a course and
+        should respect publication status.
+        """
+        *draft_organizations, _other_draft = OrganizationFactory.create_batch(3)
```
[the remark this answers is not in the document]
sampaccoud: At this moment yes. But in the code that follows I use `draft_organizations` to link them to the course whereas `_other_draft` ends up not being linked to any course. 

This is to test that organizations are returned because they are linked to the course and not all organizations are returned.

tests/apps/search/test_indexers_subjects.py
sampaccoud: hmm more liike a copy/pasta failure from `organizations"... Maybe I can write `ma première thématique`...

tests/apps/courses/test_models_course.py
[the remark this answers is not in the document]
mbenadda: Oh I see, hence the _ prefix as it’s just for unpacking. Maybe this needs a comment to make sure no one removes it by mistake

src/richie-front/js/types/Course.ts
sampaccoud: why doesn't it break the build?

mbenadda: I’m thinking this might not be linted because of declaration merging on interfaces.

src/richie/apps/courses/models/course.py
sampaccoud: the public course run is the instance representing the published course run as opposed to the draft one.

By published I mean that it needs to be currently publicly visible.

The catch is that a page in DjangoCMS can be unpublished... The public object is created the first time a draft object is published. But then of course this `public` instance will still exist if the object is then unpublished (in all languages)... so a page can have a public version of itself in database but not be `published` in any language...  
I'm going to try to rephrase this docstring.

src/richie/apps/search/indexers/courses.py
sampaccoud: in the backend I was trying to put the id first, then dates, then the remaining fields in alphabetical order. WDYT?

src/richie/apps/courses/models/course.py
mbenadda: Ok thanks for the clarification, I'm not as familiar with Django CMS so it was not obvious to me 👍

src/richie/apps/search/indexers/courses.py
mbenadda: IMO it's easier to look through props when they are alphabetically ordered unless there's a clear reason to chose another order (or categories like you did with the mappings).

Anyway, it's not a big deal either.

PR #452, opened 2019-01-29
Add healthcheck

jmaupetit:
## Purpose

As mentioned in https://github.com/openfun/arnold/issues/69 we need a way to check our application's status. We choose to follow Mozilla's Dockerflow integration pattern.

## Proposal

- [x] add `dockerflow` new dependency
- [x] configure `dockerflow`
- [x] fix the CI-generted `version.json` file path


rmoch:
Django dockerflow implementation documentation: 
https://github.com/mozilla-services/python-dockerflow/blob/master/docs/django.rst

PR #457, opened 2019-01-31
Add link to course run on search glimpses

mbenadda:
## Purpose

We could not link to the courses / course runs from the course glimpses in seach before actual courses from the models were indexed in elasticsearch.

Ditto for the courses in suggestions.

## Proposal

Now that we have a real index, we can trivially add those links by using the new `absolute_url` field on Courses.

~NB: rebased on #450~


mbenadda:
Ping @sampaccoud

PR #459, opened 2019-02-01
Improve full text search experience

### LINKED ISSUE #453: Improve user experience for search queries
## Bug Report

**Expected behavior/code**
When I type something in the search bar and tap `Enter` (or click on the suggested keyword search in the autocomplete drop-down), the query parameter should be added to the querystring.
When I empty the search bar and tap `Enter`, the query param should be removed from the querystring.

**Actual Behavior**
The querystring is not updated when interacting with the search bar.

**Steps to Reproduce**
1. load the search page localhost:8070/en/courses/
2. type "foo" in the search bar
3. either:
    a. click `Enter`
    b. click on "Search for foo in courses..." in the suggestion dropdown
3. Notice that the querystring has not changed
4. Now click on any filter in the left side bar
5. Notice that the query parameter `query=foo` is now added to the querystring... but it should have been added as soon as step 3.
6. Now remove `foo` from the search bar
7. Tap `Enter`
8. Notice that the `query=foo` is still in the querystring
9. Now click on any filter in the left side bar
10. Notice that `query=foo` has now been removed... but it should have been removed as soon as step 8.

mbenadda:
## Purpose

We were adding filters (such as facets or related resources) to the URL and history, but not full text search queries. They were only added to the URL when eg. a filter forced an update.

We also noticed the UI was no picking up the "query" param from the URL. We then added a call in the `<SearchSuggestField />` constructor to get it to initialize its state.

Finally, we needed to add a way for users to easily remove the full text query.

## Proposal

- [x] add the relevant action to the `fullTextSearch` handler to it adds full text searches to the history;
- [x] pick up full text search queries from the URL at startup;
- [x] delete the current full text search and update history when the user presses enter on an empty input.

Closes #453.

### DIFF
--- a/src/richie-front/js/components/SearchSuggestField/SearchSuggestField.tsx
+++ b/src/richie-front/js/components/SearchSuggestField/SearchSuggestField.tsx
@@ -17,6 +17,7 @@ import {
 } from '../../types/searchSuggest';
 import { commonMessages } from '../../utils/commonMessages';
 import { handle } from '../../utils/errors/handle';
+import { getSearchParam } from '../../utils/indirection/getSearchParam';
 import { location } from '../../utils/indirection/location';
 import { getSuggestionsSection } from '../../utils/searchSuggest/getSuggestionsSection';
 import { suggestionHumanName } from '../../utils/searchSuggest/suggestionHumanName';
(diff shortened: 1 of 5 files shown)


src/richie-front/js/components/SearchSuggestField/SearchSuggestField.tsx
```diff
@@ -230,17 +243,18 @@ export class SearchSuggestFieldBase extends React.Component<
       // TypeScript incorrectly infers the type of the Autosuggest suggestions prop as SearchSuggestion, which
       // would be correct if we did not use sections, but is incorrect as it is.
       <Autosuggest
-        suggestions={suggestions as any}
+        getSectionSuggestions={suggestionsFromSection as any}
         getSuggestionValue={getSuggestionValue}
-        highlightFirstSuggestion={true}
+        highlightFirstSuggestion={value.length > 2}
```
sampaccoud: I found it a bit disturbing that nothing happens when I press `enter` after typing only one letter.

PR #460, opened 2019-02-01
♻️(subjects) rename subject to category to prepare for meta-categories

sampaccoud:
## Purpose

A recurrent feature request for Richie is to allow multi-dimensional categories:
- we want to group categories below meta-categories (like for example
  subject, difficulty level, type...)
- we want to allow taxonomies (e.g. algebra and geometry are sub-
  categories of mathematics)

The first step to add this possibility is to rename `subjects` to `categories`.

## Proposal

- [x] search and replace all occurrences of the word `subject` by the word `category` and `subjects` by `categories`, respecting different cases: subject, Subject, SUBJECT ;
- [x] review replacements one-by-one in the `diff` and correct false positives ;
- [x] massage until all tests and linters pass.


src/richie-front/js/components/Search/Search.spec.tsx
```diff
@@ -10,9 +10,9 @@ import { Search } from './Search';
 describe('components/Search', () => {
   it('renders the filters pane and the list of courses', () => {
     const orgSpy = jasmine.createSpy('OrganizationSpy');
-    const subjSpy = jasmine.createSpy('SubjectsSpy');
+    const subjSpy = jasmine.createSpy('CategoriesSpy');
```
mbenadda: `subjSpy`

src/richie-front/js/data/categories/reducer.spec.ts
```diff
@@ -1,25 +1,25 @@
 import { modelName } from '../../types/models';
-import { subjects } from './reducer';
+import { categories } from './reducer';
 
-describe('data/subjects reducer', () => {
+describe('data/categories reducer', () => {
   const subj43 = {
```
mbenadda: `subj43` etc.

src/richie-front/js/data/genericReducers/resourceById/resourceById.spec.ts
```diff
@@ -4,20 +4,20 @@ import { byId } from './resourceById';
 describe('data/genericReducers/resourceById reducer', () => {
   const subj43 = {
```
mbenadda: `subj43` etc.

PR #462, opened 2019-02-01
🐛(organization) organization plugin logo too big

2018D:
## Purpose

Added sass value to cap the logo image size when it's in the plugin.
Removed unused value.

### DIFF
--- a/src/richie/apps/courses/templates/courses/plugins/_organization_plugin.scss
+++ b/src/richie/apps/courses/templates/courses/plugins/_organization_plugin.scss
@@ -30,12 +30,12 @@ $richie-organization-plugin-title-height: 1.1 * $headings-line-height *
     border: $richie-organization-plugin-border-hover;
     cursor: pointer;
   }
-  &__image {
-    @extend .card-img-top;
-  }
   &__body {
     @extend .card-body;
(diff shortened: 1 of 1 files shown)


src/richie/apps/courses/templates/courses/plugins/_organization_plugin.scss
```diff
@@ -30,12 +30,12 @@ $richie-organization-plugin-title-height: 1.1 * $headings-line-height *
     border: $richie-organization-plugin-border-hover;
     cursor: pointer;
   }
-  &__image {
-    @extend .card-img-top;
-  }
   &__body {
     @extend .card-body;
     @extend .text-center;
+    img {
```
mbenadda: maybe `& > img` to pinpoint the image as accurately as possible

2018D: What do you mean ? I am not sure I understand.

mbenadda: I just meant to use a more precise selector. We're not targeting any image somewhere inside the card but the one image directly inside the `__body`

PR #464, opened 2019-02-04
Reorganize the project's file structure

mbenadda:
## Purpose

Like we did on openfun/marsha, we're adopting the practice of grouping all source code related files into `src/`, split between frontend & backend files.

This helps us keep the project root as clean and simple as possible and should make it easier for observers and new contributors to get around in the project.

Related files such as linter config, dependencies, etc. should live alongside their code in those src/{part} folder.

## Proposal

- [x] Move all frontend related files into `src/frontend` (from `src/richie-front` and `./`)
- [ ] Move all backend related files into `src/backend`


mbenadda:
I'm up for a review on this one 💁‍♂️

mbenadda:
Ok I think `Makefile` commands work now.

I read https://docs.docker.com/engine/reference/builder/#dockerignore-file but I'm not sure what `.dockerignore` is really for or if I should update it.

jmaupetit:
The `.dockerignore` file lists all paths that should be ignored when copying context data to a Dockerimage. I'll check this while handling the next step.

jmaupetit:
@sampaccoud ping me here when you've merged your current PRs.

jmaupetit:
@mbenadda I thought I would pursue your work here, no?

PR #475, opened 2019-02-07
Allow building category trees

sampaccoud:
## Purpose

We want to allow building category taxonomies in the CMS:
- multiple types of categories e.g. `Subject` versus `Level`
- nested categories e.g. `subject / maths / algebra` versus `level / expert`

## Proposal

We propose to use the DjangoCMS page tree to build the taxonomy. Everything is already there to do it... so we just had to:

- [x] refactor the demo site generator to generate an example of category taxonomy,
- [x] limit to tree leaves the categories that can be linked via a plugin,
- [x] improve `str` display of a category to show the whole ancestors tree,
- [x] add support for category taxonomies in the course search index (filtering on a parent of the category that is linked to the course should still select the course),
- [x] add support for taxonomy to the category search index (add tree information to allow the frontend to display category filters: find what are the meta categories, what are the first level categories in each meta category, what are the children of each first level category, etc...)

Some side clean-up was made in the process:

- [x] rename `public` to `published` in tests to make a clear difference between a `published draft object` and its `public` sibling,
- [x] use the page ID instead of the page extension ID for search indexes as they are much easier to get when retrieving a categories ancestors,
- [x] improve recursive page creation in the demo site generator.

### DIFF
--- a/src/richie/apps/core/management/commands/create_demo_site.py
+++ b/src/richie/apps/core/management/commands/create_demo_site.py
@@ -39,17 +39,17 @@ NB_PERSONS = 10
 NB_CATEGORIES = 8
 PAGE_INFOS = {
     "home": {
-        "content": {"en": "Home", "fr": "Accueil"},
+        "title": {"en": "Home", "fr": "Accueil"},
         "in_navigation": False,
         "kwargs": {"template": "richie/fullwidth.html"},
     },
     "news": {
-        "content": {"en": "News", "fr": "Actualités"},
+        "title": {"en": "News", "fr": "Actualités"},
         "in_navigation": True,
         "kwargs": {"template": "richie/fullwidth.html"},
     },
     "courses": {
-        "content": {"en": "Courses", "fr": "Cours"},
+        "title": {"en": "Courses", "fr": "Cours"},
         "in_navigation": True,
         "kwargs": {
             "reverse_id": Course.ROOT_REVERSE_ID,
@@ -57,7 +57,7 @@ PAGE_INFOS = {
         },
     },
     "categories": {
-        "content": {"en": "Categories", "fr": "Catégories"},
+        "title": {"en": "Categories", "fr": "Catégories"},
         "in_navigation": False,
         "kwargs": {
             "reverse_id": Category.ROOT_REVERSE_ID,
@@ -65,7 +65,7 @@ PAGE_INFOS = {
         },
     },
     "organizations": {
-        "content": {"en": "Organizations", "fr": "Etablissements"},
+        "title": {"en": "Organizations", "fr": "Etablissements"},
         "in_navigation": True,
         "kwargs": {
             "reverse_id": Organization.ROOT_REVERSE_ID,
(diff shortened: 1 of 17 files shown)


src/richie/apps/courses/helpers.py
```diff
@@ -0,0 +1,55 @@
...
+                    ],
+                },
+                {"title": "Level"},
+            ]
+
+        page (cms.models.pagemodel.Page): Instance of a Page below which the category
+            tree is created.
+
+    Returns:
+        generator[courses.models.Category]: yield only the leaf categories of the created tree.
```
mbenadda: I'm not sure we want to only be tagging leaf categories on courses. I would think we should just exclude the top-level category (meta) and allow tagging with all others.

For example because the taxonomy might not always be 100% complete for everything that exists anywhere. Otherwise we might force users to create `Misc.` children which would not be a great look on the courses that bear it.

src/richie/apps/core/management/commands/create_demo_site.py
```diff
@@ -102,6 +102,86 @@
...
+        "title": {"en": "Level", "fr": "Niveau"},
+        "children": [
+            {"title": {"en": "Beginner", "fr": "Débutant"}},
+            {"title": {"en": "Advanced", "fr": "Avancé"}},
+            {"title": {"en": "Expert", "fr": "Expert"}},
+        ],
+    },
+    {
+        "title": {"en": "Subject", "fr": "Subjet"},
+        "children": [
```
mbenadda: I think it might indeed make more sense to have a more general and uncontroversial taxonomy for the demo site.

src/richie/apps/courses/models/category.py
```diff
@@ -83,12 +83,22 @@ class CategoryPluginModel(PagePluginMixin, CMSPlugin):
     """
     Category plugin model handles the relation between CategoryPlugin
     and their Category instance
+
+    We only propose to link category objects:
+    - that are a leaf in the category tree,
```
mbenadda: Same as in the demo site, I don't really think this is the best idea

src/richie/apps/courses/models/course.py
```diff
@@ -121,6 +121,33 @@ def get_categories(self):
             .distinct()
         )
 
+    def get_root_to_leaf_category_pages(self):
+        """
+        Build a query retrieving all pages linked to the category or one of its ancestors,
+        excluding the meta category itself.
+
+        This is usefull to build the course Elasticsearch index. When category is linked to a
```
mbenadda: useful*

src/richie/apps/courses/models/course.py
```diff
@@ -121,6 +121,33 @@ def get_categories(self):
...
+        course, we associate all its ancestors (excluding the meta category) to the course so
+        that a search for courses in an ancestor category also returns the course.
+        """
+        category_query = self.get_categories()
+
+        # 1. We want the pages related to a category directly of the course
```
mbenadda: directly related to a category?

src/richie/apps/courses/helpers.py
sampaccoud: good question... I have the feeling that this may depend on the plateform operator... Maybe we could condition this extra filtering to a setting? WDYT @2018D?

sampaccoud: I propose to make the limit filter in models/course.py optional but leave the demo site as it is ie only leaf categories linked to courses.

src/richie/apps/courses/models/category.py
sampaccoud: see explanation above.

sampaccoud:
@mbenadda ready for review. I took your comments into account in 3 fixup commits.

PR #477, opened 2019-02-08
♻️(search) refactor languages into a dynamic filter

mbenadda:
## Purpose

We first implemented the languages filter as a hardcoded filter with predefined choices in `defaults.py`.

This was not perfect as it prevented us from using all the possible language choices from `settings.ALL_LANGUAGES`.

## Proposal

Instead, we can use a simple "terms" in ElasticSearch and treat the field exactly like we already do categories and organizations. 

The new `FILTERS_DYNAMIC` name also reflects the fact that it's not just resources.

### DIFF
--- a/tests/apps/search/test_indexers_courses.py
+++ b/tests/apps/search/test_indexers_courses.py
@@ -263,11 +263,12 @@ class CoursesIndexersTestCase(TestCase):
     @patch(
         "richie.apps.search.indexers.courses.FILTERS_HARDCODED",
         new={
-            "language": {
+            "availability": {
                 "choices": {
-                    lang: [{"term": {"language": lang}}] for lang in ["en", "fr"]
+                    choice: [{"term": {"availability": choice}}]
+                    for choice in ["coming_soon", "current"]
                 },
-                "field": MultipleChoiceField,
+                "field": ChoiceField,
             }
         },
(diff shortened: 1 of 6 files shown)


tests/apps/search/test_indexers_courses.py
```diff
@@ -747,10 +825,20 @@ def test_indexers_courses_build_es_query_combined_search(self):
         request = SimpleNamespace(
             query_params=QueryDict(
                 query_string="categories=42&categories=84&query=these%20phrase%20terms&limit=2&"
-                + "language=fr&"
+                + "languages=fr&availability=current&"
                 + "start={start}&end={end}".format(start=start, end=end)
             )
         )
+
+        # Search fragments that are repeated in the query
```
mbenadda: this is fine

iu

tests/apps/search/test_indexers_courses.py
```diff
@@ -263,11 +263,12 @@ def test_indexers_courses_format_es_object_for_api(self):
     @patch(
         "richie.apps.search.indexers.courses.FILTERS_HARDCODED",
         new={
-            "language": {
+            "availability": {
```
sampaccoud: I'll just trust you on this file's diff!

PR #478, opened 2019-02-08
🐛(jobs) fix circle CI upload string with new file structure

mbenadda:
## Purpose

As we refactored the app to move frontend related files into src/frontend, we inadvertently broke the strings upload script run by the CI. We did not notice it before as it only runs on master, not on feature branches.

## Proposal

To fix this, we decided to improve the file structure there too, and properly manage the dependencies in a package.json (as opposed to piggy-backing onto the frontend's dependencies).


jmaupetit:
Nice one!

PR #482, opened 2019-02-11
🐛(course) update fragment_organization_main_logo.html

2018D:
## Purpose

Repair a problem when adding a new course page from scratch.

## Proposal

In the absence of an organization, don't try and show anything as the main organization logo.

### DIFF
--- a/src/richie/apps/courses/templates/courses/cms/fragment_organization_main_logo.html
+++ b/src/richie/apps/courses/templates/courses/cms/fragment_organization_main_logo.html
@@ -1,17 +1,19 @@
 {% load cms_tags static %}{% spaceless %}
-{% if current_page.publisher_is_draft %}
-    {% if organization.check_publication is True %}
+{% if organization %}
+    {% if current_page.publisher_is_draft %}
+        {% if organization.check_publication is True %}
+            {% show_placeholder "logo" organization.public_extension.extended_object as logo %}
+        {% else %}
+            {% show_placeholder "logo" organization.extended_object as logo %}
+        {% endif %}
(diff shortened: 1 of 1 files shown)


src/richie/apps/courses/templates/courses/cms/fragment_organization_main_logo.html
```diff
@@ -1,17 +1,19 @@
 {% load cms_tags static %}{% spaceless %}
-{% if current_page.publisher_is_draft %}
-    {% if organization.check_publication is True %}
+{% if organization %}
```
sampaccoud: Have you thought of putting this "if" in the course_detail.html template to wrap the call to this fragment?

2018D: Not really... is it better in your opinion ?

2018D: i feel that because the fragment is called this way:
`{% include "courses/cms/fragment_organization_main_logo.html" with organization=course.get_main_organization %}`
that the check has to be done after. But then again we could call `if course.get_main_organization` before the include.

sampaccoud: Yeah you're right. By the way I think the main organization will never by null because it is a required field in the creation wizard. WDYT?

2018D: It's just in case someone goes and delete the organization plugin. Especially if there's only one.

PR #486, opened 2019-02-12
Surface courses with categories/organizations even in full-text searches

mbenadda:
## Purpose

We need to be able to surface courses with organization X or category Y when a user does a full-text search for X or Y.
Currently, this does not happen if X and/or Y do not appear in the syllabus or title of the course.

## Proposal

- [x] Index categories & organization names on courses
- [x] Add a query element to the full text search fragment to all search the text in these new fields

### DIFF
--- a/src/richie/apps/search/indexers/courses.py
+++ b/src/richie/apps/search/indexers/courses.py
@@ -77,6 +77,14 @@ class CoursesIndexer:
         "title.*",
     ]
 
+    # Define the scoring boost (in ElasticSearch) related value names receive when using
+    # full-text search.
+    # For example, when a user searches for "Science" in full-text, it should match any
+    # course whose category contains "Science" or a related word, albeit with a lower
+    # score than courses that include it in their title or description.
+    # This lower score factor is the boost value we get or set here.
+    fulltext_search_filter_matching_boost = 0.05
+
     scripts = {
         # The ordering process first splits the courses into four groups, with further ordering
         # inside each one of those groups.
@@ -308,9 +316,9 @@ class CoursesIndexer:
             "title": get_best_field_language(source["title"], best_language),
         }
 
-    @staticmethod
+    @classmethod
     # pylint: disable=R0912, R0914
-    def build_es_query(request):
+    def build_es_query(cls, request):
         """
         Build an ElasticSearch query and its related aggregations, to be consumed by the ES client
         in the Courses ViewSet
@@ -385,10 +393,27 @@ class CoursesIndexer:
                         param,
                         [
                             {
-                                "multi_match": {
-                                    "fields": ["description.*", "title.*"],
-                                    "query": value,
-                                    "type": "cross_fields",
+                                "bool": {
+                                    "should": [
+                                        {
+                                            "multi_match": {
+                                                "fields": ["description.*", "title.*"],
+                                                "query": value,
+                                                "type": "cross_fields",
+                                            }
+                                        },
+                                        {
+                                            "multi_match": {
+                                                "boost": cls.fulltext_search_filter_matching_boost,
+                                                "fields": [
+                                                    "categories_names.*",
+                                                    "organizations_names.*",
+                                                ],
+                                                "query": value,
+                                                "type": "cross_fields",
+                                            }
+                                        },
+                                    ]
                                 }
                             }
                         ],
--- a/src/richie/apps/search/indexers/courses.py
+++ b/src/richie/apps/search/indexers/courses.py
@@ -3,6 +3,7 @@ ElasticSearch course document management utilities
 """
 from collections import defaultdict, namedtuple
 from datetime import MAXYEAR
+from functools import reduce
 
 from django.conf import settings
 from django.db.models import Prefetch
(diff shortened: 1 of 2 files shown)


src/richie/apps/search/indexers/courses.py
```diff
@@ -199,6 +201,26 @@ def get_data_for_es(cls, index, action):
...
+                        lambda acc, title: {
+                            **acc,
+                            title["language"]: acc[title["language"]] + [title["title"]]
+                            if acc.get(title["language"])
+                            else [title["title"]],
+                        },
+                        [
+                            title
+                            for category in categories
+                            for title in category.extended_object.title_set.values()
```
sampaccoud: This would also publish titles that are unpublished... See my proposal above.

src/richie/apps/search/indexers/courses.py
```diff
@@ -184,8 +185,9 @@ def get_data_for_es(cls, index, action):
                 )
 
             course_runs = course.get_course_runs()
+            categories = course.get_categories()
```
sampaccoud: I think you should retrieve the related published `title` objects and the `public_extention_id` in the same query. It should look something like this:
```suggestion
from django.db.models import Prefetch
...
categories = course.get_categories().prefetch_related(
    Prefetch(
        "extended_object__title_set",
        to_attr="published_titles",
        queryset=Title.objects.filter(published=True).only("language", "title"),
    )
)
.only("public_extension", "published_titles")
.distinct()
```

See example usage here: https://github.com/openfun/richie/blob/master/src/richie/apps/search/indexers/categories.py#L61

src/richie/apps/search/indexers/courses.py
```diff
@@ -211,18 +233,26 @@ def get_data_for_es(cls, index, action):
...
                         )
                         if id is not None
                     ],
+                    # Index the names of organizations to surface them in full text searches
+                    "organizations_names": reduce(
+                        lambda acc, title: {
+                            **acc,
+                            title["language"]: acc[title["language"]] + [title["title"]]
+                            if acc.get(title["language"])
+                            else [title["title"]],
```
sampaccoud: You don't need to do this:
```
>>> from collections import defaultdict
>>> d = defaultdict(list)
>>> d['a'] + ['b']
['b']
```

mbenadda:
@sampaccoud ready for review

src/richie/apps/search/indexers/courses.py
```diff
@@ -199,6 +234,24 @@ def get_data_for_es(cls, index, action):
                         language: course_run.extended_object.get_absolute_url(language)
                         for language in titles.keys()
                     },
+                    "categories": [
+                        str(pk) for pk in categories.values_list("pk", flat=True)
```
sampaccoud: Now that you call only, we use only, we should stick to the instance:
```suggestion
                        str(category.pk) for category in categories
```

src/richie/apps/search/indexers/courses.py
```diff
@@ -216,12 +269,21 @@ def get_data_for_es(cls, index, action):
...
+                        lambda acc, title: {
+                            **acc,
+                            title["language"]: acc[title["language"]] + [title["title"]]
+                            if acc.get(title["language"])
+                            else [title["title"]],
+                        },
+                        [
+                            title
+                            for organization in organizations
+                            for title in organization.extended_object.title_set.values()
```
sampaccoud:
```suggestion
                            for title in organization.published_titles
```

src/richie/apps/search/indexers/courses.py
```diff
@@ -183,9 +192,35 @@ def get_data_for_es(cls, index, action):
                     simple_text.body
                 )
 
+            # Make sure we get title information for categories in the same request
+            categories = (
```
sampaccoud: this is not retrieving categories but pages related to categories. So maybe?
```suggestion
            category_pages = (
```

src/richie/apps/search/indexers/courses.py
```diff
@@ -183,9 +192,35 @@ def get_data_for_es(cls, index, action):
...
+            organizations = (
+                course.get_organizations()
+                .prefetch_related(
+                    Prefetch(
+                        "extended_object__title_set",
+                        to_attr="published_titles",
+                        queryset=Title.objects.filter(published=True),
+                    )
+                )
+                .only("public_extension", "extended_object")
```
sampaccoud: I think you can remove `public_extension` because the code below to get the list of organization ids is making another query anyway.

src/richie/apps/search/indexers/courses.py
```diff
@@ -199,6 +234,22 @@ def get_data_for_es(cls, index, action):
                         language: course_run.extended_object.get_absolute_url(language)
                         for language in titles.keys()
                     },
+                    "categories": [str(category.pk) for category in category_pages],
```
sampaccoud:
```suggestion
                    "categories": [str(page.pk) for page in category_pages],
```

src/richie/apps/search/indexers/courses.py
```diff
@@ -199,6 +234,22 @@ def get_data_for_es(cls, index, action):
...
+                    "categories_names": reduce(
+                        lambda acc, title: {
+                            **acc,
+                            title.language: acc[title.language] + [title.title]
+                            if acc.get(title.language)
+                            else [title.title],
+                        },
+                        [
+                            title
+                            for category in category_pages
```
sampaccoud: page
```suggestion
                            for page in category_pages
```

src/richie/apps/search/indexers/courses.py
```diff
@@ -199,6 +234,22 @@ def get_data_for_es(cls, index, action):
...
+                            for title in category.published_titles
```
sampaccoud:
```suggestion
                            for title in page.published_titles
```

PR #494, opened 2019-02-18
🐛(person) fix Person.get_full_name when person_title is empty

sveetch:
## Purpose

"get_full_name" method expected "person_title" was allways filled
although this field is not required resulting in an AttributeError
on page which display full name (at least the course detail).

## Proposal

Method has been fixed to support empty "person_title" and a test
has been modified to check this case.

### DIFF
--- a/src/richie/apps/persons/models.py
+++ b/src/richie/apps/persons/models.py
@@ -110,8 +110,12 @@ class Person(BasePageExtension):
         """
         Return person's full name
         """
-        return "{person_title} {first_name} {last_name}".format(
-            person_title=self.person_title.title,
+        person_title = (
+            "{:s} ".format(self.person_title.title) if self.person_title else ""
+        )
+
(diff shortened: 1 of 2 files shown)


src/richie/apps/persons/models.py
```diff
@@ -110,10 +110,14 @@ def get_full_name(self):
         """
         Return person's full name
         """
-        return "{person_title} {first_name} {last_name}".format(
-            person_title=self.person_title.title,
-            first_name=self.first_name,
-            last_name=self.last_name,
+        if self.person_title:
+            return "{person_title} {first_name} {last_name}".format(
+                person_title=self.person_title.title,
```
sampaccoud: Maybe avoid repeating ourselves with:
```suggestion
                person_title="{:s} ".format(self.person_title.title) if self.person_title else "",
```

Issue #476, opened 2019-02-08
Generate the filters side bar dynamically


sampaccoud:
## Feature Request

**Is your feature request related to a problem or unsupported use case? Please describe.**
PR https://github.com/openfun/richie/pull/475 introduces configurable nested categories.

The next feature we need to land is to build the filter side panel dynamically in the frontend according to what filters are configured in the backend.

**Describe the solution you'd like**
We brainstormed and propose to open an endpoint on the backend that exposes a precise definition of how the filters are configured. It could look like this:

```
GET /api/v1.0/filters

[
    {
        # hardcoded like in richie/apps/search/defaults.py
        name: "new",
        # This would make a filter without header (just a checkbox `new` here)
        human_name: None, 
        is_drill_down: true,
        sort_mode: tree, # Respect the sorting defined in the CMS page tree
        values: [ 
            {
                name: new,
                human_name: Nouveauté,
                children: []                
            },
        ],
    },
    {
        # hardcoded like in richie/apps/search/defaults.py
        name: "availability",
        human_name: "Disponibilité",
        is_drill_down: true,
        sort_mode: tree,
        values: [
            {
                name: coming_soon,
                human_name: A venir,
                children: []                
            },
            {
                name: current,
                human_name: Commencé,
                children: []                
            },
            {
                name: open,
                human_name: Ouvert à l'inscription,
                children: []                
            },
    },
    {
        # retrieved from categories page tree
        name: "subjects",
        human_name: "Subjets",
        is_drill_down: false,
        sort_mode: count, # Sort by the number of results for each facet
        values: [
            {
                name: mathematics,
                human_name: Mathématiques,
                children: [
                    {
                        name: algebra,
                        human_name: Algèbre,
                        children: [],
                    },
                    {
                        name: geometry,
                        human_name: Géométrie,
                        children: [],
                    },
                    ...
                ]                
            },
            {
                name: arts,
                human_name: Arts,
                children: [
                    {
                        name: literature,
                        human_name: Littérature,
                        children: []
                    },
                    {
                        name: painting,
                        human_name: Peinture,
                        children: []
                    },
                    ...
                ]
            },
            ...
        ],
    },
    {
        # retrieved from categories page tree
        name: "levels",
        human_name: "Niveau",
        is_drill_down: yes,
        sort_mode: tree,
        values: [
            {
                name: beginner,
                human_name: Beginner,
                children: []
            },
            {
                name: advanced,
                human_name: Avancé,
                children: []
            },
            {
[...]

Issue #454, opened 2019-01-29
Deduplicate course results in autosuggest


sampaccoud:
## Bug Report

**Expected behavior/code**
Course suggestions in autocomplete should not be duplicated.

**Actual Behavior**
We are indexing course runs since PR https://github.com/openfun/richie/pull/443. It means that we now have several records for a given course (as many as there are course runs). When the title of a course matches a search query, the course runs are proposed one after the other (because they match just as much as their sibling...).

**Steps to Reproduce**
1. Go to the search page http://localhost:8070/en/courses/
2. Start typing the name of one of the courses you see in results and that has several course runs (just try one if you're not sure)
3. Notice that the autocomplete suggestions include all the course runs of the course one above the other. Duplicates...

**Possible Solution**
Deduplicate results server side (in the autocomplete view for courses?).

**Alternative**
Deduplicate client side...

sampaccoud:
This issue is fixed by PR https://github.com/openfun/richie/pull/496. Indexing course runs presented too many drawbacks that could not be solved satisfactorily (duplicate results that can' be solved withouy raising pagination issues). 

We decided to go back to indexing courses. Course runs are nox indexed as a nested field. Filtering on nested fields is much harder but we managed to do what we needed to do thanks to a painless script. Up to now this is the best solution we found.

Issue #425, opened 2018-11-19
Add a field to indicate the languages in which a course is available.


sampaccoud:
## Feature Request

**Is your feature request related to a problem or unsupported use case? Please describe.**
We need to be able to indicate, on the course syllabus, the languages in which the course content is available.

**Describe the solution you'd like**
I think this field should be on the course run object because each course run could be available in different languages. In particular, edX does not support multilingual course content, so the way to do multilingual courses could be to open several course runs, each one for a specific language.

Technically, the `languages` field could be a Django `ArrayField` of `CharField` with a list of ISO 639 languages as choices.

**Discovery, Documentation, Adoption, Migration Strategy**

- [x] The languages field should be required to create a course run,
- [x] We should add a field to select multiple languages on the CMS wizard form to create new course runs,
- [x] The list of languages should be displayed and editable on the course pages and on the course run pages,
- [x] ~Language should be a filter on the search page. It should be linked to the dates of the related course run so that when we filter by a language and a status, we only see the courses that have a course in this status for the selected language.~ The language filter on the search page should be made a facet.

**Do you want to work on it through a Pull Request?**
:+1:

mbenadda:
This seems like a fine solution to me 👍.

We'll just have to be careful to return the "most correct" course run per language & display it in the "most correct" language available when we generate search results for courses with a language filter.

robertgerinlajoie:
Hi, good point. Some observations.

1- I take for granted an equivalent search page like the edX one, see https://www.edx.org/course/?, and where we can search for course langage and other fields.

2- to be more precise, we should use internally only the ISO-639-1, the 2 letters code, and not mention the country specific locale. And as a parameter for Ritchie admin, we should be able to specify the preferred suggested langage (french, english, ...) in front of the complete langage list;

sampaccoud:
Hi @robertgerinlajoie! 
Some elements of answer:

1. yes, that's what we have in mind. Contrary to the search engine you link, we will be able to filter on one but also on several languages,
2. Whether you use ISO 639-1 two-letter codes or country+language locales is up-to-you in Django. It depends on what you declare in the `LANGUAGES` setting. The preferred language in Django is also configurable via the `LANGUAGE_CODE` setting and could be either a two-letter language or a full locale if I'm not mistaken.

[...]

PR #496, opened 2019-02-20
Switch to indexing courses instead of course runs

sampaccoud:
## Purpose

Indexing course runs lead to a lot of duplicate content on the frontend.

When searching for a course with its name, we were most of the time seeing all the course runs of the same course side by side. 

## Proposal

- [x] Index courses instead of course runs,
- [x] index course runs as "nested" fields in a course document,
- [x] rewrite filter queries on availability,
- [x] rewrite filter queries on languages,
- [x] rewrite sorting script in [Painless](https://www.elastic.co/guide/en/elasticsearch/reference/master/modules-scripting-painless.html),
- [x] take into account eventual availability filters applied in sorting script,
- [x] take into account level of availability filtering in the sorting script,
- [x] take into account eventual language filters applied in the sorting script,
- [x] compute course state as a script field on the basis of related course runs,
- [x] refactor organization and categories viewsets/indexers/forms to be coherent with courses.

### Related issues

Fixes https://github.com/openfun/richie/issues/507
Fixes https://github.com/openfun/richie/issues/476
Fixes https://github.com/openfun/richie/issues/454
Fixes https://github.com/openfun/richie/issues/425


2018D:
it's showing all cms available language now instead of course available language only
Screenshot_2019-03-08 Cours

sampaccoud:
@2018D It's normal that course runs can be of any language, not only the languages in which the site is translated. What is not really realistic is that the demo picks random languages in the whole set of possible languages.

I pushed the following improvements:
- limit number of languages in course runs of the demo site,
- filter out language facets with a count of 0,
- sort language facets by the number of results,
- sort filters according to the position defined in settings (languages should come last).

Can you test again and tell me how I can still improve this PR?

2018D:
"filter-out language facets with a count of 0" was the missing item for me.

Here's some new feedback:

- The textual search box works well.👍

- However the filtering menu on the left doesn't update the available course list. It seems to have the data and the menu reacts when using it, but the courses showing up aren't updating. 

- The fact that you can only filter one item per filter sections is now very obvious 👍 , however it doesn't tell you which item you've selected. All items of the section are marked with an x. It could benefit from hidding unselected item like on edx.org. They add a selected tag.
Screenshot_2019-03-11 Courses

2018D:
On the course card. When it says on-going, It's also indicating the min date:
Screenshot_2019-03-14 Courses
No date would be better in this case. That or on-going until [end] date.

sampaccoud:
@mbenadda the point raised by @2018D seems to be frontend related. To go further, we need to uniformize the way this date/text/CTA is displayed with the course glimpse rendered in the CMS pages. Looks like a job for you :wink:

2018D:
I think it's due to the use of {None} according to comment at src/richie/apps/courses/models/course.py line 58. The none is interpreted as min date by the frontend.

mbenadda:
Oh nice catch @2018D! Is `course.state.datetime` nullable @sampaccoud?

sampaccoud:
@mbenadda Yes: course.state.datetime is null for states that don't need to display a date (see https://github.com/openfun/richie/pull/496/files#diff-130a37291b0a06e9aee2f7bfd45f1460R25). See the CMS course glimpses for examples.

src/richie/apps/search/filter_definitions/base.py
```diff
@@ -0,0 +1,382 @@
...
+                    {"count": count, "human_name": values[name], "name": name}
+                    for name, count in facet_counts
+                ],
+            }
+        }
+
+
+class NestingWrapper(BaseFilterDefinition):
+    """
+    A filter definition to aggregate children filter definitions under a nested query.
```
mbenadda: ... This is necessary to use filters and aggregations on values that live in nested fields. Such as any information that is linked not directly to courses but to course runs (a nested field of courses).

src/richie/apps/search/filter_definitions/base.py
```diff
@@ -0,0 +1,382 @@
...
+            },
+            self.filter_definitions.values(),
+            {},
+        )
+
+    def get_query_fragment(self, data):
+        """
+        Aggregate query fragments from the nested children in a list of mappings.
+
+        Returns:
```
mbenadda: `--------`

src/richie/apps/search/filter_definitions/base.py
```diff
@@ -0,0 +1,382 @@
...
+                        {"nested": {"path": self.path, "query": nested_query}}
+                    ],
+                }
+            ]
+            if queries
+            else []
+        )
+
+    # pylint: disable=arguments-differ
+    def get_aggs_fragment(self, queries, data, *args, **kwargs):
```
mbenadda: No docstring? Is this not a pylint error?

tests/apps/search/test_query_courses.py
```diff
@@ -0,0 +1,612 @@
+"""
```
mbenadda: This is the test suite that cannot be ran on its own `bin/pytest tests/apps/search/test_query_courses.py`

tests/apps/search/test_elasticsearch_features.py
```diff
@@ -1,373 +0,0 @@
-"""
```
mbenadda: 👍

src/richie/apps/search/indexers/courses.py
```diff
@@ -69,8 +70,17 @@ class CoursesIndexer:
     form = CourseSearchForm
 
     scripts = {
+        # We use the same `state` script for several use cases as the code is mostly common and
```
mbenadda: You have probably thought of this but just in case

To me those scripts look like python strings. So maybe you could simply concatenate the common part with the 2 different parts that handle the return values to make 2 different scripts without repeating the common part?

src/richie/apps/search/indexers/courses.py
```diff
@@ -79,33 +78,49 @@ class CoursesIndexer:
...
+        # -------------- NOW (current timestamp) --------------
+        #     Bucket 0: Courses that are ongoing and open
+        #      sorted by ascending end of enrollment date
+        #         < 1 x datetime.MAXYEAR distance >
         #
-        #  < ~1 x datetime.MAXYEAR distance >
+        # --------------   1 x datetime.MAXYEAR  --------------
+        #      Bucket 1: Courses that are future and open
+        #          sorted by ascending start date
+        #         < 2 x datetime.MAXYEAR distance >
```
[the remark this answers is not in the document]
sampaccoud: Let's remove them then.

src/richie/apps/search/indexers/courses.py
sampaccoud: Yes I thought of it but it would make the script very difficult to read/understand? The performance loss appears negligeable (just one more if... a question of nano seconds?).

mbenadda: I made the suggestion more as a way to make this easier to understand than out of performance considerations.

Anyway, there's no clean answer, it was just a suggestion, you can leave it as it is 👍

src/richie/apps/search/indexers/courses.py
```diff
@@ -439,7 +544,7 @@ def build_es_query(cls, request, filters):
         )
 
     @staticmethod
-    def get_list_sorting_script():
+    def get_list_sorting_script(states=None):
```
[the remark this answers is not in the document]
sampaccoud: A Python anti-pattern. Lists (like dictionaries and others) are passed as reference and would be instantiated when the app starts if place as argument default. The list would then be muted each time the function is called. Not what we want.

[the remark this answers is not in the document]
mbenadda: I did not know this! As a usually-javascript developer, this feels counter-intuitive to me as it is the opposite behavior of JS which would give you a brand new list on each invocation.

Cheers for the explanation 👍

src/richie/apps/search/filter_definitions/base.py
sampaccoud: maybe because of the pylint rule above? anyway, let me add a docstring...

tests/apps/search/test_indexers_courses.py
```diff
@@ -1033,145 +1006,3 @@ def test_indexers_courses_build_es_query_search_with_invalid_params(self):
                 SimpleNamespace(query_params=QueryDict(query_string="limit=-2")),
                 DEFAULT_FILTERS,
             )
-
-    def test_indexers_courses_list_sorting_script(self):
```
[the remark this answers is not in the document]
sampaccoud: Yes. I battle test it in tests/apps/search/test_query_courses.py with some randomness to check that sorting works as expected.

src/richie/apps/search/indexers/courses.py
sampaccoud: I would cut the algo in 2 different places and make it harder to follow IMO.

tests/apps/search/test_query_courses.py
sampaccoud: Ok there was some incoherence in imports. I refactored a bit more :recycle:

sampaccoud:
@2018D I have added a commit that solves Issue https://github.com/openfun/richie/issues/507 both in the model and in the search index: https://github.com/openfun/richie/pull/496/commits/8284c72ae5956ae5b88b199e8221c0db307a5927.

Please test it and let us know.

PR #497, opened 2019-02-21
✨(plugins) add 'template' and 'content' field to LargeBanner plugin

sveetch:
## Purpose

LargeBanner plugin currently lack of features to be really useful in
many cases.

## Proposal

This commit add two new fields to cover much more usage case. A
'template' field to choose another template to render plugin, alike
with page templates, plugin template choices is defined from
settings. And a 'content' field which is represented by a CKeditor
instance to be able to include some markup.


sandbox/settings.py
```diff
@@ -437,6 +437,13 @@ class Base(DRFMixin, ElasticSearchMixin, Configuration):
     # CKEditor widgets/fields
     CKEDITOR_SETTINGS["toolbar_HTMLField"] = CKEDITOR_SETTINGS["toolbar_CMS"]
 
+    # Available LargeBanner plugin template choices
+    # Careful! Don't remove items that are allready used in some placeholder
```
```suggestion
    # Careful! Don't remove items that are already used in some placeholder
```

src/richie/plugins/large_banner/choices.py
```diff
@@ -0,0 +1,18 @@
...
+Large banner plugin field choices
+"""
+from django.conf import settings
+
+
+def get_template_choices():
+    """
+    Return list of template choices.
+    """
+    return settings.RICHIE_LARGEBANNER_TEMPLATES
```
sampaccoud: I don't understand why you make these 2 functions.
I propose to follow the same pattern as here: https://github.com/openfun/richie/blob/master/src/richie/apps/courses/defaults.py

Which means:
- rename the file to defaults.py
- replace these functions by:

```python
    TEMPLATES = return getattr(settings, "RICHIE_LARGEBANNER_TEMPLATES", (
        "richie/large_banner/large_banner.html", _("Default")
    ) 
```

Use this variable in your model with:
```python
choices=TEMPLATES
default=TEMPLATES[0][0],
```

src/richie/plugins/large_banner/templates/richie/large_banner/large_banner.html
```diff
@@ -10,6 +10,11 @@
                     {% thumbnail instance.background_image 768x450 crop=',0' %} 768w"
             alt="banner"/>
     {% endif %}
-    <h1 class="large-banner__title">{{ instance.title }}</h1>
-    <img src="{% thumbnail instance.logo 200x100 crop='smart' %}" class="large-banner__logo" alt="{{ instance.logo_alt_text }}"/>
+    <div class="large-banner__body">
+        <h1 class="large-banner__title">{{ instance.title }}</h1>
+        <img src="{% thumbnail instance.logo 200x100 crop='smart' %}" class="large-banner__logo" alt="{{ instance.logo_alt_text }}"/>
```
sampaccoud: should we also make the logo optional and wrap it with {% if instance.logo %}?

src/richie/plugins/large_banner/forms.py
```diff
@@ -0,0 +1,30 @@
...
+    Plugin form used to fill its content from frontend admin.
+    """
+
+    class Meta:
+        """
+        Form meta attributes
+        """
+
+        model = LargeBanner
+        widgets = {"content": TextEditorWidget}
```
rmoch: We shall have a specific CKEditor configuration for this type of field (which could easily break visual design)

rmoch: sorry: TextEditorWidget is good if this is raw text

src/richie/plugins/large_banner/templates/richie/large_banner/large_banner.html
```diff
@@ -10,6 +10,11 @@
...
+        {% if instance.content %}
+            <div class="large-banner__content">{{ instance.content|safe }}</div>
```
sampaccoud: Don't mark it safe since we are using a raw text plugin and don't accept formatting?

src/richie/plugins/large_banner/forms.py
sveetch: CKEditor won't have any real meaning if it's just raw plain text since CKEditor produces paragraph tags they would be escaped (since @sampaccoud have mentioned to remove `safe` filter).

By the way, i thing raw plain text would be useless in this plugin, in real world you never want to include Hero content without basic inline tags.

I propose to indeed make a "inline tags only" (bold, italic, link and paragraph) CKEditor config with every other features removed, also remove the "Source" panel so user won't be able to pass through our rules. Also, i will watch to have an optional setting to choose other config, third user could so use more tags if he wants to from its own richie project.

sampaccoud: But we can apply some markup via the linked template no?

sveetch: Yes but markup applied from template is just around inputed content, not inside.

src/richie/plugins/large_banner/models.py
```diff
@@ -31,6 +34,21 @@ class LargeBanner(CMSPlugin):
...
+        blank=True,
     )
-    logo_alt_text = models.CharField(max_length=255)
+    logo_alt_text = models.CharField(max_length=255, blank=True)
+    template = models.CharField(
+        _("Template"),
+        max_length=150,
+        choices=LARGEBANNER_TEMPLATES,
+        default=LARGEBANNER_TEMPLATES[0][0],
+        help_text=_("Optional template for custom look."),
```
rmoch: this is not an optional field.
help_text=_("Choose template to render large banner plugin") ?

src/richie/plugins/large_banner/models.py
```diff
@@ -31,6 +34,21 @@ class LargeBanner(CMSPlugin):
...
+    )
+    content = models.TextField(_(u"Content"), blank=True, default="")
```
rmoch: Python 3, unicode everywhere ! 😄

src/richie/plugins/large_banner/templates/richie/large_banner/large_banner.html
rmoch: we are now using a very simple ckeditor widget

src/richie/plugins/large_banner/models.py
sveetch: Sorry, wrong copy/pasta

src/richie/plugins/large_banner/models.py
sveetch: It is not optional indeed, blank=False is implicit in django field, you want the explicit way ?

rmoch: I suggested to write more precise `help_text` string. The word `optional` seems wrong here

PR #498, opened 2019-02-22
✨(plugins) add 'template'field to Section plugin

sveetch:
## Purpose

Section plugin lacks of 'template' field to be able to have
different layout from plugin instance to another.

## Proposal

This commit add a new 'template' field to choose different template
to render plugin, plugin template choices is defined from
settings.


sandbox/settings.py
```diff
@@ -437,6 +443,11 @@ class Base(DRFMixin, ElasticSearchMixin, Configuration):
     # CKEditor widgets/fields
     CKEDITOR_SETTINGS["toolbar_HTMLField"] = CKEDITOR_SETTINGS["toolbar_CMS"]
 
+    # Available Section plugin template choices
+    # Careful! Don't remove items that are allready used in some placeholder
```
```suggestion
    # Careful! Don't remove items that are already used in some placeholder
```

src/richie/plugins/section/choices.py
```diff
@@ -0,0 +1,18 @@
+"""
```
sampaccoud: In fact I would make the same comments as on PR https://github.com/openfun/richie/pull/497 so let's discuss and agree on something for the other one first.

src/richie/plugins/section/defaults.py
```diff
@@ -0,0 +1,11 @@
+"""
+Section plugin field choices
```
sampaccoud: Section plugin default settings ?

PR #502, opened 2019-02-25
♻️(frontend) rework frontend search use filters in courses response

mbenadda:
## Purpose

Filter definitions are now directly available on the response object for the course search API request. This means it is no longer necessary to fetch disparate data and aggregate it into filter definitions in the frontend.

Instead, we simply have to show whatever is returned by the server's filter definitions and let users manipulate the query parameters through our interface. Changes to the query parameters in turn trigger new API requests, which yield both updated search results, and updated information for the filters.

To achieve this, we realized we did not need the whole redux + connected components machinery and could instead combine two custom hooks with some light context use to have the frontend work as a simple yet 100% dynamic system.

This let us cut the bundle size ~30% and remove most of the complex code from the frontend. We're now doing pure react with some raw JS helpers where needed.

## Proposal

- [x] Implement the two hooks `useCourseSearch` & `useCourseSearchParams` that together handle all our search request logic;
- [x] Use these two hooks (`courseSearch` through props drilling and `courseSearchParams` through a context) to power all our search & filtering needs;
- [x] Drop `redux` and all its related logic in connected components, `saga` and `connect`.

This is point 5 in #476.

NB: this is probably better reviewed by going through the files than looking at the diff as it redefines the architecture of the app.


mbenadda:
I forgot to mention, during this PR I also changed our testing renderer from `enzyme` to [`react-testing-library`](https://testing-library.com/react) for a few reasons:

1. `enzyme` is getting stale when it comes to more recent `react` features. Not just hooks (which were in alpha for 3 months before the release mid-february) but also `context` which was released more than 6 months ago;
2. `react-testing-library` pushes developers to test components close to the way actual users access them, with a strong emphasis on accessibility.

This switch, made opportune by issues with `enzyme` and new `react` features, is also a great chance to try out this newer testing library, especially with Richie's now reduced scope when it comes to the frontend.

lunika:
I can't build the project, I have the following error:

```
$ yarn build
yarn run v1.13.0
$ tsc --noEmit && webpack
../../node_modules/@types/react/index.d.ts:2683:14 - error TS2300: Duplicate identifier 'LibraryManagedAttributes'.

2683         type LibraryManagedAttributes<C, P> = C extends React.MemoExoticComponent<infer T> | React.LazyExoticComponent<infer T>
                  ~~~~~~~~~~~~~~~~~~~~~~~~

  node_modules/@types/react/index.d.ts:2746:14
    2746         type LibraryManagedAttributes<C, P> = C extends React.MemoExoticComponent<infer T> | React.LazyExoticComponent<infer T>
                      ~~~~~~~~~~~~~~~~~~~~~~~~
    'LibraryManagedAttributes' was also declared here.

node_modules/@types/react/index.d.ts:2746:14 - error TS2300: Duplicate identifier 'LibraryManagedAttributes'.

2746         type LibraryManagedAttributes<C, P> = C extends React.MemoExoticComponent<infer T> | React.LazyExoticComponent<infer T>
                  ~~~~~~~~~~~~~~~~~~~~~~~~

  ../../node_modules/@types/react/index.d.ts:2683:14
    2683         type LibraryManagedAttributes<C, P> = C extends React.MemoExoticComponent<infer T> | React.LazyExoticComponent<infer T>
                      ~~~~~~~~~~~~~~~~~~~~~~~~
    'LibraryManagedAttributes' was also declared here.


Found 2 errors.
```

mbenadda:
You seem to have two different versions of the react typings. This should not happen as we have a resolution in `package.json` specifically to avoid this ([see the lockfile](https://github.com/openfun/richie/blob/a962dbd83844dd3f49b72600962a8a68af05b483/src/frontend/yarn.lock#L716)).

Have you tried deleting `node_modules` and reinstalling from the lockfile?

lunika:
> You seem to have two different versions of the react typings. This should not happen as we have a resolution in `package.json` specifically to avoid this ([see the lockfile](https://github.com/openfun/richie/blob/a962dbd83844dd3f49b72600962a8a68af05b483/src/frontend/yarn.lock#L716)).
> 
> Have you tried deleting `node_modules` and reinstalling from the lockfile?

Yes I tried this:

- `rm -rf node_modules`
- `yarn install --frozen-lockfile`
- `yarn build`

I still have the same error.

lunika:
I found the problem. I had twice the same `node_modules`. One at the root of the project and the other one in `src/frontend`...

src/frontend/js/components/CourseGlimpse/CourseGlimpse.tsx
```diff
@@ -36,49 +36,49 @@ const messages = defineMessages({
...
+          src={course.cover_image}
+          alt={intl.formatMessage(messages.altText, {
+            courseTitle: course.title,
+          })}
+        />
+      </div>
+      <div className="course-glimpse__content">
+        <div className="course-glimpse__content__wrapper">
+          <p className="course-glimpse__content__title">{course.title}</p>
+          <p>{(organizationMain && organizationMain.title) || ''}</p>
```
lunika: This prop looks like unused, it is set to `null` in `CourseGlimpseList` component. Did you keep to use it later ?

src/frontend/js/components/SearchSuggestField/SearchSuggestField.tsx
```diff
@@ -140,125 +122,158 @@ export async function onSuggestionsFetchRequested(
...
-    let valueFromUrl: string;
-    try {
-      valueFromUrl = getSearchParam('query') || '';
-    } catch (e) {
-      // Do not break, default to empty
-      valueFromUrl = '';
-    }
-    this.state = { suggestions: [], value: valueFromUrl };
-  }
+export const SearchSuggestFieldBase = ({
```
lunika: To be honnest I don't see the interest here to move it to a functional component.

src/frontend/js/data/useCourseSearchParams/useCourseSearchParams.ts
```diff
@@ -0,0 +1,85 @@
...
+        [action.filter.name]: computeNewFilterValue(
+          courseSearchParams[action.filter.name],
+          {
+            action: (action.type === 'FILTER_ADD' && 'add') || 'remove',
+            isDrilldown: !!action.filter.is_drilldown,
+            payload: action.payload,
+          },
+        ),
+      };
+  }
```
lunika: You should maybe add a default case ? I know It should never happen but...

src/frontend/js/components/CourseGlimpse/CourseGlimpse.tsx
mbenadda: Yes. I think this probably should come as a field on the `course` now (as we discussed with @sampaccoud previously).

I did not want to go and do that while we're in the middle of a large refactor on the backend side to avoid conflicts.

src/frontend/js/components/SearchSuggestField/SearchSuggestField.tsx
mbenadda: The main point was to drop these ridiculous `whatever.bind()` calls. I think removing this from all those functions makes them somewhat clearer.

src/frontend/js/data/useCourseSearchParams/useCourseSearchParams.ts
mbenadda: I think if we add a default then we'll no longer be warned if we add new action types and forget to handle them in the reducer (as they'll be considered handled by the default).

At the same time action types are hardcoded at all the call sites (and probably not something we'd need manual typecasting to manage), so we can be pretty sure typescript will not let us dispatch something that is not one of the existing three. (and will force us to update everything if we add a new one)

src/frontend/js/components/Search/Search.tsx
```diff
@@ -1,37 +1,44 @@
...
+                : null
+            }
+          />
         </div>
-      </div>
-    );
-  }
-}
+        {courseSearchResponse &&
+        courseSearchResponse.status === requestStatus.SUCCESS ? (
```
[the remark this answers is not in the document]
mbenadda: I'd have to change the children component to accept `false` as well as `null` because `courseSearchResponse.status === requestStatus.SUCCESS` will return `false` not `null` if the status does not match

src/frontend/js/components/Search/Search.tsx
```diff
@@ -1,37 +1,44 @@
...
+          <div className="search__results">
+            <SearchSuggestField
```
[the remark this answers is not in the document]
mbenadda: Indeed this is not ideal.

The reason I included `SearchSuggestField` in my conditional block is that it is not functional until we have received the first response from the `courses` endpoint (as we don't even know which filters we can even suggest).

I think I can improve on this by setting a min-height on the filters pane and maybe a loader on the main area?

We can probably improve on this later on with "ghost" loading elements in both parts.

src/frontend/js/data/useCourseSearchParams/useCourseSearchParams.ts
lunika: If you want an error, throw an error  after the switch. This code should never be reached.

src/frontend/js/components/Search/Search.tsx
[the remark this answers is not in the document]
lunika: I don't understand, you can write this:

```
{courseSearchResponse &&
          courseSearchResponse.status === requestStatus.SUCCESS && (
            <div className="search__results">
              <SearchSuggestField
                filters={courseSearchResponse.content.filters}
              />
              <CourseGlimpseList
                courses={courseSearchResponse.content.objects}
              />{' '}
            </div>
          )}
```

And you don't have anymore the ternary rendering null

src/frontend/js/data/useCourseSearchParams/useCourseSearchParams.ts
mbenadda: I'd rather have a compile-time warning than an error. In my opinion this is one instance where we can just rely on the compiler

src/frontend/js/components/Search/Search.tsx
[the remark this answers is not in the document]
mbenadda: Oh right sorry, I thought you meant the one for the `SearchFiltersPane` `filters` prop.

Issue #506, opened 2019-02-28
add categorie plugin to course


2018D:
## Bug Report

**Expected behavior/code**
Add a categorie filtering only published category

**Actual Behavior**
The categories aren't filtered for only published categories. You can't add any because it tries to add both published and draft.

**Steps to Reproduce**
1. Add a new category to course

**Environment**
- Richie version: master using makefile script

**Possible Solution**
Filter the category when using the plugin the same way organization is filtered.

**Additional context/Screenshots**
```
Environment:

Request Method: POST
Request URL: http://localhost:8070/fr/admin/cms/page/add-plugin/?placeholder_id=2341&plugin_type=CategoryPlugin&cms_path=/fr/cours/le-plaisir-dinnover-plus-rapidement/?edit&plugin_language=fr

Django Version: 2.1.7
Python Version: 3.6.8
Installed Applications:
('djangocms_admin_style',
 'django.contrib.auth',
 'django.contrib.contenttypes',
 'django.contrib.sessions',
 'django.contrib.admin',
 'django.contrib.sites',
 'django.contrib.sitemaps',
 'django.contrib.staticfiles',
 'django.contrib.messages',
 'cms',
 'menus',
 'sekizai',
 'treebeard',
 'djangocms_text_ckeditor',
 'filer',
 'easy_thumbnails',
 'djangocms_column',
 'djangocms_link',
 'djangocms_style',
 'djangocms_snippet',
 'djangocms_googlemap',
 'djangocms_video',
 'djangocms_picture',
 'richie',
 'richie.apps.core',
 'richie.apps.courses',
 'richie.apps.persons',
 'richie.apps.search',
 'richie.plugins.large_banner',
 'richie.plugins.plain_text',
 'richie.plugins.section',
 'richie.plugins.simple_text_ckeditor',
 'dockerflow.django',
 'parler',
 'rest_framework')
Installed Middleware:
('cms.middleware.utils.ApphookReloadMiddleware',
 'django.contrib.sessions.middleware.SessionMiddleware',
 'django.middleware.csrf.CsrfViewMiddleware',
 'django.contrib.auth.middleware.AuthenticationMiddleware',
 'django.contrib.messages.middleware.MessageMiddleware',
 'django.middleware.locale.LocaleMiddleware',
 'django.middleware.common.CommonMiddleware',
 'django.middleware.clickjacking.XFrameOptionsMiddleware',
 'dockerflow.django.middleware.DockerflowMiddleware',
[...]

PR #509, opened 2019-03-01
🐛(category) fix duplicate category

2018D:
## Purpose

Make sure there's only one version of each category found by category plugin
Fix issue #506 

## Proposal

Add an extra filter

### DIFF
--- a/src/richie/apps/courses/models/category.py
+++ b/src/richie/apps/courses/models/category.py
@@ -121,10 +121,16 @@ class Category(BasePageExtension):
 def get_category_limit_choices_to():
     """Return a query limiting the categories proposed when creating a CategoryPlugin."""
     limit_choices_to = {
+        # Joining on `cms_pages` generate duplicates for categories that are under a parent page
+        # when this page exists both in draft and public versions. We need to exclude the
+        # parent public page to avoid this duplication with the first condition.
+        # The second condition makes sure the parent is a category.
+        # The third condition filters out public category.
+        # The fourth condition makes sure only categories show up.
+        "node__parent__cms_pages__publisher_is_draft": True,  # exclude category published parent
+        "node__parent__cms_pages__category__isnull": False,  # exclude meta categories
         "publisher_is_draft": True,  # plugins work with draft instances
         "category__isnull": False,  # limit to pages linked to a category object
-        "node__parent__cms_pages__category__isnull": False,  # exclude meta categories
-        "node__parent__cms_pages__publisher_is_draft": True,  # exclude published categories
     }
     # Limit to leaf categories only if active in settings (False by default)
     if getattr(settings, "LIMIT_PLUGIN_CATEGORIES_TO_LEAF", False):
--- a/src/richie/apps/courses/models/category.py
+++ b/src/richie/apps/courses/models/category.py
@@ -124,6 +124,7 @@ def get_category_limit_choices_to():
         "publisher_is_draft": True,  # plugins work with draft instances
         "category__isnull": False,  # limit to pages linked to a category object
         "node__parent__cms_pages__category__isnull": False,  # exclude meta categories
+        "node__parent__cms_pages__publisher_is_draft": True,  # exclude published categories
     }
     # Limit to leaf categories only if active in settings (False by default)
     if getattr(settings, "LIMIT_PLUGIN_CATEGORIES_TO_LEAF", False):
--- a/src/richie/apps/courses/models/course.py
+++ b/src/richie/apps/courses/models/course.py
@@ -433,16 +433,16 @@ class CoursePluginModel(PagePluginMixin, CMSPlugin):
         on_delete=models.CASCADE,
         related_name="course_plugins",
         limit_choices_to={
-            # There's a draft and a public course attached to the draft and
-            # the public parent. Without the first condition, 4 options are
-            # availables for each course. The second condition makes sure the
-            # parent is not a course. The third option filters out public
-            # course. The fourth option makes sure only courses show up.
-            # If there's is no parent, the course is filtered out.
-            "node__parent__cms_pages__publisher_is_draft": True,
-            "node__parent__cms_pages__course__isnull": True,
-            "publisher_is_draft": True,
-            "course__isnull": False,
+            # Joining on `cms_pages` generate duplicates for courses that are under a parent page
+            # when this page exists both in draft and public versions. We need to exclude the
+            # parent public page to avoid this duplication with the first condition.
+            # The second condition makes sure the parent is not a course to exclude snapshots.
+            # The third condition filters out public course.
+            # The fourth condition makes sure only courses show up.
+            "node__parent__cms_pages__publisher_is_draft": True,  # exclude course published parent
+            "node__parent__cms_pages__course__isnull": True,  # limit to course - no snapshot
+            "publisher_is_draft": True,  # plugins work with draft instances
+            "course__isnull": False,  # limit to pages linked to a course object
         },
     )
 
--- a/src/richie/apps/courses/models/organization.py
+++ b/src/richie/apps/courses/models/organization.py
@@ -132,7 +132,10 @@ class OrganizationPluginModel(PagePluginMixin, CMSPlugin):
         Page,
         on_delete=models.CASCADE,
         related_name="organization_plugins",
-        limit_choices_to={"publisher_is_draft": True, "organization__isnull": False},
+        limit_choices_to={
+            "publisher_is_draft": True,  # plugins work with draft instances
+            "organization__isnull": False,  # limit to pages linked to a course object
+        },
     )
 
     class Meta:
--- a/tests/apps/courses/test_cms_plugins_category.py
+++ b/tests/apps/courses/test_cms_plugins_category.py
@@ -22,6 +22,33 @@ class CategoryPluginTestCase(CMSTestCase):
     Test that CategoryPlugin correctly displays a Category's page placeholders content
     """
 
+    @staticmethod
+    def test_cms_plugins_category_unique():
+        """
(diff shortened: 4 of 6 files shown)


src/richie/apps/courses/models/category.py
```diff
@@ -86,6 +86,7 @@ def get_category_limit_choices_to():
         "publisher_is_draft": True,  # plugins work with draft instances
         "category__isnull": False,  # limit to pages linked to a category object
         "node__parent__cms_pages__category__isnull": False,  # exclude meta categories
+        "node__parent__cms_pages__publisher_is_draft": True,  # exclude published meta categories
```
sampaccoud:
```suggestion
        "node__parent__cms_pages__publisher_is_draft": True,  # exclude published categories
```

sampaccoud:
@2018D Can you add a test that fails before your fix and passes after?

2018D:
The code for the new test could be refactored to other context as well. Organization plugin, course plugin, course select, organization select and so on. I'll see if i can't add them in as well.

2018D:
Added a commit with more uniqueness test for the remaining courses plugin.

src/richie/apps/courses/models/course.py
```diff
@@ -346,16 +346,10 @@ class CoursePluginModel(PagePluginMixin, CMSPlugin):
...
-            # the public parent. Without the first condition, 4 options are
-            # availables for each course. The second condition makes sure the
-            # parent is not a course. The third option filters out public
-            # course. The fourth option makes sure only courses show up.
-            # If there's is no parent, the course is filtered out.
-            "node__parent__cms_pages__publisher_is_draft": True,
-            "node__parent__cms_pages__course__isnull": True,
-            "publisher_is_draft": True,
-            "course__isnull": False,
+            "node__parent__cms_pages__publisher_is_draft": True,  # exclude course published parent
```
sampaccoud: Not really, it's more:
```suggestion
            "node__parent__cms_pages__publisher_is_draft": True,  # course has published parent
```

src/richie/apps/courses/models/course.py
```diff
@@ -346,16 +346,10 @@ class CoursePluginModel(PagePluginMixin, CMSPlugin):
         on_delete=models.CASCADE,
         related_name="course_plugins",
         limit_choices_to={
-            # There's a draft and a public course attached to the draft and
-            # the public parent. Without the first condition, 4 options are
-            # availables for each course. The second condition makes sure the
-            # parent is not a course. The third option filters out public
-            # course. The fourth option makes sure only courses show up.
-            # If there's is no parent, the course is filtered out.
```
sampaccoud: don't remove this explanation. I had to re-read it in order to understand what we were doing here.
The small comments below can be additional.

You can improve it by replacing:
```
If there's is no parent, the course is filtered out.
```
by:
```
If there's is no published parent, the course is filtered out.
```

src/richie/apps/courses/models/organization.py
```diff
@@ -131,7 +131,10 @@ class OrganizationPluginModel(PagePluginMixin, CMSPlugin):
         Page,
         on_delete=models.CASCADE,
         related_name="organization_plugins",
-        limit_choices_to={"publisher_is_draft": True, "organization__isnull": False},
+        limit_choices_to={
+            "publisher_is_draft": True,  # plugins work with draft instances
```
sampaccoud: I think we should add the condition:
```python
"node__parent__cms_pages__publisher_is_draft": True, # organization has published parent
```

tests/apps/courses/test_cms_plugins_category.py
```diff
@@ -22,6 +22,33 @@ class CategoryPluginTestCase(CMSTestCase):
     Test that CategoryPlugin correctly displays a Category's page placeholders content
     """
 
+    @staticmethod
+    def test_cms_plugins_category_unique():
```
sampaccoud: I think it would be clearer to remove this test and modify the above test. You just need to replace:
```python
self.assertIn(leaf_category.extended_object.get_title(), rendered_form)
```
by 
```python
self.assertEqual(rendered_form.count(leaf_category.extended_object.get_title()), 1)
```

and update the docstring...

tests/apps/courses/test_cms_plugins_course.py
```diff
@@ -20,6 +20,31 @@ class CoursePluginTestCase(TestCase):
     Test that CoursePlugin correctly displays a Course's page placeholders content
     """
 
+    @staticmethod
+    def test_cms_plugins_course_unique():
```
sampaccoud: I think it would be clearer to remove this test and modify the above test. You just need to replace:
```python
self.assertIn(course.extended_object.get_title(), plugin_form.as_table())
```
by 
```python
rendered_form = plugin_form.as_table()
self.assertEqual(rendered_form.count(course.extended_object.get_title()), 1)
```

and update the docstring...

tests/apps/courses/test_cms_plugins_organization.py
```diff
@@ -21,6 +21,29 @@ class OrganizationPluginTestCase(CMSTestCase):
     Test that OrganizationPlugin correctly displays a Organization's page placeholders content
     """
 
+    @staticmethod
+    def test_cms_plugins_organization_unique():
```
sampaccoud: ditto.

src/richie/apps/courses/models/organization.py
2018D: To enable the use of sub-organization 👍

2018D: Adding the condition breaks the test... i'm manually testing to understand why.

2018D: So no need for this condition. Even if the organization is a child organization, they'll appear in the list. The 2 existing conditions are sufficient.

sampaccoud: no the purpose of this is not show organizations that are below a page that is not published... as it should be considered unpublished if its parent is hidden.

2018D: Adding` "node__parent__cms_pages__publisher_is_draft": True ` breaks existing tests and does not appear to be useful functionally. Let's keep the conditions for organization as is for now.

tests/apps/courses/test_cms_plugins_organization.py
2018D: done

tests/apps/courses/test_cms_plugins_course.py
2018D: done

tests/apps/courses/test_cms_plugins_category.py
2018D: done

src/richie/apps/courses/models/course.py
2018D: done

sampaccoud: not done ;-)

tests/apps/courses/test_cms_plugins_category.py
```diff
@@ -37,22 +38,30 @@ class Meta:
                 fields = ["page"]
 
         meta_category = CategoryFactory()
+        meta_category.extended_object.publish("en")
```
sampaccoud: BTW you should be able to do: 
```python
meta_category = CategoryFactory(should_publish=True)
```

src/richie/apps/courses/models/course.py
sampaccoud: Ah I just bumped into a problem on my other PR that explains why we put this clause in the query. I think it's to avoid duplicate results that result when you join on `cms_pages` for the other clause that does it.

I'm now improving this situation on the other PR and I can propose a fixup here if you want?

2018D:
Modified some of the comments to reflect what was determined here: 
https://github.com/openfun/richie/pull/547/files#diff-130a37291b0a06e9aee2f7bfd45f1460

src/richie/apps/courses/models/course.py
```diff
@@ -431,16 +431,16 @@ class CoursePluginModel(PagePluginMixin, CMSPlugin):
...
-            "node__parent__cms_pages__publisher_is_draft": True,
-            "node__parent__cms_pages__course__isnull": True,
-            "publisher_is_draft": True,
-            "course__isnull": False,
+            # Joining on `cms_pages` generate duplicates for courses that are under a parent page
+            # when this page exists both in draft and public versions. We need to exclude the
+            # parent public page to avoid this duplication with the first condition.
+            # The second condition makes sure the parent is not a course to exclude snapshots.
+            # The third option filters out public course.
```
sampaccoud:
```suggestion
            # The third condition filters out public course.
```

src/richie/apps/courses/models/course.py
```diff
@@ -431,16 +431,16 @@ class CoursePluginModel(PagePluginMixin, CMSPlugin):
...
+            # The fourth option makes sure only courses show up.
```
sampaccoud:
```suggestion
            # The fourth condition makes sure only courses show up.
```

src/richie/apps/courses/models/category.py
```diff
@@ -86,6 +86,7 @@ def get_category_limit_choices_to():
         "publisher_is_draft": True,  # plugins work with draft instances
         "category__isnull": False,  # limit to pages linked to a category object
         "node__parent__cms_pages__category__isnull": False,  # exclude meta categories
+        "node__parent__cms_pages__publisher_is_draft": True,  # exclude published categories
```
sampaccoud:
```suggestion
        "node__parent__cms_pages__publisher_is_draft": True,  # # exclude category published parent
```

+ add the same paragraph as what you added for the course below?

Issue #507, opened 2019-02-28
add course run


2018D:
## Bug Report

**Expected behavior/code**
add a course run. Fill in the fields and move on

**Actual Behavior**
Adding a course run causes an error because the start time is used even if there isn't one.

**Steps to Reproduce**
1. Add a course run.

**Environment**
- Richie version: master using the makefile script

**Possible Solution**
Add a check on start date and return a value such as 
> No start date
if there is none

**Additional context/Screenshots**
```
Environment:


Request Method: GET
Request URL: http://localhost:8070/fr/cours/lart-de-louer-autrement/test/?edit

Django Version: 2.1.7
Python Version: 3.6.8
Installed Applications:
('djangocms_admin_style',
 'django.contrib.auth',
 'django.contrib.contenttypes',
 'django.contrib.sessions',
 'django.contrib.admin',
 'django.contrib.sites',
 'django.contrib.sitemaps',
 'django.contrib.staticfiles',
 'django.contrib.messages',
 'cms',
 'menus',
 'sekizai',
 'treebeard',
 'djangocms_text_ckeditor',
 'filer',
 'easy_thumbnails',
 'djangocms_column',
 'djangocms_link',
 'djangocms_style',
 'djangocms_snippet',
 'djangocms_googlemap',
 'djangocms_video',
 'djangocms_picture',
 'richie',
 'richie.apps.core',
 'richie.apps.courses',
 'richie.apps.persons',
 'richie.apps.search',
 'richie.plugins.large_banner',
 'richie.plugins.plain_text',
 'richie.plugins.section',
 'richie.plugins.simple_text_ckeditor',
 'dockerflow.django',
 'parler',
 'rest_framework')
Installed Middleware:
('cms.middleware.utils.ApphookReloadMiddleware',
 'django.contrib.sessions.middleware.SessionMiddleware',
 'django.middleware.csrf.CsrfViewMiddleware',
 'django.contrib.auth.middleware.AuthenticationMiddleware',
 'django.contrib.messages.middleware.MessageMiddleware',
 'django.middleware.locale.LocaleMiddleware',
 'django.middleware.common.CommonMiddleware',
 'django.middleware.clickjacking.XFrameOptionsMiddleware',
 'dockerflow.django.middleware.DockerflowMiddleware',
 'cms.middleware.user.CurrentUserMiddleware',
 'cms.middleware.page.CurrentPageMiddleware',
 'cms.middleware.toolbar.ToolbarMiddleware',
[...]

PR #510, opened 2019-03-01
🐛(course) add check for start value

2018D:
## Purpose

The compute_state function requires a start date that can be empty.
If start date is absent, make sure the returned CourseState is the
default Coming soon. Fix issue #507 

## Proposal

Add a value check for start

### DIFF
--- a/src/richie/apps/courses/models/course.py
+++ b/src/richie/apps/courses/models/course.py
@@ -307,25 +307,39 @@ class CourseRun(BasePageExtension):
           6: there are no runs at all > "coming soon": {None}
         """
         now = timezone.now()
-        if start < now:
-            if end > now:
-                if enrollment_end > now:
-                    # ongoing open
-                    return CourseState(
-                        0, _("enroll now"), _("closing on"), enrollment_end
-                    )
-                # ongoing closed
-                return CourseState(4, None, _("on-going"), None)
-            # archived
-            return CourseState(5, None, _("archived"), None)
-        if enrollment_start > now:
-            # future not yet open
-            return CourseState(2, _("see details"), _("starting on"), start)
-        if enrollment_end > now:
-            # future open
-            return CourseState(1, _("enroll now"), _("starting on"), start)
-        # future already closed
-        return CourseState(3, None, _("enrollment closed"), None)
+        # start is empty hence default value
+        state = CourseState(6, None, _("coming soon"), None)
+        if start:
+            if start < now:
+                if end:
+                    if end > now:
+                        if enrollment_end:
+                            if enrollment_end > now:
+                                # ongoing open
+                                state = CourseState(
+                                    0, _("enroll now"), _("closing on"), enrollment_end
+                                )
+                            else:
+                                # ongoing closed
+                                state = CourseState(4, None, _("on-going"), None)
+                        return state
+                    else:
+                        # archived
+                        state = CourseState(5, None, _("archived"), None)
+                    return state
+                return state
+            if enrollment_start:
+                if enrollment_start > now:
+                    # future not yet open
+                    state = CourseState(2, _("see details"), _("starting on"), start)
+                elif enrollment_end > now:
+                    # future open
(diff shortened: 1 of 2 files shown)


2018D:
Added a test to check for if start = None when a courserun is initialized. Courserun state has to have a value even if compute_state cannot be called.

src/richie/apps/courses/models/course.py
```diff
@@ -330,9 +330,12 @@ def compute_state(start, end, enrollment_start, enrollment_end):
     @property
     def state(self):
         """Return the state of the course run at the current time."""
-        return self.compute_state(
-            self.start, self.end, self.enrollment_start, self.enrollment_end
-        )
+        if self.start:
```
sampaccoud: I think you need to do this inside the `compute_state` method instead of here otherwise the `compute_state` method is still broken.

tests/apps/courses/test_models_course_run.py
```diff
@@ -22,6 +22,16 @@ def setUp(self):
         super().setUp()
         self.now = timezone.now()
 
+    def test_models_course_run_state_coming_soon(self):
+        """
+        A course run that has no value passed should return a state with
```
sampaccoud:
```suggestion
        A course run that has no start date should return a state with 
```

tests/apps/courses/test_models_course_run.py
```diff
@@ -22,6 +22,16 @@ def setUp(self):
...
         self.now = timezone.now()
 
+    def test_models_course_run_state_coming_soon(self):
+        """
+        A course run that has no value passed should return a state with 
+        "coming soon" as text.
+        """
+        course_run = CourseRunFactory(start=None)
+        self.assertEqual(
+            course_run.state, CourseState(6, None, "coming soon", None)
```
sampaccoud: In this case we should also make sure that it is ranked last by the method course.get_course_runs. I'm afraid they might come first at the moment.
Or should we exclude them from the search engine index until they are properly defined?
Another soliution could be to introduce a new state for unscheduled courses so that it's easier to rank them last.
WDYT?

src/richie/apps/courses/models/course.py
2018D: The problem is that the linter complains that there are too many return statement. Not sure how to get around this problem.

tests/apps/courses/test_models_course_run.py
2018D: Introducing a new state would be ideal. The new indexing is done on courses instead of course-run so we'll need a way to represent it anyway. What does @mbenadda think about this ?

tests/apps/courses/test_models_course_run.py
2018D: This has been commited. I'll push it after the other changes.

src/richie/apps/courses/models/course.py
2018D: Right so just insert the state into a value and return it at the end. Got it.

tests/apps/courses/test_models_course_run.py
2018D: According to info on the state property in course:
src/richie/apps/courses/models/course.py line 210

The CourseState 6 is used when there is no courserun and is the lowest priority. This means that we can use this state as well as long as there is no date entered...

sampaccoud: Good idea. 
BTW the implications of such change are deep because they impact the search engine. As we are in the middle of a big refactoring of the search engine, I would like to postpone this work...

2018D:
I found other limit case. Will need new tests and new conditions to be added. Don't merge yet

2018D:
Added a few more value-check. The course and course run page work properly even if date values are missing. Will test further to make sure that returned state due to missing value make sense. Will add test scenario for those.

2018D:
This is now fixed thanks to f44c4bf8d57f17f9d2c5f8868bd03945b49c6405

PR #511, opened 2019-03-02
💄(homepage) improve homepage layout

sveetch:
## Purpose

Homepage has no real webdesign integration and is pretty
basic.

## Proposal

This commit improve Homepage layout with some stylesheets and
template, taking funmooc.fr as a reference.

This depends from #508 merge.

### DIFF
--- a/src/frontend/scss/_colors.scss
+++ b/src/frontend/scss/_colors.scss
@@ -5,6 +5,7 @@
 $black: #000000;
 $white: #ffffff;
 $gray7: #111111;
+$gray17: #2b2b2b;
 $gray20: #333333;
 $gray40: #666666;
 $gray80: #cdcdcd;
--- a/sandbox/settings.py
+++ b/sandbox/settings.py
@@ -467,6 +467,16 @@ class Base(DRFMixin, ElasticSearchMixin, Configuration):
         "easy_thumbnails.processors.background",
     )
 
+    RICHIE_SECTION_TEMPLATES = [
+        ("richie/section/section.html", _("Default")),
+        ("richie/section/highlighted_items.html", _("Highlighted items")),
+    ]
+
+    RICHIE_LARGEBANNER_TEMPLATES = [
+        ("richie/large_banner/large_banner.html", _("Default")),
+        ("richie/large_banner/hero-intro.html", _("Hero introduction")),
+    ]
+
     LOGGING = {
         "version": 1,
         "disable_existing_loggers": True,
--- a/src/frontend/scss/_main.scss
+++ b/src/frontend/scss/_main.scss
@@ -7,6 +7,7 @@
 @import './mixins/shapes';
 @import './mixins/flexbox';
 @import './mixins/cards';
+@import './mixins/hero';
 @import './mixins/list-group';
 
 // TWITTER BOOTSTRAP V4
@@ -54,6 +55,8 @@
 @import '../../richie/apps/persons/templates/persons/cms/person_detail';
 @import '../../richie/apps/search/templates/search/search';
 @import '../../richie/plugins/large_banner/templates/richie/large_banner/large_banner';
+@import '../../richie/plugins/large_banner/templates/richie/large_banner/hero-intro';
+@import '../../richie/plugins/section/templates/richie/section/highlighted_items';
 
 // Page component styles
 @import './components/header';
--- a/src/frontend/scss/_variables.scss
+++ b/src/frontend/scss/_variables.scss
@@ -8,6 +8,16 @@
 // Bootstrap settings
 //
 
+// Add a new breakpoint "xxl" for more granular downgrading
+$grid-breakpoints: (
+  xs: 0,
+  sm: 576px,
+  md: 768px,
+  lg: 992px,
+  xl: 1200px,
+  xxl: 1900px,
+);
+
 // More space between columns
 $grid-gutter-width: 20px;
 
--- a/src/frontend/scss/mixins/_hero.scss
+++ b/src/frontend/scss/mixins/_hero.scss
@@ -1,0 +1,143 @@
+// Structural rules for Hero types
+
+// Function to return every Hero element selectors depending on methodology
+// and possibly parent.
+//
+// @arg {String} $methodology [null]
+//   If value is "bem", object element selectors will be created with BEM
+//   methodology and starting from given `$parent`. Any other value is trigger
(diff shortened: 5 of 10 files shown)


sandbox/templates/richie/section/_highlighted_items.scss
```diff
@@ -0,0 +1,94 @@
...
+$richie-section-highlights-background-odd: $gray98 !default;
+$richie-section-highlights-title-color: $gray17 !default;
+$richie-section-highlights-title-margin: 0 auto 1rem !default;
+$richie-section-highlights-title-padding: 1rem !default;
+$richie-section-highlights-title-textalign: center !default;
+$richie-section-highlights-item-gutter: 0.625rem !default;
+
+// TODO: unrelated: Better Hero on large_banner template "hero-intro.html"
+.section{
+    $base-section-selector: &;
```
mbenadda: Why bother with this? `&` is already a constant

src/frontend/scss/_colors.scss
```diff
@@ -5,6 +5,7 @@
 $black: #000000;
 $white: #ffffff;
 $gray7: #111111;
+$gray17: #2b2b2b;
```
mbenadda: Why not reuse gray20 instead of making a new arbitrary gray17?

sandbox/templates/richie/section/_highlighted_items.scss
```diff
@@ -0,0 +1,94 @@
...
+            text-align: $richie-section-highlights-title-textalign;
+            @include media-breakpoint-up(md) {
+                @include make-container();
+                @include make-container-max-widths();
+            }
+        }
+
+        #{$base-section-selector}__items {
+            display: flex;
+            flex-direction: row;
```
mbenadda: is `row` not the default direction?

sandbox/templates/richie/section/_highlighted_items.scss
sveetch: Because to be able to make full BEM class name path you need to have the previous parent name, once you enter a new rule like `&--highlights` the value of `&` become `section--highlights` but that's not what we need for such child element like `__title` etc..

src/frontend/scss/_colors.scss
sveetch: gray20 is to dark compared to gray17, at the beginning it was gray20 but i needed to better fit to the funmooc homepage which use gray17.

sandbox/templates/richie/section/_highlighted_items.scss
sveetch: Yes it is, but personally i like to be explicit because some old browser didn't always have the right defaults and leads sometime to some bugs, like for explicitly defines flex grow, shrink and basis instead of the shortcut `flex`.

Maybe we should remove it if compatibility is not an issue with some browser from ~3years ago

sandbox/templates/richie/section/_highlighted_items.scss
mbenadda: I guess the plain way I'd write this to be clear would be:
```scss
.section {
  &__title {
    // default styles

    .section--highlights & {}
  }
}
```

I guess in this case it does not really matter because there are no default styles to section titles right?

src/frontend/scss/_colors.scss
mbenadda: Ok. I was just thinking that 7 different grays is starting to feel like a lot and 17-20 is not too far (if 0 is black and 100 is white)

sveetch: Designers often use a lot of same color tones to make a design, this can be justified for backgrounds which cover a lot more surface than font and so very significant for our eye.

sandbox/templates/richie/section/_highlighted_items.scss
sveetch: It should do the trick for this context even i'm not very confident to do this everywhere since we will have so to re-open the modifier under each element. This is a personal point of view but commonly i prefer to have one way tree like.

Demonstration: https://gist.github.com/sveetch/f67a7a19678572c1f99d6462619c2e28

mbenadda: IMO the point is that the repetition is worth it if it makes the code and its intention trivial to understand.

And (more importantly) its `section__title` styling itself for various circumstances, you don't have to look around modifiers for `section__title` styles.

Wondering what you think about this gist @sampaccoud () ?

sampaccoud: IMO #pim is much clearer than #poum... title and items are higher levels of details than highlights and dummy, so they should definitely be the parents.

The $ variable name also brings an inconsistence in the syntax style that is difficult to read and understand.

sveetch:
Here is a sample screenshot of new homepage render. Note that the first introduction Hero comes from created demo-site, then after i added another Hero block with content from funmooc for demonstration purpose, this last block has been manually added and not part of content created by demo-site.

Also the screenshot image quality has been downgraded to have a reasonable image file size.

screenshot_2019-03-06 home

src/richie/apps/courses/models/course.py
```diff
@@ -16,10 +16,10 @@
 from cms.models.pluginmodel import CMSPlugin
 from filer.fields.image import FilerImageField
 
-from ...core.fields.multiselect import MultiSelectField
-from ...core.models import BasePageExtension, PagePluginMixin
 from .category import Category
 from .organization import Organization
+from ...core.fields.multiselect import MultiSelectField
+from ...core.models import BasePageExtension, PagePluginMixin
```
sampaccoud: I think you get these errors because you forgot to run `make bootstrap` after rebasing to master.

src/frontend/package.json
```diff
@@ -7,7 +7,7 @@
     "build": "tsc --noEmit && webpack",
     "generate-l10n-template": "rip json2pot 'i18n/js/**/*.json' -o i18n/frontend.pot",
     "lint": "tslint -c tslint.json 'js/**/*.ts?(x)'",
-    "prettier-write": "prettier --write 'js/**/*.+(ts|tsx|json|js|jsx)' '*.+(ts|tsx|json|js|jsx)'",
+    "prettier-write": "prettier --write 'js/**/*.+(ts|tsx|json|js|jsx)' '*.+(ts|tsx|json|js|jsx)' '**/*.+(css|scss)'",
```
jmaupetit: Can you add this change to a separated commit?

PR #518, opened 2019-03-07
💄(homepage) add homepage plugin constraints and enable LinkPlugin

sveetch:
## Purpose

Sections allowed every available plugins for inclusion while we
want to restrict available plugin to a little set. Also, Sections
need capacity to have a button link like "Read more".

## Proposal

This commits add placeholder plugins rules on homepage main
placeholder to only enable CoursePlugin, OrganizationPlugin,
CategoryPlugin and LinkPlugin inside a SectionPlugin. Also there is
a new template for LinkPlugin to correctly style a link include in
section.

### DIFF
--- a/sandbox/settings.py
+++ b/sandbox/settings.py
@@ -296,6 +296,14 @@ class Base(DRFMixin, ElasticSearchMixin, Configuration):
         "richie/homepage.html maincontent": {
             "name": _("Main content"),
             "plugins": ["LargeBannerPlugin", "SectionPlugin"],
+            "child_classes": {
+                "SectionPlugin": [
+                    "CoursePlugin",
+                    "OrganizationPlugin",
+                    "CategoryPlugin",
+                    "LinkPlugin",
+                ]
+            },
         },
         # Course detail
         "courses/cms/course_detail.html course_cover": {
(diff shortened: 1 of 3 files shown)


sampaccoud:
@sveetch maybe add a small screen shot of the link?

sandbox/settings.py
```diff
@@ -296,6 +296,14 @@ class Base(DRFMixin, ElasticSearchMixin, Configuration):
         "richie/homepage.html maincontent": {
             "name": _("Main content"),
             "plugins": ["LargeBannerPlugin", "SectionPlugin"],
+            "child_classes": {
+                "SectionPlugin": [
+                    "CoursePlugin",
+                    "OrganizationPlugin",
+                    "CategoryPlugin",
+                    "LinkPlugin",
```
sampaccoud: add person plugin?

sandbox/settings.py
```diff
@@ -458,6 +466,11 @@ class Base(DRFMixin, ElasticSearchMixin, Configuration):
         "toolbar_HTMLField": [["Undo", "Redo"], ["Bold", "Italic"], ["Link", "Unlink"]],
     }
 
+    # Additional LinkPlugin templates. Note how choice value is just a keyword
+    # instead of full template path. Value is used inside an path formatting
```
```suggestion
    # instead of full template path. Value is used inside a path formatting
```

src/richie/plugins/section/templates/richie/section/_highlighted_items.scss
```diff
@@ -92,6 +92,32 @@ $richie-section-highlights-item-gutter: 0.625rem !default;
                     margin: 0;
                 }
             }
+
+            // Make a caesura in flex flow so button row allways takes full
```
sampaccoud:
```suggestion
            // Make a caesura in flex flow so button row always takes full width
```

sandbox/settings.py
sveetch: We need to add a little bit of styles for this plugin, this will be in a new different PR.

sveetch:
screenshot_20190307_111448

sandbox/settings.py
sampaccoud: :+1:

PR #519, opened 2019-03-07
💄(project) add new template for a simple single column page

sveetch:
## Purpose

Currently there is only a "full width" page where content takes the
full width which is a little rude to make simple page like Annexes.

## Proposal

We need to add a new simple template which only include a single
column structure with a little vertical padding so we have a nicer
render for simple pages.

## Sample screenshot

screenshot_2019-03-07 about


sampaccoud:
@sveetch should we justify the text?

I think we should also make sure that the common plugins are displaying nicely WDYT?

sveetch:
@sampaccoud ok this is a bit out of scope of this PR purpose and it would produces a lot of changes. I propose to validate this PR as it and we discuss apart on this subject for another new PR.

sampaccoud:
Of course!

PR #525, opened 2019-03-12
💄(single-column) add proper basic layout for available plugins 

sveetch:
## Purpose

Single column template allow a dozen of plugins but without any
proper layout, this is not really pretty.

## Proposal

This commit adds some styles around these plugins for single column
template so it have a nicer basic layout.

## Todo

- [x] Depends from #522 merge (would work without but does not have any sense);
- [x] Decide about image;

### DIFF
--- a/sandbox/templates/djangocms_video/full-width/video_player.html
+++ b/sandbox/templates/djangocms_video/full-width/video_player.html
@@ -1,0 +1,3 @@
+<div class="embed-responsive embed-responsive-16by9">
+    {% include "djangocms_video/default/video_player.html" %}
+</div>
--- a/sandbox/settings.py
+++ b/sandbox/settings.py
@@ -490,6 +490,8 @@ class Base(DRFMixin, ElasticSearchMixin, Configuration):
     # such as "templates/djangocms_link/VALUE/link.html"
     DJANGOCMS_LINK_TEMPLATES = [("button-caesura", _("Button caesura"))]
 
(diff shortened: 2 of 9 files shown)


sveetch:
Current layout with this commit:

screen

With these contents:

placeholders

sandbox/templates/djangocms_video/full-width/video_player.html
```diff
@@ -0,0 +1,3 @@
+<div class="embed-responsive embed-responsive-16by9">
```
sampaccoud: same question as previous PR: can we put these templates and scss files in Richie? That's because we want Richie to work like this by default without having to do these overrides in your project...

sveetch: @sampaccoud for third party plugins we can't, since they use a strange template path system which does not allow to define a real path just a keyword to find it.

sampaccoud: It was agreed to move all templates/scss/static files out of the sandbox and to the core richie app in a separate PR so that it is part of a startup kit that installs Richie with nearly zero work.

sampaccoud: FYI @jmaupetit

jmaupetit: :ok_hand:

PR #526, opened 2019-03-13
Bump release to 1.0.0-beta.2

mbenadda:
### Added

- Users can create a custom taxonomy of categories in the CMS (eg. Subjects with Computer Science (incl. Algorithms & Data Structures), History, etc., and Levels with Beginner, Intermediate, etc.).
- Filters are dynamically generated by the backend to support custom meta categories (level 1 nodes in the category taxonomy) and fix structual issues with dates.
- Integrate the homepage and fill it with content in the demo-site.
- Integrate a design for the list of organizations page.

### Fixed

- Richie is compatible with MySQL, with database switching utilities and dedicated tests.
- Full-text searches also search in linked object names, such as categories & organizations.
- Make improvements to the demo site to make it more representative of real-world use.
- Don't break course pages if there is no linked organization.
- Fix i18n issues that broke the backend when using other languages than the default english.
- Prevent display glitches during loading of the courses search page.


mbenadda:
done @jmaupetit

PR #531, opened 2019-03-14
🐛(courses) add missing migrations on courses

sveetch:
## Purpose

Some changes on courses model two weeks ago triggered pending
migrations because of some changes on categorypluginmodel,
coursepluginmodel, courserun and organizationpluginmodel fields.

## Proposal

This just resets courses migrations to include change without
introducing any other changes.


sampaccoud:
I would squash all migrations and start anew as long as we are not in production...

PR #550, opened 2019-03-22
🐛(search) fix courses autocomplete

mbenadda:
## Purpose

Autocomplete for courses was broken due to recent changes in the way courses are formatted in ElasticSearch by our state script.

Courses from the autocomplete query do not use the state script and are thefore missing expected fields when they are being formatted to the shape we're returning on the API.

## Proposal

We reasoned this is not unwanted behavior *per se*: we don't need the same shape for an autocomplete item as a full item, as we're only making use of IDs, titles and, in the case of courses, absolute_url.

We can add a `format_es_document_for_autocomplete` on indexers that builds these lighter objects for the sole use of autosuggest.


tests/apps/search/test_autocomplete_courses.py
```diff
@@ -0,0 +1,149 @@
...
+        courses = [
+            {
+                "complete": {
+                    "en": slice_string_for_completion(
+                        "Artificial intelligence for mushroom picking"
+                    )
+                },
+                "course_runs": [
+                    {
+                        # A) ongoing course, next open course to end enrollment
```
sampaccoud: You should now be able to remove the course runs safely :wink: just add an empty course_runs field common to all?

tests/apps/search/test_autocomplete_courses.py
```diff
@@ -0,0 +1,149 @@
...
+        """
+        Not a test.
+        Prepare the ElasticSearch index and execute the query in it.
+        """
+
+        courses = [
+            {
+                "complete": {
+                    "en": slice_string_for_completion(
+                        "Artificial intelligence for mushroom picking"
```
sampaccoud: I'm interested! :mushroom:

tests/apps/search/test_autocomplete_categories.py
```diff
@@ -0,0 +1,111 @@
...
+                "path": "001001",
+                "title": {"en": "Ocean biking"},
+            },
+            {
+                "complete": {
+                    "en": slice_string_for_completion("Eclectic bikeshedding")
+                },
+                "id": "51",
+                "path": "001002",
+                "title": {"en": "Eclectic bikeshedding"},
```
sampaccoud: ...is what we do for a living

PR #551, opened 2019-03-22
Allow nested categories and categories split on several filters

sampaccoud:
## Purpose

For the moment, the category filter includes all categories. But since categories are now a taxonomy, we want to be able to declare several filter blocks that each target a specific part of the category taxonomy.

We also need to be able to render the taxonomy with its nested parents and children.

## Proposal

- [x] Build ES documents ID with the path and parental status of a category's related page node (we need the ID to carry information about a category's place in the taxonomy so that facet counts have all the information),
- [x] refactor organizations to use the same ids so that they work the same eventhough we don't specifically need to nest organizations,
- [x] allow targeting different portions of the category taxonomy from different filters,
- [x] allow limiting what facets are computed by the filters. By default we will only facet the top level categories and the frontend will need to send separated queries to get facet counts for children categories,
- [x] allow limiting the scope of a response to just the objects or just the filters. This will make the query lighter when requesting facet counts to display the children of a filter box.

### DIFF
--- a/src/richie/apps/search/filter_definitions/courses.py
+++ b/src/richie/apps/search/filter_definitions/courses.py
@@ -4,7 +4,10 @@ from functools import reduce
 from django import forms
 from django.conf import settings
 from django.utils import timezone, translation
+from django.utils.functional import cached_property
 from django.utils.translation import gettext as _
+
+from cms.api import Page
 
 from ..fields.array import ArrayField
 from ..indexers import ES_INDICES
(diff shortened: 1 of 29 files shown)


src/richie/apps/search/filter_definitions/courses.py
```diff
@@ -26,6 +29,26 @@ class IndexableFilterDefinition(TermsAggsMixin, TermsQueryMixin, BaseFilterDefin
     number of choices.
     """
 
+    def __init__(self, name, reverse_id=None, **kwargs):
+        self.reverse_id = reverse_id
+        super().__init__(name, **kwargs)
+
+    @cached_property
+    def aggs_include(self):
```
mbenadda: Missing a docstring

sampaccoud:
Nice ! Do you believe me if I say I didn't even start the site to check if it was working lol?

src/richie/apps/search/filter_definitions/courses.py
sampaccoud: I still don't understand why pylint ignores this file :thinking:

2018D:
I've tried refreshing data using make search-index with no succes after deleting all pages related to level categories. The level section still appears without any sub-category. After adding a new kind of meta category and some sub, they don't appear. Has anyone had the same problem ?

sampaccoud:
@2018D 
> The level section still appears without any sub-category
Yes it's normal. The level section appears because it is declared in settings (see https://github.com/openfun/richie/pull/551/files#diff-2549135b4f1fca369d13f36d820feda0R84)

> After adding a new kind of meta category and some sub, they don't appear after the make search-index command
Same issue, you need to declare your category in the settings.

We did it like this so it is possible to decide what categories appear as a filter, and what is their behavior as is it possible to define the filter definition class used and pass parameters.

PR #552, opened 2019-03-22
✨(menu) add language chooser menu

2018D:
## Purpose

A menu language have to be displayed in header so user can select
its prefered language to navigate.

## Proposal

This commit include the menu language chooser inside header menu to
list project available languages. Each language is displayed with
their language code ("fr-fr"), not their long language name (it
wouldn't fit in an horizontal menu).


src/richie/apps/core/templates/menu/language_menu.html
```diff
@@ -0,0 +1,22 @@
...
+get_languages will find all available language, including unpublished pages.
+{% endcomment %}
+{% if languages|length > 1 %}
+    {% for language in languages %}
+        {% if language.0 in current_page.get_languages %}
+            {% if lang != language.0 %}
+            <li class="topbar__menu__list__item">
+                <a class="topbar__menu__list__item__link" href="{% page_language_url language.0 %}"
+                title="{% trans "Change to language:" %} {{ language.0 }}">
+                {% if language.0 == "en" %}
```
sveetch: With `{% get_available_languages as languages %}` before the loop, this would be useless, a simple `{{ language.1 }}` will be enough to display untranslated language name as written from `settings.LANGUAGES`

sampaccoud: Anyway I think it's too long to display the full language name here... some will be longer than french or english!
What do you think about displaying the chooser as: 

EN | **FR** | **DE**   (where english is currently showing)

sveetch: Since we won't use a dropdown menu for now, definitively the short code yes.

tests/apps/core/test_language_chooser.py
```diff
@@ -0,0 +1,74 @@
...
+
+from richie.apps.core.helpers import create_i18n_page
+
+
+class LanguageChooserTests(CMSTestCase):
+    """Test the language chooser menu for available languages"""
+
+    def test_language_chooser_available_language(self):
+        """
+        Menu should allways contains every available languages from settings,
```
sampaccoud:
```suggestion
        Menu should always contain every available language from settings,
```

tests/apps/core/test_language_chooser.py
```diff
@@ -0,0 +1,74 @@
...
+        no matter page has translation or not
```
sampaccoud:
```suggestion
        no matter if the page has a translation or not
```

tests/apps/core/test_language_chooser.py
```diff
@@ -0,0 +1,74 @@
...
+            title="Language menu test",
+            slug="language-test-page",
+            template="richie/fullwidth.html",
+            published=True,
+        )
+
+        url = page.get_absolute_url(language="en")
+        response = self.client.get(url)
+
+        # Every available language are present
```
sampaccoud:
```suggestion
        # Every available language is present
```

tests/apps/core/test_language_chooser.py
```diff
@@ -0,0 +1,74 @@
...
+        for item in ["en", "fr"]:
+            self.assertContains(
+                response, page.get_absolute_url(language=item)
+            )
+
+    def test_language_chooser_available_language_with_translated_page(self):
+        """
+        Menu should contains every available languages, current language item
```
sampaccoud:
```suggestion
        Menu should contain every available language, current language item
```

tests/apps/core/test_language_chooser.py
```diff
@@ -0,0 +1,74 @@
...
+        should be marked distinct class name
```
sampaccoud:
```suggestion
        should be marked with a distinct class name
```

tests/apps/core/test_language_chooser.py
```diff
@@ -0,0 +1,74 @@
...
+            "en": "Language menu test",
+            "fr": "Test du menu de langues",
+        }
+        page = create_i18n_page(
+            content, published=True, template="richie/fullwidth.html"
+        )
+        url = page.get_absolute_url(language="fr")
+        response = self.client.get(url)
+
+        # Every available language are present
```
sampaccoud:
```suggestion
        # Every available language is present
```

Issue #504, opened 2019-02-28
"Rare" filter values cannot be removed from search


mbenadda:
## Bug Report

**Expected behavior/code**
In addition to the "top facets" (or top filter values) as defined by any given search, all currently selected filter values should be available in the `filters` field of a course search request response.

**Actual Behavior**
Selected filter values that are not in the "top facets" are not included. This makes it impossible for the user to interact with them.
This specific situation can happen if a user starts selecting filter values, picks a "rare" value that is more frequent in the subset of courses filtered by the

**Steps to Reproduce**
- select a bunch of filter values
- select a "rare" filter value that was not available in the default "top facet" but is now available because it happens to be more frequent in the subset of courses your current active filters has selected
- remove the more common filter values
- the "rare" filter value is not visible but is still enabled (see the URL)

(once we add a clean way to directly select "rare" filter values, this will become even more critical)

**Environment**
- Richie version: current master
- Platform: N/A

**Possible Solution**
`FilterDefinitions`' `get_faceted_definition` methods need to explicitly ensure any active value related to their filter is included in the slice of `filters` they are generating.

**Additional context/Screenshots**
Note: this was not an issue before as the client built the filters from complete datasets. Now that the server sends a minimal amount of data (eg #476), it needs to ensure all necessary data is included.

PR #558, opened 2019-03-27
🐛(search) make sure selected filter values are included in the result

sampaccoud:
## Purpose

On indexable filters that are using a terms aggregation with a limit of 10 facet results, we could end up with a selected filter value not being included in the top 10 facets. Since the frontend now relies on the query result to display the filter boxes, the selected facet was not visible on the site anymore and could not be unselected...

Fixes https://github.com/openfun/richie/issues/504

## Proposal

We fix this problem by forcing specific facetting on the selected filter values. We choose to merge them with the facets returned by the terms aggregation. 

Note that it can lead to more than 10 facets but we think this allows keeping a significant number of new available choices.

### DIFF
--- a/src/richie/apps/search/filter_definitions/courses.py
+++ b/src/richie/apps/search/filter_definitions/courses.py
@@ -44,12 +44,13 @@ class IndexableFilterDefinition(TermsAggsMixin, TermsQueryMixin, BaseFilterDefin
         Returns:
         --------
             string: a regex depending on filters configuration in settings and pages in the CMS:
-                - "^$" if the `reverse_id` does not correspond to any published page which will
-                    will not match any value and return an empty list of facets,
-                - ".*-0001.{4}" if the `reverse_id` points to a published page with path "0001"
-                    this will match ids of the children of this page that will be of the form
-                    P-00010001 if they have children or L-00010001 if they are leafs,
-                ' ".*" if no `reverse_id` is set (delegated to super) which will match all values.
+            - "": the empty string, if the `reverse_id` does not correspond to any published
+                page which will not match any value and return an empty list of facets,
+            - ".*-0001.{4}": if the `reverse_id` points to a published page with path "0001"
+                this will match ids of the children of this page that will be of the form
+                P-00010001 if they have children or L-00010001 if they are leafs,
+            ' ".*": if no `reverse_id` is set (delegated to super) which will
+                match all values.
         """
         if self.reverse_id:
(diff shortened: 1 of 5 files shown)


src/richie/apps/search/filter_definitions/mixins.py
```diff
@@ -37,27 +38,57 @@ def get_aggs_fragment(self, queries, data, *args, **kwargs):
...
+        # Use all the query fragments from the queries *but* the one(s) that filter on the
+        # current filter
+        filter_fragments = [
+            clause
+            for kf_pair in queries
+            for clause in kf_pair["fragment"]
+            if kf_pair["key"] is not self.name
+        ]
+
+        # Terms aggregation will return the n top facet counts amount all the values taken
```
sampaccoud:
```suggestion
        # Terms aggregation will return the n top facet counts among all the values taken
```

src/richie/apps/search/filter_definitions/mixins.py
```diff
@@ -37,27 +38,57 @@ def get_aggs_fragment(self, queries, data, *args, **kwargs):
...
+                            # filter on the current filter: we manually add back the only one that
+                            # is relevant to the current choice.
+                            "must": [{"term": {self.term: value}}]
+                            + filter_fragments
+                        }
+                    }
+                }
+                for value in data.get(self.name, [])
+                # The Elasticsearch include regex matches exact values so we must do the same
+                # by adding ^ and $ and the beginning and at the end of the pattern
```
sampaccoud:
```suggestion
                # by adding ^ (resp. $) at the beginning (resp. at the end) of the pattern.
```

src/richie/apps/search/filter_definitions/courses.py
```diff
@@ -124,7 +125,12 @@ def get_faceted_definitions(self, facets, *args, **kwargs):
         """
         Build the filter definition's values from base definition and the faceted keys in the
         current language.
-        Those provide us with the keys and counts that we just have to consume.
+        Those provide us with the keys and counts that we just have to consume. They come from:
+        - a bucket with the top facets in decreasing order of counts,
+        - specific facets for values that were select in the querystring (we must force them
+          because they may not be in the n top facet counts but we must make sure we keep it
+          so that it remains available as an option so the use sees it and can unselect it)
```
mbenadda:
```suggestion
          so that it remains available as an option so the user sees it and can unselect it)
```

PR #563, opened 2019-03-27
✨(search) implement parent & children filters in the frontend

mbenadda:
## Purpose

The existing implementation of parent & children filters in the frontend expected filter values to have a "children" value that directly contained their own children values.

We decided to go another way with standard formatted paths that give us many benefits in terms of search but needed a different implementation in the frontend as parent filters need to fetch their children when the user requests them.

This also improves scalability.

## Proposal

Update the frontend to use the paths & make the appropriate requests.

WIP: missing tests

Fixes https://github.com/openfun/richie/issues/582

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,10 +8,14 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Added
+
+- Enable parent/children filters in the filters pane in Search.
+
 ### Changed
 
 - Translations are loaded dynamically in frontend application.
-- Optimized frontend build in our official Docker image
+- Optimized frontend build in our official Docker image.
 
 ## [1.0.0-beta.4] - 2019-04-08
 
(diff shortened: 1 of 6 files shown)


src/frontend/js/components/SearchSuggestField/SearchSuggestField.tsx
```diff
@@ -224,8 +224,17 @@ export const SearchSuggestFieldBase = ({
     filterName: modelName.CATEGORIES | modelName.ORGANIZATIONS,
     payload: string,
   ) => {
+    const metaCategorySubpath = payload.substr(2, 8);
+
+    const filter = Object.values(filters)
```
lunika: This is a negative point of hooks IMO. You can't test this implementation. You can't test `addFilter` function

lunika:
I worked on the test of this PR. It is not really easy because we update a state in an async action and so react warns us with this message:

```
When testing, code that causes React state updates should be wrapped into act(...):
      
      act(() => {
        /* fire events that update state */
      });
      /* assert on the output */
```

But, the async `act` function is only the first [alpha release of react](https://github.com/facebook/react/releases/tag/v16.9.0-alpha.0). I tried to use it and other tests are failing, we will have to wait a little bit before using it, and obviously types are not updated.

mbenadda:
I think the 💩 fix should be removed from this PR.

I followed the whole async act discussion and am glad it's finally merged. Hopefully we can start using it very soon.

This PR (minus the 💩) is GTM for me but I can't approve it since I'm the one who initially opened it. You'll have to approve it yourself

PR #565, opened 2019-03-28
💄(search) fix search filters glitch on click

jmaupetit:
## Purpose

When clicking on a search filter to toggle it, search filters lists are updated producing a glitch in the search filters column.

## Proposal

We've decided to make each search filter group list a block with a fixed height to avoid this disturbing glitch (blocks jump).

An alternative would have been to define a max-height, but this would just limit the jump phenomenon that will still be disturbing when only a few filters are still available.

## Sneak peek

richie-search-filters-height


jmaupetit:
I think you made the point. After discussing with @lunika it appears that the search request needs to get modified: the payload only contains active filters. Hence we cannot toggle the state of each filter...

jmaupetit:
After discussions, my proposal is not acceptable at all. I will close this as improving search UX requires more work than just updating filters height.

PR #571, opened 2019-03-29
✨(project) add new context processor to add Site values in templates

sveetch:
## Purpose

We need to reach current site object (from Django Site framework)
from every templates and so we will be able to use site name or
domain from templates.

## Proposal

This commit add and enable a new template context processor in
settings.

### DIFF
--- a/src/richie/apps/core/context_processors.py
+++ b/src/richie/apps/core/context_processors.py
@@ -1,0 +1,53 @@
+"""
+Template context processors
+"""
+from django.conf import settings
+from django.contrib.sites.models import Site
+
+
+def get_site_metas(with_static=False, with_media=False, is_secure=False, extra=None):
+    """
+    Return metas from the current *Site* and settings
+
+    Added Site metas will be callable in templates like this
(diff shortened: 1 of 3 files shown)


src/richie/apps/core/context_processors.py
```diff
@@ -0,0 +1,53 @@
...
+
+    Optionally it can also return ``STATIC_URL`` and ``MEDIA_URL`` if needed
+    (like out of Django requests).
+    """
+    site_current = Site.objects.get_current()
+    metas = {
+        "SITE": {
+            "name": site_current.name,
+            "domain": site_current.domain,
+            "web_url": "http://%s" % site_current.domain,
```
sampaccoud: use f string?

src/richie/apps/core/context_processors.py
```diff
@@ -0,0 +1,53 @@
...
+        }
+    }
+    if is_secure:
+        metas["web_url"] = "https://%s" % site_current.domain
```
sampaccoud: ditto

PR #581, opened 2019-04-01
Standardize the Makefile

lunika:
## Purpose

In the makefile when we want to use an other task we call it using the MAKE environment variable. When it's a prerequisite we should instead use rules dependencies

## Proposal

Refactor makefile and use prequisite tasks when possible.


lunika:
> Argh, in fact you have a typo in your commit title.

Fixed

PR #583, opened 2019-04-01
Improve frontend editing for person and blog

sampaccoud:
## Purpose

It was not possible to edit the title of the blog post and person pages as easily as on the other types of pages.

## Proposal

- [x] add frontend editing of the page title on a blog post
- [x] replace the page title, on the top of a person page, by the concatenation of her title, first name and last name. Make them editable in frontend.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -83,6 +83,7 @@ compilemessages: ## compile the gettext files
 	@$(COMPOSE_RUN) -w /app/src/richie app python /app/sandbox/manage.py compilemessages
 
 demo-site:  ## create a demo site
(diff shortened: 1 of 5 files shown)


sveetch:
@sampaccoud What about the full name in the select input when you are selecting a person from Person plugin form ?

sampaccoud:
@sveetch as discussed, modifying the full name in plugin forms is not so easy because we don't have access to the Page model. To do so, we should specify a custom form with a custom field that renders how we expect. 

This should be done in another PR.

PR #586, opened 2019-04-01
🔧(blog) allow adding blog post plugins on the home page

sampaccoud:
## Purpose

When we added the blog post plugin, we forgot to add it as a possible option on the home page.

## Proposal

- [x] Add BlogPostPlugin as a possible option on the home page.


sandbox/settings.py
```diff
@@ -289,11 +289,12 @@ class Base(DRFMixin, ElasticSearchMixin, Configuration):
             "plugins": ["LargeBannerPlugin", "SectionPlugin"],
             "child_classes": {
                 "SectionPlugin": [
+                    "BlogPostPlugin"
```
sveetch: Missing ending comma

PR #566, opened 2019-03-28
I18n support

lunika:
## Purpose

We want to translate all the application at least in french and french canadian. We use crowdin to allow translators to help us. Now we have to automate when source translation are uploaded on crowdin, how we fetch latest translations and how to inject them in front and back applications

## Proposal

- [x] use crowdin-cli to interact with crowdin. We also use this CLI in the CI
- [x] remove existing tool to upload files on crowdin
- [x] create task in makefiles to ease interaction with crowdin and messages compilation
- [x] fully use `react-intl` in front application

PR #589, opened 2019-04-02
Refactor dynamic import in frontend application

lunika:
## Purpose

I'm not confident with what I made in my [PR managing i18n](https://github.com/openfun/richie/pull/566) and I don't know if we can really rely with this kind of dynamic require.
So I look for an other solution using dynamic import in a more "typescript" way.

## Proposal

- [x] use core-js 3 to have a more recent version of core-js, this is not really realted with the main purpose of this PR but it's I think better to use this new version.
- [x] use @babel/plugin-syntax-dynamic-import to have dynamic import 
- [x] refactor `index.tsx` file

There are maybe negative points here :warning:  :

- the build now have lot of files and not only one `index.js`. This is because the build import every possible file "importable" and react-intl as a lot of `locale-data` files.
- I have switch the module resolution use in the typescript config, so I don't really know with which browser our code will work now, I found nothing about this in the documentation, but with babel maybe we have no compatibility regression.

### DIFF
--- a/src/frontend/js/index.tsx
+++ b/src/frontend/js/index.tsx
@@ -38,7 +38,7 @@ document.addEventListener('DOMContentLoaded', event => {
   // Find all the elements that need React to render a component
   Array.prototype.forEach.call(
     document.querySelectorAll('.fun-react'),
-    (element: Element) => {
+    async (element: Element) => {
       // Generate a component name. It should be a key of the componentLibrary object / ComponentLibrary interface
       const componentName = startCase(
         get(element.className.match(/fun-react--([a-zA-Z-]*)/), '[1]') || '',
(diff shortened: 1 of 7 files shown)


mbenadda:
- IMO this does not change anything as far as browser support is concerned because all imports (dynamic or otherwise) are replaced by `__webpack_require__`.
- The typescript config change has no influence on that because typescript does not emit anything (it only runs as a linter and `es6` => `esnext` made this linter accept the use of lazy imports)
- Ditto the babel plugin does not transpile `import()`, it simply lets babel parse it without throwing errors

src/frontend/js/index.tsx
```diff
@@ -54,12 +54,12 @@ document.addEventListener('DOMContentLoaded', event => {
         }
 
         try {
-          addLocaleData(require(`react-intl/locale-data/${localeCode}`));
```
mbenadda: IMO `index` is not the right place to do this: it is impossible for React code to reach this code and load another locale than the one that was loaded at start. At the same time we're not aggregating locale loads between render trees because this is in the iterator.

The problems are theoretical as we don't allow locale switching right now but we should be aware of the possible implications:
- load in index (but outside the iterator) to avoid multiple loads of locale data
- load inside a component (`App?`) so language can be changed and locale data dynamically reloaded when this happens

mbenadda:
Overall I like this approach much better than the other one with `require`! This is a good idea 👍

Just one concern: do you think the hundreds of bundle slices will be handled by our static file pipeline properly and cache busted as needed? @sampaccoud maybe?

sampaccoud:
The Django ManifestStaticFileStorage back treats all static files equally so... yes they will be hashed, served and cached busted like all the other static files.

Issue #567, opened 2019-03-28
Improve styling images


sampaccoud:
## Feature Request

**Is your feature request related to a problem or unsupported use case? Please describe.**
Some configuration is missing to allow controlling the positioning of images.

**Describe the solution you'd like**
- Adding images from CKEditor:
    * remove the useless "image" button,
    * when adding an image via the "plugin" bouton:
        - allow horizontal alignment via the 3 CKEditor alignment buttons: it works but the button doesn't show it's pressed...
        - allow applying horizontal alignment on the image itself: this should result in a floating image behavior.
- Addition images directly from a plugin: allow applying horizontal alignment on the image plugin itself that should result in a floating image.

When adding a legend it should be centered below the image (it is currenlty left aligned).

sveetch:
- [ ] Update Ckeditor configs to disable ckeditor image plugin;
- [ ] Fix CKeditor text alignment button (need to add alignment styles to ckeditor css?);
- [ ] Write djangocms-picture css styles for image alignment (align-left, align-right, align-center);
- [ ] Legend (figure ? caption?);
- [ ] Ensure there is a default `max-width: 100%` on basic `img` style (Bootstrap does not do that?);

sveetch:
Done in #594

PR #594, opened 2019-04-03
💄(layout) improve styling images and CKEditor usage

sveetch:
## Purpose

Some configuration is missing to allow controlling the positioning
of images and image element (img and figure) behaviors are not
correct.

Related to issue #567

## Proposal

Here we add some basic image element styles so image have correct
basic behaviors on its alignments and with its container. Also we
disable some useless CKEditor features and fixed alignment buttons.

Note than since CKEditor edit render have its own CSS file to
customize render and since we need to use it to fix text alignment,
we forced its addition to the repository in richie static directory.

### DIFF
--- a/sandbox/settings.py
+++ b/sandbox/settings.py
@@ -446,7 +446,7 @@ class Base(DRFMixin, ElasticSearchMixin, Configuration):
         "language": "{{ language }}",
         "skin": "moono-lisa",
         "toolbarCanCollapse": False,
-        "contentsCss": "/static/css/ckeditor.css",
+        "contentsCss": "/static/richie/css/ckeditor.css",
         # Enabled showblocks as default behavior
         "startupOutlineBlocks": True,
         # Enable some plugins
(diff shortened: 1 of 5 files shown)


sveetch:
@sampaccoud this is not ready yet since for CKEditor button alignment fix i need to add a CSS file (dedicated to CKEditor in its iframe):

```
/*
CSS for CKEditor content
*/

/*
 * Add alignment classes required for alignment buttons to work right according
 * to CKEditor configuration option "justifyClasses" */
.text-left {
  text-align: left;
}

.text-right {
  text-align: right;
}

.text-center {
  text-align: center;
}
```

But CSS file are ignored from GIT, so it have to be a Sass file to build but so we need to modify `sass` action from `richie/frontend/package.json` since it is configured to build `_main.scss` only.

sveetch:
## Image elements behaviors sample

### With this structure

image-samples_structure

### Get this render

image-samples_render

PR #605, opened 2019-04-04
🐳(front) build frontend in production mode

lunika:
## Purpose

The Dockerfile at root project is used to build the production image. We
should build the frontend application in production mode.

## Proposal

- [x] add `--mode=production` to `yarn build` in Dockerfile.


lunika:
@jmaupetit I add the compression but the sourcemap will not work, the browser is not able to fetch the sources, they are not present.

PR #608, opened 2019-04-05
👷(circleci) add check CHANGELOG job

jmaupetit:
## Purpose

We need a way to remind us to update the project's `CHANGELOG` when we submitting a PR.

## Proposal

- [x] add a `check-changelog` job to the CI

One may believe that this is another annoying task, but the experience we had with Arnold's repository shows that it drastically speeds up the release cycle since almost everything has already been prepared upfront.

You may have noticed that this job **is not required** to merge a PR since the `CHANGELOG` does not have to be updated for every submitted changes.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -41,6 +41,17 @@ jobs:
           command: |
             gitlint --commits origin/master..HEAD
 
+  # Check that the CHANGELOG has been updated in the current branch
+  check-changelog:
+    machine: true
+    working_directory: ~/fun
+    steps:
+      - checkout
+      - run:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -718,6 +729,11 @@ workflows:
           filters:
             tags:
               only: /.*/
+      # Check CHANGELOG update
+      - check-changelog:
```
lunika: on arnold you ignore the master branch. Why not here ?

jmaupetit: Because I forgot. Thank you :heart:

jmaupetit:
I'll wait for the next release before merging this.

PR #615, opened 2019-04-09
🐳(frontend) improve build

jmaupetit:
## Purpose

Now that we've packaged the frontend application, we no longer require to copy the whole repository sources to compile the frontend.

## Proposal

- [x] only copy what is required to build the frontend application
- [x] improve build path consistency

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 ### Changed
 
 - Translations are loaded dynamically in frontend application.
+- Optimized frontend build in our official Docker image
 
 ## [1.0.0-beta.4] - 2019-04-08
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Changed
+
+* Optimized frontend build in our official Docker image
```
lunika: why a `*` ? We use `-` everywhere ?

jmaupetit: Good catch! I used vim and not VSCode that automatically lint my markdown sources :wink:

Dockerfile
```diff
@@ -20,11 +20,12 @@ FROM python:3.7-stretch as base
 # ---- front-end builder image ----
 FROM node:10 as front-builder
 
-# FIXME: we should only copy src/frontend, but for now compiling scss files
-# requires files from the backend sources
-COPY ./src /app/src/
+WORKDIR /builder
```
lunika: I think you can remove this instruction.

PR #635, opened 2019-04-10
Fix statics in docker and package builds

jmaupetit:
## Purpose

We've detected multiple issues related to static files packaging (Docker + Python):

- `ckeditor.css` file is missing in the Docker image
- `statics` & `locale` contents are not distributed with the python package

## Proposal

- [x] add an exception to the `.dockerignorefile` for the `ckeditor.css` particular case
- [x] persist/attach frontend builds for the backend packaging
- [x] ensure locale are also packaged

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -454,17 +454,11 @@ jobs:
   #   * you have define both the TWINE_USERNAME & TWINE_PASSWORD secret
   #     environment variables in CircleCI UI (with your PyPI credentials)
   pypi:
-    machine: true
+    docker:
+      - image: circleci/python:3.7-stretch
     working_directory: ~/fun
     steps:
       - checkout
-      - restore_cache:
-          keys:
-            - docker-debian-images-ci-{{ .Revision }}
-      - run:
-          name: Load images to docker engine
-          command: |
-            docker load < docker/images/dev/richie.tar
       # Restore built python packages
       - attach_workspace:
           at: ~/fun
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -598,6 +600,8 @@ jobs:
           root: ~/fun
           paths:
             - src/frontend/i18n/frontend.pot
+            - src/richie/static/richie/js/*.js
```
lunika: The build is not made in production mode, you will not have an optimized build. Is it impostant or not ?

Dockerfile
```diff
@@ -34,13 +34,18 @@ FROM base as back-builder
...
-# Copy distributed application's statics
-COPY --from=front-builder /builder/src/richie/static/richie /builder/src/richie/static/richie
-
 # Copy required python dependencies
 COPY setup.py setup.cfg MANIFEST.in /builder/
 COPY ./src/richie /builder/src/richie/
 
+# Copy distributed application's statics
+COPY --from=front-builder \
+    /builder/src/richie/static/richie/js/*.js \
```
lunika: You are doing this to be sure to not have `.map` files ?

.circleci/config.yml
jmaupetit: Yes it is. And it should as it will be the default build that will be shipped with the python package.

Dockerfile
jmaupetit: Not specially to avoid copying the `.map` files but rather to explicitly copy all JS files and not a folder that may override existing JS files in the statics directory (as it is the case for the CSS directory).

.circleci/config.yml
lunika: This task name is not relevant. It tests that the front application build well and we need the build in development mode.

You can create a new task or add more steps to this task to build the app in production mode.

jmaupetit: True. Done.

PR #637, opened 2019-04-11
💄(person) style persons list and fix minor layout feedbacks

sveetch:
## Purpose

The persons list page can be visited by editing the url. It should
look decent.

## Proposal

This commit add a new person list template, a new person glimpse
template and use it to represent person in list and plugin. Also fix
minor feedbacks about grey color used as global page background that
was too light compared to the one from funmooc.


sampaccoud:
@sveetch I think we should add a "teachers" item in the menu for this page... at least in the demo site... Same for categories...

sveetch:
> @sveetch I think we should add a "teachers" item in the menu for this page... at least in the demo site... Same for categories...

I enabled theses pages in create demo site

src/richie/apps/courses/templates/courses/cms/course_detail.html
```diff
@@ -48,13 +48,13 @@ <h1 class="course-detail__content__title">
 
     <div class="course-detail__aside">
 
+      {% if current_page.publisher_is_draft %}
```
sampaccoud: :+1:

PR #643, opened 2019-04-11
Simplify Richie settings by declaring as defaults those that are unlikely to require customization

sampaccoud:
## Purpose

We want to simplify Richie's settings as much as possible while still maintaining configurability.

## Proposal

I'm proposing to move the settings that are unlikely to be customized to a default pattern inside Richie.

- [x] move search settings to Richie as defaults,
- [x] refactor how the Elasticsearch client is instantiated,
- [x] move specific languages settings to Richie as defaults,
- [x] declare as defaults the settings related to Richie plugins,
- [x] minor typos and removing useless settings.

### DIFF
--- a/tests/apps/search/test_autocomplete_categories.py
+++ b/tests/apps/search/test_autocomplete_categories.py
@@ -89,11 +89,7 @@ class AutocompleteCategoriesTestCase(TestCase):
             }
             for category in categories
         ]
-        bulk(
-            actions=actions,
-            chunk_size=settings.ES_CHUNK_SIZE,
-            client=settings.ES_CLIENT,
-        )
+        bulk(actions=actions, chunk_size=500, client=settings.ES_CLIENT)
         indices_client.refresh()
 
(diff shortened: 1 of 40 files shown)


tests/apps/search/test_autocomplete_categories.py
```diff
@@ -89,11 +89,7 @@ def execute_query(self, querystring=""):
             }
             for category in categories
         ]
-        bulk(
-            actions=actions,
-            chunk_size=settings.ES_CHUNK_SIZE,
-            client=settings.ES_CLIENT,
-        )
+        bulk(actions=actions, chunk_size=500, client=ES_CLIENT)
```
jmaupetit: Why using a fixed chunk size instead of the configured/default one?

sampaccoud: This is a test so I'm testing the value passed as well so that the test would break if someone makes a change that alters the chunk size unwillingly.

PR #645, opened 2019-04-12
♻️(demo) Move the demo site to its own application

sampaccoud:
## Purpose

The command that creates the demo site is inside the core app. This was creating circular dependency issues and is a bad design because the dependency tree should be in order: 

**core <- courses <- demo**

## Proposal

- [x] Move all the code related to the demo site to a separate `demo` app,
- [x] Refactor tests and factories so that:
    - tests can use only factory boy dummy images,
    - the demo site command can use real images with a random picker.


sampaccoud:
@jmaupetit well spotted. I moved back the whole `static` folder to the `core` app as the logo is also used in our templates...

PR #646, opened 2019-04-12
WIP:🐛(project) fix cropping issue with images

sveetch:
## Purpose

There is no cropping on image from PicturePlugin as we
expected when giving 'width' and 'height' template variable
around placeholders (on logo or portrait).

## Proposal

This commit is a Proof of Concept how we could enforce sizes
in default plugin template with our context variables and easy
thumbnail (or any other thumbnailer). Also to be able to override
default plugin template, 'richie.apps.core' has been moved before
'cms' in 'settings.INSTALLED_APPS'.

This does not contain easy thumbnail usage yet and there is
some debug pointers in our own default plugin template.

### DIFF
--- a/src/richie/apps/core/templates/djangocms_picture/debug/picture.html
+++ b/src/richie/apps/core/templates/djangocms_picture/debug/picture.html
@@ -1,0 +1,85 @@
+{% load thumbnail %}
+<p style="z-index: 200; background: white;">
+    Width:{{ instance.width }}<br>
+    Height:{{ instance.height }}<br>
+    picture_size:{{ picture_size }}
+</p>
+{% if picture_link %}
+    <a href="{{ picture_link }}"
+    {% if instance.link_target %} target="{{ instance.link_target }}"{% endif %}
+    {{ instance.link_attributes_str }}>
+{% endif %}
+
+{# start render figure/figcaption #}
+{% if instance.caption_text %}
+    <figure>
+{% endif %}
+{# end render figure/figcaption #}
+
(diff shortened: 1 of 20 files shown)


src/richie/apps/core/templates/djangocms_picture/debug/picture.html
```diff
@@ -0,0 +1,85 @@
...
+
+{# start render figure/figcaption #}
+{% if instance.caption_text %}
+    <figure>
+{% endif %}
+{# end render figure/figcaption #}
+
+
+<img src="{{ instance.img_src }}"
+    alt="{% if instance.picture.default_alt_text and not instance.attributes.alt %}{{ instance.picture.default_alt_text }}{% endif %}"
```
sampaccoud: This is bugged as well and generates a duplicated `alt` attribute when `alt` is passed in `instance.attributes`:
```suggestion
    {% if instance.picture.default_alt_text and not instance.attributes.alt %}alt="{{ instance.picture.default_alt_text }}"{% endif %}
```

sveetch: Well it a bug from plugin, since i just copied and pasted its default template without changing these lines. Whatever it is just a PoC to show it works to override default template when we move up richie.apps.core just before cms. You may not be able to merge it since it is largely outdated from master.

sampaccoud:
Image cropping/resizing was revamped in PR https://github.com/openfun/richie/pull/680.

PR #655, opened 2019-04-16
Improve CI performances

jmaupetit:
## Purpose

Using the machine executor with a handcrafted docker images backup and loading takes to much time to make our CI reliable. Plus, after thinking a bit more about what we were doing, there is no point in running those tests in our Docker image as we are assessing our code, not our image.

## Proposal

- [x] move all jobs to docker executors (major speed up)
- [x] decrease the number of jobs by merging related jobs (linting mostly)

During this work, I also had to:

- [x] make the `RICHIE_ES_HOST` setting configurable in the sandbox


sandbox/settings.py
```diff
@@ -515,6 +515,11 @@ class Base(DRFMixin, Configuration):
         },
     }
 
+    # Elasticsearch
+    RICHIE_ES_HOST = values.Value(
+        "elasticsearch", environ_name="RICHIE_ES_HOST", environ_prefix=None
```
sampaccoud: Why not just:
```
RICHIE_ES_HOST = values.Value("elasticsearch")
```

jmaupetit: Because, we want it to be overridable by an environment variable not prefixed by `DJANGO_`. Am I wrong?

sampaccoud: IC. But since it is a Django setting like the others I think we should follow the same pattern.

jmaupetit: This is not the case for database-related settings, and in this case, it should be called `ES_HOST`, and thus be overridable by a `DJANGO_ES_HOST` environment variable...

sampaccoud: but DATABASES is a common Django setting where ES_HOST is just for Richie... Still I'm ok to remove the RICHIE_ if you think it's better, but think of the coherence with all the other richie settings.

jmaupetit: Don’t get me wrong: I prefer to let it as is, ie named RICHIE_ES_HOST, and let’s make all other Richie specific setting follow the same pattern, ie RICHIE_*

sampaccoud: :+1:  it should already be the case for all the settings I moved to defaults.py last week.

jmaupetit: Ok then I'll mark this as resolved.

.circleci/config.yml
```diff
@@ -64,379 +67,255 @@ jobs:
...
   test-back-mysql:
-    machine: true
+    docker:
+      - image: circleci/python:3.7-stretch
+        environment:
+          DJANGO_SETTINGS_MODULE: settings
+          DJANGO_CONFIGURATION: Test
+          DJANGO_SECRET_KEY: ThisIsAnExampleKeyForTestPurposeOnly
+          PYTHONPATH: /home/circleci/fun/sandbox
+          RICHIE_ES_HOST: localhost
```
sampaccoud: then maybe we should put localhost as the default ES host WDYT?

jmaupetit: 🤔 not sure we often map local ports to an ES cluster.

PR #656, opened 2019-04-16
Refactor SearchSuggestField to use more functional tests

mbenadda:
## Purpose

`<SearchSuggestField />` was using tests that were very "unit" oriented, testing disjointed parts with a lot of mocks to glue everything. This is not something we favor anymore as it does not make it simple to cover use-cases and actual reported bugs with new tests.

## Proposal

As we were working on fixing a bug with the autosuggest, we decided to refactor the tests to work in a more functional way by mimicking user interaction with `react-testing-library`. Our "unit" now being `<SearchSuggestField />` instead of the various functions it has and APIs of `react-autosuggest` it intereacts with.

As we did this, we noticed functions that did not need to be exported and curried anymore, and we did some light changes that simplified the actual component code too.


src/frontend/js/components/SearchSuggestField/SearchSuggestField.spec.tsx
```diff
@@ -1,305 +1,330 @@
...
+    !!element.querySelector('b') &&
+    element.querySelector('b')!.innerHTML.includes(value) &&
+    element.innerHTML.includes('in courses...');
+
   beforeEach(jest.resetAllMocks);
+
+  // Disable useless async act warnings
+  // TODO: remove this spy as soon as async act is available
+  beforeAll(() => {
+    jest.spyOn(console, 'error');
```
lunika: Beurk

PR #662, opened 2019-04-18
Use tickboxes to show active filters

### LINKED ISSUE #597: Replace crosses by checkboxes on search filters
## Feature Request

The way filters are shown when selected seems confusing. Most users understood that it was an AND when it is in fact an OR.

- [x] Replace the cross on the right by a checkbox on the left
- [x] Show facet count even for facets that are selected

Example:

mbenadda:
## Purpose

We used button with a close (X) icon to enable users to manage active filters in the filters pane of the search view. This was not the best way to help users intuitively understand what those checkboxes enable them to achieve.

## Proposal

Using checkbox, they can understand without any explanation that ticked filters are included in the results and non-ticked filters are not. This is also very helpful to make the way multi-faceted filters (OR) work obvious.

Also:
Now that our search filter groups use checkbox-label pairs to represent the available filters, we should use accessible <fieldset /> and <legend /> elements to properly associate the related checkboxes together and with their common title (which is, in this case, the filter definition human name).

Closes #597.


sampaccoud:
The`Clear 1 active filter` link disappears a fraction of a second before the filters are re-drawn creating a double-flicker effect.

mbenadda:
@sampaccoud #663

PR #667, opened 2019-04-19
🚸(demo) associate persons to organizations on the demo site

sampaccoud:
## Purpose

Courses were tagged randomly with persons and organizations without any link between the two. This resulted in all persons being linked to all organizations. We can make our demo data set more realistic
by assigning each person to an organization.

## Proposal

I pick a random organization for each person. Then, for each course I associate persons and organizations in coherence so that the team of a course is composed of persons belonging to one of the organizations that are presenting the course.


src/richie/apps/demo/defaults.py
```diff
@@ -6,13 +6,13 @@
...
     "course_organizations": 3,
-    "course_persons": 3,
+    "course_persons": 6,
     "course_subjects": 2,
     "person_subjects": 3,
     "organizations": 5,
     "licences": 5,
-    "persons": 10,
-    "blogposts": 6,
+    "persons": 100,
```
sveetch: It's a little bit too much for a demo, 50 seems more reasonable and just as relevant than 100. Demo site routine was already a little long to proceed before these changes, this is a demo site, not a stress test demo :)

sampaccoud: ok

PR #670, opened 2019-04-23
♻️(persons) link persons to organizations

sampaccoud:
## Purpose

Last week, we displayed the link between a person and an organization via their presence on a course. The problem is that a course is linked to several organizations and several persons and we are not able to differentiate which persons belong to which organization. 

## Proposal

- [x] Add a placeholder `organizations` on the person detail page,
- [x] Replace the utility method `get_persons` on the organization model to retrieve related persons from the new placeholder plugins,
- [x] Add tests and modify existing tests.

### DIFF
--- a/src/richie/apps/courses/templates/courses/cms/person_detail.html
+++ b/src/richie/apps/courses/templates/courses/cms/person_detail.html
@@ -39,6 +39,17 @@
     </div>
   </div>
 
+  {% if current_page.publisher_is_draft or not current_page|is_empty_placeholder:"organizations" %}
+  <div class="organization-glimpse-list">
+    <h2 class="organization-glimpse-list__title">{% trans "Organizations" %}</h2>
+    {% placeholder "organizations" or %}
+      <p class="organization-glimpse-list__empty">
+        {% trans "No associated organizations" %}
+      </p>
+    {% endplaceholder %}
+  </div>
+  {% endif %}
+
   {% with courses=person.get_courses %}
     {% if courses %}
       <div class="course-glimpse-list">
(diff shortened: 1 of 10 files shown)


src/richie/apps/courses/templates/courses/cms/person_detail.html
```diff
@@ -39,6 +39,17 @@ <h1 class="person-detail__title">
     </div>
   </div>
 
+  {% if current_page.publisher_is_draft or not current_page|is_empty_placeholder:"organizations" %}
+  <div class="organization-glimpse-list">
+    <h2 class="organization-glimpse-list__title">{% trans "Organizations" %}</h2>
+    {% placeholder "organizations" or %}
+      <p class="organization-glimpse-list__item organization-glimpse-list__item--empty">
```
sampaccoud: @mbenadda Not sure what is the correct markup here. This is the item to display when the list is empty...

sveetch: There is no CSS rule for empty modifier on a glimpse it's something only related to its list container.

Inside `.organization-glimpse-list` in `src/frontend/scss/objects/_organization_glimpses.scss` you may add:
```
    &__empty {
      @include sv-flex(1, 0, auto);
      padding: 1rem;
      text-align: center;
    }
```

And change your empty item HTML to:

```
<p class="organization-glimpse-list__empty">
```

 However there is no such rules in `.organization-list` also, since we conditioned many list blocks so they do not appear when empty.

sampaccoud: @sveetch done

Issue #663, opened 2019-04-18
Prevent search filters from jumping around


mbenadda:
## Bug Report

**Expected behavior/code**
The Search filters pane and its contents should be mostly steady when the user interacts with them so as not to confuse them.

**Actual Behavior**
The whole contents of the search filters pane jump around on every user interaction. They even jump around twice when moving from 0 to 1 or 1 to 0 active filters.

**Steps to Reproduce**
1. Enable filters in the search filters pane

**Environment**
- Richie version: latest

**Possible Solution**
We should probably hide neither filter values with a facet count of 0, nor the "Disable all active filters" link (or keep its space so what's below it in the pane does not jump around).

**Additional context/Screenshots**
N/A

PR #671, opened 2019-04-23
🚸(search) prevent search filters from jumping around

sampaccoud:
## Purpose

The number of facets displayed varies each time a new filter or query is applied because we only display facets that have some results. 

## Proposal

In order to prevent this annoying effect, we can display all facets and disable those that have zero results.

- [x] Make minimum document count configurable on term facets and don't exclude facets with zero results,
- [x] Stop excluding facets with zero results from choice facets (they are configurable as well...),
- [x] Display facets with zero results as disabled,
- [x] Reserve vertical space for the "Disable all active filters" link, even when it is hidden.

Fixes https://github.com/openfun/richie/issues/663

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,9 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 ### Changed
 
 - Search as the user types in the course search field. This replaces
-  autosuggest in course names with a full index search.
+  autosuggest in course names with a full index search,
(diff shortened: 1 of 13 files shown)


mbenadda:
@sampaccoud needs review (can't assign you as you're the one who opened it)

mbenadda:
Good idea @lunika, I had set it back to `default` when disabled but `not-allowed` is probably more explicit!

PR #674, opened 2019-04-24
Show results count in course search

### LINKED ISSUE #455: Show the number of results on the search page
## Feature Request

We want to see the number of results each search query/filter returns.

Display the number of results below the search bar and repeat it at the bottom of the page (only when the number of results is more than... 20?).

mbenadda:
## Purpose

We want to show the number of results a given combination of a text query (or none) and search filters returns.

## Proposal

We can simply show it at the top of the search results, and repeat it at the bottom if the list is long enough to warrant it.

Closes #455

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,8 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 - Add organizations to a person via plugins on the person detail page,
 - Show related persons on the organization detail page,
 - On a person detail page, show courses to which s.he participated and
-  blogposts s.he authored.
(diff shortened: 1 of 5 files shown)


sampaccoud:
I noticed 2 problems:

- On my laptop the bottom count would be useful as soon as the number of results is more than 8,
- The bottom count touches the last result glimpses:

image

mbenadda:
> * On my laptop the bottom count would be useful as soon as the number of results is more than 8,
Changed `20` to `8`

> * The bottom count touches the last result glimpses:
Fixed

PR #680, opened 2019-04-29
Add page roles to ease the assignment of permissions

sampaccoud:
## Purpose

We need to ease the assignment of permissions. For example, when a new organization is created, we need to give a lot of Django permissions and some DjangoCMS page permissions to an administrator in this organization.

## Proposal

My proposal is to build on top of existing Django and DjangoCMS permission functionalities.

I propose to introduce a new model called `PageRole` that links a page with:

- a user group for a given role. This allows us to then grab a user group and assign permissions to it,
- a folder in Django Filer : this allows us to limit what files each user can see/update.

For example the admin group for an organization, should automatically be assigned permissions when a new course is created under this organization.

A lot of work was necessary on the filer and images to achieve what we wanted to do:

- add a new picture plugin that is simpler for the user,
- add template tags and refactor templates to finely configure by code how images are displayed. While we were at it, we made the pictures fully responsive using `srcset`.


src/richie/apps/courses/defaults.py
```diff
@@ -3,6 +3,11 @@
 """
 from django.conf import settings
 
+ADMIN, EDITOR = "ADMIN", "EDITOR"
```
jmaupetit: I would write two separated lines to make them more explicit.

src/richie/apps/courses/models/role.py
```diff
@@ -0,0 +1,91 @@
...
+
+    def create_group(self):
+        """
+        Create a user group for this role and associate it to all default permissions defined in
+        the `PAGE_ROLES` setting.
+
+        This code should only be run once for a given page role so that permissions for
+        this group can then be edited via the Django admin.
+        """
+        assert self.group_id is None
```
jmaupetit: Did you mean?

```python
assert self.group.id is None
```

src/richie/apps/courses/models/role.py
```diff
@@ -0,0 +1,91 @@
...
+        self.group = Group.objects.create(name=str(self))
```
jmaupetit: Using `get_or_create` and allowing to execute it multiple times would allow to automatically reset default permissions if they change for any particular reason. WDYT?

src/richie/apps/courses/models/role.py
```diff
@@ -0,0 +1,91 @@
...
+    def save(
+        self, force_insert=False, force_update=False, using=None, update_fields=None
+    ):
+        """
+        Validate the object before saving it and create a group if it does not exist yet.
+        """
+        self.full_clean(exclude=["group"])
+
+        # Create the related group the first time the instance is saved
+        if not self.group_id:
```
jmaupetit: ?
```python
if not self.group.id:
```

tests/apps/courses/test_models_organization_role.py
```diff
@@ -0,0 +1,106 @@
...
+        self.assertEqual(role.group.permissions.count(), 2)
+
+        # Removing a permission
+        change_page = Permission.objects.get(codename="change_page")
+        role.group.permissions.remove(change_page)
+        self.assertEqual(role.group.permissions.count(), 1)
+        self.assertEqual(
+            [p.codename for p in role.group.permissions.all()],
+            [extra_permission.codename],
+        )
```
jmaupetit: You haven't tested that the `create_group` method is only called once on an instance. Or, if you consider my alternative, we need to check that we reset appropriate permissions once called.

src/richie/apps/core/admin.py
```diff
@@ -0,0 +1,66 @@
...
+        Optional alternative view name to compute the url of the linked object in place of the
+        standard object admin change form.
+    anchor (string):
+        Optional alternative anchor to display on the link in place of the str representation
+        of the linked object.
+
+    Returns
+    -------
+    function
+        The function that Django admin must call with the object as arguement to render the field
```
jmaupetit: `s/arguement/argument/`

src/richie/apps/core/admin.py
```diff
@@ -0,0 +1,66 @@
...
+
+        Returns
+        -------
+        string
+            The html representing the link to the object admin change view.
+
+        """
+        linked_obj = getattr(obj, field_name)
+        if linked_obj is None:
+            return "-"
```
jmaupetit: `N/A` instead?

src/richie/apps/courses/models/role.py
sampaccoud: The `group_id` field is holding the foreign key reference so we save a useless db query.

src/richie/apps/courses/models/role.py
sampaccoud: My idea was to help configure initial permissions and then let users customize them afterwards...

jmaupetit: Don't you think it would be useful to be also able to "reset" permissions at once?

sampaccoud: as discussed IRL, resetting permissions would involve more thinking because the users may have added or removed permission. We need to think of the actual use cases and find appropriate solutions.

src/richie/apps/core/admin.py
sampaccoud: I think "-" is more discreet and does the job.

tests/apps/courses/test_models_organization_role.py
sampaccoud: @jmaupetit I added some checks in the test, please take a look.

tests/plugins/simple_picture/test_models.py
```diff
@@ -0,0 +1,99 @@
...
+
+    @mock.patch.dict(
+        SIMPLEPICTURE_PRESETS, {"my-preset": {"src": {"size": (500, 500)}}}
+    )
+    @mock.patch.object(FilerThumbnailer, "get_thumbnail", return_value=DummyThumbnail())
+    def test_models_simplepicture_get_info_minimal(self, mock_thumbnail):
+        """srcset, alt and sizes are optional attributes in settings."""
+        simple_picture = SimplePictureFactory()
+        info = simple_picture.get_info("my-preset")
+        print(info)
```
sveetch: Ooops !

src/richie/apps/search/indexers/courses.py
```diff
@@ -357,15 +357,15 @@ def get_es_document_for_course(cls, course, index=None, action="index"):
...
         ):
-            # Force the image format before computing it
-            cover_image.use_no_cropping = False
-            cover_image.width = COURSES_COVER_IMAGE_WIDTH
-            cover_image.height = COURSES_COVER_IMAGE_HEIGHT
-            cover_images[cover_image.cmsplugin_ptr.language] = cover_image.img_src
+            language = cover.cmsplugin_ptr.language
+            with translation.override(language):
+                cover_images[language] = cover.get_info(
+                    "glimpse", alt=_("{:s} cover").format(titles[language])
```
mbenadda: I think this is not ideal as eg. a screen reader would just read the title of the course twice? Once for this and once for the actual title? What's the best practice here?

src/richie/apps/courses/defaults.py
```diff
@@ -8,6 +8,99 @@
...
+        "courses.delete_personpluginmodel",
+        "courses.view_personpluginmodel",
+        "courses.add_categorypluginmodel",
+        "courses.change_categorypluginmodel",
+        "courses.delete_categorypluginmodel",
+        "courses.view_categorypluginmodel",
+        "courses.add_licencepluginmodel",
+        "courses.change_licencepluginmodel",
+        "courses.delete_licencepluginmodel",
+        "courses.view_licencepluginmodel",
```
jmaupetit: Can we sort them alphabetically per section?

jmaupetit: And apply the same pattern for every permission set?

src/richie/apps/courses/models/role.py
```diff
@@ -66,6 +76,16 @@ def save(
                 name=str(self)[: Group._meta.get_field("name").max_length]
             )
 
+        # Create the related filer folder the first time the instance is saved.
+        # Why create this folder at the root and not below a parent `organization` folder?
+        # - the only way to refer to an existing parent folder is by its name... but it could be
+        #   changed by a user via the interface and break the functionality,
```
jmaupetit: :1st_place_medal:

src/richie/plugins/simple_picture/defaults.py
```diff
@@ -0,0 +1,50 @@
...
+                {  # list of tuples ({easythumbnail options}, {srcset description})
+                    "options": {"size": (1000, 1000), "crop": "smart"},
+                    "descriptor": "1000w",
+                },
+                {
+                    "options": {"size": (2000, 2000), "crop": "smart"},
+                    "descriptor": "2000w",
+                },
+            ],
+            "sizes": "100vw",  # e.g 1000px ou 100vw
```
jmaupetit: `s/ou/or/`

src/richie/plugins/simple_picture/defaults.py
```diff
@@ -0,0 +1,50 @@
...
+from django.conf import settings
+
+SIMPLEPICTURE_PRESETS = getattr(
+    settings,
+    "RICHIE_SIMPLEPICTURE_PRESETS",
+    {
+        "default": {
+            "src": {"size": (500, 500), "crop": "smart"},  # easythumbnail options
+            "srcset": [
+                {  # list of tuples ({easythumbnail options}, {srcset description})
```
jmaupetit: I don't get your comment: what tuples are you pointing to?

src/richie/plugins/simple_picture/models.py
```diff
@@ -0,0 +1,90 @@
...
+    def get_info(self, preset_name, alt=None):
+        """
+        Compute picture information for a given preset defined in settings.
+
+        A preset is of the form:
+
+        "default": {
+            # easythumbnail options
+            "src": {"size": (1000, 1000), "crop": "smart"},
+            # list of tuples ({easythumbnail options}, {srcset description})
```
jmaupetit: Idem: which tuples?

src/richie/apps/core/admin.py
```diff
@@ -1,6 +1,13 @@
 """Utility functions for admin classes."""
```
jmaupetit: Will you document this part for third-party integration and remove this commit from your PR?

sampaccoud: I was planning to just remove it and add it to our cookie cutter. It is not necessary for 99% of the sites that will use Richie... I just added it here as a POC of what we need to do for our special use case in https://github.com/openfun/fun-mooc.

src/richie/plugins/simple_picture/models.py
sampaccoud: Fixed. This was a comment left from an earlier implementation.

src/richie/plugins/simple_picture/defaults.py
sampaccoud: Fixed. This was a comment left from an earlier implementation.

src/richie/plugins/simple_picture/defaults.py
sampaccoud: Fixed. Thanks!

src/richie/apps/courses/defaults.py
sampaccoud: I don't understand your second comment.

src/richie/apps/search/indexers/courses.py
sampaccoud: You're right I simplified alts. Full optimization of alts and accessibility needs to be done more seriously in a future PR.

src/richie/apps/courses/defaults.py
sampaccoud: After taking a look at it, I propose to leave the order as it is. I had ordered it carefully by semantics. Alphabetical order would make things very difficult to follow here.

src/richie/apps/courses/settings/mixins.py
```diff
@@ -0,0 +1,20 @@
...
+    A Django Configuration mixin to set sensible defaults for the settings of Richie's
+    courses apps.
+    """
+
+    def __new__(cls):
+        """
+        Set all settings as attributes of this class so it can be used as a mixin in
+        Django Configuration.
+        """
+        for key, value in uppercase_attributes(settings).items():
```
jmaupetit: :1st_place_medal:

src/richie/apps/courses/templates/courses/cms/blogpost_detail.html
```diff
@@ -34,6 +34,7 @@ <h1 class="blogpost-detail__title">{% render_model current_page "title" %}</h1>
   {% endwith %}
 
   <div class="blogpost-detail__cover">
+    {% blocktrans with title=current_page.get_title asvar picture_alt %}{{ title }} cover{% endblocktrans %}
```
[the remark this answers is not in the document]
sampaccoud: @jmaupetit there is no more picture_alt (this was a leftover of earlier implementations) and I added a commit to test our alts thoroughly.

PR #684, opened 2019-05-06
🐛(search) fix course indexer when no related organization

sampaccoud:
## Purpose

This bug was identified in the Sentry dashboard of our https://richie.education demo site.

If a course with no organization is indexed in Elasticsearch (which should not happen if moderation is correctly done), the serializer breaks because it expects at least one organization to be declared
on the course. 

Even if this is not a desired situation, the code should not break if it happens.

# Proposal

I propose to set the highlighted organization to `null` in its serialized form if the course has no organization.

I added a test to surface the bug and secure the fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,8 +10,9 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Fixed
 
+- Fix serializer that was failing for a course indexed with no organization,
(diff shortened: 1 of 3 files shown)


jmaupetit:
Don't forget the changelog :wink:

sampaccoud:
@jmaupetit probably doesn't deserve a log WDYT?

jmaupetit:
IMO every bug fix deserves to be logged.

PR #686, opened 2019-05-09
✨(statics) allow to dynamically set webpack publicPath

lunika:
## Purpose

Webpack uses the property `output.publicPath` to import chunks. If we use
a CDN, its url should be set in the `publicPath`
but it's not possible to know it at build time. To fix this issue
the variable __webpack_public_path__ can be changed at runtime and then
used to fetch chunks. See the documentation:
https://webpack.js.org/configuration/output/#outputpublicpath
Define the settings `CDN_DOMAIN` in your project and it will be used
as base domain to load js chunks.

## Proposal

- [x] Configure webpack to dynamically set output.publicPath.
- [x] Create a context processor to set in the context the `CDN_DOMAIN` settings if defined.


sandbox/settings.py
```diff
@@ -133,6 +133,7 @@ class Base(DRFMixin, Configuration):
                     "django.template.context_processors.static",
                     "cms.context_processors.cms_settings",
                     "richie.apps.core.context_processors.site_metas",
+                    "richie.apps.core.context_processors.cdn_domain",
```
sampaccoud: Actually, you could add it to the `site_metas` context processor... it is exactly the idea behind it.
Adding this context processor to your project is required for Richie to work properly so let's keep it simple and group everything under the same processor?

src/frontend/webpack.config.js
```diff
@@ -9,15 +9,20 @@ module.exports = {
   // Currently, @babel/preset-env is unaware that using import() with Webpack relies on Promise internally.
   // Environments which do not have builtin support for Promise, like Internet Explorer, will require both
   // the promise and iterator polyfills be added manually.
+  //
+  // chunkFilename must have a unique and different name on each build
+  // this will prevent to replace existing chunk if backend static storage
```
mbenadda: ...will prevent overwriting of existing chunks...

sandbox/settings.py
lunika: :ok_hand:

PR #688, opened 2019-05-10
🔧(project) improve Makefile readability

jmaupetit:
## Purpose

While developing Riche, the project's `Makefile` became quite big and messy. It was time to invest some efforts to make it more readable / homogeneous.

## Proposal

We've updated the project's `Makefile` with the following statements in mind:

1. Every variable should be defined in the _ad hoc_ **VARIABLES** section with a relevant subsection
2. Every new rule should be defined in the _ad hoc_ **RULES** section with a relevant subsection depending on the targeted service
3. Rules should be sorted alphabetically within their section
4. When a rule has multiple dependencies, you should:
   - duplicate the rule name to add the help string (if required)
   - write one dependency per line to increase readability and diffs
5. `.PHONY` rule statement should be written after the corresponding rule

I am open to other rules to apply to make this `Makefile` more easily readable/maintainable.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,10 +8,28 @@
 # PLEASE DO NOT USE IT FOR YOUR CI/PRODUCTION/WHATEVER...
 #
 # /!\ /!\ /!\ /!\ /!\ /!\ /!\ /!\ /!\ /!\ /!\ /!\ /!\ /!\ /!\ /!\ /!\ /!\
+#
+# Note to developpers:
+#
+# While editing this file, please respect the following statements:
+#
+# 1. Every variable should be defined in the ad hoc VARIABLES section with a
+#    relevant subsection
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -33,124 +53,156 @@ COMPOSE_TEST_RUN     = $(COMPOSE) run --rm -e DJANGO_CONFIGURATION=Test
...
+lint-front: \
+  lint-front-tslint \
+  lint-front-prettier
+.PHONY: lint-front
+
+lint-front-tslint: ## lint TypeScript sources
+	@$(YARN) lint
+.PHONY: lint-front-tslint
+
+lint-front-prettier: ## run prettier over js/jsx/json/ts/tsx files -- beware! overwrites files
```
lunika: should be before `lint-front-tslint`

Issue #664, opened 2019-04-18
Add an index for persons


sampaccoud:
## Feature Request

**Is your feature request related to a problem or unsupported use case? Please describe.**
We would like to be able to search persons the same way we search organizations and categories.

**Describe the solution you'd like**
- [x] Create a new index for persons
- [x] Add persons to the autocomplete suggestions in the main course search
- [x] Add a field `persons_names` in the courses index to allow matching courses with the name of a person in their team
- [x] Add a persons filter to the search pane

PR #690, opened 2019-05-10
Index persons themselves & on courses

mbenadda:
## Purpose

We want users to be able to find courses by persons (in the UI) and to find persons themselves through the API (for now).

## Proposal

- [x] Create a persons index that opens up the possibility to make a person search, and that makes available persons autocompletion right now.
- [x] Index related persons on courses to users can find courses by related persons through full text search.

Handles boxes 1 and 3 from #664

### DIFF
--- a/src/richie/apps/search/forms.py
+++ b/src/richie/apps/search/forms.py
@@ -315,11 +315,7 @@ class ItemSearchForm(SearchForm):
         # Build a query that matches on the name field if it was handed by the client
         full_text = self.cleaned_data.get("query")
         if full_text:
-            query = {
-                "query": {
-                    "match": {"title.fr": {"query": full_text, "analyzer": "french"}}
-                }
-            }
+            query = {"query": {"match": {"title.fr": {"query": full_text}}}}
         # Build a match_all query by default
         else:
             query = {"query": {"match_all": {}}}
--- a/src/richie/apps/search/forms.py
+++ b/src/richie/apps/search/forms.py
@@ -221,6 +221,18 @@ class CourseSearchForm(SearchForm):
                                             "type": "cross_fields",
                                         }
                                     },
+                                    {
+                                        "multi_match": {
+                                            "boost": getattr(
+                                                settings,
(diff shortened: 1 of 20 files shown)


src/richie/apps/search/forms.py
```diff
@@ -317,7 +329,7 @@ def build_es_query(self):
         if full_text:
             query = {
                 "query": {
-                    "match": {"title.fr": {"query": full_text, "analyzer": "french"}}
+                    "match": {"title.fr": {"query": full_text, "analyzer": "simple"}}
```
sampaccoud: this would still not be accurate for organizations and categories. Actually we should just remove the analyzer spec from this query so that it just uses the same analyzer as specified in the mapping for each object...

tests/apps/search/test_query_courses.py
```diff
@@ -1073,6 +1081,19 @@ def test_query_courses_match_all_no_filter_pages(self, *_):
         self.assertEqual(content["filters"]["levels"]["values"], [])
         self.assertEqual(content["filters"]["organizations"]["values"], [])
 
+    def test_query_courses_by_related_person(self, *_):
+        """
+        Full-text search appropriately returns the list of courses that match a given
+        related person name even through a text query.
+        """
+        courses_definition, content = self.execute_query("query=boulgakov")
+        # Keep only the courses that contain the word "artificial"
```
sampaccoud: that are linked to persons whose name contains "boulgakov"?

mbenadda:
@sampaccoud I think there may be something wrong with the way I index persons or the way I index person IDs. I think I'm not indexing the same IDs in both cases.

I tried replacing `"persons": [str(person.id) for person in persons]` with `"persons": [str(person.extended_object_id) for person in persons]` in the courses indexer but no luck so far.

Maybe you have a pointer here?

sampaccoud:
Yes!
`course.get_persons()` always returns the draft instance (because it's the one pointed by the plugin. Maybe we should make it clearer in the docstring?).
So in the course indexer we should write:
```python
"organizations": [person.public_extension.id for person in persons]
```

PR #693, opened 2019-05-13
Add command to initialize site structure

sampaccoud:
## Purpose

We had a management command to create full demo sites. But all Richie need to have required pages under which the different objects will be created (courses, organizations, persons, blog posts...). We need a command line to create only these required pages.

## Proposal

I refactored the management command that creates a demo site and added a new command line to create just the site structure. I placed it in the `courses` app because all sites need it.

### DIFF
--- a/src/richie/apps/courses/defaults.py
+++ b/src/richie/apps/courses/defaults.py
@@ -6,3 +6,53 @@ from django.conf import settings
 PAGE_EXTENSION_TOOLBAR_ITEM_POSITION = getattr(
     settings, "RICHIE_PAGE_EXTENSION_TOOLBAR_ITEM_POSITION", 4
 )
+
+# For each type of page we define:
+#   - the `reverse_id` of the page under which pages should be created via the wizard,
+#   - the template to be used when creating a new page of this type,
+BLOGPOSTS_PAGE = {
+    "reverse_id": "blogposts",
+    "template": "courses/cms/blogpost_detail.html",
+}
+CATEGORIES_PAGE = {
+    "reverse_id": "categories",
+    "template": "courses/cms/category_detail.html",
+}
+COURSERUNS_PAGE = {"template": "courses/cms/course_run_detail.html"}
+COURSES_PAGE = {"reverse_id": "courses", "template": "courses/cms/course_detail.html"}
+ORGANIZATIONS_PAGE = {
+    "reverse_id": "organizations",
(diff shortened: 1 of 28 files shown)


src/richie/apps/courses/defaults.py
```diff
@@ -6,3 +6,50 @@
 PAGE_EXTENSION_TOOLBAR_ITEM_POSITION = getattr(
     settings, "RICHIE_PAGE_EXTENSION_TOOLBAR_ITEM_POSITION", 4
 )
+
+BLOGPOSTS_PAGE = {
```
jmaupetit: Can you add a comment to this section explaining the purpose of those defaults?

src/richie/apps/demo/management/commands/create_demo_site.py
```diff
@@ -33,10 +33,10 @@
     SINGLECOLUMN_CONTENT,
     SUBJECTS_INFO,
 )
-from ...helpers import create_categories, recursive_page_creation
+from ...helpers import create_categories
 from ...utils import pick_image
 
-logger = logging.getLogger("richie.commands.core.create_demo_site")
+logger = logging.getLogger("richie.demo.commands.create_demo_site")
```
jmaupetit: Maybe we should adopt a logger naming convention such as `richie.{{ purpose }}.xxx` and not follow the related file path, WDYT?

```suggestion
logger = logging.getLogger("richie.commands.demo.create_demo_site")
```

src/richie/apps/courses/management/commands/richie_init.py
```diff
@@ -0,0 +1,22 @@
...
+import logging
+
+from django.contrib.sites.models import Site
+from django.core.management.base import BaseCommand
+
+from richie.apps.core.helpers import recursive_page_creation
+
+from ...defaults import PAGES_INFO
+
+logger = logging.getLogger("richie.core.commands.richie_init")
```
jmaupetit:
```suggestion
logger = logging.getLogger("richie.commands.core.richie_init")
```

src/richie/apps/courses/defaults.py
sampaccoud: done

src/richie/apps/demo/management/commands/create_demo_site.py
sampaccoud: I changed to `richie.commands.demo.create_demo_site` as discussed.

PR #696, opened 2019-05-15
🐛(demo) fix and test the create_demo_site management command

sampaccoud:
## Purpose

PR https://github.com/openfun/richie/pull/693 broke the `create_demo_site` management command. 

It's not the first time we break this command because it is not tested. This is historical because it takes too long to run.

## Proposal

I fixed the command and added a test to secure it, bringing it down to 4-5 seconds by mocking the number of objects that are created. It's still too long but at least we can be sure the command works.


jmaupetit:
Don't forget the CHANGELOG!

sampaccoud:
@jmaupetit this time there's really no change log because it's fixing a bug introduced in the same release.

PR #697, opened 2019-05-15
Add persons to course search filters and autosuggest

### LINKED ISSUE #664: Add an index for persons

mbenadda:
## Purpose

Allows users to filter courses in course Search by related persons. Said persons are also now part of the suggestions in the main search bar.

## Proposal

- [x] refactor resource names & suggestions on the frontend to allow new existing kinds of objects (persons now, different kinds of categories soon)
- [x] have suggestions on the API specify the kind of object they're suggesting
- [x] enable the persons filter by adding the filter definition to our `defaults`
- [x] suggest values from the person filters in the main search bar

Closes #664.

NB: we still have some refactoring to do to make filters in autosuggest completely dynamic, which will help remove hardcoded stuff from the frontend and use `settings`/`defaults` from the backend instead.


src/richie/apps/search/filter_definitions/courses.py
```diff
@@ -202,6 +202,15 @@ def get_faceted_definitions(self, facets, *args, **kwargs):
         }
 
 
+class PersonsFilterDefinition(IndexableFilterDefinition):
+    """Persons are a special case of terms-based filter as they do not use MPTT paths."""
```
sampaccoud: Is it easy to make persons use IndexableFilterDefintion and the 2 others use a derived class that adds agg filtering?
Keep in mind that we will soon refactor nested indexables to not use path anymore...

mbenadda:
@sampaccoud I fixed the tests and addressed your comment.

I'm not sure how necessary it was as that bit will be refactored away sometime soon as you say, but it was an easy fix so I did it anyway.

sampaccoud:
@mbenadda nice! It will definitely be easier to get started with refactoring than the way it was.

src/frontend/js/components/SearchSuggestField/getSuggestionsSection.ts
```diff
@@ -15,7 +13,7 @@ import { handle } from '../../utils/errors/handle';
  * @param sectionTitleMessage MessageDescriptor for the title of the section that displays the suggestions.
  * @param query The actual payload to run the completion search with.
  */
-export const getSuggestionsSection = async <Kind extends modelName>(
+export const getSuggestionsSection = async <Kind extends string>(
```
lunika: I don't know if it's still needed to capture `Kind`. It's just a string now.

src/frontend/js/settings.ts
```diff
@@ -1,15 +1,13 @@
-import { modelName } from './types/models';
-
-export const API_ENDPOINTS = {
+const API_ENDPOINTS = {
```
lunika: Why are you keeping this variable ? It is not exported and not used at all. You should maybe remove it ?

mbenadda: Good catch!

PR #703, opened 2019-05-23
Autocompletion is dynamic based on filter definitions

mbenadda:
## Purpose

Autocompletion for filters in courses search was hardcoded through strings in an array in the `<SearchSuggestField />` component.

This is not appropriate as we want to make richie easy to personalize, and it also breaks down if a user disables the filter related to one of the hardcoded autocompletion filters.

## Proposal

We want to make it dynamic based on backend settings and actual CMS content.

- [x] categories have a convenience method to retrieve the meta category
- [x] categories are indexed with a "kind" field to differentiate eg. subjects from levels
- [x] API endpoints are open for eg subjects and levels and filter search & autocomplete results based on kind
- [x] autocomplete on the frontend operates only on filters which are marked `is_autocompletable` on the API


src/richie/apps/courses/models/category.py
```diff
@@ -45,6 +45,24 @@ def __str__(self):
...
+        """
+        # Shorcut to the category's page node
+        node = self.extended_object.node
+
+        return Category.objects.filter(
+            extended_object__publisher_is_draft=False,
+            extended_object__node__path__in=[
+                node.path[0:pos] for pos in range(0, len(node.path), node.steplen)[1:]
+            ],
+            extended_object__node__parent__cms_pages__publisher_is_draft=False,
```
sampaccoud: add a comment that this is to avoid duplicates due to the join on cms_pages?

src/richie/apps/courses/models/category.py
```diff
@@ -45,6 +45,24 @@ def __str__(self):
...
+        """
+        Get the meta category the current category falls under. Meta-categories are a special kind
+        of categories that are not linked to courses but help us organize categories into
+        different kinds that are handled separately.
+        """
+        # Shorcut to the category's page node
+        node = self.extended_object.node
+
+        return Category.objects.filter(
+            extended_object__publisher_is_draft=False,
```
sampaccoud: Second thoughts:
1) it is akward to always return the public version: we should probably return the draft meta category for a draft category, and the published meta category for a published category...
2) we should probably not return a public version if it is not published in any language. This would allow us to not index a published category placed below a meta category that is not published? something like what is done here: https://github.com/openfun/richie/blob/master/src/richie/apps/courses/models/course.py#L141

sampaccoud: FTR, point 2 is actually a more general issue that must be addressed in another PR: https://github.com/openfun/richie/issues/704

src/richie/apps/courses/models/category.py
```diff
@@ -45,6 +45,26 @@ def __str__(self):
...
+        """
+        Get the meta category the current category falls under. Meta-categories are a special kind
+        of categories that are not linked to courses but help us organize categories into
+        different kinds that are handled separately.
+        """
+        # Shorcut to the category's page node
+        node = self.extended_object.node
+
+        self_is_draft = self.extended_object.publisher_is_draft
+        return Category.objects.filter(
```
sampaccoud:
```suggestion
        return Category.objects.get(
```

src/richie/apps/courses/models/category.py
```diff
@@ -45,6 +45,26 @@ def __str__(self):
...
+            extended_object__publisher_is_draft=self_is_draft,
+            extended_object__node__path__in=[
+                node.path[0:pos] for pos in range(0, len(node.path), node.steplen)[1:]
+            ],
+            extended_object__node__parent__cms_pages__category__isnull=True,
+            # Avoid duplicates from the join on cms_pages
+            extended_object__node__parent__cms_pages__publisher_is_draft=self_is_draft,
+        )[0]
```
sampaccoud:
```suggestion
        )
```

src/richie/apps/search/indexers/categories.py
```diff
@@ -101,6 +106,12 @@ def get_es_document_for_category(cls, category, index=None, action="index"):
         # Shorcut to the category's page node
         node = category.extended_object.node
 
+        # Find the meta category the current category falls under
+        try:
+            kind = category.get_meta_category().extended_object.reverse_id
+        except IndexError:
```
sampaccoud:
```suggestion
        except Category.DoesNotExit:
```

+ add comment
`
The category as no meta which means it is a meta itself. Let's give it a None kind.
`

tests/apps/courses/test_models_category.py
```diff
@@ -54,6 +54,60 @@ def test_models_category_str(self):
...
+            "Subjects",
+            "courses/cms/category_detail.html",
+            "en",
+            parent=not_a_category_page,
+            published=True,
+        )
+        meta_category = CategoryFactory(
+            extended_object=meta_category_page, should_publish=True
+        )
+
```
sampaccoud: Add:
```python
with self.assertRaises(Category.DoesNotExit):
    meta_category.get_meta_category()
```

PR #707, opened 2019-05-24
⚡️(helpers) make recursive_page_creation idempotent

jmaupetit:
## Purpose

When deploying Richie, we need to run management commands in jobs. We need those jobs to be idempotent. The recently introduced `richie_init` management command is not.

## Proposal

- [x] ensure that the `recursive_page_creation` helper can be run multiple times without failing or creating duplicates

### DIFF
--- a/tests/apps/core/test_helpers.py
+++ b/tests/apps/core/test_helpers.py
@@ -5,6 +5,7 @@ import random
 from unittest import mock
 
 from django.contrib.auth.models import Permission
+from django.contrib.sites.models import Site
 from django.test import TestCase
 from django.test.utils import override_settings
 
@@ -12,7 +13,12 @@ from cms.api import Page
 from cms.test_utils.testcases import CMSTestCase
 
 from richie.apps.core.factories import PermissionFactory
-from richie.apps.core.helpers import create_i18n_page, get_permissions
+from richie.apps.core.helpers import (
+    create_i18n_page,
+    get_permissions,
+    recursive_page_creation,
+)
+from richie.apps.courses.defaults import PAGES_INFO
 
 
 class GetPermissionsHelpersTestCase(TestCase):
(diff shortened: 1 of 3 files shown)


tests/apps/core/test_helpers.py
```diff
@@ -219,3 +225,120 @@ def test_helpers_create_i18n_page_published(self, mock_publish):
...
+        self.assertEqual(Page.objects.filter(reverse_id="organizations").count(), 2)
+        self.assertEqual(Page.objects.filter(reverse_id="persons").count(), 2)
+        self.assertEqual(Page.objects.count(), 12)
+
+    def test_helpers_recursive_page_creation_recursiveness(self):
+        """Test embedded pages creation."""
+        site = Site.objects.get(id=1)
+
+        pages_info = PAGES_INFO.copy()
+        pages_info.update(
```
sampaccoud: This is quite a slow test. So I think you should test only one embedded page here. The above test for the full PAGES_INFO is also testing on too many pages.

tests/apps/core/test_helpers.py
```diff
@@ -219,3 +225,120 @@ def test_helpers_create_i18n_page_published(self, mock_publish):
...
+        self.assertEqual(Page.objects.filter(reverse_id="persons").count(), 2)
+        self.assertEqual(Page.objects.count(), 12)
+
+    def test_helpers_recursive_page_creation_can_be_run_multiple_times(self):
+        """Ensure we can run the recursive_page_creation helper multiple times with the same
+        parameters without failing or creating new pages."""
+        site = Site.objects.get(id=1)
+        self.assertEqual(Page.objects.count(), 0)
+        recursive_page_creation(site=site, pages_info=PAGES_INFO)
+        recursive_page_creation(site=site, pages_info=PAGES_INFO)
```
sampaccoud: this is quite slow to run so maybe you can just add it at the end of the previous test? I would also reduce this test to 1 or 2 pages...

jmaupetit: Ok, I'll keep a single test of the full `PAGES_INFO` and use smaller pages set for other tests.

PR #717, opened 2019-06-03
🐛(permissions) fix page permissions that were making pages private

sampaccoud:
## Purpose

The `PagePermission` model is actually a 2-headed snake... it can act as a permission if `can_view` is False but acts as a view restriction if `can_view` is True.

Said differently, the purpose of the `can_view` flag is not to define that some staff users have the right to view the page... its purpose is to define who can view the page on the frontend.

What we want here is a permission on admin pages for our staff users so `can_view` should be False.

## Proposal

- [x] write tests to surface the problem,
- [x] set `can_view` to False on all our page permissions.

### DIFF
--- a/src/richie/apps/courses/defaults.py
+++ b/src/richie/apps/courses/defaults.py
@@ -89,7 +89,7 @@ COURSE_ADMIN_ROLE = {
         "can_publish": False,
         "can_change_permissions": False,
         "can_move_page": True,
-        "can_view": True,
+        "can_view": False,  # beware: can_view = True makes it a view restriction...
         "grant_on": 5,  # page and descendants
     },
     "course_folder_permissions": {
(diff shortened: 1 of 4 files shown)


src/richie/apps/courses/defaults.py
```diff
@@ -89,7 +89,7 @@
         "can_publish": False,
         "can_change_permissions": False,
         "can_move_page": True,
-        "can_view": True,
+        "can_view": True,  # beware: can_view = True makes it a view restriction...
```
jmaupetit: So this means that course page access is restricted; anonymous users can search and filter courses from the courses page view, but cannot click on a page to get a detailed view of a course?

sampaccoud: sorry that's a typo... these defaults are not tested by my test because the test includes a mock. I fixed it.

PR #731, opened 2019-06-19
♻️(persons) replace firstname, lastname and person title by page title

sampaccoud:
## Purpose

Differentiating a eperson's name into first name and last name is a bad practice. 

Second, our first_name, last_name and person_title fields were not kept up to date with the page title. 

Finally, what we are doing here is a website, not a directory... so the page title is a perfect field
to store the full person name...

## Proposal

In this commit, I remove some models and fields but don't reflect it in database migrations. This will need to be done in a further release so that blue/green deployments don't break.


sampaccoud:
@mbenadda done https://github.com/openfun/richie/issues/732

PR #737, opened 2019-06-24
Improve label for child filter toggle

mbenadda:
## Purpose

The child filter toggle in the course search was deemed compliant, but the label for the toggle is not very explicit, especially for a screen reader user that cannot see its context.

## Proposal

Instead of "Show child filters", we can have it say "Show more filters for Subject X", which is way more explcit. Ditto for the other toggle position.

### DIFF
--- a/src/frontend/js/components/SearchFilterValueParent/SearchFilterValueParent.tsx
+++ b/src/frontend/js/components/SearchFilterValueParent/SearchFilterValueParent.tsx
@@ -12,13 +12,13 @@ import { SearchFilterValueLeaf } from '../SearchFilterValueLeaf/SearchFilterValu
 
 const messages = defineMessages({
   ariaHideChildren: {
-    defaultMessage: 'Hide child filters',
+    defaultMessage: 'Hide additional filters for {filterValueName}',
     description:
       'Accessibility message for the button to hide children of the current filter',
     id: 'components.SearchFilterValueParent.ariaHideChildren',
   },
   ariaShowChildren: {
-    defaultMessage: 'Show child filters',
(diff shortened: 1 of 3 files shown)


src/frontend/js/components/SearchFilterValueParent/SearchFilterValueParent.tsx
```diff
@@ -12,13 +12,13 @@ import { SearchFilterValueLeaf } from '../SearchFilterValueLeaf/SearchFilterValu
...
   ariaHideChildren: {
-    defaultMessage: 'Hide child filters',
+    defaultMessage: 'Hide additional filters for {filterValueName}',
     description:
       'Accessibility message for the button to hide children of the current filter',
     id: 'components.SearchFilterValueParent.ariaHideChildren',
   },
   ariaShowChildren: {
-    defaultMessage: 'Show child filters',
+    defaultMessage: 'Show more filters for {filterValueName}',
```
sampaccoud: or subfilters ? I'm thinking `sous-filtres` in french?

mbenadda: "Show more filters for Mathematics" // "Voir plus de filtres pour Mathématiques"
"Show Mathematics subfilters" // "Voir les sous-filtres de Mathématiques"

The first one sounds simpler to me (even though it's less precise to the ears of Richie-experts like ourselves)

sampaccoud: According to the poll, your proposal is the best!

PR #740, opened 2019-06-26
Feedback from support week 26

sampaccoud:
## Purpose

This work follows feedback from our support team and from the accessibility audit.

## Proposal

- Add a new placeholder "catch phrase" on the course,
- Extend the plain text and simple text plugins to allow configuring their max length and CKEditor on specifc placeholders,
- Configure max length and CKEditor on course, person and blog post placeholders, 
- solve padding issues on the course licence,
- change the person bio to a plain text plugin,
- update README to the latest situation
- fix empty page on the annex link at the bottom of the sitemap


src/frontend/scss/templates/courses/cms/_course_detail.scss
```diff
@@ -12,7 +12,7 @@ $richie-course-detail-title-underline-border: 1px solid $gray40 !default;
 
 $richie-course-detail-content-width-large: 70% !default;
 $richie-course-detail-content-margin: 0 !default;
-$richie-course-detail-content-padding: 0 !default;
+$richie-course-detail-content-padding: 0 0 30px 0 !default;
```
mbenadda: You should use `rem`

src/frontend/scss/templates/courses/plugins/_licence_plugin.scss
```diff
@@ -20,8 +20,13 @@ $richie-licence-plugin-description-fontcolor: $gray40 !default;
   }
 
   &__wrapper {
+    margin-left: 10px;
```
mbenadda: Same here, better to use `rem`

src/frontend/scss/templates/courses/plugins/_licence_plugin.scss
```diff
@@ -20,8 +20,13 @@ $richie-licence-plugin-description-fontcolor: $gray40 !default;
...
     @include media-breakpoint-up($richie-licence-plugin-breakpoint) {
-      @include sv-flex(1, 0, calc(100% - #{$richie-licence-plugin-logo-width}));
+      @include sv-flex(
+        1,
+        0,
+        calc(100% - #{$richie-licence-plugin-logo-width + 10})
```
mbenadda: This `10` comes from the margin above right? It should then be set in a variable and reused

sampaccoud: @mbenadda my problem is that the logo width is expressed in px so I get the error:

> "Incompatible units: 'rem' and 'px'."

sampaccoud: So I left px as discussed.

PR #749, opened 2019-07-02
🐛(courses) fix getting objects related to a course via plugins

sampaccoud:
## Purpose

On the course model we have getters to retrieve persons, categories and organizations related to a course. 

They were not filtered by language so it was mixing up results between languages. This was especially buggy when retrieving the main organization.

## Proposal

I added a filter on the language to the 3 concerned getters on the course model: `get_organizations`, `get_categories` and `get_persons`.


sampaccoud:
@jmaupetit Normally no impact since this join was already done?

PR #758, opened 2019-07-11
Add icon categories

sampaccoud:
## Purpose

We want to be able to tag courses with icons.
We also want to be able to display a chosen icon on the course glimpse.

## Proposal

I chose to use the existing category page to do icons. 

- add an `icon` placeholder to the category page
- add a `course_icons` placeholder to the course page
- add icons to the category and course search indexes
- add icons to the demo site

While working on this feature I got to fix and improve a few things:

- fix the `get_placeholder_plugins` template tag that was failing when the placeholder was missing
- free course getters from placeholder names on all objects
- free course related objects getters from placeholder names


src/richie/apps/search/indexers/courses.py
```diff
@@ -370,7 +373,22 @@ def get_es_document_for_course(cls, course, index=None, action="index"):
         ):
             language = cover.cmsplugin_ptr.language
             with translation.override(language):
-                cover_images[language] = get_picture_info(cover, "glimpse")
+                cover_images[language] = get_picture_info(cover, "cover")
+
+        # Prepare icon
```
mbenadda:
```suggestion
        # Prepare the related category icon
```

src/richie/apps/demo/management/commands/create_demo_site.py
```diff
@@ -173,6 +180,9 @@ def create_demo_site():
                 ),
                 random.choice(levels),
             ],
+            fill_icons=random.sample(
```
mbenadda: IMO not every course should have an icon?

src/richie/apps/courses/models/category.py
```diff
@@ -26,6 +26,8 @@ class Category(BasePageExtension):
     page that presents the thematic.
     """
 
+    color = models.CharField(max_length=10, blank=True, null=True)
```
mbenadda: I think 10 is too short a length. If we want to use eg. colors with transparency like `rgba(255,255,255,0.75)` it would not hurt to have a `max_length` of eg. 25

sampaccoud: The color is stored as hexadecimal. With transparency it is of the form #24c4a2aa, so 9 characters.

This being said, the default html5 widget I used does not seem to handle transparency and we don't have the use case for it at the moment.

PR #762, opened 2019-07-12
👷(linters) lint backend source code with bandit

sampaccoud:
## Purpose

We want to lint our Python source code with [bandit](https://github.com/PyCQA/bandit).

## Proposal

- Add bandit as dev dependency
- Run bandit in the CI as a linter
- Add a make target to run bandit as a linter
- Fix or ignore warnings in order to get all green on all severity levels (low, medium and high)


jmaupetit:
Also, I think it is worth to be logged in the project changes.

PR #772, opened 2019-07-30
Move the search template title into react

mbenadda:
## Purpose

Currently the Django template is responsible for displaying the page title on the search results, like on any other page. This is good in that it allows this title to be managed in the CMS. However, it is
troublesome in terms of page structure as we want to handle the search filters pane more cleanly.

## Proposal

To keep the title in the CMS while improving our page structure, we decided to pass the page title to the React frontend via a data attribute in the Django template. This lets us cleanly insert it in the relevant place in our markup.

In our opinion, this is a better compromise than hardcoding it in React (as we lose CMS management of the title, which should not happen if we can avoid it), or adding hacks to the CSS which can be more harmful in the longer term.

### DIFF
--- a/src/frontend/js/components/Search/Search.tsx
+++ b/src/frontend/js/components/Search/Search.tsx
@@ -11,7 +11,11 @@ import { SearchFiltersPane } from '../SearchFiltersPane/SearchFiltersPane';
 import { SearchLoader } from '../SearchLoader/SearchLoader';
 import { SearchSuggestField } from '../SearchSuggestField/SearchSuggestField';
 
-export const Search = () => {
+interface SearchProps {
+  pageTitle?: string;
+}
+
+export const Search = ({ pageTitle }: SearchProps) => {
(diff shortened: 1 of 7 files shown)


mbenadda:
Check this out @deodat

src/frontend/js/components/Search/Search.tsx
```diff
@@ -30,20 +34,23 @@ export const Search = () => {
...
-            <CourseGlimpseList
-              courses={courseSearchResponse.content.objects}
-              meta={courseSearchResponse.content.meta}
-            />{' '}
-          </div>
-        ) : (
-          <SearchLoader />
-        )}
+        <div className="search__results">
+          {pageTitle ? <h1 className="search__results__title">{pageTitle}</h1> : null}
```
lunika: Why not doing this:
```
{pageTitle && (<h1 className="search__results__title">{pageTitle}</h1>)}
```

PR #777, opened 2019-08-06
Scale facet filters

mbenadda:
## Purpose

In the search filters pane, we display the top N (by default 10) values for each filter, ordered by facet count. It makes sense to show the values with the most matching courses more prominently, but users might want to filter by other values, especially on Richie instances that host a large number of courses.

## Proposal

Add a button on said search filters that opens a modal with a search field that is specialized to that filter.

This will let users easily add any value for a given filter through a full-text search.

### DIFF
--- a/src/frontend/js/components/SearchFilterGroup/SearchFilterGroup.tsx
+++ b/src/frontend/js/components/SearchFilterGroup/SearchFilterGroup.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 
 import { FilterDefinition } from '../../types/filters';
+import { SearchFilterGroupModal } from '../SearchFilterGroupModal';
 import { SearchFilterValueLeaf } from '../SearchFilterValueLeaf/SearchFilterValueLeaf';
 import { SearchFilterValueParent } from '../SearchFilterValueParent/SearchFilterValueParent';
 
@@ -30,5 +31,8 @@ export const SearchFilterGroup = ({ filter }: SearchFilterGroupProps) => (
         ),
       )}
     </div>
+    {filter.is_autocompletable ? (
+      <SearchFilterGroupModal filter={filter} />
+    ) : null}
   </fieldset>
 );
--- a/src/frontend/js/components/SearchFilterGroupModal/index.tsx
+++ b/src/frontend/js/components/SearchFilterGroupModal/index.tsx
@@ -4,7 +4,7 @@ import {
   FormattedMessage,
   MessageDescriptor,
 } from 'react-intl';
-import Modal from 'react-modal';
+import ReactModal from 'react-modal';
 
 import { fetchList } from '../../data/getResourceList/getResourceList';
 import { CourseSearchParamsContext } from '../../data/useCourseSearchParams/useCourseSearchParams';
@@ -42,6 +42,14 @@ const messages = defineMessages({
     id: 'components.SearchFilterGroupModal.moreOptionsButton',
   },
 });
+
+// The `setAppElement` needs to happen in proper code but breaks our testing environment.
+// This workaround is not satisfactory but it allows us to both test <SearchFilterGroupModal />
+// and avoid compromising accessibility in real-world use.
+const isTestEnv = typeof jest !== 'undefined';
+if (!isTestEnv) {
+  ReactModal.setAppElement('#modal-exclude');
+}
 
 export const SearchFilterGroupModal = ({
   filter,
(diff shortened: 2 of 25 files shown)


src/frontend/js/components/SearchFilterGroupModal/index.tsx
```diff
@@ -0,0 +1,141 @@
...
+  const [error, setError] = useState(null as Nullable<
+    FormattedMessage.MessageDescriptor
+  >);
+
+  // We need the current course search params to get the facet counts
+  const [coursesSearchParams, dispatchCourseSearchParamsUpdate] = useContext(
+    CourseSearchParamsContext,
+  );
+
+  useAsyncEffect(async () => {
```
lunika: When the modal is closed and the reopen, it has the same state. I don't know if this is want you want but you should maybe reset the state otherwise

src/frontend/js/components/SearchFilterGroupModal/index.tsx
```diff
@@ -22,6 +22,14 @@ const messages = defineMessages({
   },
 });
 
+// The `setAppElement` needs to happen in proper code but breaks our testing environment.
+// This workaround is not satisfactory but it allows us to both test <SearchFilterGroupModal />
+// and avoid compromising accessibility in real-world use.
+const isTestEnv = !!jest;
```
lunika: This is not working, I have an error in my browser: `ReferenceError: jest is not defined`

src/frontend/js/components/SearchFilterGroup/SearchFilterGroup.tsx
```diff
@@ -30,5 +31,8 @@ export const SearchFilterGroup = ({ filter }: SearchFilterGroupProps) => (
         ),
       )}
     </div>
+    {filter.is_autocompletable ? (
+      <SearchFilterGroupModal filter={filter} />
```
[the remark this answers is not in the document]
mbenadda: I think so too. The thing is we have no way to know that from the client side right now. I was planning to handle this in another PR as it's mostly backend

src/frontend/js/components/SearchFilterGroupModal/index.tsx
```diff
@@ -0,0 +1,141 @@
...
+      offset: '0',
+      query,
+    });
+
+    if (searchResponse.status === requestStatus.FAILURE) {
+      setValues([]);
+      return setError(messages.error);
+    }
+
+    const facetResponse = await fetchList('courses', {
```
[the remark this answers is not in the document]
mbenadda: Yes, the only way to have facet counts is to make a request on the courses endpoint. The scope prevents the server from actually getting and returning all the courses objects

src/richie/apps/core/templates/richie/base.html
```diff
@@ -9,66 +9,70 @@
...
-                    </li>
-                </ul>
-            </div>
-        </nav>
-        {% endblock body_header %}
 
-        <div class="body-content {% block bodycontent_classes %}{% endblock bodycontent_classes %}">
-        {% block content %}{% endblock content %}
-        </div>
+        {# Wrap all body content in an ID'd element to easily remove it from view when a modal is open #}
```
sampaccoud: I think you should explain that this is important for accessibility.

tests/apps/search/test_query_categories.py
```diff
@@ -0,0 +1,134 @@
...
+                        "id": "8421",
+                        "is_meta": False,
+                        "logo": "en/logo",
+                        "nb_children": 0,
+                        "path": "8421",
+                        "title": "Beginner",
+                    }
+                ],
+            },
+        )
```
sampaccoud: Maybe add accents to your content to make the same "diacritic insensitive" tests as what you did on the autocomplete endpoint?

sampaccoud: You should also test the other way around ie that
```python
content = self.execute_query(kind="subjects", querystring="query=Bégînnêr")
```
returns ID 8399.

mbenadda: Done 👍

tests/apps/search/test_query_persons.py
```diff
@@ -0,0 +1,119 @@
...
+]
+
+
+@mock.patch.object(  # Avoid messing up the development Elasticsearch index
+    PersonsIndexer,
+    "index_name",
+    new_callable=mock.PropertyMock,
+    return_value="test_persons",
+)
+class OrganizationsQueryTestCase(TestCase):
```
sampaccoud: Copy/pasta

mbenadda:
Note: part of #376, along with #790

PR #779, opened 2019-08-12
Paginate course search

### LINKED ISSUE #754: Add pagination to the search page
## Feature Request

I recently built a site with >500 courses. With no filters applied, there are too many results. 
On a good laptop with a good connection, it works fine but it will be problematic on mobiles, old laptops or slow connections...

We need to:

- add Google style pagination to search results,
- show less results per page for narrow screens and more on large screens.

We could consider infinite scroll but it would raise many more questions:

- what happens to the footer
- what happens to the filter side panel

mbenadda:
## Purpose

On large richie instances, there can be hundreds of courses available to display. As the default view of the search page is non-filtered, the page could be slow to load with hundreds of courses to load from the API and display in complex glimpses with their cover images.

## Proposal

We can paginate the view down to a useful level (say, 20 courses by default) which keeps the page reasonably long and deals with performance issues.

- [x] add a pagination to the course search view, linked with the course search params
- [x] scroll to top when the user changes pages

Closes #754

### DIFF
--- a/src/frontend/js/components/PaginateCourseSearch/index.spec.tsx
+++ b/src/frontend/js/components/PaginateCourseSearch/index.spec.tsx
@@ -1,0 +1,200 @@
+import '../../testSetup';
+
+import { cleanup, fireEvent, render } from '@testing-library/react';
+import React from 'react';
+import { IntlProvider } from 'react-intl';
+
+import { PaginateCourseSearch } from '.';
+import { CourseSearchParamsContext } from '../../data/useCourseSearchParams/useCourseSearchParams';
+
+describe('<PaginateCourseSearch />', () => {
+  beforeEach(jest.resetAllMocks);
+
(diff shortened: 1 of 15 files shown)


src/frontend/js/components/PaginateCourseSearch/index.spec.tsx
```diff
@@ -0,0 +1,201 @@
...
+import { cleanup, fireEvent, render } from '@testing-library/react';
+import React from 'react';
+import { IntlProvider } from 'react-intl';
+
+import { PaginateCourseSearch } from '.';
+import { CourseSearchParamsContext } from '../../data/useCourseSearchParams/useCourseSearchParams';
+
+describe('<PaginateCourseSearch />', () => {
+  beforeEach(jest.resetAllMocks);
+  afterEach(cleanup);
```
lunika: Once rebased you will not need it anymore.

PR #786, opened 2019-08-19
💄(glimpse) add template for organization plugin

2018D:
## Purpose

We want to allow showing a difference between founding members and new partners by changing the logo size.

## Proposal

Add a ~template~ variant option to differentiate between various types of organizations.


src/richie/apps/courses/migrations/0006_auto_20190807_2043.py
```diff
@@ -0,0 +1,34 @@
...
+    dependencies = [("courses", "0005_auto_20190717_0827")]
+
+    operations = [
+        migrations.AlterUniqueTogether(
+            name="persontitletranslation", unique_together=set()
+        ),
+        migrations.RemoveField(model_name="persontitletranslation", name="master"),
+        migrations.RemoveField(model_name="person", name="first_name"),
+        migrations.RemoveField(model_name="person", name="last_name"),
+        migrations.RemoveField(model_name="person", name="person_title"),
```
sampaccoud: @2018D these modifications are not related to your PR so don't include them.
We are planning to remove these fields in the next BC release.

src/richie/apps/courses/migrations/0006_auto_20190807_2043.py
```diff
@@ -0,0 +1,34 @@
...
+                    ("courses/plugins/organization_small.html", "Small"),
+                ],
+                default="courses/plugins/organization.html",
+                help_text="Optional template for custom look.",
+                max_length=150,
+                verbose_name="Template",
+            ),
+        ),
+        migrations.DeleteModel(name="PersonTitle"),
+        migrations.DeleteModel(name="PersonTitleTranslation"),
```
sampaccoud: ditto

src/richie/apps/courses/cms_plugins.py
```diff
@@ -27,9 +28,10 @@ class OrganizationPlugin(CMSPluginBase):
     """
 
     cache = True
+    fieldsets = ((None, {"fields": ["page", "template"]}),)
     model = OrganizationPluginModel
     module = PLUGINS_GROUP
-    render_template = "courses/plugins/organization.html"
+    render_template = ORGANIZATION_TEMPLATES[0][0]
```
sampaccoud: maybe this is not necessary since you are overriding it below?

src/richie/apps/courses/templates/courses/cms/fragment_organization_small_glimpse.html
```diff
@@ -0,0 +1,26 @@
+{% load i18n cms_tags extra_tags thumbnail %}
```
sampaccoud: since this is a massive copy/paster, have you tried factorizing it and forcing a context variable like it's done for categories (tag, icon...)?

src/richie/apps/courses/cms_plugins.py
2018D: It's to force a default and make sure that previously added plugins still work.

src/richie/apps/courses/templates/courses/cms/fragment_organization_small_glimpse.html
2018D: i have not. I'll look into it.

src/richie/apps/courses/cms_plugins.py
sampaccoud: But you're adding a field to the model with a default... and anyway you always override it so this should be useless.

src/richie/apps/courses/templates/courses/cms/fragment_organization_small_glimpse.html
2018D: I still need to call a different template because it's optional and not based on context.
I now use the fragment_organization more efficiently.

src/richie/apps/courses/cms_plugins.py
2018D: It's required for the build. Copied a comment from section_plugin indicating it.

2018D:
Added a test to verify if the templates are rendered correctly.

src/richie/apps/courses/cms_plugins.py
```diff
@@ -27,9 +28,12 @@ class OrganizationPlugin(CMSPluginBase):
     """
 
     cache = True
+    fieldsets = ((None, {"fields": ["page", "template"]}),)
     model = OrganizationPluginModel
     module = PLUGINS_GROUP
-    render_template = "courses/plugins/organization.html"
+    # Required from CMSPluginBase signature but not used since we override it
+    # from render()
```
sampaccoud: :+1:

src/richie/apps/courses/templates/courses/plugins/organization_small.html
```diff
@@ -0,0 +1,10 @@
+{% load cms_tags i18n %}
+{% spaceless %}
+{% if current_page.publisher_is_draft or instance.check_publication is True %}
+  {% with organization_page=relevant_page.organization.extended_object %}
+    <a class="organization-glimpse__small organization-glimpse organization-glimpse--link{% if organization_page.publisher_is_draft is True %} organization-glimpse--draft{% endif %}" href="{{ organization_page.get_absolute_url }}">
```
mbenadda: The link is part of the template (it's inside the "component boundary") and should not be taken out. I previously fixed this because we had links sitting outside the component boundary.

If you want to apply a different class on some use sites it should be done through a prop you pass to the glimpse template

sampaccoud: @2018D good catch, and we already do it for the category here: https://github.com/openfun/richie/blob/master/src/richie/apps/courses/templates/courses/plugins/category_plugin.html

You should use the same pattern and pass a "form_factor" prop.

sampaccoud:
@2018D I merged PR https://github.com/openfun/richie/pull/817 which added a [migration 0006](https://github.com/openfun/richie/blob/master/src/richie/apps/courses/migrations/0006_add_program.py) in the `courses` app.

Remember to rename your migration to 0007 when rebasing on master.

sampaccoud:
Following discussions in this PR, I have refactored it to respect the glimpse. Instead of overriding the template, we should allow forcing a form factor via the context.

@mbenadda can you see if it now fits your requirements?

PR #787, opened 2019-08-21
Bump version to 1.7.0

mbenadda:
### Added

- Paginate the course search results view.
- Add a CSS class to move content offscreen so it's only available for users of assistive technologies.

### Changed

- Allow category icons to be displayed on course detail and not on glimpse.
- Autosuggestion endpoints are diacritics insensitive.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,15 +8,17 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+## [1.7.0] - 2019-08-21
+
 ### Added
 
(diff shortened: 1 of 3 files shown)


sampaccoud:
Maybe we should precise to what this applies:
> Add a CSS class to move content offscreen so it's only available for users of assistive technologies.

This new feature is missing?

> allow icons to display on course detail and not on glimpse

mbenadda:
> Maybe we should precise to what this applies

I added this more as a tool for developers using Richie than anything else. Maybe I can remove it then

> This new feature is missing?

It was not in the changelog, I can add it

PR #790, opened 2019-08-23
Hide "More options" in search pane when it is irrelevant

mbenadda:
## Purpose

There is no point in showing the "More options" button or opening the related modal for a filter group when the related filter is not searchable or all possible values are already displayed.

## Proposal

- [x] add a `has_more_values` field on filter definitions that lets consumers of the API know if there are any more values to be found than those included in the current response;
- [x] add a `is_searchable` field on filter definitions;
- [x] update the frontend to only show the `More options` button on a search filter group when it is actually possible to search for more options


mbenadda:
Note: finishes items 3 & 4 from #376.

src/richie/apps/search/filter_definitions/base.py
```diff
@@ -207,12 +210,30 @@ def get_faceted_definitions(self, facets, *args, **kwargs):
...
         # for each facet, we derive the value and the count:
         # eg. for facet key `availability@coming_soon`, the value is `coming_soon`
-        facet_counts_dict = {
-            key.split("@")[1]: facet["doc_count"]
+        facet_counts = [
+            (key.split("@")[1], facet["doc_count"])
             for key, facet in facets.items()
             if "{:s}@".format(self.name) in key  # 6 times faster than startswith
             and facet["doc_count"] >= self.min_doc_count
-        }
```
sampaccoud: In fact, we have a bug here (already existing) on a edge case:

> if a selected facet has a count of less than `min_doc_count` it will be left out and shouldn't

I think we can solve the bug and simplify the code below by working on the condition in the dict comprehension. I didn't test it but maybe something like:
```python
facet_name = "{:s}@".format(self.name)
sorted_facets = sorted(facets.items(), key=lambda f: f[1]["doc_count'])
facet_counts_dict = {
    key.split("@")[1]: facet["doc_count"]
    for i, (key, facet) in enumerate(sorted_facets)
    if facet_name in key  # 6 times faster than startswith 
    and key in data[self.name]
    or facet["doc_count"] >= self.min_doc_count and  i < facet_counts_limit  # or i <= maybe?
}
has_more_values = len(facet_counts_dict) > facet_counts_limit

File: src/richie/apps/search/filter_definitions/courses.py [hunk 2]
```diff
@@ -143,16 +225,55 @@ def get_faceted_definitions(self, facets, *args, **kwargs):
...
+        elif len(key_count_pairs) <= base_facet_limit:
+            has_more_values = False
+        # We are somewhere between one more than the base limit and the requested number: this is
+        # ambiguous because it depends on the placement of our active values (which are always
+        # included no matter the circumstances).
+        # If at least one of the values beyond base limit is not an active value, we know we're not
+        # going to show it. Therefore there are more values.
+        elif [
+            value
+            for value in [key_count_pair[0] for key_count_pair in key_count_pairs][
```
sampaccoud: I think one check can do it all:
```python
has_more_values = any(filter(lambda c: c[0] not in data[self.name], key_counts_pairs[base_facet_limit:]))
```
Because:
```
if len(key_count_pairs) == base_facet_limit + control_length:
    then there is necessarily 1 of the extra values that is not a selected facet (the +1) so it is True

if len(key_count_pairs) <= base_facet_limit:
    then key_counts_pairs[base_facet_limit:] is equal to [] and it is False
```

src/richie/apps/search/filter_definitions/courses.py
```diff
@@ -143,16 +228,48 @@ def get_faceted_definitions(self, facets, *args, **kwargs):
...
+        #          active values, we are going to show all of them and therefore there is nothing
+        #          more to request.
+        has_more_values = any(
+            filter(
+                lambda c: c[0] not in data[self.name],
+                key_count_pairs[base_facet_limit:],
+            )
         )
 
+        # Now that we're used our extra values to determine if there are more values to be fetched,
```
sampaccoud:
```suggestion
        # Now that we've used our extra values to determine if there are more values to be fetched,
```

src/richie/apps/search/filter_definitions/courses.py
```diff
@@ -170,15 +287,26 @@ def get_faceted_definitions(self, facets, *args, **kwargs):
             self.name: {
                 # We always need to pass the base definition to the frontend
                 "base_path": self.base_page.node.path if self.base_page else None,
+                # If values are removed due to `min_doc_count`, we are not returning them, and
+                # therefore by definition our filter `has_more_values`.
+                "has_more_values": has_more_values
+                or any(
+                    count < self.min_doc_count for key, count in key_count_map.items()
```
sampaccoud:
```suggestion
                    count < self.min_doc_count for count in key_count_map.values()
```

PR #792, opened 2019-08-26
Relocate core app static files to make fonts work and release as a fix 1.7.2

sampaccoud:
## Purpose

In Release 1.7.1, font files are not references correctly in the css files.

## Proposal

Fix path to fonts in css so new fonts are found in static files.


src/frontend/scss/_fonts.scss
```diff
@@ -1,32 +1,31 @@
 @font-face {
   font-family: 'Exo 2';
-  src: url('/static/richie/fonts/exo2-medium-webfont.woff2') format('woff2'),
-    url('/static/richie/fonts/exo2-medium-webfont.woff') format('woff');
+  src: url('/static/fonts/exo2-medium-webfont.woff2') format('woff2'),
```
jmaupetit: My 2 cents:

1. Why not moving Django statics to `src/richie/apps/core/static/richie/{fonts,images}` to be more consistent with front-end build targets (css+js)?
2. What about using a relative path as it's done for the icomoon font?
```suggestion
$richie-fonts-path: '../fonts' !default;
// [...]
  src: url('#{$richie-fonts-path}/exo2-medium-webfont.woff2') format('woff2'),
```

sampaccoud: @jmaupetit Good idea. I did it. Check it out!

PR #802, opened 2019-08-28
Improve error messages in the "more options" modal in filters

mbenadda:
## Purpose

The "More options" modal for filter values was making request to the server as soon as the user started to type, and immediately showed a cryptic error message as it received an error for the server, because textual search only works in our API endpoints with 3 characters or more.

## Proposal

Stop sending useless API requests when users input 1 or 2 characters and show a helpful message in the modal instead.

### DIFF
--- a/src/frontend/js/components/SearchFilterGroupModal/index.spec.tsx
+++ b/src/frontend/js/components/SearchFilterGroupModal/index.spec.tsx
@@ -146,6 +146,12 @@ describe('<SearchFilterGroupModal />', () => {
       content => content.startsWith('Value #84') && content.includes('(12)'),
     );
 
+    // User starts typing, less than 3 characters
+    fetchMock.resetHistory();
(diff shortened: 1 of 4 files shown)


src/frontend/js/components/SearchFilterGroupModal/index.spec.tsx
```diff
@@ -146,6 +146,12 @@ describe('<SearchFilterGroupModal />', () => {
       content => content.startsWith('Value #84') && content.includes('(12)'),
     );
 
+    // User starts typing, less than 3 characters
+    fetchMock.resetHistory();
+    fireEvent.change(field, { target: { value: 'us' } });
+    expect(fetchMock.called()).toEqual(false);
+    getByText('Type 3 characters or more to start searching.');
```
sampaccoud: or `at least`?

PR #808, opened 2019-08-29
Improve and complete strings in pagination

mbenadda:
## Purpose

Visible pagination does not contain a lot of strings, but we have a few more strings for accessibility helpers that provide context along with the page numbers.

We previously attempted to do this "smartly" through the HTML structure of our interpolated strings. This is not ideal as it works for english and some other languages, but fails for other languages with a different grammar.

Also, we forgot a bunch of strings, which resulted in hardcoded english in our template.

## Proposal

Use a more straightforward structure for our JSX and put all content in strings, repeating it instead of concatenating them.

### DIFF
--- a/src/frontend/js/components/PaginateCourseSearch/index.tsx
+++ b/src/frontend/js/components/PaginateCourseSearch/index.tsx
@@ -4,16 +4,47 @@ import { defineMessages, FormattedMessage } from 'react-intl';
 import { CourseSearchParamsContext } from '../../data/useCourseSearchParams/useCourseSearchParams';
 
 const messages = defineMessages({
-  currentlyReading: {
-    defaultMessage: 'Currently reading',
+  currentlyReadingLastPageN: {
+    defaultMessage: 'Currently reading last page {page}',
     description:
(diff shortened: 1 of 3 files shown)


src/frontend/js/components/PaginateCourseSearch/index.tsx
```diff
@@ -4,17 +4,47 @@ import { defineMessages, FormattedMessage } from 'react-intl';
...
     description:
-      'Accessibility helper in pagination, shown next to the current page number (Currently reading Page N)',
-    id: 'components.PaginateCourseSearch.currentlyReading',
+      'Accessibility helper in pagination, shown next to the current page number when it is the last page',
+    id: 'components.PaginateCourseSearch.currentlyReadingLastPageN',
+  },
+  currentlyReadingPageN: {
+    defaultMessage: 'Currently reading page {page}',
+    description:
+      'Accessibility helper in pagination, shown next to the current page number when it is not the last page',
```
sampaccoud:
```suggestion
      'Accessibility helper in pagination, shown next to the current page number when it is not the last page.',
```

PR #809, opened 2019-08-29
Release 1.8.1

mbenadda:
### Fixed

- Fix "More options" modal for searchable filter values not based on MPTT paths (eg. "Person" in the sandbox environment).
- Improve pagination UX by increasing contrast between the active page number and others, and making button clickable areas larger.
- Stop making failed API requests and instead show a helpful error message in the "More options" modal for filter values when the user inputs their first 1 or 2 characters.
- Improve pagination strings for localization.

### Security

- Update the mixin-deep and set-value packages to safe versions.

- [x] update the repo copy of translations

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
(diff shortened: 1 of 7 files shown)


sampaccoud:
You also need to update the. MO files since they are committed.

PR #812, opened 2019-08-30
Add a standalone search field component

mbenadda:
## Purpose

Reuse parts of what we built for <SearchSuggestField /> to make <RootSearchSuggestField />, a React-powered course search field that can live anywhere in Richie, independently from <Search />.

## Proposal

Since most of its features are pretty different from those in the original <SearchSuggestField />, we found it preferable to create a new component, and try to reuse as much of the actual logic as possible.

- [x] refactor `<SearchSuggestField />` to make parts reusable
- [x] add `absolute_url` to course suggestions
- [x] create a new standalone course search field

### DIFF
--- a/src/frontend/js/common/searchFields/getSuggestionsSection.spec.ts
+++ b/src/frontend/js/common/searchFields/getSuggestionsSection.spec.ts
@@ -6,7 +6,7 @@ import { getSuggestionsSection } from './getSuggestionsSection';
 const mockHandle: jest.Mock<typeof handle> = handle as any;
 jest.mock('../../utils/errors/handle');
 
-describe('utils/searchSuggest/getSuggestionsSection', () => {
+describe('common/searchFields/getSuggestionsSection', () => {
   afterEach(() => {
     fetchMock.restore();
   });
--- a/src/frontend/js/common/searchFields/index.tsx
+++ b/src/frontend/js/common/searchFields/index.tsx
@@ -1,0 +1,98 @@
+import React from 'react';
+
+import { APICourseSearchResponse } from '../../types/api';
+import {
+  SearchAutosuggestProps,
+  SearchSuggestion,
+  SearchSuggestionSection,
+} from '../../types/Suggestion';
+import { handle } from '../../utils/errors/handle';
+import { getSuggestionsSection } from './getSuggestionsSection';
+
+/**
+ * `react-autosuggest` callback to get a human string value from a Suggestion object.
+ * @param suggestion The relevant suggestion object.
+ */
+export const getSuggestionValue: SearchAutosuggestProps['getSuggestionValue'] = suggestion =>
+  suggestion.title;
+
+/**
+ * `react-autosuggest` callback to render one suggestion.
+ * @param suggestion Either a resource suggestion with a model name & a machine name, or the default
+ * suggestion with some text to render.
+ */
+export const renderSuggestion: SearchAutosuggestProps['renderSuggestion'] = suggestion => (
+  <span>{suggestion.title}</span>
+);
+
+/**
+ * `react-autosuggest` callback to build up the list of suggestions and sections whenever user
+ * interaction requires us to create or update that list.
+ * @param filters The general filters object as returned by the API on a course search.
+ * @param setSuggestions The suggestion setter method for the component using our helper.
+ * @param incomingValue The current value of the search suggest form field.
+ */
+export const onSuggestionsFetchRequested = async (
+  filters: APICourseSearchResponse['filters'],
+  setSuggestions: (suggestions: SearchSuggestionSection[]) => void,
+  incomingValue: string,
+) => {
+  if (incomingValue.length < 3) {
+    return setSuggestions([]);
+  }
+
+  // Fetch the suggestions for each resource-based section to build out the sections
+  let sections: SearchSuggestionSection[];
+  try {
+    sections = (await Promise.all(
+      Object.values(await filters)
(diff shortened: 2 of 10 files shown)


src/frontend/js/common/searchFields/getSuggestionsSection.spec.ts
```diff
@@ -6,7 +6,7 @@ import { getSuggestionsSection } from './getSuggestionsSection';
 const mockHandle: jest.Mock<typeof handle> = handle as any;
 jest.mock('../../utils/errors/handle');
 
-describe('utils/searchSuggest/getSuggestionsSection', () => {
+describe('common/searchFields/getSuggestionsSection', () => {
```
lunika: in the `getSuggestionSection` module there is one unused import.

src/frontend/js/common/searchFields/index.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+import { APICourseSearchResponse } from '../../types/api';
+import { Suggestion, SuggestionSection } from '../../types/Suggestion';
+import { handle } from '../../utils/errors/handle';
+import { getSuggestionsSection } from './getSuggestionsSection';
+
+/**
+ * Define the kind of suggestions our `<SearchSuggestField />` component supports.
+ */
+type SearchSuggestion = Suggestion<'categories' | 'organizations' | 'persons'>;
```
lunika: Why not adding all types declaration in the `types` directory ?

src/frontend/js/common/searchFields/index.tsx
```diff
@@ -0,0 +1,111 @@
...
+ */
+export const getSuggestionValue: SearchAutosuggestProps['getSuggestionValue'] = suggestion =>
+  suggestion.title;
+
+/**
+ * `react-autosuggest` callback to render one suggestion.
+ * @param suggestion Either a resource suggestion with a model name & a machine name, or the default
+ * suggestion with some text to render.
+ */
+export const renderSuggestion: SearchAutosuggestProps['renderSuggestion'] = suggestion => (
```
lunika: Do you consider this function as a react component ? If not, I think you don't need to import React in this module and as ther is no jsx you can rename the file in `.ts`

src/frontend/js/common/searchFields/index.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+/**
+ * Helper to pick out the relevant filter for a suggestion from the general filters object.
+ * @param filters The general filters object as returned by the API on a course search.
+ * @param suggestion The suggestion with which we're doing some work (eg. the suggestion the user selected).
+ */
+export const getRelevantFilter = (
+  filters: APICourseSearchResponse['filters'],
+  suggestion: SearchSuggestion,
+) => {
```
lunika: Some tests are missing for this module, no ?

src/frontend/js/components/RootSearchSuggestField/index.tsx
```diff
@@ -0,0 +1,174 @@
...
+   * @param suggestion `suggestion` as key to an anonymous object: the suggestion the user picked.
+   */
+  const onSuggestionSelected: SearchAutosuggestProps['onSuggestionSelected'] = async (
+    _,
+    { suggestion },
+  ) => {
+    // Course suggestions are treated differently: when the user finds a course they're interested in,
+    // we're just sending them to that's course page instead of the course search page.
+    if (isCourseSuggestion(suggestion)) {
+      return location.replace(suggestion.absolute_url);
```
lunika: Do you think it's important to keep the session history ? If yes you should use `location.assign`.

src/frontend/js/components/RootSearchSuggestField/index.tsx
```diff
@@ -0,0 +1,174 @@
...
+    }
+
+    // Get the relevant filter and move to the course search view with the selected suggestion as an
+    // active filter value.
+    const filter = getRelevantFilter(await filters(), suggestion);
+    location.replace(
```
lunika: Same here about session history.

src/frontend/js/common/searchFields/index.tsx
mbenadda: This function uses JSX and therefore needs `.tsx` and `import React`. Are you suggesting changes?

src/frontend/js/components/RootSearchSuggestField/index.tsx
mbenadda: good catch!

src/frontend/js/common/searchFields/index.tsx
mbenadda: Not sure they're very relevant since everything is tested through the two search field components. There's no new code here, it was just moved around.

src/frontend/js/common/searchFields/index.tsx
lunika: No forget my comment 🙂

src/frontend/js/common/searchFields/index.tsx
lunika: testing it direcly in a dedicated test is better IMO but if you think it's enough like this we can continue this way.

mbenadda: This is just glue code that is very specific to our autocomplete components and IMO it's barely worth it to even split it out to reuse like I did. I think it's better to not test this and test the actual components users will interact with.

mbenadda:
Can I get an approval @sampaccoud @lunika ? @sveetch will integrate this search component with the nabber in another PR

PR #822, opened 2019-09-13
➕(layout) add 'dj-pagination' dependency and use it on blogposts and persons

sveetch:
## Purpose

Previously page listings objects lacked of any pagination. This prooved to be a flaw with some lists which have largely too many objects. However since these pages are CMS pages, we don't have possibility to implement pagination in their views.

## Proposal

This commit uses [dj-pagination](https://github.com/pydanny/dj-pagination) which provides template tags to make pagination in place directly in templates without manipulating anything from any view.

This adds a new dependency, a new template, a new setting, a new sass object source and involves blogpost and person list templates changes.


sveetch:
- [ ] Reproduce search pagination CSS from ".paginate-course-search" into the new ".pagination" object which will be shared through every lists;
- [ ] Update dj-pagination templates to fit to this CSS;
- [ ] Update search components to use the ".pagination" object;

sveetch:
@sampaccoud ok so now, there is only a single '.pagination' CSS object shared through every pages and components with the same behavior and display.

sveetch:
@sampaccoud also note this add new translatable string for pagination (last page, currently reading page, etc..) to collect in PO files.

src/richie/apps/core/templates/richie/pagination.html
```diff
@@ -0,0 +1,57 @@
...
+            {% endif %}
+            </span>
+            <span aria-hidden="true">{% if page == 1 %}Page {{ page }}{% else %}{{ page }}{% endif %}</span>
+          </span>
+        </li>
+      {% else %}
+        {% if page == 1 %}
+          <li class="pagination__list__item">
+            <a href="{{ request.path }}{% if getvars %}?{{ getvars|slice:"1:" }}{% endif %}" class="pagination__list__item__page-number">
+              <span class="offscreen">{% blocktrans %}Page {{ page }}{% endblocktrans %}</span>
```
mbenadda: This is redundant: we can just keep one "Page 1" and remove the `offscreen` class and `aria-hidden` attribute

mbenadda:
Also this might just be my opinion but I think this version of the pagination shows way too many pages by default

Capture d’écran 2019-09-17 à 15 13 03

The one I did in react uses `1, n-2, n-1, n, n+1, n+2, last` which seems more reasonable to me

src/richie/apps/core/templates/richie/pagination.html
sveetch: This is not really redundant as first element has a special trick to ignore the "page=1" argument when building url such as it create something like `/mylist/` instead of `/mylist/?page=1` or  `/mylist/?sessionid=foo` instead of `/mylist/?page=1&sessionid=foo` if there was other argument. 

Also i don't understand the point about to remove offscreen class and the span since i reproduced the exact behavior from search pagination.

sveetch:
@mbenadda i have added some settings to control the window and margin between showed pages and here is what i've ended to:

pages

src/richie/apps/core/templates/richie/pagination.html
mbenadda: > Also i don't understand the point about to remove offscreen class and the span since i reproduced the exact behavior from search pagination.

This is the entirety of what I was talking about.

The search pagination has a mistake here: if the content is exactly the same, it should exist only once, and not once with `aria-hidden` and once `offscreen`

mbenadda:
Nice job on the pagination size 👍

src/richie/apps/core/templates/richie/pagination.html
sveetch: Ok, removed those two spans around text from first page item.

PR #831, opened 2019-09-19
📝(project) add a file at the root of the project to document upgrades

sampaccoud:
## Purpose

We need a place where all important instructions to upgrade the project are listed. 

## Proposal

We think an UPGRADE.md file at the root of the project, and following the same format as the CHANGELOG.md file, is a good way to make this information available to all.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,0 +1,40 @@
+# Upgrade
+
+All instructions to upgrade this project from one release to the next will be
+documented in this file. Upgrades must be run sequentially, meaning you should
+not skip minor/major releases while upgrading (fix releases can be skipped).
+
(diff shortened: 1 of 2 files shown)


UPGRADE.md
```diff
@@ -0,0 +1,48 @@
...
+
+The format is inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+Before any upgrade, collect static files and run database migrations:
+
+    ```bash
+    $ make collectstatic
+    $ make migrate
+    ```
```
jmaupetit: You don't need to indent this code block.

PR #833, opened 2019-09-20
Use SVG icons instead of an icon font

mbenadda:
## Purpose

Using an icon font to include icons from the Icomoon free icon set had some drawbacks in terms of maintainability and accessibility.
We decided we'd rather user SVG icons instead.

## Proposal

We just replaced the existing icons as used through the icon font with the same icons as embedded SVGs through <use>. We also removed some icons that were loaded even though they are not used anywhere.


sampaccoud:
Don't forget the changelog as it is impacting users who overrode the templates.

PR #836, opened 2019-09-21
💄(layout) enable submenu in footer main menu

sveetch:
## Purpose

It has been required that footer main menu should be able to display a second
level menu if there is one.

## Proposal

This commit changed footer_menu.html template to enable to list direct
children of first level items. Note that this only display well if all first
level items have children, else it could seem a little bit empty.

### DIFF
--- a/src/richie/apps/core/templates/menu/footer_menu.html
+++ b/src/richie/apps/core/templates/menu/footer_menu.html
@@ -1,4 +1,4 @@
-{% load cms_tags %}{% spaceless %}
+{% load menu_tags %}{% spaceless %}
 
 {% for child in children %}
     {% with children_slug=child.get_menu_title|slugify %}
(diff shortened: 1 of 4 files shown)


sveetch:
### Screenshots for menu with second level children

footer_with_submenu large

footer_with_submenu small

### Screenshots for menu without second level children

(the previous behavior for basic horizontal first level is still the same)

footer_without_submenu large

footer_without_submenu small

src/richie/apps/core/templates/menu/footer_menu.html
```diff
@@ -8,12 +8,26 @@
...
             <a class="body-footer__menu__list__item__link" href="{{ child.attr.redirect_url|default:child.get_absolute_url }}">{{ child.get_menu_title }}</a>
-            {% comment %}Dropdown menu are not yet implemented since there is no need about it for now{% endcomment %}
-            {% comment %}{% if child.children %}
-            <ul>
-                {% show_menu from_level to_level extra_inactive extra_active template "" "" child %}
-            </ul>
-            {% endif %}{% endcomment %}
+
+            {% with subchildren=child.children %}
+                {% if subchildren %}
```
sampaccoud: They're actually called `grand_children` :laughing:

src/richie/apps/core/templates/menu/footer_menu.html
```diff
@@ -8,12 +8,26 @@
             {% if child.sibling %} body-footer__menu__list__item--sibling{% endif %}
             {% if child.descendant %} body-footer__menu__list__item--descendant{% endif %}">
             <a class="body-footer__menu__list__item__link" href="{{ child.attr.redirect_url|default:child.get_absolute_url }}">{{ child.get_menu_title }}</a>
```
sampaccoud: can you wrap this in a block so it can be removed by overriding the template and not rewriting the whole thing?

sveetch: What line(s) are we talking about to move to a block ?

Also this for a usecase where someone would create a "myproject/menu/footer_menu.html" which inherits from richie footer_menu and then change base.html to use this path on templatetag "show_menu_belowid" in footer ?

sveetch: Finally we let down this since it is simpler to override the whole template since it can be overriden directly from "base.html".

Issue #824, opened 2019-09-14
Course runs are referenced in sitemap.xml


sveetch:
## Bug Report

**Expected behavior/code**
Since course run pages are not desired to be scanned from search engines, their urls should not be referenced in "sitemap.xml".

**Actual Behavior**
"sitemap.xml" does reference all course run pages.

**Steps to Reproduce**
1. Open https://richie.education/sitemap.xml
2. Search for a course run url (using a pattern like `/session-` should work)
3. And you can see they are all present

Warning, the sitemap.xml from the demo is huge and not human readable (everything on a same line). If you want to read it completely, save it as a file and put its source to https://htmlformatter.com/, then update previously saved file.

**Environment**
- Richie version: 1.8.3
- Platform: Any

sveetch:
It will need to subclass `cms.sitemaps.CMSSitemap` in a new custom class which will override `CMSSitemap.items()` to filter out course run pages.

sampaccoud:
This issue was solved by PR https://github.com/openfun/richie/pull/1202 as course runs are not CMS pages anymore.

PR #837, opened 2019-09-21
🐛(sitemap) filter out Course run items from "sitemap.xml"

sveetch:
## Purpose

Currently, the generated "sitemap.xml" contains all course run pages but they are not desired for SEO so they should not be included in Sitemap file.

This is related to opened issue #824 

## Proposal

This commit add a new custom Sitemap generator for cms pages which inherit from the one from DjangoCMS just to filter out Course run pages.

### DIFF
--- a/src/richie/apps/core/sitemaps.py
+++ b/src/richie/apps/core/sitemaps.py
@@ -1,0 +1,16 @@
+"""
+Sitemap generators for contents
+"""
+from cms.sitemaps import CMSSitemap
+
+
+class CourseRunFreeSitemap(CMSSitemap):
+    """
+    A CMS pages sitemap which is free from Course run items.
+    """
+
+    def items(self):
+        titles = super().items()
+
+        # Exclude all page which have a non null course run attribute
+        return titles.filter(page__courserun__isnull=True)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Fixed
+
+- Filter out Course runs from "sitemap.xml".
+
 ## [1.11.0] - 2019-10-11
 
 ### Added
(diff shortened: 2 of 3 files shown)


sveetch:
Note than this is currently a proof of concept since filtering from new custom cms page sitemap generator stand on demo behavior which starts every course run slug with `session` which is probably not solid behavior in real life.

However, it works to reduce sitemap size, on my local demo i'm going from 946 items to 490 items.

src/richie/apps/core/sitemaps.py
```diff
@@ -0,0 +1,18 @@
+"""
+Sitemap generators for contents
+"""
+from cms.sitemaps import CMSSitemap
+
+
+class CourseRunFreeSitemap(CMSSitemap):
+    """
+    A CMS pages sitemap which are free of Course run items.
```
sampaccoud:
```suggestion
    A CMS pages sitemap which is free from Course run items.
```

src/richie/apps/core/sitemaps.py
```diff
@@ -0,0 +1,18 @@
...
+    """
+
+    def items(self):
+        titles = super().items()
+
+        # Currently a very dummy filter standing on demo behavior which create
+        # every course run with a slug starting with "sesssion". This is until
+        # how to correctly filter out course run from a cms page Title queryset.
+        return titles.exclude(slug__startswith="session")
```
sampaccoud: The query you're looking for should be something like:
```suggestion
        return titles.filter(page__course_run__isnull=True)
```

sampaccoud:
Add a small test to check that run is not included in the rendered sitemap?

src/richie/apps/core/sitemaps.py
sveetch: Nice, i need to test this, thanks.

sveetch: Ok now we have a clean way to exclude course run and it works nice. 👌

sveetch:
@sampaccoud yeah, now i have to find time to create a test to cover this.

sampaccoud:
This problem was solved by PR https://github.com/openfun/richie/pull/1202 as course runs are not CMS pages anymore.

Issue #825, opened 2019-09-15
All pages have HTML validation errors and warnings


sveetch:
## Bug Report

**Expected behavior/code**
All pages should pass the HTML validation without any error and possibly no warnings.

**Actual Behavior**
Take almost any page from a Richie website and give to the validator tool. An example:

https://validator.w3.org/nu/?showsource=yes&showoutline=yes&doc=https%3A%2F%2Frichie.education%2Ffr%2Fcategories%2Fniveau%2F

**Steps to Reproduce**
1. Copy a reachable (no localhost only) Richie url;
2. Paste it to the validator tool https://validator.w3.org/nu/
3. And scroll down to the report

**Environment**
- Richie version: 1.8.3
- Platform: Any

**Possible Solution**
Here are distinct messages which was found on pages.

- Error on `srcset` on `<img>` tags, caused by last ending comma in attribute value;
- Error on `sizes` on `<img>` tags, it seems every size item expect a leading media condition;
- Error on missing `alt` attribute on some `<img>` tags, we need to find template which build these tags to add the missing attribute, even if it's empty it is required;
- Error on `href` value on links with a `mailto` content which is not escaped and leading to include quote characters in attribute quotes;
- Error on obsolete `frameborder` attribute on the “iframe” element for teaser video course content, this is an old attribute commonly keeped for Internet Explorer <= 9, but since a long time we just can adjust iframe border from CSS (which is very often resolved with a global `border:0;` on iframe element);
- Warning on `type` attribute on `<script>` tags. Just need to remove them it's a deprecated attribute since HTML5, it's probably only in the base.html template;
- Warning on missing `lang` attribute on `<html>` tag. This is a useful thing for SEO and it's easy to implement, we already define this value from OG tags and it have to be done once in `base.html` template.

sveetch:
Missing `lang` attribute on `<html>` should have been resolved with 1.9.0.

sveetch:
I get it for the wrong value in `srcset` wichh end on a comma. It is because in thumbnail parts in templates we do something like this:

```
{% thumbnail picture 200x200 crop upscale subject_location=subject_location %} 200w,
{% if width >= 400 %}{% thumbnail picture 400x400 crop upscale subject_location=subject_location %} 400w,{% endif %}
{% if width >= 600 %}{% thumbnail picture 600x600 crop upscale subject_location=subject_location %} 600w{% endif %}
```

[...]

PR #838, opened 2019-09-21
👌(layout) fix HTML validation warnings and errors

sveetch:
## Purpose

Issue #825 has reported some warnings and errors from HTML validator in almost all pages which should be resolved for sanity.

## Proposal

This pull request will fix every reported warnings and errors from issue except the one about missing `lang` attribute which has already been fixed in a previous commit about SEO.

- [x] Warning about deprecated `type` attribute on `<script>` tags;
- [x] Error on `srcset` attribute value on `` tags;
- [x] Error on missing `alt` attribute on some `` tags;
- [x] Error on `href` value on links with `mailto` content;
- [x] Error on `sizes` attribute value on `` tags;
- [x] Error on obsolete `frameborder` attribute on `iframe`;
- [x] Warning about multiple `<h1>` on the same page (homepage);
- [x] Error on `<button>` inside `<a>` on course glimpses.
- [x] Error on invalid `role` value on pagination bullet list;

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Fixed
 
+- Add missing attribute "alt" on some "<img>" tags in templates,
 - Remove deprecated attribute "type" from some "<script>" tags in templates,
 - In the sandbox, make API calls work behind an htaccess by removing Basic
   Auth fallback,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Fixed
 
+- Fix invalid "button" tag inside a "a" tag inside course glimpse,
 - Fix invalid attributes on "iframe" tag from djangocms_video plugin,
 - Fix invalid syntax in "sizes" attribute value on "<img>" tags,
 - Add url escaping on variables inside links from on social network badges,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Fixed
 
+- Fix invalid syntax in "sizes" attribute value on "<img>" tags,
 - Add url escaping on variables inside links from on social network badges,
 - Fix invalid syntax in "srcset" attribute value on "<img>" tags,
 - Add missing attribute "alt" on some "<img>" tags in templates,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Fixed
 
+- Remove deprecated attribute "type" from some "<script>" tags in templates,
 - In the sandbox, make API calls work behind an htaccess by removing Basic
   Auth fallback,
 - Fix React pagination role-element pair. Pagination should now be seen as a
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,7 +37,20 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 - In the sandbox, make API calls work behind an htaccess by removing Basic
   Auth fallback,
 - Fix React pagination role-element pair. Pagination should now be seen as a
-  navigational element by screen readers.
+  navigational element by screen readers,
+- Fix add plugin to team and organization on fragment_course_content template
+  when they are empty,
+- Remove deprecated attribute "type" from some "<script>" tags in templates,
+- Add missing attribute "alt" on some "<img>" tags in templates,
+- Fix invalid syntax in "srcset" attribute value on "<img>" tags,
+- Add url escaping on variables inside links on social network badges,
+- Fix invalid syntax in "sizes" attribute value on "<img>" tags,
+- Fix invalid attributes on "iframe" tag from djangocms_video plugin,
+- Fix invalid "button" tag inside a "a" tag inside course glimpse,
+- Fix invalid "role" value on pagination list with a "nav" tag around bullet
+  list instead,
+- Fix warning about multiple "h1" tags on homepage, section template will
+  always have a default level title to 2.
 
 ## [1.11.0] - 2019-10-11
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,8 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Fixed
 
+- Fix invalid "role" value on pagination list with a "nav" tag around bullet
+  list instead,
 - Fix invalid "button" tag inside a "a" tag inside course glimpse,
 - Fix invalid attributes on "iframe" tag from djangocms_video plugin,
 - Fix invalid syntax in "sizes" attribute value on "<img>" tags,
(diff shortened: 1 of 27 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,15 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Fixed
+
+- Remove deprecated attribute "type" from some "<script>" tags in templates.
+- Add missing attribute "alt" on some "<img>" tags in templates.
+- Fix invalid syntax in "srcset" attribute value on "<img>" tags.
+- Add url escaping on variables inside links from on social network badges.
```
sampaccoud:
```suggestion
- Add url escaping on variables inside links on social network badges.
```

src/richie/apps/core/templates/djangocms_video/default/video_player.html
```diff
@@ -0,0 +1,33 @@
+{% load i18n cms_tags %}
```
mbenadda: > Unfortunaly due to order of "INSTALLED_APPS" setting, this does
not work. "djangocms_video" is loaded before "richie.core" and so the first
win the loading match. However i altered locally my settings to test this
fix correctly fix the validation issue.

Then we should change the order in `INSTALLED_APPS` to have the correct order?

sveetch: This would resolve our issue and all other with plugins and external applications, yes.

But i'm not sure enough about consequences. From what i see at `src/richie/apps/core/templates/`, it should be safe but i request someone else to validate this.

It validated, from what i talked with samuel, the same change should be applied to project which embed richie like your "ongoing" funmooc.fr reboot with richie and so to fun coorporate, so their template directory should be applied on top of richie which should be on top of cms apps and plugins.

sampaccoud: @mbenadda @sveetch after reviewing this, it appears that the proposed change only affects the sandbox... it is thus something that has to be done in each project using Richie and not here. I propose to remove this commit and merge the rest now. WDYT?

sveetch: @sampaccoud There is still two issue to fix which are about homepage and course glimpses

sveetch:
@sampaccoud remaining issues has been solved plus two new ones, everything is valid except the ones from "iframes" tags. I propose to keep my commit about them since it contains the correct fix and should automatically popup once install apps reordering has been done.

src/richie/apps/courses/templates/courses/cms/fragment_course_glimpse.html
```diff
@@ -45,7 +45,7 @@
           </p>
           {% if course_state.call_to_action %}
           <div class="course-glimpse__footer__cta">
-              <button class="button">{{ course_state.call_to_action|capfirst }}</button>
```
mbenadda: I disagree with this change.

Generally, as a rule, if it looks like a button it should be a `<button>` not a `<span class="button">`.

I understand how this is is specific situation where it is inside an `<a>` but I'd need something more convincing than an HTML validator warning to override the rule I state above.

sveetch: This is not a warning, it is an error. And this is not "an HTML validator", this is THE validator, we are talking about Nu Html Checker which is the official W3C validator which strictly follow the HTML reference.
 
The button element in this context has only a layout meaning. Semantically and accessibility talking it was wrong. Alike you don't put a `<a>` in another  `<a>`, you must not put  `<button>` in  `<a>`.

The choice of `<span>` is because it's the less annoying element since `.button` class does everything itself, but we may use a simple `<div>` if it troubles you (but not `<p>` since it possibly involves some default margin/padding/font settings from bootstrap typography.

mbenadda: To be completely frank, I think HTML validation only has value when it allows us to detect issues that actual users (incl. assistive technology users) face. Beyond that, in my opinion HTML being valid or not is not something to be concerned about.

I searched for data about whether it's better for accessibility tools to keep a real `<button>` here or resort to the `<span>` workaround and could not find anything. Do you have anything on this?

sveetch: HTML validation is relevant since it valid browser doest not switch in quirk mode to automatically resolve wrong html.

I don't have any links about `<button>` is better than a `<span>` in our context, since the button was here only for layout interest since it lives inside a link and so the whole glimpse is clickable, it does not change anything to a span or a div or something else, it even not have any accessibility issue, it's just cosmetic.

`button` element has been introduced to use an element which have default browser style and implicit aria role just like a input button that must be inside a form, so the `button` element can be used outside a form.

mbenadda: I'm still not convinced about the relevance of HTML validation, and we don't have any information on what's the better choice here for accessibility. It's not very important since everything is wrapped in an `<a>` anyway.

Let's merge this 🤝

src/richie/apps/core/templates/richie/pagination.html
```diff
@@ -4,53 +4,55 @@
...
-        {% if page == 1 %}
-          <li class="pagination__list__item">
-            <a href="{{ request.path }}{% if getvars %}?{{ getvars|slice:"1:" }}{% endif %}" class="pagination__list__item__page-number">
-              {% blocktrans %}Page {{ page }}{% endblocktrans %}
-            </a>
-          </li>
-        {% else %}
-          <li class="pagination__list__item">
-            <a href="?page{{ page_suffix }}={{ page }}{{ getvars }}" class="pagination__list__item__page-number">
+  <nav aria-labelledby="pagination-label">
```
mbenadda: HTML validation is really showing some value here 👍

I need to do the same change in the react-generated equivalent

src/richie/apps/core/templates/djangocms_video/default/video_player.html
```diff
@@ -0,0 +1,33 @@
...
+{% comment %}
+This is a copy of original template from plugin just to clean <iframe> from
+obsolete attribute "frameborder" and invalid "allowfullscreen" attribute value.
+{% endcomment %}
+
+{% if instance.embed_link %}
+    {# show iframe if embed_link is provided #}
+    <iframe src="{{ instance.embed_link_with_parameters }}" {{ instance.attributes_str }} allowfullscreen></iframe>
```
sampaccoud: @sveetch In your commit message you say that allowfullscreen is invalid but you are keeping it here?

sveetch: @sampaccoud it does not seems so, i thinks it's typo error from me in my commit message. But finally i checked from MDN and it does seems it could be replaced by `allow="fullscreen"`  because `allowfullscreen` is not deprecated but marked as "historical attribute".

sampaccoud: I propose to keep `allowfullscreen`.

sampaccoud:
@mbenadda @sveetch I finalized the PR following our discussions:
- rebase to master
- reorder apps in the INSTALLED_APPS setting of the sandbox, and test the demo site manually to check that everything works fine and all overrides are still effective,
- amend commit about the `djangocms_video` plugin to remove your reserves about the ordering of apps in INSTALLED_APPS.

It should be ready to merge if you agree.
@sveetch maybe you can run your validator one last time on this PR to check that the rebase went well.

sveetch:
@sampaccoud launched my validator on sample pages with update branch, it's ok, 100% valid :+1:

sampaccoud:
@sveetch Nice! Let's merge it.

Issue #840, opened 2019-09-23
Allow nesting organizations (by for example level of partnership)


sampaccoud:
## Feature Request

**Is your feature request related to a problem or unsupported use case? Please describe.**
We want to be able to rank organizations by membership level.

**Describe the solution you'd like**
I thought of allowing to nest organization pages below bucket pages. For example:
```
- Organization
    * Premium
        + Premium university 1
        + Premium university 2
    * Regular
        + Regular university 1
        + Regular university 2
```

The Premium and Regular pages would be assigned to the existing `organization_list.html` template.

We should add a new template `organization_list_nested.html` that would list all universities below section titles "Premium" and "Regular". Note that this logic could work with deeper nesting...

**Describe alternatives you've considered**
I have considered doing it manually with the existing "single_column.html` template but we have too many organization and the placeholder widget would be very hard to manipulate with 150 organization...

sampaccoud:
This issue was solved by 3 Pull Requests: https://github.com/openfun/richie/pull/858, https://github.com/openfun/richie/pull/858 and https://github.com/openfun/richie/pull/858.

What we did is to allow tagging organizations by category. Then allow inserting multiple organizations in a page via a plugin targeting them via their category. And finally allow specifying different variants of organization glimpse when inserting organizations via a plugin so that we can for example display `premium` organizations bigger than `standard` organizations.

PR #844, opened 2019-09-25
Add categories to organizations

sampaccoud:
## Purpose

We want to be able to group organizations by their membership level (https://github.com/openfun/richie/issues/840).

## Proposal

This PR is the first of two steps to deliver this feature by allowing to tag organizations with categories.

We think this is a better solution than what is proposed in issue https://github.com/openfun/richie/issues/840 because:

- all other types of pages are already taggable with categories,
- this enables tagging organizations by any other dimensions,
- we don't alter the path of organization pages with their partnership level.

While working on this feature I had the opportunity to make some refactoring:

- factorize method to get page extensions related to a category as it was repeated numerous times,
- improve tests for the person detail view.

The second step after this PR will be to introduce a plugin allowing to embed in any page, all the organizations linked to a given category.


tests/apps/courses/test_models_category.py
```diff
@@ -254,6 +255,64 @@ def test_models_category_get_blogposts_several_languages(self):
...
+        retrieved_organizations = category.get_organizations()
+
+        with self.assertNumQueries(2):
+            self.assertEqual(set(retrieved_organizations), set(organizations))
+
+        with self.assertNumQueries(0):
+            for organization in retrieved_organizations:
+                self.assertEqual(
+                    organization.extended_object.prefetched_titles[0].title, "my title"
+                )
```
jmaupetit: :1st_place_medal:

PR #847, opened 2019-10-01
Add plugin to embed organization from category

sampaccoud:
## Purpose

We want to be able to add a list of organizations on a page, by just specifying their category.

## Proposal

- [x] add a new plugin which takes a caegory as input field. The plugin displays the list of all organizations that are linked to the targeted category,
- [x] allow specifying the glimpse variant that should be used,

While working on this feature, I had the chance to improve a few things:

- [x] sort objects related to a category by their tree path so that administrators can control the order from the pages tree,
- [x] include descendants in related objects (so that the science category page gets all maths courses as related pages),
- [x] display related organizations on the category detail page,
- [x] improve integration tests on the category detail page (especially checking the presence and publication status of all related objects),
- [x] in factories, share the code that automatically creates related pages,
- [x] in factories, remove anti-pattern that creates related pages from the targeted page.


src/richie/apps/courses/models/organization.py
```diff
@@ -241,4 +241,39 @@ def __str__(self):
...
+        max_length=50,
+        choices=defaults.ORGANIZATION_GLIMPSE_VARIANT_CHOICES,
+        help_text=_("Optional glimpse variant for a custom look."),
+        blank=True,
+        null=True,
+    )
+
+    class Meta:
+        db_table = "richie_organizations_for_category_plugin"
+        verbose_name = _("organizations for category plugin")
```
mbenadda:
```suggestion
        verbose_name = _("organizations by category plugin")
```

IMO it's easier to understand this way, eg. when you first encounter the plugin.

We could keep the name in the code unchanged as that's more of a bother to change

src/richie/apps/courses/models/category.py
```diff
@@ -67,7 +67,7 @@ def get_meta_category(self):
             extended_object__node__parent__cms_pages__publisher_is_draft=self_is_draft,
```
mbenadda: linnked => linked in commit message

sampaccoud:
>  it would seem odd that only organizations can be listed by category

It would be easy to add similar plugins for other objects but I didn't do it because we didn't encounter the use case. Now that we have the pattern, it will be easy to add.

> it would seem odd that we can only list by category, and not by eg. organization

I didn't get your point. List organizations by organizations? You already have an OrganizationPlugin that allows to add organizations one by one.

src/richie/apps/courses/models/organization.py
sampaccoud: Not a problem to change it in the code as it is not yet merged... So ok I will change it everywhere.

sampaccoud:
@mehdi oh yeah. We could add a plugin but it is much less likely that we will need this. Categories are created with the purpose to group objects so when we want to display a list of objects we can just create a category for that... Organizations are a defined thing and already have their page. Displaying related objects... We'll see.

mbenadda:
Ok then, I will come back after the build passes (and we get another reviewer maybe) to validate this ✨

PR #857, opened 2019-10-14
Wrap course detail sections in blocks

sampaccoud:
## Purpose

The projects using richie as a dependency should be able to override a precise portion of the course detail view without having to override the whole template.

This will ease maintenance on such projects.

## Proposal

- [x] wrap each section in the course detail template in a template block,
- [x] rename the `title` block in html head to `head_title` to avoid collision with the new H1 `title` block.

While working on this PR I had the chance to rename `course-detail__aside__run` selector to `course-detail__aside__runs` in `course_detail.html` template to reflect the fact that there are many course runs.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -20,6 +20,19 @@ $ make migrate
   template(s) using icons from the icon font need to be updated to use SVG
   icons too.
 
+## 1.10.x to 1.11.x
+
+## Before switching
+
+- If you override `_course_detail.scss`, the `course-detail__aside__run` selector was renamed
+  to `course-detail__aside__runs` in our `course_detail.html` template.
+
+## 1.9.x to 1.10.x
+
+### Before switching
+
+- For each item in the [changelog](./CHANGELOG.md), check that your css overrides have no impact.
+
 ## 1.8.x to 1.9.x
 
 ### Before switching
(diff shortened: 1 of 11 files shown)


UPGRADE.md
```diff
@@ -20,6 +20,19 @@ $ make migrate
...
+## Before switching
+
+- If you override `_course_detail.scss`, the `course-detail__aside__run` selector was renamed
+  to `course-detail__aside__runs` in our `course_detail.html` template.
+
+## 1.9.x to 1.10.x
+
+### Before switching
+
+- For each item in the [changelog](./CHANGELOG.md), check that your css override have no impact.
```
jmaupetit: `s/override/overrides/`

UPGRADE.md
```diff
@@ -20,6 +20,19 @@ $ make migrate
   template(s) using icons from the icon font need to be updated to use SVG
   icons too.
 
+## 1.10.x to 1.11.x
+
+## Before switching
+
+- If you override `_course_detail.scss`, the `course-detail__aside__run` selector was renamed
+  to `course-detail__aside__runs` in our `course_detail.html` template.
```
jmaupetit: :1st_place_medal:

PR #859, opened 2019-10-15
Override react components

mbenadda:
## Purpose

This PR makes it possible to build Richie's frontend from a dependents project (ie. by installing the `richie-education` `npm` package.

The purpose of this is to enable users to override components in the frontend to more easily customize the UI for their portal built with Richie.

## Proposal

- [x] make it possible to run the Richie build from a dependent (`devDependencies` -> `dependencies`, polyfill through `import`, change entrypoints)
- [x] standardize import paths (with absolute import paths & `index` filenames)
- [x] update our `webpack` config and add CL args to enable dependent builds with overrides

Dependents can now run their own build with overrides by specifying a JSON settings file through `--richie-setings foo.json`  and using the `--richie-dependent-build` flag to enable dependent-specific module resolution paths.

**Documentation will follow in a subsequent PR as we need to kickstart it using proper tooling and this is out of the scope of this PR.**


src/frontend/js/components/PaginateCourseSearch/index.spec.tsx
```diff
@@ -1,11 +1,11 @@
-import '../../testSetup';
+import 'testSetup';
 
-import { cleanup, fireEvent, render } from '@testing-library/react';
+import { fireEvent, render } from '@testing-library/react';
 import React from 'react';
 import { IntlProvider } from 'react-intl';
 
+import { CourseSearchParamsContext } from 'data/useCourseSearchParams/useCourseSearchParams';
 import { PaginateCourseSearch } from '.';
```
lunika: Why not also moving this relative path ?

src/frontend/js/components/SearchSuggestField/SearchSuggestField.spec.tsx
```diff
@@ -1,14 +1,14 @@
...
+import 'testSetup';
 
 import { fireEvent, render, wait } from '@testing-library/react';
 import fetchMock from 'fetch-mock';
 import React from 'react';
 import { IntlProvider } from 'react-intl';
 
-import { CourseSearchParamsContext } from '../../data/useCourseSearchParams/useCourseSearchParams';
+import { CourseSearchParamsContext } from 'data/useCourseSearchParams/useCourseSearchParams';
 import { SearchSuggestField } from './SearchSuggestField';
```
lunika: In this commit some paths are still relative. I know they only concern tests and the tested component for for a seek of consistency we should maybe also use absolute paths ?

src/frontend/webpack.config.js
```diff
@@ -44,4 +67,15 @@ module.exports = {
       { enforce: 'pre', test: /\.js$/, loader: 'source-map-loader' },
     ],
   },
+
+  plugins: [
+    // Use module replacement to override any number of Richie components as defined in the settings
+    ...Object.entries(overrides).map(
+      entry =>
+        new webpack.NormalModuleReplacementPlugin(
+          new RegExp(`components\/${entry[0]}$`),
```
lunika: Did you did that to force only component overriding ?

mbenadda: Yes! This is the result of a discussion with @sampaccoud, and is also why we moved to absolute paths. It makes it way easier for dependents to override components

src/frontend/js/components/SearchSuggestField/SearchSuggestField.spec.tsx
mbenadda: I think generally absolute paths are better but when we're importing stuff from the same directory IMO it's part of the same module and it makes more sense to import it directly. The same goes for tests.

Now of course this is debatable but I think simplicity trumps consistency here.

PR #860, opened 2019-10-16
🐛(sandbox) make API calls work behind an htaccess

sampaccoud:
## Purpose

The DRF API calls were responding with a 403 status code for anonymous users, and were working for authenticated users. This was due to a Basic Auth fallback option in the default configuration of DRF's authentication backends.

For an anonymous user:
- The first authentication backend in the list of DRF's default backends (Django session authentication) does not try to authenticate because it does not see any cookie,
- The second backend (Basic Authentication) tries to authenticate because it sees an Authentication header. But this header is coming from the Nginx htaccess (and is not known by Django), so the authentication fails with a 403.

Removing the Basic Authentication backend allows anonymous users to be really considered as anonymous in Django.

## Proposal

- Remove Basic Auth fallback option from DRF's authentication backends,
- Add a test to validate that the site now works behind an htaccess.

Another alternative would have been that nginx does not pass the Basic Authentication headers. This can be done in [arnold](https://github.com/openfun/arnold) but we also don't want to allow basic authentication on our site so it had to be removed here as well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,8 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Fixed
 
+- In the sandbox, make API calls work behind an htaccess by removing Basic
+  Auth fallback,
 - Fix React pagination role-element pair. Pagination should now be seen as a
   navigational element by screen readers.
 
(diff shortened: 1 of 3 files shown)


jmaupetit:
This is a backport of https://github.com/openfun/fun-mooc/pull/41

PR #862, opened 2019-10-18
Add cookiecutter template

sampaccoud:
## Purpose

We want to make it as straightforward as possible for developpers to bootstrap their own Richie project. 

## Proposal

I propose to integrate a Cookiecutter template inside the Richie repository, together with a command to generate a new project in seconds with guidance.

- [x] Add a full Cookiecutter template (copied from https://github.com/openfun/fun-mooc)
- [x] Extend our Docker Compose and make configurations to generate the Cookiecutter project in a single command,
- [x] Add documentation about bootstrapping a new project with Cookiecutter,
- [x] Run our linters on the Cookiecutter template to our make configuration and to the CircleCI configuration.

## Loopholes

- [x] No linting on the frontend
- [x] pylint can't run on the Cookiecutter template
- [x] duplicate code between the Cookiecutter template and the sandbox. We could imagine using the Cookiecutter template to build the sandbox but it is not straightforward as the way Richie is installed defers.


Makefile
```diff
@@ -219,6 +224,15 @@ test-back: ## run back-end tests
...
 .PHONY: test-back
 
+# -- Cookie cutter
+
+start-project: ## bootstrap a new project based on Richie using Cookiecutter
+	@$(COMPOSE_RUN_COOKIECUTTER) \
+	    "/app/$$(basename $(PWD))/cookiecutter" \
+	    -o /app \
+		--config-file "/app/$$(basename $(PWD))/.cookiecutterrc"
+.PHONY: start-project
```
jmaupetit: Should be tested in the CI. WDYT?

Makefile
```diff
@@ -219,6 +224,15 @@ test-back: ## run back-end tests
 	@DB_PORT=$(DB_PORT) bin/pytest
 .PHONY: test-back
 
+# -- Cookie cutter
+
+start-project: ## bootstrap a new project based on Richie using Cookiecutter
```
jmaupetit: Also: not a big fan of this strategy with the docker-compose volume trick that forces to create a repository a the same level/path as a cloned Richie project.

cookiecutter/{{cookiecutter.module_name}}/.circleci/config.yml
```diff
@@ -0,0 +1,384 @@
...
+          name: Publish images
+          command: |
+            # Display either:
+            # - DOCKER_TAG: master (Git branch)
+            # or
+            # - DOCKER_TAG: 1.1.2 (Git tag v1.1.2)
+            echo "DOCKER_TAG: ${DOCKER_TAG} (Git ${RELEASE_TYPE}${CIRCLE_TAG})"
+            docker push ~/{{cookiecutter.dockerhub_account}}/{{cookiecutter.module_name}}:${DOCKER_TAG}
+            if [[ -n "$CIRCLE_TAG" ]]; then
+              docker push ~/{{cookiecutter.dockerhub_account}}/{{cookiecutter.module_name}}:latest
```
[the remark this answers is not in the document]
sampaccoud: lol I think a find & replace went wrong :face_with_head_bandage:

sampaccoud:
> $ docker run --rm -v /path/to/project:/app fundocker/cookiecutter gh:openfun/richie

But this doesn't work with Cookiecutter... because the template is nested in the repository... we would need to customize the Docker image to know how to go look for the template inside richie's code... 

If we want to make it easy I would prefer to publish a small click command line tool that does this.

jmaupetit:
> But this doesn't work with Cookiecutter... because the template is nested in the repository.

I know, and I'm surprised that it's not possible. Do you think we can propose a PR to change this in CookieCutter's repository (i.e. adding a new option in the CLI to handle this)?

Makefile
sampaccoud: @jmaupetit not sure what I can do better for this command inside richie... generate the project inside richie would be worse ? Asking for the location is not a possbile in a make command... I could make a bin but it would be less straightforward... Do you have something in mind?

Makefile
sampaccoud: Yes I will open an issue for this as this PR is already too big.

Makefile
jmaupetit: Yes. IRL discussion is on-going...

sampaccoud:
Closing this PR because the way we generate richie sites has changed a lot (see https://www.github.com/openfun/richie-site-factory) and the cookie cutter template will be very different and probably not be in this repository.

sampaccoud:
This feature was finally implemented here: https://github.com/openfun/richie/pull/1682

PR #863, opened 2019-10-18
👷(circleci) fix test-back-mysql following mysql image upgrade

sampaccoud:
## Purpose

A new `mysql:5.6` Docker image was published that broke our CI because our fun user was not allowed to create a database anymore. It is necessary because tests run on a specific database called `test_richie` that it creates if it is missing. 

## Proposal

Making sure the database is already created in the container when it starts fixes the bug.

I also changed the name of the database for Posgresql so that it doesn't need to create another one either...

I did not include a changelog because there is no change on Richie, only this repository's CI.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -194,7 +194,7 @@ jobs:
         environment:
           MYSQL_ROOT_PASSWORD:
           MYSQL_ALLOW_EMPTY_PASSWORD: yes
-          MYSQL_DATABASE: richie
+          MYSQL_DATABASE: test_richie
           MYSQL_USER: fun
           MYSQL_PASSWORD: pass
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -194,7 +194,7 @@ jobs:
         environment:
           MYSQL_ROOT_PASSWORD:
           MYSQL_ALLOW_EMPTY_PASSWORD: yes
```
jmaupetit: You can either choose to also change this line as follow:
```suggestion
          MYSQL_ALLOW_EMPTY_PASSWORD: true
```

... or postpone it for a future contribution.

sampaccoud: Actually I removed it because it is already the default value.

PR #872, opened 2019-10-25
🐛(website) fix documentation build [REVERT]

mbenadda:
This reverts commit 225ea8822e6c3b6b5387bc7759e1b1cf46da7f8f.

The change to the deployment command was unnecessary as the issues we were encountering were caused by an invalid token (which was not understood as such by us).

We can just revert them and everything works as intended with a valid token.


jmaupetit:
There is a gitmoji for that :rewind: !

PR #879, opened 2019-10-31
Implement login buttons in React

mbenadda:
## Purpose

Caching pages and content will be essential in Richie. We don't want to make caching pages difficult or impossible because we just need a small part of the menu with information about the user and their login status.

Still, we need to show Login/Signup buttons to anonymous users and a small account menu to logged in users.

## Proposal

We decided to handle this part of the page with a React component. This way, we can safely cache the page and rely on our React component to find out the login status of the current user and show them the appropriate UI.

- [x] create a `whoami` API endpoint (using a standard DRF `ViewSet`)
- [x] convert our `<SearchLoader />` to a more generic `<Spinner />` and make it accessible
- [x] enable login by using Django admin & the contrib auth logout view
- [x] Add a `<UserLogin />` component to log the user in or show their login status


sandbox/urls.py
```diff
@@ -11,6 +11,7 @@
 
 from cms.sitemaps import CMSSitemap
 
+from richie.apps.core.urls import urlpatterns as core_urlpatterns
```
lunika: s/retriete/retrieve/ In your commit message

src/richie/apps/core/api.py
```diff
@@ -0,0 +1,45 @@
...
+    API endpoints to access users.
+    """
+
+    # pylint: disable=no-self-use,unused-argument
+    def list(self, request, version):
+        """
+        List requests for users. We do not provide an API to get a list of registered users at
+        this time, and therefore return 501 Not Implemented.
+        """
+        return Response(status=501)
```
lunika: Did we plan to implement them in the future ? If we don;t know we should probably return a 404

tests/apps/core/test_api.py
```diff
@@ -0,0 +1,47 @@
...
+from django.contrib.auth.models import User
+from django.test import TestCase
+
+
+class ApiUsersViewSetTestCase(TestCase):
+    """Test requests on user-related API endpoints."""
+
+    def test_viewsets_users_retrieve(self):
+        """The expected endpoint to retrieve users by ID is not implemented."""
+        response = self.client.get(f"/api/v1.0/users/42/")
```
lunika: Why are you using a f-string here ?

src/frontend/js/components/UserLogin/index.tsx
```diff
@@ -0,0 +1,109 @@
...
+   * `user` is:
+   * - `undefined` when we have not made the `whoami` request yet;
+   * - `null` when the user is anonymous or the request failed;
+   * - a user object when the user is logged in.
+   */
+  const [user, setUser] = useState<Maybe<Nullable<User>>>(undefined);
+
+  useAsyncEffect(async () => {
+    try {
+      const response = await fetch('/api/v1.0/users/whoami/', {
```
lunika: the `/api/v1.0` part of the endoint is not in a constant ? The day we will increment the api version we will have to edit all files ? You should probably open an issue for this ?

mbenadda: IMO it's not a big deal given the size of Richie's JS codebase but sure I can open an issue.

src/richie/apps/core/api.py
```diff
@@ -0,0 +1,45 @@
...
+from .serializers import UserSerializer
+
+
+class UsersViewSet(ViewSet):
+    """
+    API endpoints to access users.
+    """
+
+    # pylint: disable=no-self-use,unused-argument
+    def list(self, request, version):
```
sampaccoud: I think you should return 405 http codes for methods that are not supported. Isn't it what DRF returns if you just don't declare any list method on the class?

src/richie/apps/core/api.py
```diff
@@ -0,0 +1,45 @@
...
+    def retrieve(self, request, pk, version):
+        """
+        Single item requests for users. We do not provide an API to get registered user by id
+        at this time, and therefore return 501 Not Implemented.
+        """
+        return Response(status=501)
+
+    @action(detail=False)
+    # pylint: disable=no-self-use,invalid-name,unused-argument
+    def whoami(self, request, version):
```
sampaccoud: from django.contrib.auth.decorators import login_required
from django.utils.decorators import method_decorator 

```suggestion
    @method_decorator(login_required)
    def whoami(self, request, version):
```

mbenadda: Does this not redirect to a login page? (this is what I understand form the docs https://docs.djangoproject.com/en/2.2/topics/auth/default/#the-login-required-decorator)

src/richie/apps/core/api.py
mbenadda: IMO `405` is the wrong code to return here. It's not as if we support `POST` or `PUT` or whatever on the endpoints.

If `501` is not what we want then we're probably better off with `404` as @lunika suggests?

mbenadda: DRF returns a `404` if the methods are not implemented. I'll just do the same here.

mbenadda:
Update: I can't return a `404` because the request is then further handle by Django's matching, ending up with a `302` (probably as a redirect to localized assets or homepage?)

src/richie/apps/core/api.py
sampaccoud: so just remove these methods from you class?

src/richie/apps/core/api.py
sampaccoud: I'm not sure but your test should tell. The correct way to do it otherwise should be to raise the `NotAuthenticated` exception: https://www.django-rest-framework.org/api-guide/exceptions/#notauthenticated

mbenadda: - the `@method_decorator(login_required)` does a `302` redirect which is not what we want on an API endpoint

- `raise NotAuthenticated` returns a `403` which seems less appropriate than a `401` as this route **is intended to be used by unauthenticated users**

mbenadda:
Update 2: I just removed `list` & `retrieve` and dropped the tests. No point in specifying & testing endpoints that do not exist.

src/frontend/js/components/UserLogin/index.spec.tsx
```diff
@@ -0,0 +1,80 @@
...
+    loginUrl: '/login',
+    logoutUrl: '/logout',
+    signupUrl: '/signup',
+  };
+
+  beforeEach(fetchMock.restore);
+
+  it('gets and renders the user name and a log out button', async () => {
+    fetchMock.get('/api/v1.0/users/whoami/', {
+      full_name: 'Decimus Iunius Iuvenalis',
```
lunika: nice name

PR #880, opened 2019-10-31
🔥(styles) remove custom "Exo 2" webfont from the project

mbenadda:
## Purpose

Earlier, we added the "Exo 2" webfont to Richie. In our opinion, this was not appropriate as Richie is meant to be an Open Source project, and styling the resulting portals should be the responsibility of dependents.

Exo 2 is a very distinctive font, that therefore should not be the default
font for a project like Richie.

## Proposal

 We're just removing it from the statics and our CSS font definitions.


jmaupetit:
Why not using system fonts instead "à la github"? https://css-tricks.com/snippets/css/system-font-stack/

PR #886, opened 2019-11-07
Handle unexpected calls on categories catchall API routes

mbenadda:
## Purpose

The categories ViewSet sits on a catchall url so it can receive requests for any "kind" of category in an intuitive way. This has a drawback though: unexpected requests might end up in our categories
routes.

## Proposal

We simply want to make sure those are handled properly and return eg. 404 errors, and do not raise 500s or return any unwanted data.

### DIFF
--- a/src/richie/apps/search/viewsets/categories.py
+++ b/src/richie/apps/search/viewsets/categories.py
@@ -6,6 +6,7 @@ from django.utils.translation import get_language_from_request
 
 from elasticsearch.exceptions import NotFoundError
 from rest_framework.decorators import action
+from rest_framework.exceptions import NotFound
 from rest_framework.response import Response
 from rest_framework.viewsets import ViewSet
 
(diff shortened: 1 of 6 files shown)


tests/apps/search/test_viewsets_categories.py
```diff
@@ -181,14 +164,29 @@ def test_viewsets_categories_search_with_invalid_params(self):
...
         self.assertEqual(response.status_code, 400)
         self.assertTrue("limit" in response.data["errors"])
+
+    def test_viewsets_categories_catchall_retrieve(self):
+        """
+        Error case: unrelated requests end up in the categories ViewSet.
+        Since the categories ViewSet sits on a catchall URL, we need to make sure it returns
+        proper errors and does not raise eg. a 500 when called from an unexpected URL.
+        """
+        # The request is for users instead of categories
```
sampaccoud:
```suggestion
        # The request is for a user instead of categories
```

src/richie/apps/search/viewsets/categories.py
```diff
@@ -71,7 +74,7 @@ def list(self, request, version, kind):
         return Response(response_object)
 
     # pylint: disable=no-self-use,invalid-name,unused-argument
-    def retrieve(self, request, pk, version):
+    def retrieve(self, request, pk, version, kind):
```
sampaccoud: not sure why you are changing this and why it is still working :man_shrugging:?

src/richie/apps/search/viewsets/categories.py
```diff
@@ -38,16 +38,19 @@ def list(self, request, version, kind):
...
+                _source=getattr(self._meta.indexer, "display_fields", "*"),
+                index=self._meta.indexer.index_name,
+                doc_type=self._meta.indexer.document_type,
+                body=query,
+                # Directly pass meta-params through as arguments to the ES client
+                from_=offset,
+                size=limit or getattr(settings, "RICHIE_ES_PAGE_SIZE", ES_PAGE_SIZE),
+            )
+        except NotFoundError:
+            return Response(status=404)
```
sampaccoud: did you try raising NotFound https://www.django-rest-framework.org/api-guide/exceptions/#notfound ?

Issue #888, opened 2019-11-08
Accessibility — WCAG AA / RGAA Compliance


mbenadda:
## Richie is not accessible

Richie is not RGAA/WCAG compliant. This is a problem for us because we want to support all users. This of course starts, but does not end, with compliance with those norms.

An professional audit of Richie was done in mid-2019 that pointed out a lot of issues. Some of them have been fixed since then. This issue exists to list the work that remains to be done to achieve RGAA/WCAG AA compliance.

Note: this issue should evolve over time to track our progress towards an accessible Richie.

### Text contrast

Colors in Richie are somewhat all over the place and most of the color pairings outside of black-on-white fail the minimal required 4.5 contrast ratio (although a value closer to 7 is preferable for regular-sized text).

This is in part dependent on theming for the final result but Richie needs a general overhaul of its color system, and needs to have acceptable color pairings by default.

- [ ] overhaul color management in Richie to simplify the use of accessible pairs
- [ ] make sure Richie comes with accessible colors by default

PR #889, opened 2019-11-08
Fix logo link in base footer

mbenadda:
## Purpose

The logo link in the footer used an empty alt text on the image, which is the only content for the link. This is incorrect as there is no text at all on the link for accessibility tech users.

## Proposal

We can just add `{{ SITE.name }}` like on the logo in the header.

Part of our effort on #888

### DIFF
--- a/src/richie/apps/core/templates/richie/base.html
+++ b/src/richie/apps/core/templates/richie/base.html
@@ -95,7 +95,7 @@
                 </div>
                 <div class="body-footer__brand">
                     <a href="/" class="body-footer__brand__link">
-                        <img src="{% static "richie/images/logo.png" %}" class="body-footer__brand__logo" alt="">
+                        <img src="{% static "richie/images/logo.png" %}" class="body-footer__brand__logo" alt="{{ SITE.name }}">
                     </a>
                     {% include "social-networks/footer-badges.html" %}
                 </div>


src/richie/apps/core/templates/richie/base.html
```diff
@@ -95,7 +95,7 @@
                 </div>
                 <div class="body-footer__brand">
                     <a href="/" class="body-footer__brand__link">
-                        <img src="{% static "richie/images/logo.png" %}" class="body-footer__brand__logo" alt="">
+                        <img src="{% static "richie/images/logo.png" %}" class="body-footer__brand__logo" alt="{{ SITE.name }}">
```
sampaccoud: add a small test to explain this and secure it?

mbenadda: IMO this is pretty basic stuff and there's not tests for the footer AFAIK. Should I start a new test file for the footer to test this?

sampaccoud: Or add a Django template comment to explain why alt must be set as the site name?
I'm really afraid that these improvements get lost by developpers not aware about accessibility.

sampaccoud: ok then :man_shrugging:

PR #891, opened 2019-11-08
Fix language menu accessibility

mbenadda:
## Purpose

Link titles, which are necessary for language switching links to provide context on what those links actually do, were broken due to an incorrect use of `{% blocktrans %}`.

## Proposal

We used blocktrans variables to fix them, and took this opportunity to improve the link titles by adding a different one for the currently selected language.

Part of our effort on #888


tests/apps/core/test_language_chooser.py
```diff
@@ -58,6 +61,8 @@ def test_language_chooser_available_language_with_translated_page(self):
                 'topbar__menu__list__item--en">'
             ),
         )
+        self.assertContains(response, "en, switch to Anglais")
```
sampaccoud: This will break once the translations become available right? Why not include the translations in this PR and make the correct test directly?

mbenadda: Done

tests/apps/core/test_language_chooser.py
```diff
@@ -67,3 +72,4 @@ def test_language_chooser_available_language_with_translated_page(self):
                 'topbar__menu__list__item--active">'
             ),
         )
+        self.assertContains(response, "fr, currently in Français")
```
sampaccoud: should be "actuellement en Français" or something like that?

Issue #870, opened 2019-10-25
Dockerize the Richie documentation & website


mbenadda:
## Feature Request

**Is your feature request related to a problem or unsupported use case? Please describe.**
The documentation website for Richie uses `yarn` for local development and build. This is fine for frontend developers but might be a problem for developers who prefer to operate everything inside containers.

It also means users need to use `yarn` if they wish to see the documentation website locally (as opposed to the published online version).

**Describe the solution you'd like**
A new entry can be added in `docker-compose` to allow (at least) `yarn start` which serves the docs website, and could go further to allow work on the docs (with eg. `yarn version`)/

**Describe alternatives you've considered**
N/A

**Discovery, Documentation, Adoption, Migration Strategy**
We can then add a line in the main `README.md` that points to the way the docs website can be served locally.

**Do you want to work on it through a Pull Request?**
🚪🏃‍♂️

mbenadda:
This work should also encompass `Makefile` tasks related to the docs website, as well as one for to run prettier on it, just like we have for the frontend.

PR #893, opened 2019-11-12
Run prettier on website dir

mbenadda:
## Purpose

We were not running prettier on the JS and JSON files in the website directory, and had no CI job to do that.

## Proposal

- [x] add a CI job for prettier linting on the website dir
- [x] run prettier write on the website dir to fix existing issues


mbenadda:
I don't think it's worth it as we don't have an image that operates on the website directory.

This can probably be done if/when someone enables docker for the other website related tasks like building it, running it etc.

mbenadda:
(I added a comment to this effect on #870)

PR #895, opened 2019-11-13
Rework search view layout

mbenadda:
## Purpose

The Search component was placing the title and search field inside a `search__results` div and pushing the filters pane to the very top of the page.

## Proposal

We now have a dedicated `search__header` area that helps us make the layout cleaner and lighter while keeping everything as easy to access as before.

Note: we took this opportunity to remove react test warnings on the `<Search />` component.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Changed
 
+- Rework `<Search />` layout to make it lighter and simpler.
 - Remove the "Exo 2" font from Richie. Use system `sans-serif` instead.
 - Rename the `fun-react` class used to Django-React interop to `richie-react`.
 - Improve documentation on Django-React interop.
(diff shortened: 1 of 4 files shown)


jmaupetit:
@mbenadda can you add snapshots to your PR description please? :pray:

mbenadda:
I can't screenshot every state and every device format. Here are are 2 examples:

### Before

before large

### After

after large

### Before

before small

### After

after small

mbenadda:
After speaking with @sampaccoud this morning we decided on a larger refactor that will enable us to split the Search bar into a separate component. This should provide what you need.

In the meantime I think we can fo forward with this one

jmaupetit:
After though?

mbenadda:
The way this is implemented, while it works for Richie right now, does not suit us going forward.

In any case, it will be made mostly irrelevant soon by the larger refactor I was alluding to earlier. We decided to drop the layout changes and just keep the small improvement commit in #896.

PR #898, opened 2019-11-15
Add an API endpoint to get static search filter definitions

mbenadda:
## Purpose

We're refactoring our course search to make it more modular. One key part of this is to allow components to use search/autosuggest based on filters without making full course search requests first.

## Proposal

- [x] rework filter definitions to add a method that returns the static definition for a filter
- [x] all filter definitions have a `base_path` property
- [x] add an API route with the static filter definitions

While doing this, we noticed a missing trailing slack on the `bootstrap-elasticsearch` API url, and decided to add it.

### DIFF
--- a/src/richie/apps/search/views.py
+++ b/src/richie/apps/search/views.py
@@ -4,9 +4,12 @@ from django.core import management
 from django.http import HttpResponse
 from django.utils.encoding import force_text
 from django.utils.translation import ugettext_lazy as _
+from django.views.decorators.cache import cache_page
 
 from rest_framework.decorators import api_view
 from rest_framework.response import Response
+
+from .filter_definitions import FILTERS
 
 
 @api_view(["POST"])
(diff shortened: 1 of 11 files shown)


src/richie/apps/search/views.py
```diff
@@ -24,3 +26,19 @@ def bootstrap_elasticsearch(request, version):
 
     messages.info(request, _("The search index was successfully bootstrapped"))
     return Response({})
+
+
+@api_view(["GET"])
+# pylint: disable=unused-argument
+def filter_definitions(request, version):
```
sampaccoud: Since this only changes when we change settings, it could be calculated only once at startup instead of calling it each time the view is called.

mbenadda:
@sampaccoud done! I ended up just adding cache to the filter definitions view

src/richie/apps/search/views.py
```diff
@@ -24,3 +27,20 @@ def bootstrap_elasticsearch(request, version):
 
     messages.info(request, _("The search index was successfully bootstrapped"))
     return Response({})
+
+
+@api_view(["GET"])
+@cache_page(60 * 60 * 2)
```
sampaccoud: maybe add a test that mock `get_static_definitions` and ensures that it is not called on the second request to this url?

mbenadda: Done 👍

PR #900, opened 2019-11-18
Fix language chooser

sampaccoud:
## Purpose

The language chooser was showing languages even when they were marked as "not public" in the settings.

## Proposal

- Remove the template tag that gets languages for the language chooser and rely on the languages variable already prepared by the language chooser template tag. It respects the public status of languages.
- Refactor the language chooser menu to include the `<ul>` tag so we can avoid having an empty `<ul>` on sites that have only one public language.

### DIFF
--- a/src/richie/apps/core/templates/menu/language_menu.html
+++ b/src/richie/apps/core/templates/menu/language_menu.html
@@ -1,22 +1,26 @@
 {% load i18n menu_tags cms_tags %}{% spaceless %}
-{% get_current_language as current_language_code %}
+
 {% if languages|length > 1 %}
+  {% get_current_language as current_language_code %}
+  <ul class="topbar__menu__list topbar__menu__list--aside">
     {% for code, name in languages %}
-    {% get_language_info for code as lang %}
-        {% if code == current_language_code %}
-            <li class="topbar__menu__list__item topbar__menu__list__item--language topbar__menu__list__item--{{ code }} topbar__menu__list__item--active">
-                <a class="topbar__menu__list__item__link" href="{% page_language_url code %}"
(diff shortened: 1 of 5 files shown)


src/richie/apps/core/templates/richie/base.html
```diff
@@ -63,9 +63,7 @@
                     <ul class="topbar__menu__list">
                         {% show_menu 0 100 0 0 "menu/header_menu.html" %}
                     </ul>
-                    <ul class="topbar__menu__list topbar__menu__list--aside">
-                        {% language_chooser "menu/language_menu.html" %}
-                    </ul>
+                    {% language_chooser "menu/language_menu.html" %}
```
mbenadda: This is great, nice catch

The template boundaries were incorrect before IMO 👍

src/richie/apps/core/templates/menu/language_menu.html
```diff
@@ -1,22 +1,29 @@
 {% load i18n menu_tags cms_tags %}{% spaceless %}
 {% get_current_language as current_language_code %}
 {% if languages|length > 1 %}
-    {% for code, name in languages %}
+  {% for code, name in languages %}
+    {% if forloop.first %}
+      <ul class="topbar__menu__list topbar__menu__list--aside">
```
mbenadda: Can't you just put this inside the `if languages|length > 1`?

sampaccoud: But it is already no?

mbenadda: But I mean the `{% if forloop.first %}` and last are not necessary right?

sampaccoud: You damn right!

PR #901, opened 2019-11-19
Refactor search for modularity

mbenadda:
## Purpose

We want to unplug `<SearchSuggestField />` from `<Search />` so we can extract it and put it somewhere else on the page. Then, they can interact through the URL query string.

## Proposal

One key dependency between `<SearchSuggestField />` and the rest of `<Search />` is that `<SearchSuggestField />` needs the filter definitions to configure itself.

This can be solved, and we can improve `<RootSearchSuggestField />` in the process, by using our new `filter-definitions` API endpoint to directly get static filter definitions from those components.

- [x] update types to add newly available static filter definitions
- [x] create a hook with a simple API so components can access the static filter definitions
- [x] use the new hook in `<SearchSuggestField />` and `<RootSearchSuggestField />`

### DIFF
--- a/src/frontend/js/data/useStaticFilters/index.tsx
+++ b/src/frontend/js/data/useStaticFilters/index.tsx
@@ -1,0 +1,54 @@
+import { useRef, useState } from 'react';
+
+import { FilterDefinition, StaticFilterDefinitions } from 'types/filters';
+import { Nullable } from 'utils/types';
+import { useAsyncEffect } from 'utils/useAsyncEffect';
+
+// Our search and autosuggestion pipeline operated based on filter definitions. Obviously, we can't filters
+// courses by courses, but we still need a filter-definition-like config to run courses autocompletion.
+const coursesConfig: FilterDefinition = {
+  base_path: null,
+  human_name: 'Courses',
+  is_autocompletable: true,
+  is_searchable: true,
+  name: 'courses',
+  position: 99,
+};
+
+/**
+ * Hook to provide static filter definitions to components as a promise while abstracting away
+ * data-fetching & caching logic.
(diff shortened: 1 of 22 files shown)


src/frontend/js/data/useStaticFilters/index.tsx
```diff
@@ -0,0 +1,48 @@
...
+const coursesConfig: FilterDefinition = {
+  base_path: null,
+  human_name: 'Courses',
+  is_autocompletable: true,
+  is_searchable: true,
+  name: 'courses',
+  position: 99,
+};
+
+export const useStaticFilters = (includeCoursesConfig = false) => {
```
lunika: Hoe to use this filter is not obvious for me. Maybe specifying the return value can help to use it ?

mbenadda: done, let me know if it's clearer now

mbenadda:
I don't think there's anything that warrants a changelog entry here. Nothing impacts Richie users nor end users.

src/frontend/js/data/useStaticFilters/index.tsx
lunika: Not really, `useStaticFilters` does not return a `Promise`, it returns a function and this function returns a `Promise`. This is important to specify it to understand how to use it, without an example or reading the function it's not obvious.

mbenadda: Oh right, my bad! Fixing this right now 😅

lunika:
and richie's developers ? Since we allow the possibility to modify/extend it I think any code modification should be noticed in the changelog.

mbenadda:
I think we're never gonna agree on this... I added a changelog entry

PR #903, opened 2019-11-21
✨(project) add a command to lint only what has changed since master

sampaccoud:
## Purpose

The backend linters have become very slow, especially pylint. When working on a branch, we don't need to lint the whole project, but only what has changed. 

## Proposal

The idea of this command is to run the linters only on the files pointed by a `git diff --name-only master HEAD`.

### DIFF
--- a/bin/lint-back-diff
+++ b/bin/lint-back-diff
@@ -1,0 +1,15 @@
+#!/usr/bin/env bash
+source "$(dirname "${BASH_SOURCE[0]}")/_config.sh"
+
+DIFF_FILES="$(git diff --name-only master | grep -e .py$ | grep -v migrations | tr "\n" " ")" || true
+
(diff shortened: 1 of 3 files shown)


bin/lint-back-diff
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+source "$(dirname "${BASH_SOURCE[0]}")/_config.sh"
+
+DIFF_FILES="$(git diff --name-only master HEAD | grep -e .py$ | grep -v migrations | tr "\n" " ")" || true
```
jmaupetit: I suggest you the following that allows to run back-end linters even on un-committed changes (not only committed changes compared to `master`):

```suggestion
DIFF_FILES="$(git diff --name-only master | grep -e .py$ | grep -v migrations | tr "\n" " ")" || true
```

Issue #902, opened 2019-11-20
Main organisation selection


2018D:
## Bug Report

**Expected behavior/code**
The main organisation used to always be the first in the list of organization within a course.

**Actual Behavior**
The main organization is the first overall in the list of organizations.

**Environment**
Richie version: master

**Possible Solution**
Uncomment line  https://github.com/openfun/richie/blob/daee32dce1b7212c0ace9e827b60c749368bd669/src/richie/apps/courses/models/course.py#L287 of src/richie/apps/courses/models/course.py
to bring back the original behaviour.

PR #905, opened 2019-11-21
🐛(courses) fix main organization broken by a comment & a missing test

sampaccoud:
## Purpose

The main organization method was broken since we inadvertently commented a line in the code. 

## Proposal

The real problem is that this was not pinpointed by a breaking test. I fixed the method and improved the related test to make sure that it fails when the ordering clause is commented.

Fixes https://github.com/openfun/richie/issues/902

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,9 +8,19 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Added
+
+- Add an API endpoint to get static versions of the filter definitions.
+
 ### Changed
 
 - Change the way frontend search field components are configured.
+
+### Fixed
+
+- Main organization after an ordering clause was inadvertently commented,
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -8,18 +8,27 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Added
```
mbenadda: I think your rebase broke the CHANGELOG :)

bin/lint-back-diff
```diff
@@ -1,15 +1,17 @@
 #!/usr/bin/env bash
```
mbenadda: in commit message `would raise useless`

CHANGELOG.md
mbenadda: My bad! Mistakes were made earlier and you're actually fixing it 👍

PR #906, opened 2019-11-22
Refactor frontend localization

mbenadda:
## Purpose

As we were refactoring our frontend bootstrapping code, we were hindered by the way localization works for frontend components:

- each component has its own `data-locale`
- polyfills and message loading logic is executer on a per-component basis (but always loads everything)
- we needed conversion between languages, BCP47/RFC5646 locales and ISO15897 locales

## Proposal

Instead we propose to harmonize localization in the frontend through a single source of truth: the `lang` attribute on the `<html>` tag that we already need to be accessible.

- [x] rework the `react_locale` filter into `rfc_5646_locale`
- [x] use this locale on `<html lang>`
- [x] rename client side translation files so they are named after the BCP47/RFC5646 locale and not the ISO15897 locale
- [x] only run the localization/polyfills code once for the whole app
- [x] update docs & `UPGRADE.md` to reflect these changes


src/frontend/js/index.tsx
```diff
@@ -60,8 +60,11 @@ document.addEventListener('DOMContentLoaded', event => {
         .join('');
```
sampaccoud: in commit message: 
s/communicated/communicates

src/frontend/js/index.tsx
```diff
@@ -60,8 +60,11 @@ document.addEventListener('DOMContentLoaded', event => {
...
       // Sanity check: only attempt to access and render components for which we do have a valid name
       if (isComponentName(componentName)) {
-        // Determine a lang (localeCode) based on the `data-locale` attribute
-        const locale = element.getAttribute('data-locale') || 'en';
+        // Determine an ISO15897 locale from a BCP47/RFC5646 locale (eg. en-US to en_US)
+        const locale = (element.getAttribute('data-locale') || 'en-US').replace(
```
sampaccoud: In Django languages are set to "en-us" and not "en-US" so maybe you should still force capitalize on the last 2 characters?

src/richie/apps/core/templatetags/rfc_5646_locale.py
```diff
@@ -1,28 +1,31 @@
...
+def rfc_5646_locale(value):
     """
-    Converts a language (simple ISO639-1 or full language with regions) to an ISO15897 locale
-    that is supported by the React frontend.
+    Converts a language (simple ISO639-1 or full language with regions) to a BCP47/RFC5646 locale.
+    This will allow us to use this locale on the <html lang> attribute directly.
     """
-    value_locale = to_locale(value)
+    p = value.find("_")
+    if p >= 0:
```
sampaccoud: could be written as:
```python
if "_" in value:
```

tests/apps/core/test_templatetags_rfc_5646_locale.py
```diff
@@ -0,0 +1,51 @@
...
+
+    @override_settings(RFC_5646_LOCALES=["en-US", "fr-CA", "fr-FR", "es-ES"])
+    def test_templatetags_search_tags_rfc_5646_locale_canada_first(self):
+        """
+        Languages should return the first matching BCP47/RFC5646 locale (Canada comes
+        before France).
+        """
+        # Full locales incl. region
+        self.assertEqual(rfc_5646_locale("fr_FR"), "fr-FR")
+        self.assertEqual(rfc_5646_locale("fr_CA"), "fr-CA")
```
sampaccoud: as pointed above, full languages in Django will normally be in the form: "en-us" so make sure you also support it.

tests/apps/search/test_templates_search.py
```diff
@@ -33,7 +33,6 @@ def test_templates_search_content(self):
                 (
                     r"<div[\\n ]*"
                     r'class="richie-react richie-react--search"[\\n ]*'
-                    r'data-locale="en_US"[\\n ]*'
```
sampaccoud: replace it by a test that ensures the locale is in the set in the <html> tag as expected?

src/frontend/js/index.tsx
```diff
@@ -47,11 +47,49 @@ function isComponentName(
...
+  const richieReactSpots = Array.prototype.slice.call(
     document.querySelectorAll('.richie-react'),
-    async (element: Element) => {
+  );
+
+  // Only move on with anything if there are actually components to render
+  if (richieReactSpots.length) {
+    // Determine the BCP47/RFC5646 locale to use, defaulting to en-US
+    const locale =
+      document.querySelector('html')!.getAttribute('lang') || 'en-US';
```
sampaccoud: the || 'en-US' defeats our purpose as lang is required to be set?

UPGRADE.md
```diff
@@ -16,9 +16,11 @@ $ make migrate
 
 ## Unreleased
 
+- If you reused any React components in Django templates through `richie-react`, you should know that the `data-locale` attribute on them is no longer used. Instead, they rely on `<html lang>` to pick up their current locale, which should already be set in all documents anyway.
```
sampaccoud: but most importantly : if you override the richie/base.html template, make sure you port to your template the modification we made on the `lang` attribute of the <html> element.

docs/django-react-interop.md
```diff
@@ -12,15 +12,18 @@ We needed a convention that enables us to easily mark those areas of the page th
 
 We decided to use a specific CSS class name along with its modifiers. We reserve the `richie-react` class and its modified children for this purpose.
 
-Additionally, components including internationalized data or strings need to know which locale to use. All uses of this `richie-react` class can be paired with a data-attribute specifying the locale. If not defined, `en` is used as a default.
+Additionally, components including internationalized data or strings need to know which locale to use. They will pick up the locale specified on the `<html>` element through the `lang` attribute.
```
sampaccoud: maybe clearer with:
specified on the `lang` attribute of the `<html>` element

src/frontend/js/index.tsx
sampaccoud: See here: https://github.com/django/django/blob/master/django/conf/global_settings.py#L51

src/frontend/js/index.tsx
mbenadda: You're right, it should probably just break in this case

src/frontend/js/index.tsx
mbenadda: This is done by `rfc_5646_locale`. We already need the capitalization to be done for the `<html lang>` because otherwise it's not a valid BCP47/RFC5646 locale.

src/richie/apps/core/templatetags/rfc_5646_locale.py
mbenadda: we're reusing this index 2 times in the line right below. IMO it's ok like this?

lunika:
You also have to modify crowdin configuration to handle the new locale format.

UPGRADE.md
```diff
@@ -16,9 +16,12 @@ $ make migrate
...
+- As a result, the `lang` attribute on `<html>`, which is a requirement for an accessible page anyway, is now an absolute requirement. Users who overrode the base template and removed this need to add it again.
+
```
sampaccoud: One more thing : users who overrode the `REACT_LOCALES` setting (for example to change priority order) should rename the setting to `RFC_5646_LOCALES` and replace their locales by FRC5646 locales in this setting value.

mbenadda:
@lunika the changes this PR makes to `crowdin/config.yml` are not enough? It works on my machine when I download the translations now

lunika:
> @lunika the changes this PR makes to `crowdin/config.yml` are not enough? It works on my machine when I download the translations now

Yes ot's enough sorry, I didn't saw the modification...

PR #914, opened 2019-11-26
🌐(frontend) do not report errors related to en-US l10n files

### LINKED ISSUE #913: Frontend should not generate errors when it does not find `en-US` l10n files
## Bug Report

**Expected behavior/code**
Richie's frontend default strings are for the `en-US` locale, and we should not see errors for not finding specific localization files for this locale.

**Actual Behavior**
Richie's frontend reports an error that it fails to find a localization file for the `en-US` locale (when the site is loaded with this locale).

**Steps to Reproduce**
Open any page with locale as `en-US`, look at the console.

**Environment**
1.14+

**Possible Solution**
Since we don't create localization files for `en-US`, we should not attempt to fetch them or report error when they are not found.
Webpack will not make requests (and not generate 404s) because it maps all available modules during the build. However, it still produces an error when we `import()` the `en-US` module that does not exist.

We should report those errors when they are for any locale (since all others do need the localization file) except `en-US` which should instead just produce a helpful `console.log`.

N/A

mbenadda:
## Purpose

Richie's frontend operates on the assumption that en-US is the default language and is contained in the default messages as opposed to localization files that are lazily loaded for other languages.

We don't want to add a special case that does not import() an l10n file for en-US because we don't want to prevent Richie users from creating their own and loading it.

## Proposal

Therefore, we decided to load it if it's there, but not report an error that is in the default and most common case, just the app working as intended.

Closes #913

### DIFF
--- a/src/frontend/js/index.tsx
+++ b/src/frontend/js/index.tsx
@@ -91,7 +91,19 @@ document.addEventListener('DOMContentLoaded', async event => {
     try {
       translatedMessages = await import(`./translations/${locale}.json`);
     } catch (e) {
-      handle(e);
+      // Richie implicitly uses 'en-US' as its default locales: it does not provide localization files and strings
+      // for this locale but instead relies on them being the default messages throughout the code.
+      // We therefore do not want to report errors for a missing locale file that is not expected to exist in stock
+      // Richie. We still want errors if another locale is expected but missing, and we still want to log something
+      // to the console so Richie users who provide their own 'en-US' strings are warned if they are not loaded.
+      if (locale === 'en-US') {
+        // tslint:disable:no-console
+        console.log(
+          'No localization file found for default en-US locale, using default messages.',
+        );
+      } else {
+        handle(e);
+      }
     }
 
     richieReactSpots.forEach((element: Element) => {


src/frontend/js/index.tsx
```diff
@@ -91,7 +91,19 @@ document.addEventListener('DOMContentLoaded', async event => {
...
-      handle(e);
+      // Richie implicitly uses 'en-US' as its default locales: it does not provide localization files and strings
+      // for this locale but instead relies on them being the default messages throughout the code.
+      // We therefore do not want to report errors for a missing locale file that is not expected to exist in stock
+      // Richie. We still want errors if another locale is expected but missing, and we still want to log something
+      // to the console so Richie users who provide their own 'en-US' strings are warned if they are not loaded.
+      if (locale === 'en-US') {
+        // tslint:disable:no-console
+        console.log(
+          'No localization file found for default en-US locale, using default messages.',
```
sampaccoud: add a test if possible?

mbenadda: Unfortunately I’m not sure how we could go about testing this since it’s just webpack internal machinery.

The index is not tested. I specifically left this code there when I refactored and made the root component due to the challenge of testing this

PR #919, opened 2019-12-05
Add tooling for e2e testing and use it to run automated a11y checks

mbenadda:
## Purpose

As part of our work to make Richie accessible (in #888), we decided to add automated accessibility checks. Although they only surface a small part of all possible problems, we believe they can have value in preventing regressions and giving us a "cheap" win (because automated checks are so much easier than human audits).

## Proposal

- [x] add `Cypress` and create a folder for our end-to-end tests
- [x] integrate `axe` (a11y checks engine) with `Cypress`
- [x] write smoke tests to just load views and run the a11y checks
- [x] add some basic documentation on how to run the checks (users should rely directly on `Cypress` and `axe` docs for more in-depth information)


mbenadda:
Note: I'm not entirely satisfied of the way I reach some pages with clicks on given strings or elements with certain CSS classes (it is pretty brittle) but I'm not sure if it's possible to do better until we add pre-defined fixtures or a more deterministic version of the demo site specifically for end-to-end tests.

PR #920, opened 2019-12-06
Update our JS to support IE11 and old browsers

mbenadda:
## Purpose

Richie is not IE11 ready. One of the major parts that did not work with outdated browsers was the JS, which needed to be built specifically to support it.

## Proposal

- [x] add a browserslist with our target browsers
- [x] build dependencies that can't run in IE11
- [x] add required polyfills for IE11
- [x] update `base.html` JS so it runs in IE11

### DIFF
--- a/src/richie/apps/core/templates/richie/base.html
+++ b/src/richie/apps/core/templates/richie/base.html
@@ -110,26 +110,26 @@
         {% render_block "js" %}
         <script src="{% static 'richie/js/index.js' %}"></script>
         <script>
-        document.addEventListener('DOMContentLoaded', () => {
-            // Get all topbar burger elements
-            const $topbarBurgers = Array.prototype.slice.call(document.querySelectorAll('.topbar__hamburger'), 0);
-
-            // Check if there are any navbar burgers
-            if ($topbarBurgers.length > 0) {
+            document.addEventListener("DOMContentLoaded", function() {
+              // Get all topbar burger elements
+              const topbarBurgers = Array.prototype.slice.call(
+                document.querySelectorAll(".topbar__hamburger"),
+                0
+              );
+              // Check if there are any navbar burgers
+              if (topbarBurgers.length > 0) {
                 // Add a click event on each of them
-                $topbarBurgers.forEach( el => {
-                    el.addEventListener('click', () => {
-                        // Get the target from the "data-target" attribute
-                        const target = el.dataset.target;
-                        const $target = document.getElementById(target);
-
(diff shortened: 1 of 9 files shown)


src/richie/apps/core/templates/richie/base.html
```diff
@@ -110,26 +110,29 @@
...
-                $topbarBurgers.forEach( el => {
-                    el.addEventListener('click', () => {
-                        // Get the target from the "data-target" attribute
-                        const target = el.dataset.target;
-                        const $target = document.getElementById(target);
+                $topbarBurgers.forEach(function(el) {
+                  el.addEventListener("click", function() {
+                    // Get the target from the "data-target" attribute
+                    const target = el.dataset.target;
+                    const $target = document.getElementById(target);
```
lunika:
```suggestion
                    const target = document.getElementById(el.dataset.target);
```

src/richie/apps/core/templates/richie/base.html
```diff
@@ -110,26 +110,29 @@
...
                 // Add a click event on each of them
-                $topbarBurgers.forEach( el => {
-                    el.addEventListener('click', () => {
-                        // Get the target from the "data-target" attribute
-                        const target = el.dataset.target;
-                        const $target = document.getElementById(target);
+                $topbarBurgers.forEach(function(el) {
+                  el.addEventListener("click", function() {
+                    // Get the target from the "data-target" attribute
+                    const target = el.dataset.target;
```
lunika: you can remove this line. this comment for next line.

src/richie/apps/core/templates/richie/base.html
lunika: Also rename `$target` in `target`. We are not doint php ;-)

src/richie/apps/core/templates/richie/base.html
```diff
@@ -110,26 +110,29 @@
         {% render_block "js" %}
         <script src="{% static 'richie/js/index.js' %}"></script>
         <script>
-        document.addEventListener('DOMContentLoaded', () => {
-            // Get all topbar burger elements
-            const $topbarBurgers = Array.prototype.slice.call(document.querySelectorAll('.topbar__hamburger'), 0);
+            document.addEventListener("DOMContentLoaded", function() {
+              // Get all topbar burger elements
+              const $topbarBurgers = Array.prototype.slice.call(
```
lunika: why having a `$` at the beginning of your variable name ?

mbenadda:
I fixed the base.html JS issues you pointed out 👍

PR #923, opened 2019-12-07
Add categories on blogpost list page

sampaccoud:
## Purpose

We want to add a row on top of the news page with a link to each category that is tagged on one or more blog posts.

## Proposal

We introduce a new template tag to do this. The template tag is generic and can be used to retrieved all categories linked to any group of pages via a `CategoryPlugin`.

After improving the demo site data, the feature looks like this:

<p align="center">

</p>

While working on this, I had the opportunity to fix a few bugs and refactor a few things:
- [x] Improve responsiveness of footer logos
- [x] Improve responsiveness of person categories
- [x] Fix placeholders on a page created with PageFactory
- [x] Restructure categories glimpse/plugin for modularity and harmonization with other types of pages
- [x] Rename "form_factor" to "variant" as we had started doing elsewhere for clarity
- [x] Make "get_placeholder_plugins" template tag fail nicely when the page lookup passed as argument returns no page


UPGRADE.md
```diff
@@ -16,7 +16,13 @@ $ make migrate
 
 ## Unreleased
 
-- Users who make use of `<SearchSuggestField />` or `<RootSearchSuggestField />` in their own templates through `richie-react` need to update all the call sites: the `context` prop is now required for both of them. See the documentation for more details on `context`.
+- All occurences of the `form_factor` variable have been renamed to `variant` for clarity.
```
sampaccoud: You forgot to mention here about the refactoring of categories glimpse/plugin.

src/frontend/scss/objects/_blogpost_glimpses.scss
```diff
@@ -80,6 +80,29 @@ $richie-blogpost-glimpse-foot-cta-background-hover: $firebrick6 !default;
     padding: $richie-blogpost-glimpse-list-title-padding;
     text-align: $richie-blogpost-glimpse-list-title-textalign;
   }
+
+  &__categories {
```
mbenadda: You have a small problem in there. `category-plugin-tag` which are displayed inside this div are spaced with a margin-left, except for the `first-of-type`.

With wrapping elements, lines 2 and later are off to the right. This could be solved by just applying the margin on the right?

src/richie/apps/demo/defaults.py
```diff
@@ -16,7 +16,8 @@
     "licences": 5,
```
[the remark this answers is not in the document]
sampaccoud: Good point! The thing is that in real life our news are tagged with categories specific to news and there are less than 10 of them so my changes to the demo are reflecting what we the reality is.

If some other users tag their news with many different categories, you are right that this could blow up but I would like to leave it up to them to decide what we should do in this case :laughing:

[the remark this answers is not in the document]
mbenadda: Ok then, I can get behind this reasoning!

src/frontend/scss/objects/_blogpost_glimpses.scss
sampaccoud: Nicely caught :muscle: Your solution was applied.

PR #926, opened 2019-12-10
🐳(tests) update docker-compose to run e2e tests

mbenadda:
## Purpose

Update docker-compose, using cypress' own docker image, to start the development server and run end-to-end tests.

We also added a Makefile command to do the same.

NB: as noted in the docs

> This will launch the tests using `cypress run`. We do not provide a way to use `cypress open` through docker as this requires dependencies on the host. In our opinion users who need `cypress open` are better off just using `cypress` natively.


sampaccoud:
@mbenadda On the contrary, I think the fact that dependencies are required is a use case for a docker container so that we don't have to install them on our systems...

mbenadda:
Did you read the link in the docs? It specifically needs dependencies on the host to be able to run `cypress open` in a docker-container. (and those are different based on the OS eg. OS X will not need the same as Linux).

IMO it's useless, might as well install `cypress` on the host in this case.

PR #930, opened 2019-12-16
🐛 Fix an issue with get_picture_info when there is no image

mbenadda:
## Purpose

Our simple plugin's `get_picture_info()` helper operated on the assumption that an Picture has a picture attribute, with a FileImage.

Depending on how actual users use the back-office, it is possible to create Picture instances that are missing this. This caused the helper to raise an error and crash the ES index generation.

## Proposal

We added a test to reproduce this issue and solved it by simply bailing out at the top if there is no image as `get_picture_info()` has pretty much nothing to return then.

### DIFF
--- a/src/richie/apps/search/indexers/courses.py
+++ b/src/richie/apps/search/indexers/courses.py
@@ -376,7 +376,9 @@ class CoursesIndexer:
         ):
             language = cover.cmsplugin_ptr.language
             with translation.override(language):
-                cover_images[language] = get_picture_info(cover, "cover")
+                picture_info = get_picture_info(cover, "cover")
+                if picture_info:
+                    cover_images[language] = picture_info
 
         # Prepare the related category icon
         icon_images = {}
@@ -393,8 +395,9 @@ class CoursesIndexer:
                 cmsplugin_ptr__position=0,
             ):
                 with translation.override(language):
+                    picture_info = get_picture_info(icon, "icon") or {}
                     icon_images[language] = {
-                        **get_picture_info(icon, "icon"),
+                        **picture_info,
                         "color": plugin_model.page.category.color,
                         "title": plugin_model.page.get_title(),
                     }
--- a/tests/apps/search/test_indexers_courses.py
+++ b/tests/apps/search/test_indexers_courses.py
@@ -9,6 +9,7 @@ from django.test import TestCase
 import pytz
 from cms.api import add_plugin
 
+from richie.apps.courses.cms_plugins import CategoryPlugin
 from richie.apps.courses.factories import (
     CategoryFactory,
     CourseFactory,
@@ -20,6 +21,7 @@ from richie.apps.courses.models import CourseState
 from richie.apps.search.indexers.categories import CategoriesIndexer
 from richie.apps.search.indexers.courses import CoursesIndexer
 from richie.apps.search.indexers.organizations import OrganizationsIndexer
+from richie.plugins.simple_picture.cms_plugins import SimplePicturePlugin
 
 
 class CoursesIndexersTestCase(TestCase):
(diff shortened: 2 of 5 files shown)


tests/apps/search/test_indexers_courses.py
```diff
@@ -356,6 +358,70 @@ def test_indexers_courses_get_es_documents_no_end_no_enrollment_end(self):
...
+        course = CourseFactory(should_publish=True)
+        course_icons_placeholder = (
+            course.extended_object.get_public_object()
+            .placeholders.filter(slot="course_icons")
+            .first()
+        )
+        # And a picture with no image to make sure the indexer does not break
+        cover_picture = PictureFactory(picture=None)
+        # Make sure we associate the picture with the course through the icons placeholder
+        cmsplugin_ptr = CMSPlugin()
```
mbenadda: @sampaccoud this is the test I'm having trouble with. I need both a `Picture` and a `CategoryPluginModel ` instances with the same `CMSPlugin` on them (or at least something that will match on the queries for icons on the indexer).

As you can see above I managed it with the cover image query but there was no `CategoryPluginModel` there.

As a small aside, I'm also wondering how in one of the icon queries the placeholder slot is named `icon` and in the other it's `course_icons. Maybe I missed something there.

src/richie/apps/search/indexers/courses.py
```diff
@@ -393,11 +395,13 @@ def get_es_document_for_course(cls, course, index=None, action="index"):
...
-                        **get_picture_info(icon, "icon"),
-                        "color": plugin_model.page.category.color,
-                        "title": plugin_model.page.get_title(),
-                    }
+                    picture_info = get_picture_info(icon, "icon")
+                    if picture_info:
+                        icon_images[language] = {
+                            **picture_info,
+                            "color": plugin_model.page.category.color,
+                            "title": plugin_model.page.get_title(),
```
sampaccoud: be careful you should still set the color and title even if the picture is not there!

tests/apps/search/test_indexers_courses.py
```diff
@@ -356,6 +358,73 @@ def test_indexers_courses_get_es_documents_no_end_no_enrollment_end(self):
             indexed_courses[0]["course_runs"][0]["enrollment_end"].year, 9999
         )
 
+    def test_indexers_courses_get_es_document_no_image_cover_picture(self):
+        """
+        ES document is created withot errors when a cover image for the course is
```
sampaccoud:
```suggestion
        ES document is created without errors when a cover image for the course is
```

tests/apps/search/test_indexers_courses.py
```diff
@@ -356,6 +358,73 @@ def test_indexers_courses_get_es_documents_no_end_no_enrollment_end(self):
...
+        self.assertEqual(len(indexed_courses), 1)
+        self.assertEqual(
+            indexed_courses[0]["_id"],
+            str(course.extended_object.get_public_object().id),
+        )
+        self.assertEqual(indexed_courses[0]["cover_image"], {})
+
+    def test_indexers_courses_get_es_document_no_image_icon_picture(self):
+        """
+        ES document is created withot errors when a icon image for the course is
```
sampaccoud:
```suggestion
        ES document is created without errors when an icon image for the course is
```

tests/apps/search/test_indexers_courses.py
```diff
@@ -356,6 +358,73 @@ def test_indexers_courses_get_es_documents_no_end_no_enrollment_end(self):
...
+        indexed_courses = list(
+            CoursesIndexer.get_es_documents(index="some_index", action="some_action")
+        )
+
+        self.assertEqual(len(indexed_courses), 1)
+        self.assertEqual(
+            indexed_courses[0]["_id"],
+            str(course.extended_object.get_public_object().id),
+        )
+        self.assertEqual(indexed_courses[0]["cover_image"], {})
```
sampaccoud: copy/pasta fail as the idea was to test the icon, not the cover image.

PR #932, opened 2019-12-19
✨ Add category info on courses

mbenadda:
## Purpose

In order to implement more complex/full-featured designs on the frontend, notably for course glimpses, we want to have more category information available on courses objects as returned by the Search API.

This includes the placeholder slot that links said category to the course, information on the category itself as well as its meta and parent.

## Proposal

- [x] update the `Course` model's `get_categories` to annotate returned categories with placeholder slot data
- [x] index more category data with courses (new `categories_data` field)
- [x] return said additional category data on the API (WIP)
- [ ] use `category_data` in the frontend
- [ ] remove now redundant `icon` field on courses


mbenadda:
I think this is ready for review.

Also, do you know what the best practice is regarding different behavior of queries between Postgres & mysql in tests? Should try and write my test around it so it works anyway? Use a flag and make different assertions based on the database?

src/richie/apps/courses/models/course.py
```diff
@@ -310,6 +310,10 @@ def get_categories(self, language=None):
         return (
             Category.objects.filter(**filter_dict)
             .select_related("extended_object")
+            .annotate(
+                placeholder_slot=F(f"{selector:s}__placeholder__slot"),
+                placeholder_position=F(f"{selector:s}__position"),
```
sampaccoud: can you take the opportunity to use f strings as well in the above code?

src/richie/apps/search/indexers/courses.py
```diff
@@ -380,27 +381,70 @@ def get_es_document_for_course(cls, course, index=None, action="index"):
...
+                category_data["parent_title"] = {
+                    title["language"]: title["title"]
+                    for title in parent_category_page.title_set.values()
+                }
+
+            return category_data
+
+        # Generate categories data as an OrderedDict so we can deduplicate the categories by pk
+        # (use as keys) while keeping them in the same order
+        categories_data = OrderedDict()
```
sampaccoud: you don't need this. Dictionaries are ordered since python 3.7.

src/richie/apps/search/indexers/courses.py
```diff
@@ -380,27 +381,70 @@ def get_es_document_for_course(cls, course, index=None, action="index"):
...
+        for category in course.get_categories():
```
sampaccoud: I think we need the category data to be available on the course model as well so that we can do the same thing on the DjangoCMS glimpse as we do in the React glimpse.

And maybe it is interesting to be able to get a category's data from the category model too?

mbenadda: You're right about the course model.

For the category model, placeholder info is irrelevant there I think. As for the meta & parent info, it's not too hard to get from a category instance right now.

sampaccoud: seems legit

sampaccoud:
@mbenadda about your question on db specific tests, I would write the default test for postgresql and provide a specific test for other db...

mbenadda:
I solved this issue by providing a deterministic order for placeholder info with this https://github.com/openfun/richie/pull/932/files#diff-130a37291b0a06e9aee2f7bfd45f1460R321

IMO this is not going to be expensive since having 2 placeholders is already an edge case. Ordering between 2 items in some rare instances is probably acceptable to have a consistent behavior between databases?

PR #934, opened 2019-12-26
🥅 Add sentry to report frontend errors

### LINKED ISSUE #113: Add & configure raven.js
## Purpose

As mentioned in #112, we need to integrate and configure [Raven.js](https://github.com/getsentry/raven-js) to also catch front-end application errors [1, 2].

## Proposal

- [ ] load the `raven` template tag library in the base Django template
- [ ] define a `sentry_public_dsn` JavaScript variable in a `script` block from the Django base template (using `{% sentry_public_dsn %}`)
- [ ] add `raven.js` to front-end dependencies and build
- [ ] configure the client via `Raven.config(sentry_public_dsn).install()`
- [ ] wrap front-end app using `Raven.context`

## References

[1] https://docs.sentry.io/clients/python/integrations/django/#using-with-raven-js
[2] https://docs.sentry.io/clients/javascript/

mbenadda:
## Purpose

Errors emanating from Richie's frontend are being dropped right now as our generic error handler is just a wrapper for a `console.error` call. We want to catch them and get data on them.

## Proposal

Use Sentry to pick them up.

- [x] make release, environment & sentry_dsn available in `settings` (as we need to pass those to the Sentry sdk)
- [x] create a Django context processor with the frontend context that includes these and what previously existed as a frontend context
- [x] pass it to the frontend through `base.html` and the `window`
- [x] do report frontend errors through Sentry
- [x] separate errors between the backend & frontend using Sentry scope

Closes #113.


sandbox/settings.py
```diff
@@ -145,6 +141,7 @@ class Base(DRFMixin, RichieCoursesConfigurationMixin, Configuration):
                     "django.template.context_processors.static",
                     "cms.context_processors.cms_settings",
                     "richie.apps.core.context_processors.site_metas",
+                    "richie.apps.core.context_processors.frontend_context",
```
sampaccoud: This will require an entry in UPGRADE.md for users handling their own settings.

src/richie/apps/core/context_processors.py
```diff
@@ -57,3 +60,21 @@ def site_metas(request):
...
+    """
+    Context processor to add the context as expected by the frontend to the template context.
+    """
+    return {
+        "FRONTEND_CONTEXT": json.dumps(
+            {
+                "context": {
+                    "assets": {"icons": static("richie/icons.svg")},
+                    "environment": getattr(settings, "ENVIRONMENT", ""),
+                    "release": getattr(settings, "RELEASE", ""),
```
sampaccoud: Would it be possible to get the release from the frontend code? I'm thinking of someone that would customize the frontend and so run a different release number in the frontend and in the backend.

mbenadda: Done 👍

PR #936, opened 2020-01-03
🌐 Make licences translatable

mbenadda:
## Purpose

The `name` and `content` fields on the Licence model contain user-facing text. We need both of these to be translatable in the back-office and localized for end users.

The first step is to . We're then deprecating the existing, untranslated fields and
replacing them with `TranslatedFields` that interact with the new
translation model through parler.

## Proposal

- [x] deprecate existing `name` and `content` fields, keeping them around to avoid breaking blue-green deployments
- [x] introduce a `Django-parler` translation model for `Licence`
- [x] add translated `name` and `content` fields to `Licence`
- [x] add tests for internationalization of `Licence` model & admin


src/richie/apps/courses/migrations/0013_migrate_data_translated_licence_fields.py
```diff
@@ -0,0 +1,49 @@
...
+from django.conf import settings
+from django.core.exceptions import ObjectDoesNotExist
+from django.db import migrations
+
+
+def forwards_func(apps, schema_editor):
+    Licence = apps.get_model("courses", "Licence")
+    LicenceTranslation = apps.get_model("courses", "LicenceTranslation")
+
+    for object in Licence.objects.all():
```
sampaccoud: object is a reserved word in Python, so maybe call it:
```suggestion
    for licence in Licence.objects.all():
```

src/richie/apps/courses/migrations/0013_migrate_data_translated_licence_fields.py
```diff
@@ -0,0 +1,49 @@
...
+            language_code=settings.LANGUAGE_CODE,
+            name=object.name_deprecated,
+        )
+
+
+def backwards_func(apps, schema_editor):
+    Licence = apps.get_model("courses", "Licence")
+    LicenceTranslation = apps.get_model("courses", "LicenceTranslation")
+
+    for object in Licence.objects.all():
```
sampaccoud:
```suggestion
    for licence in Licence.objects.all():
```

src/richie/apps/courses/migrations/0013_migrate_data_translated_licence_fields.py
```diff
@@ -0,0 +1,49 @@
...
+        # Try default translation
+        return translations.get(language_code=settings.LANGUAGE_CODE)
+    except ObjectDoesNotExist:
+        try:
+            # Try default language
+            return translations.get(language_code=settings.PARLER_DEFAULT_LANGUAGE_CODE)
+        except ObjectDoesNotExist:
+            # Maybe the object was translated only in a specific language?
+            # Hope there is a single translation
+            return translations.get()
```
sampaccoud: Why not get the first one? 
```suggestion
            return translations.first()
```

src/richie/apps/courses/models/course.py
```diff
@@ -613,6 +620,19 @@ def __str__(self):
         )
 
 
+class LicenceTranslation(TranslatedFieldsModel):
+    master = models.ForeignKey(
+        Licence, models.SET_NULL, related_name="translations", null=True
```
sampaccoud: Is this recommended by Parler ?

I would have made this field required and deleted this record if the mater object is being deleted. WDYT?

tests/apps/courses/test_admin_licence.py
```diff
@@ -0,0 +1,82 @@
...
+
+        # Create a licence
+        licence = LicenceFactory()
+
+        # Get the admin list view
+        url = reverse("admin:courses_licence_changelist")
+        response = self.client.get(url, follow=True)
+
+        # Check that the page includes the name field
+        self.assertContains(response, licence.name, status_code=200)
```
sampaccoud: if you create a licence in several languages you can check that it diplays in the current language.

src/richie/apps/courses/models/course.py
mbenadda: Parler does not recommend anything here, in fact they omit this mandatory argument in their own docs.

What you're suggesting is the `models.CASCADE` strategy right? https://docs.djangoproject.com/en/3.0/ref/models/fields/#django.db.models.CASCADE

mbenadda:
All done @sampaccoud  ✨

src/richie/apps/courses/models/course.py
sampaccoud: yep

mbenadda: That’s what I did

src/richie/apps/courses/migrations/0013_migrate_data_translated_licence_fields.py
```diff
@@ -0,0 +1,49 @@
...
+    try:
+        # Try default translation
+        return translations.get(language_code=settings.LANGUAGE_CODE)
+    except ObjectDoesNotExist:
+        try:
+            # Try default language
+            return translations.get(language_code=settings.PARLER_DEFAULT_LANGUAGE_CODE)
+        except ObjectDoesNotExist:
+            # Maybe the object was translated only in a specific language?
+            # Hope there is a single translation
```
sampaccoud:
```suggestion
            # Take the first existing translation
```

src/richie/apps/courses/models/course.py
```diff
@@ -588,26 +589,57 @@ def __str__(self):
...
+
+class LicenceTranslation(TranslatedFieldsModel):
+    """
+    Licence Translation model.
+
+    Django parler model linked to the Licence to internationalize the fields.
+    """
+
+    master = models.ForeignKey(
+        Licence, models.CASCADE, related_name="translations", null=True
```
sampaccoud: Make it required:
```suggestion
        Licence, models.CASCADE, related_name="translations"
```

Don't forget to change it in the migration as well.

Issue #939, opened 2020-01-07
Invalid alias name error on bootstrap_elasticsearch


sampaccoud:
## Bug Report

**Expected behavior/code**
The `bootstrap_elasticsearch` command should always create a new index, build it and finally point to it with an alias.

**Actual Behavior**
In some cases, the `bootstrap_elasticsearch` command fails with the following stack trace:
```
Traceback (most recent call last):
  File "manage.py", line 14, in <module>
    execute_from_command_line(sys.argv)
  File "/usr/local/lib/python3.7/site-packages/django/core/management/__init__.py", line 381, in execute_from_command_line
    utility.execute()
  File "/usr/local/lib/python3.7/site-packages/django/core/management/__init__.py", line 375, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/usr/local/lib/python3.7/site-packages/django/core/management/base.py", line 323, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/usr/local/lib/python3.7/site-packages/django/core/management/base.py", line 364, in execute
    output = self.handle(*args, **options)
  File "/usr/local/lib/python3.7/site-packages/richie/apps/search/management/commands/bootstrap_elasticsearch.py", line 29, in handle
    regenerate_indexes(logger)
  File "/usr/local/lib/python3.7/site-packages/richie/apps/search/index_manager.py", line 120, in regenerate_indexes
    dict(actions=actions_to_create_aliases + actions_to_delete_aliases)
  File "/usr/local/lib/python3.7/site-packages/elasticsearch/client/utils.py", line 84, in _wrapped
    return func(*args, params=params, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/elasticsearch/client/indices.py", line 490, in update_aliases
    "POST", "/_aliases", params=params, body=body
  File "/usr/local/lib/python3.7/site-packages/elasticsearch/transport.py", line 353, in perform_request
    timeout=timeout,
  File "/usr/local/lib/python3.7/site-packages/elasticsearch/connection/http_urllib3.py", line 236, in perform_request
    self._raise_error(response.status, raw_data)
  File "/usr/local/lib/python3.7/site-packages/elasticsearch/connection/base.py", line 162, in _raise_error
    status_code, error_message, additional_info
elasticsearch.exceptions.RequestError: RequestError(400, 'invalid_alias_name_exception', 'Invalid alias name [richie_courses], an index exists with the same name as the alias')
```

[...]

PR #941, opened 2020-01-08
♻️ Improve ES index management command and tests

mbenadda:
## Purpose

While trying to solve #939, we looked at the `index_manager` tests.

We noticed they were using a lot of mocks such that they provided very little confidence in the code actually working as intended.

## Proposal

We replaced two heavily mocked tests with an entirely non-mocked one that should cover the same code surface with more confidence.

We also touched up on the code in the index_manager module, eg. fixing a small bug we encountered and replacing format calls with f strings.

This did not find us the bug as we had hoped it might, but it's an improvement nonetheless.


mbenadda:
Replaced by #945 which actually fixes the bug.

PR #943, opened 2020-01-09
🔧 Variabilize ES index names through settings

### LINKED ISSUE #940: Allow running several Richie websites on the same Elasticsearch cluster
## Feature Request

We would like to allow hosting several Richie websites on the same Elasticsearch cluster. It is not possible at the moment because the name of the search index and of the index alias are not configurable and will clash.

Add a setting to allow configuring the name of the search index and of the index alias.
This setting should default to what is currently used as names.

mbenadda:
## Purpose

ElasticSearch index names are hardcoded in their respective indexers.

There is a `richie_` prefix that should prevent conflicts with other applications running on the same ElasticSearch instance, but that still prevents more than one instance of Richie to run on the same ElasticSearch, as indices would share names.

## Proposal

We can make this changeable through the settings and an environment variable, using `richie_`as a default so it's not a breaking change for existing users.

Closes #940.

Note: rebased on #941 to avoid pointless conflicts as it affects the same files.

### DIFF
--- a/sandbox/settings.py
+++ b/sandbox/settings.py
@@ -296,6 +296,9 @@ class Base(DRFMixin, RichieCoursesConfigurationMixin, Configuration):
     RICHIE_ES_HOST = values.Value(
         "elasticsearch", environ_name="RICHIE_ES_HOST", environ_prefix=None
     )
+    RICHIE_ES_INDICES_PREFIX = values.Value(
+        default="richie", environ_name="RICHIE_ES_INDICES_PREFIX", environ_prefix=None
+    )
 
     # Sentry
     SENTRY_DSN = values.Value(None, environ_name="SENTRY_DSN")
--- a/src/richie/apps/search/defaults.py
+++ b/src/richie/apps/search/defaults.py
@@ -2,11 +2,15 @@
 Import custom settings and set up defaults for values the Search app needs
 """
 from django.conf import settings
+from django.utils.functional import lazy
 from django.utils.translation import ugettext_lazy as _
 
 # Elasticsearch
 ES_CHUNK_SIZE = 500
 ES_PAGE_SIZE = 10
+
+# Use a lazy to enable easier testing by not defining the value at bootstrap time
+ES_INDICES_PREFIX = lazy(lambda: settings.RICHIE_ES_INDICES_PREFIX)()
 
 # Define the scoring boost (in ElasticSearch) related value names receive when using
 # full-text search.
--- a/src/richie/apps/search/index_manager.py
+++ b/src/richie/apps/search/index_manager.py
@@ -11,7 +11,7 @@ from elasticsearch.exceptions import NotFoundError, RequestError
 from elasticsearch.helpers import bulk
 
 from . import ES_CLIENT
-from .defaults import ES_CHUNK_SIZE
+from .defaults import ES_CHUNK_SIZE, ES_INDICES_PREFIX
 from .indexers import ES_INDICES
 from .text_indexing import ANALYSIS_SETTINGS
 
@@ -107,11 +107,13 @@ def regenerate_indices(logger):
         for index, alias in indices_to_unalias
     ]
 
-    # Identify orphaned indices
+    # Identify orphaned indices that belong to our own app.
     # NB: we *must* do this before the update_aliases call so we don't immediately prune
     # version n-1 of all our indices
     useless_indices = [
-        index for index, details in existing_indices.items() if not details["aliases"]
+        index
+        for index, details in existing_indices.items()
+        if index.startswith(str(ES_INDICES_PREFIX)) and not details["aliases"]
     ]
 
     # Replace the old indices with the new ones in 1 atomic operation to avoid outage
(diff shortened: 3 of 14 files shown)


src/richie/apps/search/index_manager.py
```diff
@@ -112,15 +113,14 @@ def regenerate_indexes(logger):
     # NB: we *must* do this before the update_aliases call so we don't immediately prune
     # version n-1 of all our indexes
     useless_indexes = [
-        index for index, details in existing_indexes.items() if "aliases" not in details
+        index for index, details in existing_indexes.items() if details["aliases"] == {}
```
sampaccoud: so "aliases" is always in details right?

src/richie/apps/search/indexers/categories.py
```diff
@@ -26,7 +27,7 @@ class CategoriesIndexer:
     """
 
     document_type = "category"
-    index_name = "richie_categories"
+    index_name = f"{ES_INDICES_PREFIX}categories"
```
sampaccoud: May be nicer to write `f"{ES_INDICES_PREFIX}_categories"` with ES_INDICES_PREFIX = "richie".

src/richie/apps/search/index_manager.py
```diff
@@ -112,15 +113,14 @@ def regenerate_indexes(logger):
     # NB: we *must* do this before the update_aliases call so we don't immediately prune
     # version n-1 of all our indexes
     useless_indexes = [
-        index for index, details in existing_indexes.items() if "aliases" not in details
```
sampaccoud: you must only delete indexes that start with `ES_INDICES_PREFIX`.

mbenadda:
Note: this PR probably needs to handle this use case https://github.com/openfun/richie/pull/945#discussion_r366865601.

mbenadda:
Note: Not sure we have to account for the special case from #945 after all, as it seems impossible to detect from the code in `regenerate_indices`.

sandbox/settings.py
```diff
@@ -296,6 +296,9 @@ class Base(DRFMixin, RichieCoursesConfigurationMixin, Configuration):
     RICHIE_ES_HOST = values.Value(
         "elasticsearch", environ_name="RICHIE_ES_HOST", environ_prefix=None
     )
+    RICHIE_ES_INDICES_PREFIX = values.Value(
+        "richie", environ_name="RICHIE_ES_INDICES_PREFIX", environ_prefix=None
```
sampaccoud: You need to set a default of `None` if you want the code in `src/richie/apps/search/defaults.py ` to work:
```suggestion
        "richie", environ_name="RICHIE_ES_INDICES_PREFIX", environ_prefix=None
```

Maybe add a test to validate precedence on index name?

mbenadda: I'm not sure how to test this since it's something that runs at bootstrap time. Also your suggested change seems not to be working but I see what you wanted to suggest 👍

src/richie/apps/search/defaults.py
```diff
@@ -2,12 +2,24 @@
...
+    ES_DEFAULT_INDICES_PREFIX = (
+        Site.objects.get_current().domain.replace(".", "_").replace("-", "_")
+    )
+# pylint: disable=broad-except
+except Exception:
+    ES_DEFAULT_INDICES_PREFIX = "richie"
+ES_INDICES_PREFIX = getattr(
+    settings, "RICHIE_ES_INDICES_PREFIX", ES_DEFAULT_INDICES_PREFIX
+)
+
```
sampaccoud: Actually the tests are broken. Maybe what we want is:
```python
ES_INDICES_PREFIX = (
    settings.RICHIE_ES_INDICES_PREFIX or
    Site.objects.get_current().domain.replace(".", "_").replace("-", "_") or 
    "richie"
)
```

sandbox/settings.py
```diff
@@ -296,6 +296,9 @@ class Base(DRFMixin, RichieCoursesConfigurationMixin, Configuration):
     RICHIE_ES_HOST = values.Value(
         "elasticsearch", environ_name="RICHIE_ES_HOST", environ_prefix=None
     )
+    RICHIE_ES_INDICES_PREFIX = values.Value(
+        default=None, environ_name="RICHIE_ES_INDICES_PREFIX", environ_prefix=None
```
```suggestion
        default="richie", environ_name="RICHIE_ES_INDICES_PREFIX", environ_prefix=None
```

src/richie/apps/search/defaults.py
```diff
@@ -2,12 +2,16 @@
...
 from django.conf import settings
+from django.utils.functional import lazy
 from django.utils.translation import ugettext_lazy as _
 
 # Elasticsearch
 ES_CHUNK_SIZE = 500
 ES_PAGE_SIZE = 10
 
+# Use a lazy to enable easier testing by not defining the value at bootstrap time
+ES_INDICES_PREFIX = lazy(lambda: settings.RICHIE_ES_INDICES_PREFIX or "richie")()
```
sampaccoud:
```suggestion
ES_INDICES_PREFIX = lazy(lambda: settings.RICHIE_ES_INDICES_PREFIX)()
```

PR #945, opened 2020-01-13
🐛 Fix invalid alias name bug during `regenerate_indexes`

### LINKED ISSUE #939: Invalid alias name error on bootstrap_elasticsearch

mbenadda:
## Purpose

Fixes #939.

In some instances, Richie can end up with an ElasticSearch that contains broken indices, with each index's expected alias as an index name. This can happen for example if the signals that are tasked with updating courses, organizations, etc. are happening in a state where the `bootstrap_elasticsearch` command was not run.

ES then implicitly creates the index for the object we're trying to update.

This caused `bootstrap_elasticsearch` to fail when attempting to create an alias with the same name as an existing index.

## Proposal

We replaced two heavily mocked tests with an entirely non-mocked one that should cover the same code surface with more confidence.

We also touched up on the code in the index_manager module, eg. fixing a small bug we encountered and replacing format calls with f strings.

We then created a test that reproduces the bug in #939. Since the existing index with a problematic name, is broken by nature (it was created with no content, no mappings, no scripts etc.) and a general indices regeneration was triggered, we decided to just delete the broken index and move on with the operation in this case.

### DIFF
--- a/src/richie/apps/search/index_manager.py
+++ b/src/richie/apps/search/index_manager.py
@@ -36,18 +36,17 @@ def get_indexes_by_alias(existing_indexes, alias):
             yield index, alias
 
 
-def perform_create_index(indexable, logger):
+def perform_create_index(indexable, logger=None):
     """
     Create a new index in ElasticSearch from an indexable instance
     """
     indices_client = IndicesClient(client=ES_CLIENT)
     # Create a new index name, suffixing its name with a timestamp
-    new_index = "{:s}_{:s}".format(
-        indexable.index_name, timezone.now().strftime("%Y-%m-%d-%Hh%Mm%S.%fs")
-    )
+    new_index = f"{indexable.index_name:s}_{timezone.now():%Y-%m-%d-%Hh%Mm%S.%fs}"
 
     # Create the new index
-    logger.info('Creating a new Elasticsearch index "{:s}"...'.format(new_index))
+    if logger:
+        logger.info(f'Creating a new Elasticsearch index "{new_index:s}"...')
     indices_client.create(index=new_index)
 
     # The index needs to be closed before we set an analyzer
@@ -112,7 +111,7 @@ def regenerate_indexes(logger):
     # NB: we *must* do this before the update_aliases call so we don't immediately prune
     # version n-1 of all our indexes
     useless_indexes = [
-        index for index, details in existing_indexes.items() if "aliases" not in details
+        index for index, details in existing_indexes.items() if not details["aliases"]
     ]
 
     # Replace the old indexes with the new ones in 1 atomic operation to avoid outage
@@ -120,7 +119,6 @@ def regenerate_indexes(logger):
         dict(actions=actions_to_create_aliases + actions_to_delete_aliases)
     )
 
-    # Cleanup step: do prune older indexes that are now useless
     for useless_index in useless_indexes:
         # Disable keyword arguments checking as elasticsearch-py uses a decorator to list
         # valid query parameters and inject them as kwargs. I won't say a word about this
@@ -130,7 +128,7 @@ def regenerate_indexes(logger):
         indices_client.delete(index=useless_index, ignore=[400, 404])
 
 
-def store_es_scripts(logger):
+def store_es_scripts(logger=None):
     """
     Iterate over the indexers listed in the settings, import them, and store the scripts
     they define on their "scripts" key in ElasticSearch
@@ -138,5 +136,5 @@ def store_es_scripts(logger):
     for indexer in ES_INDICES:
         for script_id, script_body in indexer.scripts.items():
             if logger:
-                logger.info('Storing script "{:s}"...'.format(script_id))
+                logger.info(f'Storing script "{script_id:s}"...')
             ES_CLIENT.put_script(id=script_id, body=script_body)
--- a/src/richie/apps/search/index_manager.py
+++ b/src/richie/apps/search/index_manager.py
@@ -1,5 +1,5 @@
 """
-ElasticSearch indexes utilities.
+ElasticSearch indices utilities.
 """
 from functools import reduce
 
@@ -26,12 +26,12 @@ def richie_bulk(actions):
     )
 
 
-def get_indexes_by_alias(existing_indexes, alias):
+def get_indices_by_alias(existing_indices, alias):
     """
     Get existing index(es) for an alias. Support multiple existing aliases so the command
     always 'cleans up' the aliases even if they got messed up somehow.
     """
-    for index, details in existing_indexes.items():
+    for index, details in existing_indices.items():
         if alias in details.get("aliases", {}):
             yield index, alias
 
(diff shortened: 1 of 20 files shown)


src/richie/apps/search/index_manager.py
```diff
@@ -36,18 +36,19 @@ def get_indexes_by_alias(existing_indexes, alias):
...
+def perform_create_index(indexable, logger=None):
     """
     Create a new index in ElasticSearch from an indexable instance
     """
     indices_client = IndicesClient(client=ES_CLIENT)
     # Create a new index name, suffixing its name with a timestamp
-    new_index = "{:s}_{:s}".format(
-        indexable.index_name, timezone.now().strftime("%Y-%m-%d-%Hh%Mm%S.%fs")
+    new_index = (
+        f"{indexable.index_name:s}_{timezone.now().strftime('%Y-%m-%d-%Hh%Mm%S.%fs'):s}"
```
sampaccoud:
```suggestion
        f"{indexable.index_name:s}_{timezone.now():%Y-%m-%d-%Hh%Mm%S.%fs'}"
```

src/richie/apps/search/index_manager.py
```diff
@@ -112,15 +113,34 @@ def regenerate_indexes(logger):
     # NB: we *must* do this before the update_aliases call so we don't immediately prune
     # version n-1 of all our indexes
     useless_indexes = [
-        index for index, details in existing_indexes.items() if "aliases" not in details
+        index for index, details in existing_indexes.items() if details["aliases"] == {}
```
sampaccoud: Maybe more robust and nicer ({} creates an empty object for an akward comparison?):
```suggestion
        index for index, details in existing_indexes.items() if not details.get("aliases")
```

src/richie/apps/search/index_manager.py
```diff
@@ -112,15 +113,34 @@ def regenerate_indexes(logger):
...
-    indices_client.update_aliases(
-        dict(actions=actions_to_create_aliases + actions_to_delete_aliases)
-    )
+    def perform_aliases_update():
+        try:
+            indices_client.update_aliases(
+                dict(actions=actions_to_create_aliases + actions_to_delete_aliases)
+            )
+        except RequestError as exception:
+            # This operation can fail if an index exists with the same name as an alias we're
```
sampaccoud: I thought of a case which might be problematic. 

Let's say you have a Richie site running with the default index name and  you add a new site to the same ES cluster. If you don't customize the index name, the aliases will be in conflict and this code will attempt to drop the index of the existing site without warning... Not sure what happens in this case as it is an alias and not an index but we must check and maybe add a test? I might also be missing something...

src/richie/apps/search/index_manager.py
mbenadda: This will not work as `details["aliases"]` is always defined. When the index is "orphaned" it is an empty object

src/richie/apps/search/index_manager.py
mbenadda: Not sure how we would have a conflict here, as we'd have eg `richie_courses` from the first default install and then `prefixed_richie_courses` from the other one

mbenadda: (Note that this PR does not include the changes related to index name variabilization)

src/richie/apps/search/index_manager.py
sampaccoud: it will work because {} verifies `not details["aliases"]`.
if it is always defined, then:
```suggestion
        index for index, details in existing_indexes.items() if not details["aliases"]
```

src/richie/apps/search/index_manager.py
sampaccoud: sorry I forgot to mention that in my scenario, you would start the second project with the default index name as well (because you didn't know about this setting). You first index would at risk of this kind of negligence.

mbenadda: Yes, it would be broken to try and run two projects without prefixing. However, it is currently broken and this PR does not change that.

Maybe there's a way to mitigate this, IMO it should then be addressed in the next PR that deals with variabilization of index names?

src/richie/apps/search/index_manager.py
mbenadda: Thanks, this works great! Done 👍

tests/apps/search/test_index_manager.py
```diff
@@ -155,165 +158,129 @@ def get_es_documents(self, index, action="index"):
...
-        )
-        self.indices_client.delete.assert_called_with(
-            ignore=[400, 404], index="richie_orphan"
-        )
+        # Create all our indices from scratch
+        # Use a mocked timezone.now to  check the names of our indices as they include a datetime
+        creation1_datetime = datetime(2010, 1, 1, tzinfo=timezone.utc)
+        creation1_string = creation1_datetime.strftime("%Y-%m-%d-%Hh%Mm%S.%fs")
+        with mock.patch.object(timezone, "now", return_value=creation1_datetime):
+            regenerate_indexes(None)
```
sampaccoud: You should probably search and replace `indexes` for `indices` on the whole code base! ^^

tests/apps/search/test_index_manager.py
```diff
@@ -155,165 +158,129 @@ def get_es_documents(self, index, action="index"):
...
-                            "alias": "richie_persons",
-                        }
-                    },
-                ]
-            }
+        # Create a course and trigger a signal to index it. This will create a
+        # broken "richie_courses" index
+        course = CourseFactory(should_publish=True)
+        update_course(course.extended_object, "en")
+
```
sampaccoud: I think you should add a check that the index was actually created.

Issue #946, opened 2020-01-13
Limit the number of courses displayed on a category page


sampaccoud:
## Bug Report

**Expected behavior/code**
Computing a category page should always be fast
The number of related courses displayed should not exceed what is Ux friendly.

**Actual Behavior**
Categories display courses related to all their descendants. So for the top categories, it tries to retrieve... all the courses! On a big catalog, this is too much for a good Ux and even fails to compute with a timeout if the number of courses is really high.

**Steps to Reproduce**
1. Generate a demo site with a big number of courses: https://github.com/openfun/richie/blob/master/src/richie/apps/demo/defaults.py#L7
2. Visit the page of one of the top categories

**Environment**
- Richie version: 1.6.1
- Platform: Docker

**Possible Solution**

- [ ] Limit the number of related courses retrieved (Maybe add a [slice template filter](https://docs.djangoproject.com/en/dev/ref/templates/builtins/#slice) in the [category detail template](https://github.com/openfun/richie/blob/fb0725cc3f9d78f9e3fdc192c7170b3060cfb08a/src/richie/apps/courses/templates/courses/cms/category_detail.html#L86)?)
- [ ] Add a link "see more courses" that redirects to the search page with the category filter activated.
- [ ] Do the same for:
    * Related organizations on the detail category page
    * Related courses on the detail organization page
    * Etc.

PR #948, opened 2020-01-17
⚡️ Limit related courses on other pages

mbenadda:
## Purpose

Per #946.

Richie currently has scaling problems where large numbers of related items automatically loaded on some pages can cause them to be very slow. Eg. a category page on a Richie instance with ~500 courses can take up to 30 seconds to load.

This is unacceptable. Users probably don't even want to see hundred-items-long lists of courses or other objects on a related page.

## Proposal

- [x] Add settings/defaults/a context manager to limit the lengths of automatic lists of related objects
- [x] Create a filter to slice queries for related items through this new setting+context manager
- [x] Move `get_es_id` from categories indexer to category model so we can more easily build search URLs that filter by category
- [x] Ditto for organizations `get_es_id`
- [x] Limit the number of related courses on category pages
- [x] ... on organization pages
- [x] ... on person pages

### DIFF
--- a/src/richie/apps/courses/templates/courses/cms/category_detail.html
+++ b/src/richie/apps/courses/templates/courses/cms/category_detail.html
@@ -1,5 +1,5 @@
 {% extends "richie/fullwidth.html" %}
-{% load cms_tags extra_tags i18n thumbnail %}
+{% load cms_tags extra_tags first_n i18n thumbnail %}
 
 {% block meta_opengraph_contextuals %}
     <meta property="og:title" content="{{ current_page.get_title|truncatechars:65 }}">
@@ -83,7 +83,7 @@
   {% endif %}
 {% endwith %}
 
-{% with courses=category.get_courses %}
+{% with courses=category.get_courses|first_n:PAGES_SETTINGS.MAX_RELATED_COURSES_ON_DETAILS_PAGES %}
   {% if courses %}
     <section class="course-glimpse-list">
       <h2 class="course-glimpse-list__title">{% trans "Related courses" %}</h2>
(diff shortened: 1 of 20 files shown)


src/richie/apps/courses/templates/courses/cms/category_detail.html
```diff
@@ -92,6 +92,17 @@ <h2 class="course-glimpse-list__title">{% trans "Related courses" %}</h2>
           {% include "courses/cms/fragment_course_glimpse.html" %}
         {% endif %}
       {% endfor %}
+      {% if courses.count == PAGES_SETTINGS.MAX_RELATED_COURSES_ON_DETAILS_PAGES %}
+        {% if category.get_meta_category %}
```
sampaccoud: why not show the search page filtered on this category? It seems to be what you naturally want to see and not all sibling categories.

src/richie/apps/courses/templatetags/first_n.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+@register.filter("first_n")
+def first_n_filter(value, arg):
+    """
+    Picks the first N items in a list.
+    The reason this exists over stock "slice" is to allow using a variable to define
+    the number of items to pick.
+    """
+    return value[slice(0, arg)]
```
sampaccoud: I find it very complicated to add a context manager just for this (remember that all projects using Richie will have to add our context managers to their settings for it to work... so we should avoid adding new ones).
Maybe considering the setting as a default and using it directly here can make all this easier. If our use cases get more complex later we will think of another way but for now it seems good enough?
Something like:
```suggestion
@register.filter("limit_related")
def limit_related(value, arg=None):
    arg = arg or getattr(settings, "RICHIE_LIMIT_RELATED_OBJECTS", 20)
    return value[slice(0, arg)]
```

src/richie/apps/courses/templates/courses/cms/category_detail.html
mbenadda: I'm not sure what you're suggesting here. It seems to me the link is precisely pointing to `the search page filtered on this category`

sampaccoud: ok! I just realized why you need to get the id of the meta to do that :-/

mbenadda:
This is handled better by #992.

PR #952, opened 2020-01-22
♻️(backend) refactor PersonFactory to use PageFactory

mbenadda:
## Purpose

Our object factories in Richie, used for tests and to generate the demo site, use calls to `create_page` to create related pages (`extended_object`) for all objects.

Since then, we built a `PageFactory` and `TitleFactory` that allows us to create related CMS pages more cleanly and more explicitly.

This is the beginning of our work on refactoring all our factories away from those direct `create_page` calls and towards `PageFactory` use.

## Proposal

- [x] Create a new base class to replace `PageExtensionDjangoModelFactory` using `PageFactory` as a `SubFactory` instead of an ad-hoc call to `create_i18n_page`
- [x] Refactor `PersonFactory` to be based on this new version;
- [ ] ditto for `OrganizationFactory`...
- [ ] `CategoryFactory`...
- [ ] `BlogPostFactory`...
- [ ] `ProgramFactory`...
- [ ] `CourseFactory`...
- [ ] `CourseRunFactory`.

- [] item 1...
- [] item 2...


src/richie/apps/demo/management/commands/create_demo_site.py
```diff
@@ -197,9 +197,9 @@ def create_demo_site():
             random.randint(1, NB_OBJECTS["person_organizations"]),  # nosec
         )
         person = PersonFactory(
-            page_in_navigation=True,
-            page_languages=["en", "fr"],
-            page_parent=pages_created["persons"],
+            extended_object__in_navigation=True,
+            extended_object__parent=pages_created["persons"],
+            extended_object__title__title="en",
```
sampaccoud:
```suggestion
            extended_object__title__language="en",
```

tests/apps/courses/test_factories_person.py
```diff
@@ -18,7 +18,7 @@ def test_factories_person_portrait(self):
         PersonFactory should be able to generate plugins with a realistic portrait
         for several languages.
         """
-        person = PersonFactory(page_languages=["fr", "en"], fill_portrait=True)
+        person = PersonFactory(extended_object__languages="fr,en", fill_portrait=True)
```
sampaccoud: I think this creates an akward page because its `languages` field is not coherent with its related title objects. Maybe not a big deal if the test passes...

tests/apps/search/test_indexers_persons.py
```diff
@@ -26,14 +27,30 @@ def test_indexers_persons_get_es_documents(self, _mock_picture):
...
+            extended_object__title__language="en",
+            extended_object__title__title="my first person",
             fill_portrait=True,
-            page_title={"en": "my first person", "fr": "ma première personne"},
             should_publish=True,
         )
+        TitleFactory(
+            page=person1.extended_object, language="fr", title="ma première personne"
+        )
+        add_plugin(
```
sampaccoud: I think it will be easier to understand the code in this case if we don't use `fill_portrait` in the first place.

tests/apps/courses/test_factories_person.py
mbenadda: Actually it's a relevant question. If we use the correct way to build out the person & page, there will be no portrait for the second language (unless added manually).

I guess I could refactor these tests to reflect the fact that we don't expect the factory to generate multilingual portraits?

PR #960, opened 2020-02-28
✨(blogposts) add method on blogpost model to retrieve related blogposts

sampaccoud:
## Purpose

We want to display, in the side column of the blogpost detail page, a list of related blogposts. 

## Proposal

I propose to add a method to retrieve the list of blogposts that share at least one category with the current blogpost.

### DIFF
--- a/src/richie/apps/courses/models/blog.py
+++ b/src/richie/apps/courses/models/blog.py
@@ -2,6 +2,7 @@
 Declare and configure the models for the blog part
 """
 from django.db import models
+from django.utils import translation
 from django.utils.translation import ugettext_lazy as _
 
 from cms.api import Page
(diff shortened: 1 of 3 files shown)


src/richie/apps/courses/models/blog.py
```diff
@@ -30,6 +31,52 @@ def __str__(self):
...
+        return (
+            BlogPost.objects.filter(
+                extended_object__publisher_is_draft=is_draft,
+                extended_object__placeholders__cmsplugin__language=language,
+                **selector,
+            )
+            .exclude(id=self.id)
+            .select_related("extended_object")
+            .distinct()
+            .order_by("extended_object__node__path")
```
mbenadda: It's just a suggestion here, but should we generally order blog posts by date instead of node path?

tests/apps/courses/test_models_blogpost.py
```diff
@@ -22,3 +23,57 @@ def test_models_blogpost_str(self):
...
         with self.assertNumQueries(1):
             self.assertEqual(str(blogpost), "Blog Post: My first article")
+
+    @staticmethod
+    def _attach_category(blogpost, category):
+        """Not a test. Utility method to easily attach a category to a blogpost."""
+        placeholder = blogpost.extended_object.placeholders.get(slot="categories")
+        add_plugin(placeholder, "CategoryPlugin", "en", page=category.extended_object)
+
+    def test_models_blogpost_get_related_blogposts_plugin_on_draft(self):
```
mbenadda: It's not `on_draft` since you're also testing public blogposts?

src/richie/apps/courses/models/blog.py
sampaccoud: On the "blogpost" main page they are sorted by publication date... but here I thought sorting by path would allow changing the order to convey importance.
Anyway, if you don't voluntarily change the order, the blogposts will be sorted by date because they are created when you want to publish them...

tests/apps/courses/test_models_blogpost.py
sampaccoud: Good catch. I intially had 2 tests ;-)

src/richie/apps/courses/models/blog.py
mbenadda: But you would have to reorder them in the tree when that does not affect their order on the blogpost page. That does not seem very intuitive to me.

> Anyway, if you don't voluntarily change the order, the blogposts will be sorted by date because they are created when you want to publish them...

Maybe I'm just nitpicking then. This is ok 😅

sampaccoud: We could decide to order the blogpost page by importance as well... This is a very small change in the template. We could make it a setting if it proves necessary. We will see what the users say.

PR #963, opened 2020-03-03
Fix ghost placeholders

sampaccoud:
## Purpose

Some ghost placeholders appear in the side toolbar on pages that make use of our template tags pulling content from a placeholder on another page.

## Proposal

We thought we could use the same placeholder template tags either on the current page as editable placeholders or on another page to only display content from another related page. It worked but created some ghost placeholders (placeholders from the other page appeared in the side bar of the current page and were luckily not functional).

This happens because DjangoCMS creates a placeholder in the side toolbar for any template tag in the page that derives from their `Placeholder` template tag.

We fix it here by using different placeholders for each of these 2 cases.

This is a heavy refactoring but with no impact on existing sites.


2018D:
I'm satisfied with the review.

Issue #415, opened 2018-11-05
Add a link on the course run page to easily go back to the related course page


sampaccoud:
## Feature Request

https://github.com/openfun/richie/pull/410 is adding a specific page for each course run. We need to add a link to easily go back to the related course.

**Describe the solution you'd like**
I was thinking to add this link as a breadcrumb on top of the page content zone.

**Describe alternatives you've considered**
Add a simple button?

**Discovery, Documentation, Adoption, Migration Strategy**
This breadcrumb would be available on the search page, the course page and the course run page. The snapshot page which can exist between the course page and the course run page should not materialize as a step in the breadcrumb (it's content is already displayed on the course run page):
> search > my course tile > course run 1

sampaccoud:
@2018D do we consider that the breadcrumb has solved this issue?

2018D:
Yes it has.

PR #964, opened 2020-03-03
💄(layout) unique breadcrumb layout for course run page

2018D:
## Purpose

Is an answer to issue #415 where we need a way to return to the course-run-detail page from the course run page including from snapshoted course run.

## Proposal

create a special breadcrumb layout for course run pages to
take into account snapshots.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 ### Changed
 
 - Improve ElasticSearch `regenerate_indexes` tests.
+- Improve breadcrumb on course run page by creating a specific version.
 
 ## [1.16.2] - 2019-12-18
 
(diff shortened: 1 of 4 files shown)


src/richie/apps/courses/templates/courses/cms/fragment_course_breadcrumb.html
```diff
@@ -0,0 +1,20 @@
+{% if current_page.parent_page.parent_page.course%}
```
sampaccoud: you already have a change to fork when including the template that is more natural and doesn't involve any query so it would be easier to create a separate template `fragment_course_run _breadcrumb.html`?

2018D: Right 👍 I'll refactor in a few minutes.

2018D: New version is up.

sampaccoud:
@2018D I wanted to add tests to your PR and realized that both cases are for course runs. I misunderstood while reviewing because of the name of the template. 
So I reverted to your earlier implementation, changed the name of the template and added tests to verify and secure the breadcrumb.

Can you check it out?

2018D:
I'm happy with the way it turned out. Thank you for making the tests.

PR #971, opened 2020-03-13
✨(glimpse) add new plugin "glimpse"

sveetch:
## Purpose

Users need a way to include some content almost everywhere and
especially in section items with some layout variant they can choose.

## Proposal

This new plugin will have optional title and image and a required
content text.

Every variants will be managed within a single template.

## Todo parts

- [x] Model + migration
- [x] Plugin
- [x] Factory + tests
- [x] Template
- [x] Sass source for layout integration
- [x] Screenshots for layout sample
- [x] Update Changelog


sveetch:
@sampaccoud Are you sure about raw text only in content ? There is actually not any way to add a link in this glimpse, is it ok ?

sampaccoud:
@sveetch I confirm raw text. The way to add a link should be an optional field in the model.

sveetch:
## Screenshot samples

### Plugin form

glimpse_plugin-form

### Dummy background to demonstrate structure limits

glimpse_sketchs

### Clean screenshots for each responsive breakpoint

glimpse_clean_large

glimpse_clean__medium

glimpse_clean__small

sampaccoud:
@sveetch I would crop the image on the square card to the same format as our course glimpse... with a template tag and easy thumbnail to build a srcset.

src/richie/plugins/glimpse/cms_plugins.py
```diff
@@ -0,0 +1,56 @@
...
+
+        * If model instance has an empty "variant" attribute, use the "variant"
+          value from parent template context, default to None if not set;
+        * If model instance has a not null "variant" attribute use it, no
+          matter what the value from parent template context is;
+        """
+        if not instance.variant:
+            return context.get("variant", None)
+
+        return instance.variant
```
sampaccoud:
```suggestion
        return instance.variant or context.get("variant")
```

src/richie/plugins/glimpse/models.py
```diff
@@ -0,0 +1,86 @@
...
+from django.utils.translation import ugettext_lazy as _
+
+from cms.models.fields import PageField
+from cms.models.pluginmodel import CMSPlugin
+from filer.fields.image import FilerImageField
+
+from .defaults import GLIMPSE_VARIANTS
+
+
+@python_2_unicode_compatible
```
sampaccoud: We don't support Python 2 in richie anyway...
```suggestion
```

src/richie/plugins/glimpse/models.py
```diff
@@ -0,0 +1,86 @@
...
+    )
+
+    def __str__(self):
+        return Truncator(self.title).words(6, truncate="...")
+
+    def get_link(self):
+        """
+        Return attached link if any.
+
+        In case of "link_url" and "link_page" both not empty, "link_url" is
```
sampaccoud:
```suggestion
        In case the "link_url" and "link_page" fields are both not empty, "link_url" is
```

src/richie/plugins/glimpse/models.py
```diff
@@ -0,0 +1,86 @@
...
+        always used.
+        """
+        link = False
```
sampaccoud: Returning a boolean for a link seems broken. Why not return "None"?
```suggestion
        link = None
```

src/richie/plugins/glimpse/models.py
```diff
@@ -0,0 +1,86 @@
...
+        Return if the possible link is to be open in a blank target
+        (new window/tab) or not.
+
+        Blank target is only true if object has a value for "link_url"
+        attribute.
+        """
+        if self.link_url:
+            return True
+
+        return False
```
sampaccoud:
```suggestion
        return bool(self.link_url)
```

src/richie/plugins/glimpse/templates/richie/glimpse/glimpse.html
```diff
@@ -0,0 +1,38 @@
+{% load thumbnail %}{% spaceless %}
+
+{% with variant=variant|default:"card_square" %}
+
+    {% if instance.get_link %}
+        <a href="{{ instance.get_link }}"{% if instance.is_blank_target %} target="blank"{% endif %}
```
sampaccoud: I think target blank  takes a "_"? Apparently not safe to use without the following rel attribute:
```suggestion
        <a href="{{ instance.get_link }}"{% if instance.is_blank_target %} target="_blank" rel=”noopener noreferrer”{% endif %}
```

src/richie/plugins/glimpse/templates/richie/glimpse/glimpse.html
```diff
@@ -0,0 +1,38 @@
...
+        class="glimpse glimpse--{{ variant }}">
+    {% else %}
+        <div class="glimpse glimpse--{{ variant }}">
+    {% endif %}
+        {% if instance.image %}
+        <div class="glimpse__media">
+            <img src="{% thumbnail instance.image 200x200 crop upscale subject_location=instance.image.subject_location %}" alt="">
```
sampaccoud: Can you make it an srcset like we do elsewhere?
Also for the card variant, the image could be rectangular like on the course or program?

tests/plugins/glimpse/test_cms_plugins.py
```diff
@@ -0,0 +1,201 @@
...
+        The glimpse model methods "get_link" and "is_blank_target" should
+        return the right value depending glimpse has no link, an internal link
+        or an external link.
+        """
+        page = PageFactory(title__title="Internal link")
+
+        glimpse_no_link = GlimpseFactory()
+        glimpse_external_link = GlimpseFactory(link_url="http://perdu.com")
+        glimpse_internal_link = GlimpseFactory(link_page=page)
+
```
sampaccoud: test also when both link_url and link_page fields are set?

sveetch:
Due to some problems with branch rebase between master and new_integration, this has finally be directly ported into new_integration branch.

PR #973, opened 2020-03-14
🐛Fix an issue with full-text search in course search

mbenadda:
## Purpose

Course search had an issue where filters were lost in some cases when the user typed in a full-text search. This was due to the use of a ref in the <SearchSuggestField /> component which resulted in an outdated function with an old closure being used.

## Proposal

Remove the ref which as a premature attempt at optimization.

We had some issues when trying to build tests to show this issue was resolved. We had to refactor the tests to handle history mocking differently and stop mocking our own History providing hook,
mocking only the browser's interface.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,8 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 - Make sure the course search API shows the 1st related organization by
   placeholder position as highlighted organization for a course instead of
   the first organization by node path.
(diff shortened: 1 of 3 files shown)


jmaupetit:
:wave: @mbenadda can you explain why you choose to close this PR please?

mbenadda:
It was a mistake, and Github would not let me reopen it. We merged https://github.com/openfun/richie/pull/974 instead

PR #976, opened 2020-03-18
✨(blogposts) Add a variant option for blogpost glimpses

sampaccoud:
## Purpose

We want to be able to display blogpost glimpse in several form factor variants.

## Proposal

I copy/pasted what was already done for [courses](https://github.com/openfun/richie/pull/858) and [organization](https://github.com/openfun/richie/pull/786).


mbenadda:
(I think `lint-git` is just failing due to capital A in "Add")

PR #979, opened 2020-04-03
🐛 Empty the search input field when the filters are cleared

### LINKED ISSUE #978: search input is not cleared properly
## Bug Report

**Expected behavior/code**
On the search page, clicking on "Clear x active filter(s)" should clear the search input field if it was one of the active filters

**Actual Behavior**
The search input field is not cleared. The parameter is only removed from the query string and the search results are actualized, but the search term remains in the search input field.

**Steps to Reproduce**
1. Go to the search page
2. Type of full text search term
3. Notice that the search is effective
4. Click on "clear 1 active filter" on the left
5. Notice that the search input field is not cleared.

**Environment**
- Richie version: 1.16.2

mbenadda:
## Purpose

Richie's search engine has a "Clear active filters" link that removes all current filters from the search, including full-text search.

Until now however, it did not clear the actual text in the input field, it only removed the query from the search.

## Proposal

The <SearchSuggestField /> component has no way to know another component did a "FILTER_RESET" or changed the query.

Instead of adding costly plumbing for this, we decided <SearchSuggestField /> can instead just 
clear its own value when it notices the previous courseSearchParams had a query and the incoming ones have not.
We just have to use a ref to keep courseSearchParams N-1 around and do the comparison.

We also added a test to ensure this fixed behavior persists through future changes.

Fixes #978.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,8 @@ Versioning](https://semver.org/spec/v2.0.0.html).
   the first organization by node path.
 - Fix an issue in Course Search that removed existing filters in some cases
   when using full text search.
+- Properly clear the search input when using the "clear fields" button on the
+  course search view.
 
 ### Changed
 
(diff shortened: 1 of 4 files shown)


mbenadda:
FYI the build front failing seems like an issue in circle ci (that persists in the deps cache) which should not happen after we merge #981 and get new deps.

PR #980, opened 2020-04-03
🐛 Fix an issue with whitespace in short search strings

mbenadda:
## Purpose

Our full-text search API endpoint only accepts strings of 3 characters or more. This does not include white space. For instance, "ric" is a valid search query, but "ri " is not.

This was not reflected in our frontend check, which just looked at the string length.

## Proposal

We just have to trim the string before doing this check to ensure the query is valid. Also add a test for this behavior.


mbenadda:
FYI the build front failing seems like an issue in circle ci (that persists in the deps cache) which should not happen after we merge #981 and get new deps.

PR #988, opened 2020-04-15
🔖(chore) bump version to 2.0.0-beta.0

sampaccoud:
### Added

- Add new plugin "Glimpse" to include some basic content with some variant
  form factor available.
- Add new plugin "NestedItem" to implement structured nestable and foldable
  content items.
- Add Styleguide view only available on `/styleguide/` in debug mode.
- Add full favicon set to cover every usages (browser, mobile app, etc..).

### Changed

- Major templates/css refactor to bring a complete new design for the demo.

### Removed

- The `section_cadenced` and `section_list` templates from the section plugin.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -558,7 +558,7 @@ jobs:
           command: |
             BACKEND_VERSION=$(cat setup.cfg | grep "version" | cut -d" " -f3 | cut -d"." -f1,2)
             echo "richie version: ${BACKEND_VERSION}"
-            cat website/versions.json | grep "${BACKEND_VERSION}"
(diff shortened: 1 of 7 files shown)


website/versions.json
```diff
@@ -1 +1 @@
-["1.17", "1.16", "1.15", "1.14", "1.13", "1.12"]
+["2.0", "1.17", "1.16", "1.15", "1.14", "1.13", "1.12"]
```
mbenadda, line 1: I don't think we should publish docs for beta versions as they are subject to change, and at the same time this will duplicate work to add/update docs for 2.0

sampaccoud: Ok I removed the doc release for now.

mbenadda:
Nice change 👍👍

PR #991, opened 2020-04-17
🐛(section) remove useless dependency to nesteditem migration

sveetch:
## Purpose

Migration "0005_migrate_sectionlist_to_nesteditem" has been done
planning to migrate section with deprecated template to a nesteditem
which was finally canceled, but there is still a dependency to
nesteditem initial migration that will cause trouble on migration on 
existing database.

## Proposal

We just remove the dependancy that should fix the issue without any change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,11 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 - Enforce course glimpse footer to be on a single line.
 
+### Fixed
+
+- Fix section migration 0005 that was depending on NestedItem initial
+  migration.
+
 ## [2.0.0-beta.1] - 2020-04-16
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,11 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Fixed
+
+- Fix section migration "0005_migrate_sectionlist_to_nesteditem" that was
+  depending from NestedItem initial migration.
```
sampaccoud, line 14:
```suggestion
  depending on NestedItem initial migration.
```

sampaccoud, line 14:
```suggestion
- Fix section migration 0005 that was depending on NestedItem initial migration.
```

PR #992, opened 2020-04-17
✨(courses) limit number of related objects on all detail pages

sampaccoud:
## Purpose

On a real site, some pages may have a large number of related objects.

For example a category may be related to more that 100 courses. In these cases, the page was too slow to load and too busy for a good user experience. 

## Proposal

We use dj-pagination to paginate related objects thanks to its multi-pagination feature and with a smart use of html element ids.

Fixes https://github.com/openfun/richie/issues/946

This PR is building on top of PR https://github.com/openfun/richie/pull/948 and proposes to use `dj-pagination`.


src/frontend/scss/components/templates/richie/section/_section.scss
```diff
@@ -1,3 +1,18 @@
+section {
+  // Ensure caesura always takes full width with some space around
+  // within flex grid
+  .button-caesura {
+    flex-grow: 1;
+    flex-shrink: 0;
+    flex-basis: calc(100% - 1rem);
```
sveetch, line 7: Use shortcut instead:

`@include sv-flex(1, 0, calc(100% - 1rem));`

mbenadda:
Closes #948

Issue #732, opened 2019-06-20
Remove first_name, last_name and person_title fields on person model


sampaccoud:
## Feature Request

**Is your feature request related to a problem or unsupported use case? Please describe.**
In PR https://github.com/openfun/richie/pull/731 : 
- The first_name, last_name and person_title fields were removed from the person model,
- The PersonTitle model was removed.

The corresponding database migrations were not done in order in ensure smooth blue/green deployments.

**Describe the solution you'd like**
The database migrations needs to be done in a future release v1.4.0 forcing the sequential application of release v1.3.x (in which PR https://github.com/openfun/richie/pull/731 was done) before v1.4.0 in order to be sure that the version of the code running on the blue stack does not required these fields when the migrations of v1.4.0 will remove them from the database.

PR #993, opened 2020-04-17
🗃(courses|plugins) liquidate pending migrations in view of v2.0.0

sampaccoud:
## Purpose

v2.0.0 is a breaking release. We take this opportunity to liquidate all pending migrations that remove fields.

## Proposal

I ran `makemigrations` on the whole project. After this PR, there will not be any more pending migrations.

Fixes https://github.com/openfun/richie/issues/732


sampaccoud:
@mbenadda ok for CHANGELOG. Nothing special to do in the UPGRADE imo.

Issue #769, opened 2019-07-24
Add blocks to templates


sampaccoud:
## Feature Request

**Is your feature request related to a problem or unsupported use case? Please describe.**
Several projects have expressed the need to make minor customizations to a template. For the moment, they most often need to override the whole template because we are lacking blocks in our templates.

**Describe the solution you'd like**
Add a lot of blocks in all our templates, one per functional section.

PR #995, opened 2020-04-17
🎨(blogposts) add blocks to each part of the blogpost detail template

sampaccoud:
## Purpose

In different projects using Richie, we need to override parts of the blog post detail page.

## Proposal

Add blocks around each part of the blogpost detail page to make it easier to override it.

Part of https://github.com/openfun/richie/issues/769

### DIFF
--- a/src/richie/apps/courses/templates/courses/cms/blogpost_detail.html
+++ b/src/richie/apps/courses/templates/courses/cms/blogpost_detail.html
@@ -14,118 +14,146 @@
 
 
 {% block subheader_content %}
-<div class="subheader__container">
-    <h1 class="subheader__title">{% page_attribute "page_title" request.current_page.parent_page %}</h1>
-</div>
+    <div class="subheader__container">
+        <h1 class="subheader__title">{% page_attribute "page_title" request.current_page.parent_page %}</h1>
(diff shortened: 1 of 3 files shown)


src/richie/apps/courses/templates/courses/cms/blogpost_detail.html
```diff
@@ -32,97 +34,123 @@ <h1 class="subheader__title">{% page_attribute "page_title" request.current_page
...
-                </div>
-
-                <div class="blogpost-detail__body">
-                    {% placeholder "body" or %}
-                        <p class="blogpost-detail__empty">
-                            {% trans "No body content" %}
-                        </p>
-                    {% endplaceholder %}
-                </div>
+                {% block cover %}
```
mbenadda, line 93: This is very much a nitpick but you should probably either always indent or never indent after a block declaration.

Given 4-space indentation is already a little bit painful to work with wrt. excessive indentation all around, I'd lean towards no indentation after block declarations 😇

sampaccoud: Done! The diff is now unreadable ^^

PR #1005, opened 2020-04-24
✨ Improve query analyzer language selection

### LINKED ISSUE #997: Manage textual search query analysis language
## Feature Request

This is somewhat between a bug report and a feature request.

Up to this point, we're indexing content using analyzers matching its language. Then, when a textual query hits, we're searching in all languages for fields we're searching in.

This can cause unwanted results, most notably because eg. a French query will be analyzed by an English analyzer to search in the English version of the field (in addition to being analyzed by a French analyzer to search in the French version of the same field). The bug here happens when there are stop words in the query: they are removed by French analysis but kept in by the English analysis, which matches a lot more documents than we'd like.

This is especially problematic because we have our own sorting algorithm for courses so the irrelevant results generated by this issue can be at the top of the search results.

Searching in all languages for a field is not just an oversight. We actually think it can support more use cases and be an affordance that helps users find what they're looking for (if we think of a user that landed on the wrong language for the portal, or a user from a country where more than one language is routinely used).

**This means we do not want to just force the site language onto the query analysis, or add an API parameter to control it.**

Instead, we can use NLP to try and guess the language in the query and use the appropriate analyzer if we have one and if the guess is conclusive enough. We can also use the current website language to guide this guess and serve as a default of last resort.

Practically, this means doing this guessing work in the request cycle and specifying the analyzer on the query instead of relying on the one listed in our mappings.

As a bonus, we might look into weighing results for languages depending on our best guess for a language and the current site language.

See above for discussion on other strategies.

This should be transparent for users, resolving an issue around cases where they get bogus results due to stop word analysis across languages, and generally providing the best results we reasonably can.
This does not touch indices so the upgrade is transparent there too.

Currently working on it 👍

mbenadda:
## Purpose

Each field defines its own default search analyzer in its mapping (or defaults to the content analyzer if a specific search query analyzer is not defined). This is appropriate in principle but it
breaks down when we attempt to do multilingual searches: search queries are then analyzed using the field's analyzer or search analyzer, which might be in the wrong language, and give terrible results for the search query.

To prevent this from happening, we decided to enforce an analyzer at query time. If we can find an appropriate analyzer, it can then be used before searching even in indices in other languages.

## Proposal

To best find this appropriate analyzer, we let the frontend client pass us a preferred query language, which it can determine by guessing and offering guesses to the user to help them, or other UI elements.

What we suggest:
- do not change the base UI at all;
- guess query language in the frontend (using the same Java `langdetect` lib as ported to javascript instead of python);
- when our confident guess differs from the current language **and** we can offer a better analyzer (eg. the language we guessed is supported by the current Richie instance), show a toast to the user offering this: _It seems like your search query is in Spanish, would you like to search in Spanish?_;
- if the user does change the query language, we then display a language selector next to the search input.

This way we can be helpful wrt. query language and irrelevant results, without breaking search behavior due to the inherent unreliability of attempting to guess the language of a very small text sample. We also don't pollute the UI for something that 95-99% of users will never need.

In the absence of a cue by the frontend (which would only be there for less than 5% of queries), we then default to the current UI language.

In any case, we can keep searching in all languages' versions of the fields and be sure we have the best analyzer we could offer.

## Alternatives considered

We started out trying to just automatically detect the query language and using that guess to choose the most correct analyzer without any user input besides their text query. We wanted to use `langdetect` for this.

`langdetect` is a great library, but it cannot provide reliable guesses in the absence of enough data. The author of the original library (in Java) notes [here](http://www.slideshare.net/shuyo/language-detection-library-for-java) that it is not reliable for tweets which are too short (at 100-500 characters), and that for automatic detection to be worthwhile, it needs to be very reliable.

They echo [this note](https://www.elastic.co/fr/blog/multilingual-search-using-language-identification-in-elasticsearch) from Elastic that notes ElasticSearch's own language detection should not be used for search queries as it's unreliable for text shorter than 50 characters.

Additionally, we observed as we attempted it that it fails for common queries attempted in Richie. More importantly, it produced unexpected behavior as the guesses changed when using search-as-the-user-types: if the analyzer changes through different languages as the user types their query, the search results will also flicker between vastly different result sets instead of progressively narrowing to our best results.

## TODO

- [x] set an analyzer for full-text queries in the backend;
- [x] ~accept a new query string parameter on search API endpoints so the frontend can let us use the user's preferred language (as opposed to the current UI language);~
- [ ] ~help the user select an appropriate query language in the frontend;~

Fixes #997.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,11 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Fixed
+
+- Fix an issue where analyzing course search queries with irrelevant analyzers
+  caused very low relevance results in course search.
+
 ## [2.0.0-beta.4] - 2020-04-20
 
 ### Fixed
--- a/src/richie/apps/search/defaults.py
+++ b/src/richie/apps/search/defaults.py
@@ -12,6 +12,11 @@ ES_PAGE_SIZE = 10
 # Use a lazy to enable easier testing by not defining the value at bootstrap time
 ES_INDICES_PREFIX = lazy(lambda: settings.RICHIE_ES_INDICES_PREFIX)()
 
+# Define which analyzer should be used for each language
+QUERY_ANALYZERS = getattr(
+    settings, "RICHIE_QUERY_ANALYZERS", {"en": "english", "fr": "french"}
+)
+
 # Define the scoring boost (in ElasticSearch) related value names receive when using
 # full-text search.
 # For example, when a user searches for "Science" in full-text, it should match any
--- a/tests/apps/search/test_viewsets_categories.py
+++ b/tests/apps/search/test_viewsets_categories.py
@@ -148,8 +148,9 @@ class CategoriesViewsetsTestCase(TestCase):
                             {"term": {"kind": "subjects"}},
                             {
                                 "multi_match": {
-                                    "query": "Science",
+                                    "analyzer": "english",
                                     "fields": ["title.*"],
+                                    "query": "Science",
                                 }
                             },
                         ]
--- a/src/richie/apps/search/forms.py
+++ b/src/richie/apps/search/forms.py
@@ -6,10 +6,11 @@ from functools import reduce
 
 from django import forms
 from django.conf import settings
+from django.utils.translation import get_language
 
 import arrow
 
-from .defaults import RELATED_CONTENT_BOOST
+from .defaults import QUERY_ANALYZERS, RELATED_CONTENT_BOOST
 from .filter_definitions import FILTERS, AvailabilityFilterDefinition
 
 # Instantiate filter fields for each filter defined in settings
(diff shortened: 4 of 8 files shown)


mbenadda:
## Update

After some consideration, we decided that trying to guess the language in the frontend and adding a query parameter was too much work and "trying too hard" in terms of a UX we have not seen elsewhere.

For now, we can fix what needs fixing by just analyzing queries in the current active language. If we see problems with this approach, those problems will help guide us to the best way to handle language in query analysis later on.

CHANGELOG.md
```diff
@@ -8,6 +8,13 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Fixed
+
+- Analyze text search queries with an analyzer for the current UI language
+  as opposed to the default analyzer for each language field. This should fix
+  issues where analyzing queries with multiple analyzers led to irrelevant
+  results.
```
sampaccoud, line 16: Maybe too long for a CHANGELOG that is starting to grow out of control!

src/richie/apps/search/defaults.py
```diff
@@ -12,6 +12,11 @@
 # Use a lazy to enable easier testing by not defining the value at bootstrap time
 ES_INDICES_PREFIX = lazy(lambda: settings.RICHIE_ES_INDICES_PREFIX)()
 
+# Define which analyzer should be used for each language
+QUERY_ANALYZERS_BY_LANGUAGE = getattr(
+    settings, "RICHIE_QUERY_ANALYZERS_BY_LANGUAGE", {"en": "english", "fr": "french"}
```
sampaccoud, line 17: BY_LANGUAGE may not be necessary. Like the CMS_LANGUAGES setting is in fact a complex dict, you could just call it `QUERY_ANALYZERS`

tests/apps/search/test_viewsets_categories.py
```diff
@@ -148,8 +148,9 @@ def test_viewsets_categories_search(self, mock_search):
                             {"term": {"kind": "subjects"}},
                             {
                                 "multi_match": {
-                                    "query": "Science",
+                                    "analyzer": "english",
```
sampaccoud, line 151: I didn't understand what changes in the code lead to this change of behavior?

mbenadda: This one here: https://github.com/openfun/richie/pull/1005/files#diff-848b539601c898e16a2f4cc46dbcf50fR319

PR #1009, opened 2020-05-03
Program glimpses and minor integration changes

sveetch:
## Purpose

Program glimpse layout have not been changed since it has been added
to Richie, it was an old basic square glimpse. Fun-mooc mockups have
a new clean layout for this glimpse.

Also during integration on various richie projects some minor things to
resolve have been found.

## Proposal

Back-port the glimpse layout from fun-mooc and adapt it to Richie theme.

A lot of minor and (almost all) backward compatible changes:

- Every theme property should now be safe to be null even if they are used to construct a css property or a pseudo element;
- New theme property `body-content.base-color` to set the base body element font color in global variables;
- Migrated google font loading to new version (v2);
- Correctly disable the useless block "subheader" in homepage template;
- Fixed decoration svg "arc" and "curve" which did have a tiny glitch;
- Removed deprecated style rule ".old_user-login" from "UserLogin" component;
- New boolean setting "$body-padding-fix" to control usage of "content-padding-fix" mixin related to footer decoration;
- Fixed ".button-caesura" to shrink when text is too long to avoid horizontal scrollbar;
- Fixed breadcrumb item to shrink when text is too long to avoid horizontal scrollbar;
- New theme property `hero-intro.title-color` to set a specific color on hero title;
- Fixed vertical alignment for blogpost detail date and authors;
- Alternative logo (for footer) "logo-white" has been renamed to a more generic name "logo-alt";


sampaccoud:
@sveetch can you explain what is the idea behind making padding switch configurable. Seems to generate a lot of `ifs` so can you confirm that it is worth it?

sveetch:
> @sveetch can you explain what is the idea behind making padding switch configurable. Seems to generate a lot of `ifs` so can you confirm that it is worth it?

This is related to footer image decoration which induce a padding fix, so when you disable the decoration, the padding fix should not be enabled else it produce a large useless padding. This is alike the if around some pseudo element which depend on theme properties, if you disable these theme parts, it should not produce any useless space.

Be assured this is not a start to turn every padding/margin/etc.. into variables.

PR #1017, opened 2020-05-15
Add memcached as the default Django cache backend

mbenadda:
## Purpose

We started out planning to enable caching in Django-CMS and add tests to ensure everything is cached as we expect.

As it turns out, cache was already enabled (it is on by default in Django-CMS) and we pretty much had nothing much to do once we figured that out.

We also had a lot of trouble figuring our meaningful tests to add, especially considering that Django-CMS already tests their cache strategy.

## Proposal

As a result, we're just adding a better cache backend in memached instead of RAM. We don't think Django's caching middleware is relevant here as Django-CMS manually handles it by calling cache getters/setters directly.


mbenadda:
We want to add redis instead of memcached as a cache backend. Closing this as we'll make another PR with redis.

PR #1034, opened 2020-06-07
Course run enrollment

mbenadda:
## Purpose

One of the key goals of Richie is to sit in front of one or more LMSes and allow users to discover courses and enroll in them easily. The first part is already taken care of, but there is currently no way to enroll in a course directly from Richie. Users need to be redirected to the LMS itself and do the enrollment there.

We propose a new frontend component, `<CourseRunEnrollment />`, which is shown on the "course detail" and "course run detail" views, and displays the relevant actions for users depending on the course run states, user login & enrollment states, and actual enrollments in underlying LMSes.

To that end, we decided to go with an `Enrollment` model in Richie itself. This is the first step towards having Richie piloting enrollments through to different backend LMSes. It's also a performance issue as it avoids painful round-trips through more API endpoints on the LMSes themselves.

NB: This pull request does not interoperate with any LMS yet. It lays the groundwork for managing enrollments in Richie. Enrollments in the actual LMSes can then be developed by integrating with this new enrollment model.

## Proposal

- [x] create an `Enrollment` model (with its testing factory)
- [x] open `course-runs` & `enrollments` endpoints (so the frontend can get state information)
- [x] add an `enroll` API endpoint to create new enrollments
- [x] build a frontend component to actually enroll users in course runs
- [x] integrate with the existing course & course run detail views
- [x] add a feature flag to toggle management of enrollments in Richie

Necessary asides:

- [x] enable live debugging the Python backend through `docker-compose` settings
- [x] fix a bug where the `<Spinner />` was broken due to missing CSS since v2
- [x] improve `<Spinner />` a11y with `aria-labelledby` passthrough
- [x] add test utils (factories & deferred, will be generalized to the codebase in another PR)


src/richie/apps/courses/models/enrollment.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+class Enrollment(models.Model):
+    """
+    Enrollment model. Links a user to the course runs they have enrolled in.
+
+    In most cases, actual enrollment also needs to happen in the actual LMS behind Richie.
+    Still, we need a simple and fast way to look up and manage user enrollments in
+    Richie itself, which is tough to achieve if every action necessitates a bunch of
+    API calls to different LMSes.
```
sampaccoud, line 18: Reading this made me realize that we need a feature switch on this so that it is still possible to use Richie without integration with LMS. In this mode, the course run just provides a link to the course on the LMS.
This will also allow us to merge this PR without breaking things until the user dashboard and connection backend are also available?

src/richie/apps/courses/models/enrollment.py
```diff
@@ -0,0 +1,70 @@
...
+        on_delete=models.CASCADE,
+        related_name="enrollments",
+        related_query_name="enrollment",
+    )
+    course_run = models.ForeignKey(
+        verbose_name=_("course run"),
+        help_text=_("course run the user enrolled in"),
+        to=CourseRun,
+        on_delete=models.CASCADE,
+        related_name="+",
```
sampaccoud, line 44: could be interesting to get the users enrolled on a course run? course_run.users_enrolled?

src/richie/apps/courses/models/enrollment.py
mbenadda: Good call, I had not thought about this 👍

docker-compose.yml
```diff
@@ -35,6 +35,8 @@ services:
     depends_on:
```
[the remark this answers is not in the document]
mbenadda: It's so I can put `pdb` breakpoints in the Python code and then `docker attach` to them. Without this it just prints out the `pdb` stuff and exits

src/richie/apps/courses/models/enrollment.py
mbenadda: Should this be a separate app then? So if you don't use Richie enrollments you don't even have to create db tables and everything?

src/richie/apps/courses/models/enrollment.py
```diff
@@ -0,0 +1,70 @@
...
+    )
+
+    mode = models.CharField(
+        verbose_name=_("mode"),
+        help_text=_(
+            "a mode is a kind of enrollment, course runs may support different ones"
+        ),
+        max_length=255,
+        default="HONOR",
```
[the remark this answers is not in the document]
mbenadda: I think choices would be way better (or even a foreign key to another model if need be) but I'm not sure what to put there right now. Should I just remove this field and only discuss it if/when an explicit need arises in Richie?

docker-compose.yml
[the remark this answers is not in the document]
mbenadda: Yes, it already worked in tests. This is indeed to have access to `pdb` when making requests from the browser

src/richie/apps/courses/models/enrollment.py
sampaccoud: Yes it's probably a good idea. And the whole user dashboard, ecommerce etc. will be in this app... In any case, make sure to force your db table names so that we can move things around if the choice we make turns out to be wrong.

mbenadda:
Just adding the visualization for the state machine in the frontend component 

Capture d’écran 2020-06-08 à 11 10 18

src/richie/apps/enrollments/api.py
```diff
@@ -0,0 +1,86 @@
...
+            queryset = Enrollment.objects.filter(user=self.request.user)
+
+            if self.request.query_params.get("course_run"):
+                return queryset.filter(
+                    course_run__id=self.request.query_params.get("course_run")
+                )
+
+            return queryset
+
+        return Enrollment.objects.filter(user=-1)
```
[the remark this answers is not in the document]
mbenadda: Yes, exactly! I was not aware there was a helper for this.

src/richie/apps/courses/serializers.py
```diff
@@ -0,0 +1,22 @@
...
+class CourseRunSerializer(serializers.ModelSerializer):
+    """
+    Course run serializer. Includes state but not any nested object or CMS page information.
+    """
+
+    class Meta:
+        model = CourseRun
+        fields = [
+            "id",
+            "resource_link",
```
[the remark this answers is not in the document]
mbenadda: It's already exposed on course run & course pages if you don't use the enrollments app. Then again, I don't think I use it anywhere on the frontend so I could remove it

src/frontend/js/components/CourseRunEnrollment/_styles.scss
```diff
@@ -0,0 +1,46 @@
+.course-run-enrollment {
```
lunika, line 1: in commit message

`s/It turns our this button/It turns out this button/`

src/frontend/js/components/CourseRunEnrollment/index.spec.tsx
```diff
@@ -0,0 +1,214 @@
+import { act, render, screen, fireEvent, wait } from '@testing-library/react';
```
lunika, line 1:
```suggestion
import { act, render, screen, fireEvent } from '@testing-library/react';
```

src/frontend/js/components/CourseRunEnrollment/index.tsx
```diff
@@ -0,0 +1,330 @@
...
+    id: 'components.CourseRunEnrollment.enroll',
+  },
+  enrolled: {
+    defaultMessage: 'You are enrolled in this course run',
+    description:
+      'Help text for users who see the "Go to course" CTA on course run enrollment',
+    id: 'components.CourseRunEnrollment.enrolled',
+  },
+  enrollmentClosed: {
+    defaultMessage: 'Enrollment in this course run is closed as the moment',
```
lunika, line 28:
```suggestion
    defaultMessage: 'Enrollment in this course run is closed at the moment',
```

src/richie/apps/core/context_processors.py
```diff
@@ -32,6 +32,9 @@ def site_metas(request):
             "domain": site_current.domain,
             "web_url": f"{protocol:s}://{site_current.domain:s}",
         },
+        "RICHIE_APPS": {
```
lunika, line 35: in commit message

`s/if an optional Richie app is installed our not/if an optional Richie app is installed
or not/`

mbenadda:
Everything worth keeping and updating (and that was not already merged) from this PR has been rebased in #1144.

PR #1046, opened 2020-06-10
👌(person) grid constraint for person 'main_content' placeholder

sveetch:
## Purpose

Placeholder "main_content" from person detail template have never
been really tested and so was not correctly integrated.

## Proposal

This commit just place this placeholder inside its own block element
with width constraint so it stick to the grid constraint.


src/richie/apps/courses/templates/courses/cms/person_detail.html
```diff
@@ -74,7 +74,13 @@ <h1 class="subheader__title">{% render_model current_page "title" %}</h1>
...
 {% with header_level=2 person=current_page.person %}
 <div class="person-detail">
-    {% placeholder "maincontent" %}
+    {% if not current_page|is_empty_placeholder:"maincontent" %}
+    <div class="person-detail__maincontent person-detail__block">
+        <div class="person-detail__row">
+            {% placeholder "maincontent" %}
+        </div>
+    </div>
+    {% endif %}
```
sampaccoud, line 83: Can you add a plugin in the demo to materialize this change?

PR #1048, opened 2020-06-12
✨(dropdown) add new layout component "dropdown"

sveetch:
## Purpose

A need for a dropdown component emerge from language menu which is
required to be a dropdown. Since we don't have any Framework
component available, we need to implement it first.

## Proposal

This commit adds this new component including the library "dropmic"
with a Javascript and Sass sources both modified to fit our
requirements (BEM names). A dropdown sample is available from
Styleguide page.

Since dropdown component has been added, we can integrate the language menu
with it and as required add it also in header.

This update also some theme properties related from header and footer groups to
include their own CTA dropdown colors, update styleguide and language menu.

header
footer


2018D:
Nice! i started looking into version 2 of richie this week. I was thinking that a dropdown language chooser would be a nice to have. I'm really happy to see you guys also thought the same 😄

sampaccoud:
For coherence, we want to stick to React for js components. This PR is closed in favor of https://github.com/openfun/richie/pull/1107.

PR #1051, opened 2020-06-16
👌(layout) make login/logout CTA more visible

sveetch:
## Purpose

Login and logout CTA in header was less visible than the ones from mockups.

## Proposal

We just change a theme color and make them uppercase as from mockups.


sampaccoud:
@sveetch can you add a snapshot to show it around?

sveetch:
@sampaccoud here there are some screenshot:

register

logout

PR #1052, opened 2020-06-16
👌(course) make course glimpse category cartouche less "aggressive"

sveetch:
## Purpose

In course glimpses, category cartouche have too much padding and font is in
bold, opposed to mockup which was mocking a less bigger cartouche without
bolded font.

## Proposal

Here we just fix these minor issue in course glimpse Sass.

A screenshot for sample (left for this PR display and right for current invalid layout):

cartouche

### DIFF
--- a/src/frontend/scss/objects/_course_glimpses.scss
+++ b/src/frontend/scss/objects/_course_glimpses.scss
@@ -98,7 +98,7 @@ $course-glimpse-content-padding-sides: 0.7rem !default;
       $border-width: 0,
       $font-size: 0.8rem,
       $line-height: 1.1,
-      $font-weight: bold
+      $font-weight: $font-weight-normal
     );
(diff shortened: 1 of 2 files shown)


src/frontend/scss/objects/_course_glimpses.scss
```diff
@@ -202,7 +208,7 @@ $course-glimpse-content-padding-sides: 0.7rem !default;
     @include sv-flex(1, 0, auto);
     margin: 0;
     padding: 0.75rem 0;
-    font-weight: bold;
+    font-weight: $font-weight-bold;
```
sampaccoud, line 211: Why not directly pass the css font-weight value?
```suggestion
    font-weight: $font-weight;
```

sveetch: Font thickness can change depending the font families, so this way with variables, a project could use another font and set their correct thickness values.

PR #1053, opened 2020-06-17
👌(layout) add template block around footer slogan

sveetch:
## Purpose

We need a template block around footer slogan so projects can change
it or even disable it if needed.

## Proposal

This just update the base.html to put the slogan in a template block
"body_footer_slogan".

### DIFF
--- a/src/richie/apps/core/templates/richie/base.html
+++ b/src/richie/apps/core/templates/richie/base.html
@@ -126,7 +126,9 @@
                         </a>
                     </div>
                     <div class="body-footer__insert">
-                        <p class="body-footer__title">{% trans "Life-changing learning!" %}</p>
+                        {% block body_footer_title %}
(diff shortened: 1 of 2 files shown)


src/richie/apps/core/templates/richie/base.html
```diff
@@ -126,7 +126,9 @@
                         </a>
                     </div>
                     <div class="body-footer__insert">
-                        <p class="body-footer__title">{% trans "Life-changing learning!" %}</p>
+                        {% block body_footer_slogan %}
+                            <p class="body-footer__title">{% trans "Life-changing learning!" %}</p>
+                        {% endblock body_footer_slogan %}
```
sampaccoud, line 131: Well for coherence, since it is called title:
```suggestion
                        {% block body_footer_title %}
                            <p class="body-footer__title">{% trans "Life-changing learning!" %}</p>
                        {% endblock body_footer_title %}
```

Issue #1067, opened 2020-06-30
The remove active filters on the search page breaks the search input field


sampaccoud:
## Bug Report

**Expected behavior/code**
The "Remove 1 active filter"  link on the search page side panel should not break the search input.

**Actual Behavior**
After clearing active filters the search input does not take input anymore.

**Steps to Reproduce**
1. Load the search page
2. Filter on any facet
3. Click on "remove 1 active filter"
4. Notice that it is no more possible to input a search query

**Environment**
- Richie version: 2.0.0-beta.8
- Platform: Docker

PR #1069, opened 2020-07-01
⏪(frontend) revert how SearchSuggestField clears itself

lunika:
## Purpose

This commit revert commit 1bfb2d7. This commit introduces a bug explain
in #1067. To be sure to fix it a test was added reproducing the issue
explanation. This commit is temporary, a better fix can be found but we
need a quick fix now.

## Proposal

- [x] revert commit 1bfb2d7
- [x] add a test reproducing #1067


lunika:
Closed in favor of #1070

PR #1079, opened 2020-07-10
🔨 Enable live-debugging in app container

mbenadda:
## Purpose

During development, it is generally useful to enable live-debugging in the running container. By default, docker/compose prevents us from attaching to the debugger.

## Proposal

Enabling these two settings simply allows `docker attach` to connect to eg. the live python debugger and should not have any side-effects anywhere else as this docker-compose config is only used in development.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -35,6 +35,8 @@ services:
     depends_on:
       - "${DB_HOST:-postgresql}"
       - "elasticsearch"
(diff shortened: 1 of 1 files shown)


jmaupetit:
Wouldn't it be useful to add a "debugging howto" section in the documentation? For now, I don't see what do you attach to and what's running `pdb`. :pray:

PR #1085, opened 2020-07-21
✨(core) add a plugin for django-redis that supports Redis Sentinel

madmatah:
## Purpose

We want to enable Django caching features in production and use Redis (with Sentinel support) as a cache backend.

## Proposal

Use [django-redis](https://github.com/jazzband/django-redis) library and add support for Redis Sentinel with a [custom client class](https://github.com/lamoda/django-sentinel).

- [x] Add `django-redis` dependency
- [x] Create a `richie.apps.core.cache.SentinelClient` client class based on [lamoda/django-sentinel](https://github.com/lamoda/django-sentinel) work
- [x] Configure a redis-sentinel cluster in `docker-compose.yml`
- [x] Import tests from [lamoda/django-sentinel](https://github.com/lamoda/django-sentinel) and update them to pass `make lint-back`

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -35,6 +35,7 @@ services:
     depends_on:
       - "${DB_HOST:-postgresql}"
       - "elasticsearch"
+      - redis-sentinel
     stdin_open: true
     tty: true
 
@@ -53,3 +54,32 @@ services:
       - env.d/development/crowdin
     user: "${DOCKER_USER:-1000}"
     working_dir: /app
+
+  redis-sentinel:
+    image: docker.io/bitnami/redis-sentinel:6.0-debian-10
+    depends_on:
+      - redis-primary
+      - redis-replica1
+      - redis-replica2
+    environment:
+      - REDIS_MASTER_HOST=redis-primary
+
+  redis-primary:
+    image: docker.io/bitnami/redis:6.0-debian-10
+    environment:
+      - ALLOW_EMPTY_PASSWORD=yes
+      - REDIS_REPLICATION_MODE=master
+
+  redis-replica1:
+    image: docker.io/bitnami/redis:6.0-debian-10
+    environment:
+      - ALLOW_EMPTY_PASSWORD=yes
+      - REDIS_REPLICATION_MODE=slave
+      - REDIS_MASTER_HOST=redis-primary
+
+  redis-replica2:
+    image: docker.io/bitnami/redis:6.0-debian-10
+    environment:
+      - ALLOW_EMPTY_PASSWORD=yes
+      - REDIS_REPLICATION_MODE=slave
+      - REDIS_MASTER_HOST=redis-primary
--- a/setup.cfg
+++ b/setup.cfg
@@ -40,6 +40,7 @@ install_requires =
     elasticsearch>=6.0.0,<7.0.0 # pyup: ignore
     social-auth-core[openidconnect]==3.2.0 # pyup: ignore
     social-auth-app-django==3.1.0 # pyup: ignore
+    django-redis>=4.11.0
 package_dir =
     =src
 packages = find:
@@ -54,6 +55,7 @@ dev =
     ipdb==0.13.3
     ipython==7.16.1
     isort==4.3.21
+    msgpack===1.0.0
     mysqlclient==2.0.1
     pylint==2.5.3
     pylint-django==2.1.0
--- a/src/richie/apps/core/cache.py
+++ b/src/richie/apps/core/cache.py
@@ -1,0 +1,131 @@
+"""
+Plugin for django-redis that supports redis sentinel.
+Credits :
+ - https://github.com/lamoda/django-sentinel
+ - https://github.com/KabbageInc/django-redis-sentinel
+"""
+
+import logging
+import random
+
+from django.conf import settings
+from django.core.exceptions import ImproperlyConfigured
+
+from django_redis.client import DefaultClient
+from redis.sentinel import Sentinel
+
+DJANGO_REDIS_LOGGER = getattr(settings, "DJANGO_REDIS_LOGGER", __name__)
+
+
+class SentinelClient(DefaultClient):
+    """
+    Sentinel client object extending django-redis DefaultClient
+    """
+
+    def __init__(self, server, params, backend):
+        """
+        Slightly different logic than connection to multiple Redis servers.
+        Reserve only one write and one read descriptor, as they will be closed on exit anyway.
+        """
+        super(SentinelClient, self).__init__(server, params, backend)
+        self._client_write = None
+        self._client_read = None
+        (
+            self._master_name,
+            self._sentinel_hosts,
+            self._database_name,
+        ) = self.parse_connection_string(server)
+        self.log = logging.getLogger(DJANGO_REDIS_LOGGER)
+
+    @staticmethod
+    def parse_connection_string(connection_string):
+        """
+        Parse connection string in format:
+            master_name/sentinel_server:port,sentinel_server:port/db_id
+        Returns master name, list of tuples with pair (host, port) and db_id
+        """
+        try:
+            master_name, servers_string, database_name = connection_string.split("/")
+            servers = [host_port.split(":") for host_port in servers_string.split(",")]
+            sentinel_hosts = [(host, int(port)) for host, port in servers]
+        except (ValueError, TypeError, IndexError):
+            raise ImproperlyConfigured("Incorrect format '%s'" % (connection_string))
+
+        return master_name, sentinel_hosts, database_name
+
+    def get_client(self, write=True, tried=(), show_index=False):
+        """
+        Method used to obtain a raw redis client.
+
+        This function is used by almost all cache backend
+        operations to obtain a native redis client/connection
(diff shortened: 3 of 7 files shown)


src/richie/apps/core/cache.py
```diff
@@ -0,0 +1,140 @@
...
+import logging
+import random
+
+from django.conf import settings
+from django.core.exceptions import ImproperlyConfigured
+
+from django_redis.client import DefaultClient
+from redis.sentinel import Sentinel
+
+DJANGO_REDIS_LOGGER = getattr(settings, "DJANGO_REDIS_LOGGER", False)
```
sampaccoud, line 17: instead of `DJANGO_REDIS_LOGGER or __name__` below, why not default to `__name__` here?

But not sure why we would want to rename the logger and instantiate it in the class rather than here? This is not a pattern we use elsewhere.

src/richie/apps/core/cache.py
```diff
@@ -0,0 +1,140 @@
...
+            try:
+                host, port = random.choice(  # nosec
+                    sentinel.discover_slaves(self._master_name)
+                )
+            except IndexError:
+                self.log.debug("no slaves are available. using master for read.")
+                host, port = sentinel.discover_master(self._master_name)
+
+        if password:
+            connection_url = "redis://:%s@%s:%s/%s" % (
```
sampaccoud, line 109: You can use f strings here ;-)

setup.cfg
```diff
@@ -54,6 +55,7 @@ dev =
     ipdb==0.13.3
     ipython==7.16.1
     isort==4.3.21
+    msgpack===1.0.0
```
sampaccoud, line 58: do we want to add this dependency or can we impose json?

src/richie/apps/core/cache.py
```diff
@@ -0,0 +1,140 @@
...
+        self._client_read = None
+        (
+            self._master_name,
+            self._sentinel_hosts,
+            self._database_name,
+        ) = self.parse_connection_string(server)
+        self.log = logging.getLogger((DJANGO_REDIS_LOGGER or __name__))
+
+    @staticmethod
+    def parse_connection_string(constring):
```
sampaccoud, line 41: Why not:
```suggestion
    def parse_connection_string(connection_string):
```

src/richie/apps/core/cache.py
```diff
@@ -0,0 +1,140 @@
...
+        Returns master name, list of tuples with pair (host, port) and db_id
+        """
+        try:
+            connection_params = constring.split("/")
+            master_name = connection_params[0]
+            servers = [
+                host_port.split(":") for host_port in connection_params[1].split(",")
+            ]
+            sentinel_hosts = [(host, int(port)) for host, port in servers]
+            database_name = connection_params[2]
```
sampaccoud, line 54: Maybe more pythonic using unpacking:
```suggestion
             master_name, servers_string, database_name = connection_string.split("/")
            servers = [
                host_port.split(":") for host_port in servers_string.split(",")
            ]
            sentinel_hosts = [(host, int(port)) for host, port in servers]
```

src/richie/apps/core/cache.py
```diff
@@ -0,0 +1,140 @@
...
+        instance.
+        """
+        self.log.debug("get_client called: write=%s", write)
+        if write:
+            if self._client_write is None:
+                self._client_write = self.connect(write)
+
+            if show_index:
+                return self._client_write, 0
+            return self._client_write
```
sampaccoud, line 75: hmm returning sometimes a tuple, sometimes just the connection. Maybe we should improve typing by returning always the same thing?

src/richie/apps/core/cache.py
```diff
@@ -0,0 +1,140 @@
...
+            self._client_read = self.connect(write)
+
+        if show_index:
+            return self._client_read, 0
+        return self._client_read
+
+    # pylint: disable=arguments-differ
+    def connect(self, write=True):
+        """
+        Creates a redis connection with connection pool.
```
sampaccoud, line 87:
```suggestion
        Create a redis connection with connection pool.
```

src/richie/apps/core/cache.py
madmatah: First, I agree with you, this is not a pattern we use elsewhere in our projects.
And this is not a choice of my own. But :

Since the `SentinelClient` is a `django-redis` add-on, it uses the same pattern than this library for logging in order to be consistent.

According to [django-redis documentation](https://github.com/jazzband/django-redis/blob/4a226ff18a9f0f8401104526fc54ce14de7f210f/README.rst#log-ignored-exceptions) :

> If you wish to specify the logger in which the exceptions are output, simply set the global variable DJANGO_REDIS_LOGGER to the string name and/or path of the desired logger. This will default to __name__ if no logger is specified and DJANGO_REDIS_LOG_IGNORED_EXCEPTIONS is True:
> 
> DJANGO_REDIS_LOGGER = 'some.specified.logger'

And it is used like this [in the djange-redis library](https://github.com/jazzband/django-redis/blob/b7d64f0a80180ecfe8900407fbdf5e4196fa7a2a/django_redis/cache.py#L59-L63)...


Since this does not break anything and because there is a sane default value, I suggest to keep it like this.

sampaccoud: > This will default to name if no logger is specified

So maybe you don't need to worry about it since "name" is what we use on our side?

setup.cfg
madmatah: This is just a dev dependency to be able to run the tests.

Let our users choose their favorite serialization library :-)

src/richie/apps/core/cache.py
madmatah: This behavior is required by the `django_redis` library.
See [here](https://github.com/jazzband/django-redis/blob/879cbc270466a2d731f369e7cfede375705a0fc5/django_redis/client/default.py#L82-L98).

I'm not comfortable with changing the method prototype without patching it in the `django_redis` library.

tests/apps/core/test_cache.py
```diff
@@ -0,0 +1,409 @@
+"""
+Test suite for the cache module in the `core` application
+
+Credits :
+ - https://github.com/lamoda/django-sentinel
```
jmaupetit, line 5: I am concerned that there is no license associated with this project. It has been removed while undeprecating the original project: https://github.com/lamoda/django-sentinel/commit/31d0d7d40f016a16b8995fa01eb85080ae7f6a09

docker-compose.yml
```diff
@@ -53,3 +54,25 @@ services:
...
+
+  redis-replica1:
+    image: docker.io/bitnami/redis:6.0-debian-10
+    environment:
+      - ALLOW_EMPTY_PASSWORD=yes
+
+  redis-replica2:
+    image: docker.io/bitnami/redis:6.0-debian-10
+    environment:
+      - ALLOW_EMPTY_PASSWORD=yes
```
madmatah, line 78: Some environment variables are missing here :
 - `REDIS_MASTER_HOST`
 - `REDIS_REPLICATION_MODE`

docker-compose.yml
```diff
@@ -53,3 +54,25 @@ services:
...
+
+  redis-sentinel:
+    image: docker.io/bitnami/redis-sentinel:6.0-debian-10
+    depends_on:
+      - redis-primary
+      - redis-replica1
+      - redis-replica2
+
+  redis-primary:
+    image: docker.io/bitnami/redis:6.0-debian-10
```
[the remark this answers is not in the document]
madmatah: Because the bitnami image simplifies the setup of a redis cluster (with the environment variable `REDIS_REPLICATION_MODE` that I forgot to set here... :man_facepalming:  )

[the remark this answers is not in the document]
jmaupetit: ok then

src/richie/apps/core/cache.py
madmatah: Sorry, I misinterpreted a part of your comment !
I set `__name__` as default value here instead of `False`.

tests/apps/core/test_cache.py
madmatah: I don't know why the LICENSE file has been removed, but I don't think it's a problem because :

- The original project code is under MIT licence 
- The new python package (`django-sentinel`) is referenced on PyPi with MIT license (see [here](https://pypi.org/project/django-sentinel/) and [here](https://github.com/lamoda/django-sentinel/blob/673696d0c5826bdca299662f1d22403d08476321/setup.py#L35))

madmatah:
@sampaccoud @jmaupetit Thank you very much for your feedbacks, I think I addressed all of them.
I kept some conversations as `Unresolved` to let you tell me if it's ok for you.

If you have time for a final review that would be great, thanks ! :pray:

setup.cfg
sampaccoud: I didn't realize it was just a dev dependency. Good with me.

tests/apps/core/test_cache.py
jmaupetit: As stated by the `setup.cfg` file, we can consider the forked project license as still MIT-based.

PR #1093, opened 2020-08-13
🐛(docker) fix inconsistent sentinel state when running tests

madmatah:
## Purpose

The `make test-back` command fails randomly with the following errors : 

  - `No master found for mymaster`
 - `Error 111 connecting to x.x.x.x:6379. Connection refused.`

This happens for the following reasons:
 - when containers are restarted by `docker-compose`, their IP change 
 - the `redis-sentinel` container is persisting the redis master IP across restarts

As a result, sentinel sometimes try to monitor the wrong container (i.e. postgresql) as the redis master and constantly fail.

## Proposal

- [x] Recreate the `redis-sentinel` container before running the tests to ensure it has a consistent state.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -35,7 +35,7 @@ services:
     depends_on:
       - "${DB_HOST:-postgresql}"
       - "elasticsearch"
-      - redis-sentinel
+
     stdin_open: true
     tty: true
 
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -35,7 +35,7 @@ services:
     depends_on:
       - "${DB_HOST:-postgresql}"
       - "elasticsearch"
-      - redis-sentinel
```
sampaccoud, line 38: If you do this, is sentinel going to start when we `make run`?

madmatah: I removed the dependency because it is only used in Test environment.
We don't want it when we `make run`.

PR #1094, opened 2020-08-13
✨(sandbox) allow to configure cache settings with environment variables

madmatah:
## Purpose

We want to be able to configure django cache settings with environment variables.

## Proposal

- [x] use django-configurations for cache settings
- [x] set default values that match the previous sandbox settings to prevent introducing a breaking change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 - Add Linkedin badge in social networks templates.
 - Add new section "septenary" with a new wave decoration.
 - Add redis sentinel cache backend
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 - Add Linkedin badge in social networks templates.
 - Add new section "septenary" with a new wave decoration.
 - Add redis sentinel cache backend
+- Allow to configure cache settings with environment variables
```
sampaccoud, line 16:
```suggestion
- Allow to configure cache and session settings with environment variables
```

madmatah: Thanks :pray:

PR #1113, opened 2020-09-02
🔧(prettier) add default rules

jbpenrath:
## Purpose
In my VS Code configuration, Prettier auto format code on save. But when I would like to modify my first django template, there was an exception due to Prettier added line break at the wrong places.

Furthermore, to avoid other surprise like these and to homogenize coding style, I just add few rules in the prettierrc based on [Airbnb](https://github.com/airbnb/javascript) and [Google](https://google.github.io/styleguide/jsguide.html) JS Style guides:
- 100 characters per line
- 2 spaces indentation
- Single quote

## Proposal
- [x] Homogenize richie prettier rules
- [x] Prettier must ignore django html templates

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,3 @@
+# Because django templates requires syntax which can't contain line break
+# Prettier must ignore all Django templates files
+src/richie/**/*.html
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,2 +1,5 @@
+printWidth: 100
 singleQuote: true
+tabWidth: 2
 trailingComma: all
+useTabs: false
--- a/src/frontend/js/common/searchFields/getSuggestionsSection.spec.ts
+++ b/src/frontend/js/common/searchFields/getSuggestionsSection.spec.ts
@@ -19,11 +19,7 @@ describe('common/searchFields/getSuggestionsSection', () => {
 
     let suggestionsSection;
     try {
-      suggestionsSection = await getSuggestionsSection(
-        'courses',
-        'Courses',
-        'some search',
-      );
+      suggestionsSection = await getSuggestionsSection('courses', 'Courses', 'some search');
     } catch (error) {
       fail('Did not expect getSuggestionsSection to fail');
     }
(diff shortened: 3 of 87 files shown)


lunika:
Not having `.yml` extension is valid as explain in the documentation https://prettier.io/docs/en/configuration.html

jbpenrath:
> Not having `.yml` extension is valid as explain in the documentation https://prettier.io/docs/en/configuration.html
> 
> > I also encount an error if the .prettierrc does not have .yml extension
> 
> What is this error ?

Without explicit extension, my prettier extension try to use json parser. Yet, I agree, the doc is clear about that, we could use json or yml without explicit extension. I searched in my VS Code configuration if I can set the parser somewhere but I did'nt found something interesting.

.prettierrc.yml
```diff
@@ -0,0 +1,5 @@
+printWidth: 80
```
lunika, line 1: In your commit message you suggest 100:

> I propose to use the Airbnb standard which recommends to use 100 characters per line.

jbpenrath: Honestly, I prefer 100 characters per line, 80 characters is a little bit strict sometimes. But I saw that in the richie codebase, the rule of 80 characters is used so I change it afterwards.

lunika:
You should merge your two commits IMO and keep the first one

.prettierignore
```diff
@@ -0,0 +1,3 @@
+# Because django templates requires syntax which can't contain line break
+# Prettier must ignore all Django templates files
+src/richie/**/*.html
```
lunika, line 3: Add a new line at the end of the file please

.prettierignore
```diff
@@ -0,0 +1,3 @@
+# Because django templates requires syntax which can't contain line break
```
lunika, line 1: In your commit message:

`s/s it recommended in python to respect line width of 79 characters,/It is recommened in Python to respect a line width of 79 characters,/`
`s/I propose to use the Airbnb standard/I suggest to use the Airbnb standard/`

lunika (approved):

PR #1114, opened 2020-09-02
🔧(cms) limit number of plugin in course description to one

sampaccoud:
## Purpose

There is a limit on the number of characters that this placeholder can receive, which only makes sense if the number of plugins is
also limited to one.

## Proposal

Limit number of text plugins to 1 in the course description placeholder.


jbpenrath (approved):
It's well done 😃

Issue #1103, opened 2020-08-24
Make scrolling smoother on the search page


sampaccoud:
## Feature Request

**Is your feature request related to a problem or unsupported use case? Please describe.**
When changing page on the search page, the scroll-up happens at the same time as the switch to new results, which flickers confusingly. In case the request for the new page takes time to get a response, the switch to new results can even happen after the scroll-up is finished which is even more confusing.

**Describe the solution you'd like**
Wait for the response to the new page request before triggering the scroll-up.

mbenadda:
Closed by #1117

PR #1117, opened 2020-09-04
🚸(Search) scroll after course fetching

jbpenrath:
## Purpose

Currently, on the course search page, when user filters courses, UI scrolls to top directly on filter change. But in some case (e.g: if the network is slow), new courses has not been yet retrieved and it causes flickering that may be disappointing for the user.

Related to #1103 

## Proposal
- [x] Trigger scroll up after courses have been retrieved

### DIFF
--- a/src/frontend/js/components/Search/index.spec.tsx
+++ b/src/frontend/js/components/Search/index.spec.tsx
@@ -1,27 +1,27 @@
-import { fireEvent, render, screen, waitFor } from '@testing-library/react';
+import React from 'react';
+import { act, fireEvent, render, screen, waitFor } from '@testing-library/react';
 import fetchMock from 'fetch-mock';
 import { stringify } from 'query-string';
-import React from 'react';
 import { IntlProvider } from 'react-intl';
 
 import { History, HistoryContext } from 'data/useHistory';
 import { Search } from '.';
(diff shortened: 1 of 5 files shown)


src/frontend/js/components/Search/index.spec.tsx
```diff
@@ -177,4 +181,52 @@ describe('<Search />', () => {
       );
     }
   });
+
+  it('should scroll up when filter have changed and courses are retrieved', async () => {
```
mbenadda, line 185:
```suggestion
  it('should scroll up when filters have changed and courses are retrieved', async () => {
```

src/frontend/js/components/Search/index.spec.tsx
```diff
@@ -1,27 +1,28 @@
+import React from 'react';
 import { fireEvent, render, screen, waitFor } from '@testing-library/react';
+import { act } from 'react-dom/test-utils';
```
mbenadda, line 3: Just a nitpick as it's really the same function, but we generally use the [`act` from react-testing-library](https://testing-library.com/docs/react-testing-library/api#act) for concision (one less import).

mbenadda (approved):
Looks good to me 👍

Just a small typo and a nitpick!

PR #1118, opened 2020-09-04
🐛 Use variables for language selector colors

mbenadda:
## Purpose

When the language selector is displayed in the footer/header, it must adapt to its surroundings by using appropriate colors. (otherwise it might appear black on black, or white on white). This mostly goes for the text, which appears directly on the background of the footer/header.

## Proposal

To achieve this, we can use variables from sass theming when defining language selector colors. This is only used for the button which is integrated into its parent element. The dropdown itself has its own background.


sampaccoud (approved):
seems legit

PR #1123, opened 2020-09-09
🔧(frontend): switch from tslint to eslint

jbpenrath:
## Purpose
tslint is deprecated since 2019 and typescript eslint is the main project to lint typescript project. As Eslint is the main linter in javascript world and it tooks flagship in typescript since palentir announced the deprecation of tslint, it could be interesing to switch of.

## Proposal

- [x] Switch from tslint to eslint
- [x] Make new configuration iso with the previous one
I used [eslint-config-airbnb-typescript](https://github.com/iamturns/eslint-config-airbnb-typescript) as a base and disabled rules which trigger error on our codebase and seems too strict. Then I propose to keep some rules enabled which do not pass currently because I think they help us to improve quality of code, readability and identify a11y mistakes. These rules are listed below.

- [x] Propose new lint rules and update codebase accordingly
    - [@typescript-eslint/naming-convention](https://github.com/typescript-eslint/typescript-eslint/blob/v4.1.0/packages/eslint-plugin/docs/rules/naming-convention.md)
    - [@typescript-eslint/no-unused-vars](https://github.com/typescript-eslint/typescript-eslint/blob/v4.1.0/packages/eslint-plugin/docs/rules/no-unused-vars.md)
    - [import/first](https://github.com/benmosher/eslint-plugin-import/blob/v2.22.0/docs/rules/first.md)
    - [import/order](https://github.com/benmosher/eslint-plugin-import/blob/v2.22.0/docs/rules/order.md)
    - [jsx-a11y/anchor-is-valid](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/master/docs/rules/anchor-is-valid.md)
    - [jsx-a11y/label-has-associated-control](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/master/docs/rules/label-has-associated-control.md)
Only warn but for a11y, a label must be associated to a control through `htmlFor` prop.
    - [jsx-a11y/no-static-element-interactions](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/master/docs/rules/no-static-element-interactions.md)
    - [no-empty-pattern](https://eslint.org/docs/rules/no-empty-pattern)
    - [no-nested-ternary](https://eslint.org/docs/rules/no-nested-ternary)
I only warn when this rule matched. I think nested ternary is a compact syntax but it is difficult to understand. So we accept it for compactness but linter will warn you to remember to try to find a more readable syntax.
    - [no-path-concat](https://eslint.org/docs/rules/no-path-concat)
    - [prefer-destructuring](https://eslint.org/docs/rules/prefer-destructuring)
    - [react/jsx-curly-brace-presence](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-curly-brace-presence.md)
    - [react/no-array-index-key](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-array-index-key.md)
Only warn, because sometimes it may be tricky to use something else than index but in general it avoids some problem of duplicate keys all over the app.
    - [react/self-closing-comp](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/self-closing-comp.md)


CHANGELOG.md
```diff
@@ -14,6 +14,8 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Changed
```
lunika, line 16: In your commit message use 🚨 gitmoji

src/frontend/package.json
```diff
@@ -55,6 +55,8 @@
     "@types/react-autosuggest": "10.0.0",
     "@types/react-dom": "16.9.8",
     "@types/react-modal": "3.10.6",
+    "@typescript-eslint/eslint-plugin": "^4.1.0",
```
lunika, line 58: Dependencies are always pin in our project. You must set a specific version. We have a bot dependency responsible to update them.

src/frontend/js/components/PaginateCourseSearch/index.tsx
```diff
@@ -123,7 +123,9 @@ export const PaginateCourseSearch = ({ courseSearchTotalCount }: PaginateCourseS
                 </li>
               ) : (
                 <li className="pagination__item">
-                  <a
```
mbenadda, line 126: I feel like this might be wrong.

Technically, the interaction triggers a JS handler, which infringes on the lint rule.

In practice however, clicking on a pagination is semantically closer to switching pages than doing an "action" with a button. IMO this should be a link, not a button.

src/frontend/js/components/SearchFiltersPane/index.tsx
```diff
@@ -52,10 +52,11 @@ export const SearchFiltersPane = ({
       <h2 className="search-filters-pane__title">
         <FormattedMessage {...messages.filter} />
       </h2>
-      <a
+      <span
```
mbenadda, line 55: Maybe this could just be a `button` with styles that make it look plain then?

(unlike pagination I agree this one might be better as a button)

src/frontend/js/components/UserLogin/index.tsx
```diff
@@ -73,11 +73,12 @@ export const UserLogin = ({ loginUrl, logoutUrl, signupUrl }: UserLoginProps) =>
 
   return (
     <div className="user-login">
-      {user === undefined ? (
+      {user === undefined && (
```
mbenadda, line 76: It seems to me that by writing it like this we're losing the meaning that only one of those three blocks should be shown at any time. If we add a condition to one of these it's easier to create a bug.

Maybe it's not that important. Not sure which is better, I wish javascript `if` statements were expressions :(

src/frontend/js/types/index.ts
```diff
@@ -20,5 +20,4 @@ export interface CourseState {
  * Use an empty type to make sure we do not depend on any LMS-specific fields
  * on enrollment objects, just use HTTP response codes.
  */
-// tslint:disable-next-line
```
mbenadda, line 23: 👏

src/frontend/js/components/PaginateCourseSearch/index.tsx
jbpenrath: Yep, in the case of pagination its closer but a `<a>` without `href` attribute is not something valid from W3C standard and I agree with your comment below, the best element for this case is a button. I will use a button instead of a span and update styles accordingly.

mbenadda: > a <a> without href attribute is not something valid from W3C standard

Right, I forgot about this. I guess we'd need a router to make it a real link. Let's go with button then 👍

src/frontend/js/components/PaginateCourseSearch/index.tsx
```diff
@@ -123,7 +123,8 @@ export const PaginateCourseSearch = ({ courseSearchTotalCount }: PaginateCourseS
...
+                  <button
+                    tabIndex={index}
```
mbenadda, line 127: We're not supposed to use tabIndex with positive values at all. Here the effect is those buttons are the first items in the tab order before anything else on the page, which I don't think is what we want.

I think this article sums up what I know about tab indexes pretty well: https://bitsofco.de/how-and-when-to-use-the-tabindex-attribute/

mbenadda (approved):

PR #1131, opened 2020-09-17
Improvement/sso

jbpenrath:
## Purpose

Since we use sso on Richie through edx oauth2 provider, some mechanisms have to be improved:
- Sign up link targets django admin login page. Instead it should redirect on OAuth2 provider registration page.
- When the user logs out on Richie, he is still authenticated on LMS_BACKENDS. A logout on Richie should trigger log out on LMS_BACKENDS.
- If the user is already authenticated on OAuth2 provider, it should be also authenticated on Richie.

## Proposal

- [x] Add a registration oauth2 endpoint to be able to customize sign-up url.

- [x] If user is not authenticated on Richie, check if he is authenticated from OAuth2 provider. If yes, call automatically OAuth2 authorize endpoint.

- [x] Create a logout view which logs out user from all LMS_BACKENDS linked to richie.


src/richie/apps/core/context_processors.py
```diff
@@ -32,6 +33,12 @@ def site_metas(request):
             "domain": site_current.domain,
             "web_url": f"{protocol:s}://{site_current.domain:s}",
         },
+        "SOCIAL": {
+            "registration_endpoint": getattr(
+                settings, "SOCIAL_AUTH_EDX_OAUTH2_REGISTRATION_ENDPOINT", ""
+            )
+            or reverse("admin:index")
```
slandrault, line 40: Last parameter of `getattr` is the default value, so maybe write it like
```python
getattr(
    settings, "SOCIAL_AUTH_EDX_OAUTH2_REGISTRATION_ENDPOINT", reverse("admin:index")
)
```
to remove the `or` statement and simplify your code?

jbpenrath: This syntax is less elegant but it is more resilient. In fact, getattr return default value only if the property does not exist. In the case of this settings exists, it do not return the default value.

By using `or reverse(...)`, we prevent this case.

src/frontend/js/data/useSession/index.tsx
```diff
@@ -0,0 +1,77 @@
+import React, { createContext, PropsWithChildren, useCallback, useContext, useState } from 'react';
```
lunika, line 1: You should add a module comment explaining what is this hook, why we need it, what you are doing, etc.

src/frontend/js/utils/test/factories.ts
```diff
@@ -29,3 +30,27 @@ export const UserFactory = createSpec({
   full_name: faker.fake('{{name.firstName}} {{name.lastName}}'),
   username: faker.internet.email(),
 });
+
+interface CommonDataProps {
```
lunika, line 34: Why not using the interface from `src/frontend/js/types/commonDataProps.ts` ?

src/frontend/js/types/commonDataProps.ts
```diff
@@ -6,8 +6,10 @@ import { Nullable } from 'utils/types';
  */
 export interface CommonDataProps {
   context: {
+    allowed_hosts: string[];
```
lunika, line 9: I don't find where you are using it

src/richie/apps/core/context_processors.py
```diff
@@ -32,13 +33,23 @@ def site_metas(request):
...
             {
                 "context": {
                     "csrftoken": get_token(request),
                     "environment": getattr(settings, "ENVIRONMENT", ""),
                     "release": getattr(settings, "RELEASE", ""),
                     "sentry_dsn": getattr(settings, "SENTRY_DSN", ""),
+                    "oauth2_whoami": getattr(
+                        settings, "SOCIAL_AUTH_EDX_OAUTH2_WHOAMI_ENDPOINT", None
+                    ),
+                    "allowed_hosts": getattr(settings, "ALLOWED_HOSTS", []),
```
lunika, line 52: I don't know if it's something safe to expose this setting.

src/frontend/js/utils/isSafeURI.ts
```diff
@@ -0,0 +1,66 @@
+import { location } from 'utils/indirection/window';
+
+const REDIRECT_WHITELIST =
+  (window as any).__richie_frontend_context__?.context?.redirect_whitelist || [];
```
lunika, line 4: I don't find where you add this property in the context

docker-compose.yml
```diff
@@ -1,10 +1,10 @@
-version: "3.5"
+version: '3.5'
```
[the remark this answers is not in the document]
jbpenrath: My Prettier is a little bit intrusive 🤔

src/frontend/jest.config.js
```diff
@@ -10,4 +10,5 @@ module.exports = {
   testMatch: [`${__dirname}/js/**/*.spec.+(ts|tsx|js)`],
   testURL: 'https://localhost',
   transformIgnorePatterns: ['node_modules/(?!(lodash-es)/)'],
+  coverageDirectory: '.coverage',
```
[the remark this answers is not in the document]
jbpenrath: This directory is already ignored by git. So instead to add a new directory in `.gitignore`, I preferred to update the location to store jest coverage in an already ignored directory.
```
# Unit test / coverage reports
htmlcov/
.coverage
.cache
.pytest_cache
nosetests.xml
coverage.xml
```

src/frontend/js/components/UserLogin/index.tsx
```diff
@@ -35,41 +35,54 @@ interface UserLoginProps {
   loginUrl: string;
   logoutUrl: string;
   signupUrl: string;
+  context: { [key: string]: any };
```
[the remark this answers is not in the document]
jbpenrath: You agree, I pass `oauth2_whoami` directly in props.

src/frontend/js/types/commonDataProps.ts
jbpenrath: It's a mistake. I forgot to rename `allowed_hosts` into `redirect_whitelist`

src/richie/apps/core/context_processors.py
jbpenrath: As mentionned above, I will remove `oauth2_whoami` endpoint from the context to pass it directly in props.

src/frontend/js/utils/isSafeURI.ts
jbpenrath: Related to the mistake in `src/richie/apps/core/context_processors.py`

PR #1132, opened 2020-09-17
Demo site: allow removing licences placeholders and customizing pages info

sampaccoud:
## Purpose

Some customizations that we make in themed sites using richie (cf https://github.com/openfun/richie-site-factory) are breaking the demo site. It makes it difficult to work on such sites in development if we can't generate a demo site.

## Proposal

- [x] Allow deactivating the licences placeholders without breaking the demo site
- [x] Allow customizing pages info (e.g to create a page linked to a button via its reverse id)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Changed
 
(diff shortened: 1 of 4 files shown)


jbpenrath (approved):
Good To Merge

PR #1134, opened 2020-09-18
💄(frontend) fix selector arrow color in footer

jbpenrath:
## Purpose
The language selector arrow in the footer has not the good color.
`.language-selector` was renamed into `.selector` and Footer styles had not been updated accordingly.

## Proposal
- [x] Apply `color: currentColor` to `.selector__icon`. In this way, arrow color herits of the parent color. Less code to write, easier to maintain.

### DIFF
--- a/src/frontend/scss/components/_footer.scss
+++ b/src/frontend/scss/components/_footer.scss
@@ -186,15 +186,11 @@
     display: flex;
     align-items: center;
 
-    .language-selector {
+    .selector {
       color: r-theme-val(body-footer, base-color);
 
       &__button {
         padding-left: 0;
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Fixed
 
+- Fix arrow color of language selector in footer
```
sampaccoud, line 18: You probably shouldn't add this line if the bug was introduced since the latest release...

mbenadda (approved):
GTM

PR #1139, opened 2020-09-22
🐛 (frontend) handle error invalid configuration

jbpenrath:
## Purpose

The path to get context from __richie_frontend_context__ was wrong so context variable was always undefined. In this way, Sentry was never initialize.

## Proposal

- [x] Use the correct path to get context data from `__richie_frontend_context__`

### DIFF
--- a/src/frontend/js/utils/errors/handle.spec.ts
+++ b/src/frontend/js/utils/errors/handle.spec.ts
@@ -1,0 +1,27 @@
+/**
+ * @jest-environment jsdom
+ */
+import * as Sentry from '@sentry/browser';
+
+jest.mock('@sentry/browser');
+
+describe('handle', () => {
+  (window as any).__richie_frontend_context__ = {
+    context: {
+      sentry_dsn: 'https://sentry.richie',
+      environment: 'test',
+    },
+  };
+
+  const { handle } = require('./handle');
+
+  it('should initialize sentry', () => {
+    expect(Sentry.init).toBeCalledTimes(1);
+    expect(Sentry.configureScope).toBeCalledTimes(1);
+  });
+
+  it('should report error to sentry', () => {
+    handle('An error for test');
+    expect(Sentry.captureException).toBeCalledTimes(1);
+  });
+});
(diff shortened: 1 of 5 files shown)


src/frontend/js/utils/errors/handle.spec.ts
```diff
@@ -0,0 +1,27 @@
+/**
+ * @jest-environment jsdom
+ */
+import * as Sentry from '@sentry/browser';
+
+jest.mock('@sentry/browser');
+
+describe('handle', () => {
+  (window as any).__richie_frontend_context__ = {
```
sampaccoud, line 9: In the absence of e2e tests, maybe add a test to the backend checking that this is injected to the dom? With a comment in the test to modify frontend before fixing the test?

jbpenrath: This test already exist in `tests/app/core/test_pages.py`

```py
    @override_settings(SENTRY_DSN="https://example.com/sentry/dsn")
    @override_settings(RELEASE="9.8.7")
    @override_settings(ENVIRONMENT="test_pages")
    def test_page_includes_frontend_context(self):
        """
        Create a page and make sure it includes the frontend context as included
        in `base.html`.
        """
        page = PageFactory(should_publish=True, template="richie/single_column.html")
        response = self.client.get(page.get_public_url())
        self.assertContains(response, '"environment": "test_pages"')
        self.assertContains(response, '"release": "9.8.7"')
        self.assertContains(response, '"sentry_dsn": "https://example.com/sentry/dsn"')

```

sampaccoud: :+1: Then just add the warning about modifying this test?

tests/apps/core/test_pages.py
```diff
@@ -44,6 +44,9 @@ def test_page_includes_frontend_context(self):
         """
         Create a page and make sure it includes the frontend context as included
         in `base.html`.
+
+        ⚠️ If this test fails, before fix it, identify if this change had have
```
mbenadda, line 48:
```suggestion
        ⚠️ If this test fails, before fixing it, identify if this change has had
```

mbenadda (approved):

PR #1149, opened 2020-09-30
✨(back) add LimitBrowserCacheTTLHeaders middleware

madmatah:
## Purpose

By default, DjangoCMS sends cached responses with a `Cache-control: max-age` value reflecting the server cache TTL.

The thing is : the server side cache entry get invalidated each time a user publishes new content to the site, but we cannot invalidate our client browser cache entries. That's why we want to set a long TTL on the server side, but a much lower TTL on the browser cache.

## Proposal

- [x] Add LimitBrowserCacheTTLHeaders middleware

### DIFF
--- a/src/richie/apps/core/cache.py
+++ b/src/richie/apps/core/cache.py
@@ -10,11 +10,15 @@ import random
 
 from django.conf import settings
 from django.core.exceptions import ImproperlyConfigured
+from django.utils.cache import get_max_age, patch_response_headers
+from django.utils.deprecation import MiddlewareMixin
 
 from django_redis.client import DefaultClient
 from redis.sentinel import Sentinel
 
 DJANGO_REDIS_LOGGER = getattr(settings, "DJANGO_REDIS_LOGGER", __name__)
+
+logger = logging.getLogger(__name__)
 
 
 class SentinelClient(DefaultClient):
@@ -131,3 +135,40 @@ class SentinelClient(DefaultClient):
         del self._client_read
         self._client_write = None
         self._client_read = None
+
+
+class LimitBrowserCacheTTLHeaders(MiddlewareMixin):
+    """
+    This middleware allows to define a maximum cache timeout for the cache
+    response headers (Cache-control: max-age and Expires).
+
+    It can be configured with the MAX_BROWSER_CACHE_TTL setting.
+    This value is expressed in seconds.
+
+    LimitBrowserCacheTTLHeaders should be placed at the beginning of the
+    MIDDLEWARE list, so that it'll get called last during the response phase.
+    """
+
+    # pylint: disable=no-self-use
+    def process_response(self, request, response):
+        """
+        Rewrite the "Cache-control" and "Expires headers" in the response
+        if needed.
+        """
+        try:
+            max_ttl = int(getattr(settings, "MAX_BROWSER_CACHE_TTL", None))
+            if max_ttl < 0:
+                raise ValueError("MAX_BROWSER_CACHE_TTL must be a positive integer")
+        except (ValueError, TypeError) as err:
+            logger.error(err)
+            return response
+
+        max_age = get_max_age(response)
+        if max_age is not None and max_age > max_ttl:
+            if response.has_header("Expires"):
+                # Remove the Expires response Header because patch_response_headers()
+                # adds it only if it isn't already set
+                del response["Expires"]
+            patch_response_headers(response, cache_timeout=max_ttl)
+
+        return response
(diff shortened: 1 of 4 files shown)


src/richie/apps/core/cache.py
```diff
@@ -131,3 +133,35 @@ def close(self, **kwargs):
...
+
+    # pylint: disable=no-self-use
+    def process_response(self, request, response):
+        """
+        Rewrite the Cache-control and Expires headers in the response
+        if needed.
+        """
+        max_ttl = getattr(settings, "MAX_BROWSER_CACHE_TTL", None)
+        if max_ttl is None:
+            return response
```
sampaccoud, line 158: Maybe cover also the case of invalid values?
```suggestion
try:
    max_ttl = int(max_ttl)
except Exception as err:
    logger.error(err)
    return response
```

And add a test for this case?

src/richie/apps/core/cache.py
```diff
@@ -131,3 +133,35 @@ def close(self, **kwargs):
...
+        max_age = get_max_age(response)
+        if max_age is not None and max_age > max_ttl:
+            if response.has_header("Expires"):
+                # Remove the Expires response Header because patch_response_headers()
+                # adds it only if it isn't already setI
```
sampaccoud, line 163:
```suggestion
                # adds it only if it isn't already set
```

tests/apps/core/test_cache.py
```diff
@@ -407,3 +414,93 @@ def test_version(self):
...
+            return response
+
+        with self.settings(MAX_BROWSER_CACHE_TTL=-10):
+            middleware = LimitBrowserCacheTTLHeaders(response_builder)
+            response = middleware(HttpRequest())
+            expected_expires = http_date(time.time() + 3600)
+            self.assertEqual(expected_expires, response["Expires"])
+            self.assertIn("max-age=3600", response["Cache-Control"])
+
+        with self.settings(MAX_BROWSER_CACHE_TTL="pouet"):
```
sampaccoud, line 501: :thinking:

sampaccoud (approved):

madmatah:
@sampaccoud Thanks for your feedbacks !
I removed 2 unneccessary lines after your last review, i'm ready to merge if it's ok for you :slightly_smiling_face:

src/richie/apps/core/cache.py
```diff
@@ -131,3 +135,41 @@ def close(self, **kwargs):
...
+        """
+        Rewrite the "Cache-control" and "Expires headers" in the response
+        if needed.
+        """
+        max_ttl = getattr(settings, "MAX_BROWSER_CACHE_TTL", None)
+        try:
+            max_ttl = int(max_ttl)
+            if max_ttl < 0:
+                raise ValueError("MAX_BROWSER_CACHE_TTL must be a positive integer")
+        except (ValueError, AttributeError, TypeError) as err:
```
sampaccoud, line 163: can you really get all these errors int casting?

madmatah: The `TypeError` and `ValueError` are raised by `int()`
And the `AttributeError` was for the  `getattr` that I forgot to move in the `try:` block :man_facepalming:

madmatah: Ok, here is the fixed version, with another test for the case where `MAX_BROWSER_CACHE_TTL` is not defined.

sampaccoud (approved):

src/richie/apps/core/cache.py
```diff
@@ -131,3 +135,40 @@ def close(self, **kwargs):
...
+    # pylint: disable=no-self-use
+    def process_response(self, request, response):
+        """
+        Rewrite the "Cache-control" and "Expires headers" in the response
+        if needed.
+        """
+        try:
+            max_ttl = int(getattr(settings, "MAX_BROWSER_CACHE_TTL", None))
+            if max_ttl < 0:
+                raise ValueError("MAX_BROWSER_CACHE_TTL must be a positive integer")
```
sampaccoud, line 161: @madmatah I'm now wondering if it would not have been better to use 600 as default when it is not defined in settings...

madmatah: IMO this is an opinionated default value. It should be set with full knowledge of the facts.
Maybe, doing nothing silently would have been better ?

sampaccoud: I think no because you would think that it works. My point of view was that it is no so much opionated as it is a CMS : for everyone, it is the time elapsed between a publication and the moment everyone sees it for sure. 10 minutes seemed a good default for that?

PR #1151, opened 2020-10-02
Text on program list page and ckeditor on home

sampaccoud:
## Purpose

We want to allow creating content on the programs list page.
We also need to be more lax and allow the default CKEditor configuration for all main content placeholders.

## Proposal

- [x] Use default CKEditor config for main content placeholders
- [x] Add a main content placeholder to the programs list template

### DIFF
--- a/src/frontend/scss/components/templates/courses/cms/_program_list.scss
+++ b/src/frontend/scss/components/templates/courses/cms/_program_list.scss
@@ -1,11 +1,21 @@
 .program-list {
   @include make-container-max-widths;
-  display: flex;
   margin: 0 auto;
   padding: 1rem 0;
-  flex-direction: row;
-  flex-wrap: wrap;
-  @if $body-padding-fix {
-    @include content-padding-fix;
+
+  &__block {
+    @include detail-block;
+    @if $body-padding-fix {
+      @include content-padding-fix($target: '&:last-child');
+    }
+  }
+
+  &__row {
+    @include detail-row;
+  }
+
+  &__maincontent {
+    padding-left: $grid-gutter-width;
+    padding-right: $grid-gutter-width;
   }
 }
--- a/src/richie/apps/courses/templates/courses/cms/program_list.html
+++ b/src/richie/apps/courses/templates/courses/cms/program_list.html
@@ -1,5 +1,5 @@
 {% extends "richie/fullwidth.html" %}
-{% load cms_tags i18n %}
+{% load cms_tags extra_tags i18n %}
 
 {% block subheader_content %}
 <div class="subheader__container">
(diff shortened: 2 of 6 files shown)


src/richie/apps/courses/templates/courses/cms/program_list.html
```diff
@@ -10,14 +10,26 @@ <h1 class="subheader__title">{{ current_page.get_title }}</h1>
...
+        </div>
         {% endif %}
-    {% empty %}
-        <p class="program-glimpse program-glimpse--empty">
-            {% trans "No associated programs" %}
-        </p>
-    {% endfor %}
+
+        <div class="program-list__programs program-list__row">
+            <section class="program-glimpse-list">
```
sveetch, line 21: From HTML spec, section element must have a title element (h1, h2, etc..) as a direct element (maybe even as first child). Commonly it is better to avoid using a section element because it have a strong semantical meaning except if you clearly know how to semantically structure your HTML.

src/richie/apps/courses/templates/courses/cms/program_list.html
```diff
@@ -10,14 +10,26 @@ <h1 class="subheader__title">{{ current_page.get_title }}</h1>
...
+                {% for page in current_page.get_child_pages %}
+                    {% if page.program %}
+                        {% include "courses/cms/fragment_program_glimpse.html" with program=page.program %}
+                    {% endif %}
+                {% empty %}
+                    <p class="program-glimpse program-glimpse--empty">
```
sveetch, line 27: The empty variant may not be correctly defined in your `program-list` it may not have a fine layout, i think there is a "empty" sass mixin somewhere you could use to do that.

src/richie/apps/courses/templates/courses/cms/program_list.html
sampaccoud: Good point :pray: I removed the section markup.

src/richie/apps/courses/templates/courses/cms/program_list.html
sampaccoud: I checked and it is well formatted. This element was already there and the css taking care of the empty state is already there.

src/frontend/scss/components/templates/courses/cms/_program_list.scss
```diff
@@ -1,11 +1,21 @@
...
+    @if $body-padding-fix {
+      @include content-padding-fix($target: '&:last-child');
+    }
+  }
+
+  &__row {
+    @include detail-row;
+  }
+
+  .program-list__maincontent {
```
jbpenrath, line 17:
```suggestion
  &__maincontent {
```

jbpenrath (approved):

src/frontend/scss/components/templates/courses/cms/_program_list.scss
sampaccoud: :+1:

PR #1156, opened 2020-10-07
💄(course) replace course-detail__run-cta 

2018D:
## Purpose

update fragment_course_run template to match new css class.

## Proposal

replace course-detail__run-cta in fragment_course_run template by course-run-enrollment__cta

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,11 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Fixed
+
+- Replace course-detail__run-cta in fragment_course_run template by
+  course-run-enrollment__cta
+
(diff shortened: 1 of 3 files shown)


jbpenrath (changes requested):
Hi @2018D 👋. Good catch!
But instead to restore `&__run-cta` styles in `_course_detail.scss`, you should fix class name in `fragment_course_run.html`.
In fact, since we developed React widget for enrollment, these styles have been moved in `.course-run-enrollment__cta` class.
So to fix this issue just rename `course-detail__run-cta` into `course-run-enrollment__cta` to avoid styles duplication.

2018D:
Hey @jbpenrath thank you for the quick feedback, i've amended the pr and the commit to fix it like you suggested.
Sorry for the long delay.

jbpenrath (approved):
Thank you to contribute @2018D ! 😉

PR #1159, opened 2020-10-15
🚸(plugins) improve sanitizing of input text on CKEditor plugin

sampaccoud:
## Purpose

CKEditor is submitting text with web entities (e.g. `&eacute;` for `é`) and potentially exotic characters. Our support team complained that the count of the max number of characters was wrong. 

## Proposal

This commit fixes it by normalizing and unescaping input strings. We do it for counting and also before saving it to database.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Fixed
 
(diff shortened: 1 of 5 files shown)


jbpenrath (approved):
I tried with a text of exactly 1200 characters no offset detected even with accentuated characters, bold, etc.
GTM

PR #1162, opened 2020-10-21
♻️ (frontend) remove usage of react-intl-po

jbpenrath:
## Purpose

Using react-intl-po with new version of react-intl was not working anymore. In order to remove the usage of react-intl-po we entirely our i18 workflow. Everythin is managed using react-intl cli tool. In a first step it extract messages, this messages are uploaded on crowdin and then compiled once translated.

Furthermore, we need to be allow to override translations from richie-site-factory, this refactoring is good place to put this mechanism.

## Proposal

- [x] remove usage of react-intl-po
- [x] make richie-education translations easily overridable


mbenadda (approved):
Nice job on the custom compile 👍

I think we really need a documentation section with an example of how to achieve it (maybe in [frontend overrides](https://richie.education/docs/next/frontend-overrides) ?)

docs/frontend-overrides.md
```diff
@@ -37,14 +36,70 @@ export interface CourseGlimpseProps {
...
 ```
 
 Then, your override needs to provide the same exports, explicitly a named `CourseGlimpseProps` interface and a named `CourseGlimpse` component.
 
 You also need to respect the assumptions made by other components that use your overridden version, if you are not overriding a root component.
 
 For example returning `null` might break a layout if the original component never returned such a value, etc. You also need to make sure to avoid conflict with the parameters accepted by the original component.
+
+## Override translation
+When you create an application based on richie, you can encountered two cases about translations:
```
mbenadda, line 52:
```suggestion
When you create an application based on richie, you can encounter two cases about translations:
```

docs/frontend-overrides.md
```diff
@@ -37,14 +36,70 @@ export interface CourseGlimpseProps {
...
+1. You created or overridden a react component and create new translation keys
```
mbenadda, line 53:
```suggestion
1. You created or overrode a react component and created new translation keys
```

docs/frontend-overrides.md
```diff
@@ -37,14 +36,70 @@ export interface CourseGlimpseProps {
...
+2. You just want to override a translation in a existing richie component
```
mbenadda, line 54:
```suggestion
2. You just want to override a translation in an existing richie component
```

docs/frontend-overrides.md
```diff
@@ -37,14 +36,70 @@ export interface CourseGlimpseProps {
...
+### Create new translation keys
+
+Once you created your new component with its translation keys, you have to extract them with the following command:
+```
+  formatjs extract './**/*.ts*' --ignore ./node_modules --ignore './**/*.d.ts' --out-file './i18n/frontend.json --id-interpolation-pattern '[sha512:contenthash:base64:6]' --format crowdin
+```
+
+This command extracts all translations defined in your typescript files then generates a `frontend.json` file in `i18n/` directory. This file is like a pot file, this is the base to create your translations in any language that you want.
+
+As `--format` option indicates, this command generate a file compatible with crowdin. If you want to customize this command to fit your needs, read the [formatjs/cli documentation](https://formatjs.io/docs/tooling/cli/).
```
mbenadda, line 67:
```suggestion
As `--format` option indicates, this command generates a file compatible with crowdin. If you want to customize this command to fit your needs, read the [formatjs/cli documentation](https://formatjs.io/docs/tooling/cli/).
```

docs/frontend-overrides.md
```diff
@@ -37,14 +36,70 @@ export interface CourseGlimpseProps {
...
+
+Once translations keys are extracted and your local translation are ready, you need to compile these translations. In fact, compilation process first aggregates all translation files found from provided paths then merges them with richie translations according their filename and finally generates an output formatted for `react-intl`. Below, here an example of compilation command:
```
mbenadda, line 69:
```suggestion
Once translations keys are extracted and your local translations are ready, you need to compile these translations. In fact, the compilation process first aggregates all translation files found from provided paths then merges them with richie translations according their filename and finally generates an output formatted for `react-intl`. Below, here is an example of a compilation command:
```

docs/frontend-overrides.md
```diff
@@ -37,14 +36,70 @@ export interface CourseGlimpseProps {
...
+
+```
+  node-modules/richie-education/i18n/compile-translations.js ./i18n/locales/*.json
+```
+
+This command looks for all translation files in `i18n/locales` directory then merges files found with richie translation files. You can pass several path patterns. You can also use an `--ignore` argument to ignore a particular path.
+
+### Override an existing translation keys
+
+As explain above, compilation process aggregates translations files then **merges them according their filename**. That means if you want override for example english translation, you just have to create a `en-US.json` file and redefine translation key used by Richie.
```
mbenadda, line 79:
```suggestion
As explain above, the compilation process aggregates translations files then **merges them according their filename**. That means if you want override for example the english translation, you just have to create a `en-US.json` file and redefine translation keys used by Richie.
```

docs/frontend-overrides.md
```diff
@@ -37,14 +36,70 @@ export interface CourseGlimpseProps {
...
+
+Richie uses one file per language, currently, there 4 languages supported:
+
+- English: filename is `en-US.json`
+- French: filename is `fr-FR.json`
+- Canadian french: filename is `fr-CA.json`
+- Spanish: filename is `es-ES.json`
+
+
+For example, richie uses the translation key `components.UserLogin.logIn` for the Log in button. If you want to change this label for english language, you just have to create a translation file `en-US.json` which redefines this translation key:
```
mbenadda, line 89:
```suggestion
For example, richie uses the translation key `components.UserLogin.logIn` for the Log in button. If you want to change this label for the english translation, you just have to create a translation file `en-US.json` which redefines this translation key:
```

docs/frontend-overrides.md
```diff
@@ -37,14 +36,70 @@ export interface CourseGlimpseProps {
...
+  },
+}
+```
+
+Then, for example if you put your overridden translation in i18n/overrides directory, you have to launch the compilation command below:
+```
+  node-modules/richie-education/i18n/compile-translations.js ./i18n/overrides/*.json
+```
+
+In this way, "_Authentication_" will be display as label for login button instead of "_Sign in_".
```
mbenadda, line 105:
```suggestion
In this way, "_Authentication_" will be displayed as label for login button instead of "_Sign in_".
```

mbenadda (approved):
Good job on the docs! 👍

docs/frontend-overrides.md
sampaccoud: s/two cases/two situations?

docs/frontend-overrides.md
```diff
@@ -37,14 +36,70 @@ export interface CourseGlimpseProps {
...
+
+
+### Create new translation keys
+
+Once you created your new component with its translation keys, you have to extract them with the following command:
+```
+  formatjs extract './**/*.ts*' --ignore ./node_modules --ignore './**/*.d.ts' --out-file './i18n/frontend.json --id-interpolation-pattern '[sha512:contenthash:base64:6]' --format crowdin
+```
+
+This command extracts all translations defined in your typescript files then generates a `frontend.json` file in `i18n/` directory. This file is like a pot file, this is the base to create your translations in any language that you want.
```
sampaccoud, line 65:
```suggestion
This command extracts all translations defined in your typescript files then generates a `frontend.json` file in `i18n/` directory. This file is like a pot file, this is the base to create your translations in any language you want.
```

docs/frontend-overrides.md
sampaccoud: s/according/according to

docs/frontend-overrides.md
```diff
@@ -37,14 +36,70 @@ export interface CourseGlimpseProps {
...
+
+Once translations keys are extracted and your local translation are ready, you need to compile these translations. In fact, compilation process first aggregates all translation files found from provided paths then merges them with richie translations according their filename and finally generates an output formatted for `react-intl`. Below, here an example of compilation command:
+
+```
+  node-modules/richie-education/i18n/compile-translations.js ./i18n/locales/*.json
+```
+
+This command looks for all translation files in `i18n/locales` directory then merges files found with richie translation files. You can pass several path patterns. You can also use an `--ignore` argument to ignore a particular path.
+
+### Override an existing translation keys
```
sampaccoud, line 77:
```suggestion
### Override an existing translation key
```

docs/frontend-overrides.md
sampaccoud: s/as explain/as explained

sampaccoud: s/according their filename/according to their filename

docs/frontend-overrides.md
```diff
@@ -37,14 +36,70 @@ export interface CourseGlimpseProps {
...
+
+As explain above, compilation process aggregates translations files then **merges them according their filename**. That means if you want override for example english translation, you just have to create a `en-US.json` file and redefine translation key used by Richie.
+
+Richie uses one file per language, currently, there 4 languages supported:
```
sampaccoud, line 81:
```suggestion
Richie uses one file per language. Currently, 4 languages supported:
```

docs/frontend-overrides.md
sampaccoud: don't you "just" as it may be offending for someone struggling to make it work

PR #1163, opened 2020-10-21
💄(footer) avoid stretching of footer menu items

jbpenrath:
## Purpose

Footer menu items stretchs according to space available on a row which broke
layout if there was few elements on a row.

## Proposal
- [x] Improve stretching of footer menu items
- [x] Second level nested items are now displayed in column

## Screenshot

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,12 +8,14 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+- Allow use of columns in footer menu items
+- Avoid stretching of footer menu items
+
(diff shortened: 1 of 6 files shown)


sampaccoud:
This has the result of shrinking the whole footer to the left.

image

I think what they want is that the first level of footer items stretch as they do now. But the second level should appear as a column below the first level. 

I have proposed a fixup commit to increase the number of second level items below the "richie community" first level item. This will make it more obvious in the demo site.

sampaccoud (approved):

sampaccoud:
@jbpenrath I added to the description a screenshot of the footer as it was when doing the final review. Nice :clap: !

PR #1165, opened 2020-10-23
👷(github) add automatic release on new tag

Sifflex:
Add github workflow to automatically create a new release whenever a new
tag is pushed to the repository. The release description is based on the
current CHANGELOG.

## Purpose

We wanted to make it easier to release a new version of the application.

## Proposal

Since releases are associated to a tag, we added a github workflow that calls a script everytime a new tag is pushed on the repository and creates the corresponding release.

A new github secret called API_KEY_GITHUB needs to be added to the repository for the script to work.

### DIFF
--- a/.github/workflows/autorelease.sh
+++ b/.github/workflows/autorelease.sh
@@ -1,0 +1,68 @@
+if [ $# -lt 1 ]; then
+    echo -e "Missing parameter. Usage: ./autorelease [github_token]"
+    exit 1
+fi
+
+CHANGELOG_FILE="CHANGELOG.md"
+
+git fetch --prune --unshallow --tags
+
+# Sort tags by date and pick the last one
+LAST_TAG_RAW=$(git for-each-ref --sort=creatordate --format '%(refname:short)' refs/tags | tail -1)
+
+# Remove 'v' in front of the tag
+LAST_TAG="${LAST_TAG_RAW:1}"
+
+# Find the tag's corresponding line in the CHANGELOG
+LINE_BEGIN=$(grep -n "## \[${LAST_TAG}\]" ${CHANGELOG_FILE} | cut -d : -f 1)
+
+# Start at the first paragraph instead of the title
+LINE_BEGIN=$((${LINE_BEGIN} + 2))
+
+# If there is un "Unreleased" title, the 2nd most recent tag is the 3rd title
+grep "## [Unreleased]" ${CHANGELOG_FILE}
+TAG_NUMBER=2
+if [ $? -eq 0 ]; then
+    TAG_NUMBER=3
+fi
+
+CHANGELOG_TITLES=$(grep -n "## \[" ${CHANGELOG_FILE})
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -0,0 +1,17 @@
...
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      # Checks-out your repository under $GITHUB_WORKSPACE, so the job can access it
+      - uses: actions/checkout@v2
+
+      - name: Some job
```
sampaccoud, line 16: name it something meaningful : Auto release?

.github/workflows/autorelease.sh
```diff
@@ -0,0 +1,68 @@
...
+    echo -e "Missing parameter. Usage: ./autorelease [github_token]"
+    exit 1
+fi
+
+CHANGELOG_FILE="CHANGELOG.md"
+
+git fetch --prune --unshallow --tags
+
+# Sort tags by date and pick the last one
+LAST_TAG_RAW=$(git for-each-ref --sort=creatordate --format '%(refname:short)' refs/tags | tail -1)
```
sampaccoud, line 11: You should be able to get the tag as an environment variable in the github job and pass it to this function

.github/workflows/autorelease.sh
```diff
@@ -0,0 +1,68 @@
...
+LAST_TAG="${LAST_TAG_RAW:1}"
+
+# Find the tag's corresponding line in the CHANGELOG
+LINE_BEGIN=$(grep -n "## \[${LAST_TAG}\]" ${CHANGELOG_FILE} | cut -d : -f 1)
+
+# Start at the first paragraph instead of the title
+LINE_BEGIN=$((${LINE_BEGIN} + 2))
+
+# If there is un "Unreleased" title, the 2nd most recent tag is the 3rd title
+grep "## [Unreleased]" ${CHANGELOG_FILE}
```
sampaccoud, line 23: there should always be an Unreleased tag. If not you can fail with an error

.github/workflows/autorelease.sh
```diff
@@ -0,0 +1,68 @@
...
+if [ $? -eq 0 ]; then
+    TAG_NUMBER=3
+fi
+
+CHANGELOG_TITLES=$(grep -n "## \[" ${CHANGELOG_FILE})
+LINE_END=$(echo "${CHANGELOG_TITLES}" | cut -d $'\n' -f ${TAG_NUMBER} | cut -d : -f 1)
+
+if [ -z ${LINE_END} ]; then
+    # Only one title, so the changelog goes until the end of the file
+    LINE_END=$(($(cat ${CHANGELOG_FILE} | wc -l) + 1))
```
sampaccoud, line 34: not really because there are links at the bottom of the file!

I think you can look for the next \[.*\] pattern in the file after your tag. It will match either the previous tag or a link at the bottom...

Sifflex:
Closed because of a GitHub error of mine, I will create another PR with the same purpose.

PR #1171, opened 2020-10-27
👷(github) create pull request for richie-site-factory after tagging

Sifflex:
## Purpose

Since tagging is relatively frequent, we improve the conveniency of
releasing a version of richie by automatically creating a pull request
updating all richie-related sites

## Proposal

We added a github action that is triggered everytime a new tag is created on the repository which launches a python script to create a pull request to update richie-site-factory

### DIFF
--- a/.github/workflows/deploy_richie.py
+++ b/.github/workflows/deploy_richie.py
@@ -1,0 +1,103 @@
+#! /usr/bin/python3
+
+import os
+import subprocess as cmd
+import sys
+import time
+
+from git import GitCommandError, Repo
+from github import Github, GithubException, UnknownObjectException
+
+
+def getRepoFromURl(url):
+    split = repo_url.replace("https://github.com/", "").split("/")
+    owner = None
+    try:
+        owner = g.get_organization(split[0])
+    except UnknownObjectException:
+        owner = g.get_user(split[0])
+    return owner.get_repo(split[1])
+
+
+if len(sys.argv) < 2:
+    print(f"Usage: {sys.argv[0]} VERSION GITHUB_TOKEN")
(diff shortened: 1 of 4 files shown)


.github/workflows/deploy_richie.yml
```diff
@@ -0,0 +1,21 @@
...
+
+    steps:
+      # Checks-out your repository under $GITHUB_WORKSPACE, so the job can access it
+      - uses: actions/checkout@v2
+
+      - name: Install Dependencies
+        run: pip3 install setuptools; pip3 install GitPython; pip3 install PyGithub;
+
+      # GITHUB_REF##*/ turns refs/heads/v1.3 to v1.3
+      - name: Deploy
```
sampaccoud, line 20: It's not a deployment, more an upgrade!

.github/workflows/deploy_richie.py
```diff
@@ -0,0 +1,103 @@
...
+commits = cmd.run(
+    f'cd {repo.name}; \
+                    git log --pretty="%s" {branch_name}',
+    stdout=cmd.PIPE,
+    shell=True,
+).stdout.decode("utf-8")
+last_commits = commits.split("\n")[:number_of_sites]
+
+body = ""
+for commit in last_commits:
```
sampaccoud, line 91: this is done by the bin/upgrade script if you pass the -c option.

Sifflex: The body is the one used to create the pull request because empty bodies are not allowed

sampaccoud:
github.com/openfun/richie-site-factory was deprecated and the site factory brought into the present repository with Cookiecutter.

PR #1173, opened 2020-10-29
 🚑(LMSHandler) select_lms returns None if url is blank

jbpenrath:
## Purpose

When a new course run page is created, a blank course run object is created.
But the course run page used has_connected_lms to check if the course run has an
associated known LMS. However the course run is blank, has_connected_lms crashes
because it send None value to LMSHandler.

## Proposal
- [x] `has_connected_lms` returns False directly if `course_run.resource_link` is blank

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,8 +8,14 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
-- Allow use of columns in footer menu items
-- Avoid stretching of footer menu items
+### Fixed
+  - has_connected_lms crashed when course_run is blank
+
+### Added
+  - Allow use of columns in footer menu items
+
+### Changed
+  - Avoid stretching of footer menu items
 
 ## [2.0.0-beta.16] - 2020-10-23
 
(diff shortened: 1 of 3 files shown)


src/richie/apps/courses/templatetags/extra_tags.py
```diff
@@ -216,4 +216,7 @@ def has_connected_lms(course_run):
     This enables our templates to either use the <CourseRunEnrollment /> component or a simple
     link to the course run.
     """
-    return LMSHandler.select_lms(course_run.resource_link) is not None
+    if course_run.resource_link:
+        return LMSHandler.select_lms(course_run.resource_link) is not None
```
sampaccoud, line 220: I think it's better to fix the error at its root ie in the `select_lms` method so that all other callers can benefit from the fix.

CHANGELOG.md
```diff
@@ -8,8 +8,14 @@ Versioning](https://semver.org/spec/v2.0.0.html).
```
sampaccoud, line 8: we never use `richie/` in the commit title in this project

src/richie/apps/courses/templatetags/extra_tags.py
jbpenrath: Yep of course ! 👍

sampaccoud (approved):

sampaccoud:
I think the problem is that gitllint is expecting your commit title to start with an action verb. The scope should also not contain any capital letters. I propose:
```
🚑(lms) fix "select_lms" method when the url is blank
```

PR #1174, opened 2020-10-30
⬆️(project) upgrade DjangoCMS to 3.8.0

sampaccoud:
## Purpose

[DjangoCMS 3.8.0](https://www.django-cms.org/en/blog/2020/10/28/django-cms-3-8-0-released/) was released yesterday and brings support for Django 3.1 :tada: 

## Proposal

Bump DjangoCMS dependency to 3.8.0 (and subsequently Django to 3.1.2), and fix all tests.


tests/apps/courses/test_cms_wizards_blogpost.py
```diff
@@ -315,7 +315,7 @@ def test_cms_wizards_blogpost_submit_form_invalid_slug(self):
         self.assertFalse(form.is_valid())
         self.assertEqual(
             form.errors["slug"][0],
-            "Enter a valid 'slug' consisting of letters, numbers, underscores or hyphens.",
+            "Enter a valid “slug” consisting of letters, numbers, underscores or hyphens.",
```
jbpenrath, line 318: Why did you replace single quotes?

setup.cfg
```diff
@@ -26,9 +26,9 @@ classifiers =
 include_package_data = True
 install_requires =
     arrow # pyup: ignore
-    Django<3.0.0 # pyup: ignore
+    Django # pyup: ignore
```
jbpenrath, line 29: We always install the latest django version? Should we not use a fixed version?

jbpenrath (approved):
Just few questions but it works great.

tests/apps/courses/test_cms_wizards_blogpost.py
sampaccoud: this error message is in Django's code and they seem to have replaced the single quotes to be more literal...

setup.cfg
sampaccoud: In fact we don't pin our dependencies unless necessary. It was necessary until now to limit Django to v2 but this constraint is now gone :tada:

PR #1177, opened 2020-11-03
Add flag to exclude course from catalogue

sampaccoud:
## Purpose

We want to allow excluding a course from the search page.

## Proposal

I propose to add a flag "is_listed" on the course model and index this field in Elasticsearch. While searching, I filter out from results the courses that have this flag set.

The flag can be set/unset from the "Course settings..." item in the "page" administration menu visible by staff users on any course detail page.

:information_source: Another option was to skip indexing for courses that have the "is_listed" flag unset. I abandonned this option because it was fragile. Indeed, a course that was previously listed (and thus indexed) had to be deleted from the index when it was switched to "not listed", leading to complications...


sampaccoud:
@jbpenrath I fixed the DjangoCMS 3.8 bugs and the autocomplete search. So it should be ready for further review!

src/richie/apps/courses/models/course.py
```diff
@@ -133,6 +133,11 @@ class Course(BasePageExtension):
         blank=True,
         null=True,
     )
+    is_listed = models.BooleanField(
+        default=True,
+        verbose_name=_("is listed"),
```
jbpenrath, line 138: Just a nitpick but there is no related translation.

jbpenrath (approved):
Works great ! 🎉

src/richie/apps/courses/models/course.py
sampaccoud: True! The string will only be available in Crowdin after I merge this PR :man_shrugging:

PR #1181, opened 2020-11-04
🐛(frontend) refresh js on cms-content-refresh

jbpenrath:
## Purpose

In edit mode, when user saves a modification, the new content is hot replaced
but js scripts are not reinitiliazed. In our case, as we use React, all React
components are not rendered after a content modification in Edit mode.

To solve this issue, Django-CMS provides an event "cms-content-refresh".
https://docs.django-cms.org/en/latest/topics/frontend-integration.html

## Proposal

- [x] Listen `cms-content-refresh` event and reinitialize our js each time this event is triggered.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Fixed
 
+- Reinitialize JS after saving a modification in edit mode
 - has_connected_lms crashed when course_run is blank
 - Avoid stretching of footer menu items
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Fixed
 
+- Reinitialize JS after saving a modification save in edit mode
```
sampaccoud, line 25:
```suggestion
- Reinitialize JS after saving a modification in edit mode
```

src/frontend/js/index.tsx
```diff
@@ -97,4 +97,9 @@ document.addEventListener('DOMContentLoaded', async () => {
       reactRoot,
     );
   }
-});
+}
+
+document.addEventListener('DOMContentLoaded', render);
+
+// In some case, you would like to render manually Richie components
```
sampaccoud, line 104:
```suggestion
// In some case, you would like to render Richie components manually
```

sampaccoud (approved):

PR #1185, opened 2020-11-09
🔖(patch) bump version to 2.0.0-beta.19

jbpenrath:
## Purpose
- JS comments was at the wrong place and I didn't catch up this mistake.
- The new sass compiler watch mode is not really powerful and crashes when there is a lot of files to watch.
- Selector and Spinner must use their own color palettes to be customizable on Richie Site Factory.

## Proposal

- [x] Fix comment
- [x] Go back to nodemon to watch style files changes
- [x] Add a color palette for selector and spinner components

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,20 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic
 Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [2.0.0-beta.18 - 2020-11-06
+## [Unrealeased]
+
+## [2.0.0-beta.19] - 2020-11-09
+
+### Added
+
+- Add a color palette for selector component
+- Add a color palette for spinner component
+
+### Fixed
+
+- Fix a typo in base template
+
+## [2.0.0-beta.18] - 2020-11-06
 
 ### Fixed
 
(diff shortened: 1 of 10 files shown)


sampaccoud (approved):

.circleci/config.yml
```diff
@@ -301,7 +301,7 @@ jobs:
               -wait tcp://${HOSTNAME}:3306 \
               -wait tcp://${HOSTNAME}:9200 \
               -timeout 60s \
-                ~/.local/bin/pytest
+                ~/.local/bin/pytest --disable-warnings
```
jmaupetit, line 304: Not sure it's a good idea to hide them. We need to be aware of those warnings, it's not an issue that could make the CI fail, so...

jbpenrath: That's why I do not use this option to run tests locally. But about CI, I think to remove these warnings help you to identify quickly what's wrong.

sampaccoud: +1 to remove. The warnings associated with our code have been solved and the remaining warnings are raised by our dependencies so there's nothing we can do about them.

jmaupetit: I do agree, that this is not our responsibility to fix it, but it's not worth being blind about it. :woman_shrugging:

jmaupetit: And, this globally removes warning for which we could be responsible of in future developments.

jmaupetit: If you persist in your idea to remove those warnings, you can filter them instead: https://docs.pytest.org/en/stable/warnings.html

jbpenrath: I can use this settings that ignores and **really ignores** all warnings coming from another packages than `richie` or `tests`.
```
filterwarnings = 
    ignore:::(?!(tests|richie))
```

On the other hand by using `--disable-warnings`, pytest does not ignore warnings, it does not display details of them in the stacktrace. But If you read the resume you see the number of warnings catched. And locally you can run `make test-back` and see all details.

jbpenrath: We need a consensus about this question. I created a new Pull request (#1186) for this aspect.

PR #1188, opened 2020-11-10
🐛(persons) allow adding plugins to main content when empty

sampaccoud:
## Purpose

We did not include the "maincontent" placeholder in the page when it was empty, even for staff users. As a result the placeholder
was not functional in the plugin side bar.

## Proposal

Add a condition to add the `mainbody` placedholder in the draft person detail page for staff users, even if the placeholder is empty.

### DIFF
--- a/src/richie/apps/courses/templates/courses/cms/person_detail.html
+++ b/src/richie/apps/courses/templates/courses/cms/person_detail.html
@@ -76,7 +76,7 @@
 {% block content %}{% spaceless %}
 {% with header_level=2 person=current_page.person %}
 <div class="person-detail">
-    {% if not current_page|is_empty_placeholder:"maincontent" %}
+    {% if current_page.publisher_is_draft or not current_page|is_empty_placeholder:"maincontent" %}
     <div class="person-detail__maincontent person-detail__block">
         <div class="person-detail__row">
             {% placeholder "maincontent" %}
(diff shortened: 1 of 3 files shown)


jbpenrath (approved):

tests/apps/courses/test_templates_person_detail.py
```diff
@@ -302,6 +303,15 @@ def test_templates_person_detail_maincontent_empty(self):
         self.assertContains(response, person.extended_object.get_title())
         self.assertNotContains(response, "person-detail__maincontent")
 
+       # But it should be present on the draft page
```
jbpenrath, line 306:
```suggestion
        # But it should be present on the draft page
```

src/richie/apps/courses/templates/courses/cms/person_detail.html
```diff
@@ -76,7 +76,7 @@ <h1 class="subheader__title">{% render_model current_page "title" %}</h1>
 {% block content %}{% spaceless %}
 {% with header_level=2 person=current_page.person %}
 <div class="person-detail">
-    {% if not current_page|is_empty_placeholder:"maincontent" %}
```
jbpenrath, line 79: Maybe add a entry into the CHANGELOG.

tests/apps/courses/test_templates_person_detail.py
sampaccoud: :scream_cat:

src/richie/apps/courses/templates/courses/cms/person_detail.html
sampaccoud: done

PR #1189, opened 2020-11-10
Release 2.0.0 beta.20

jbpenrath:
### Fixed

- Fix maincontent placeholder on person detail page when empty

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unrealeased]
 
+## [2.0.0-beta.20] - 2020-11-10
+
 ### Fixed
 
 - Fix maincontent placeholder on person detail page when empty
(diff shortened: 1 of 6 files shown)


sandbox/settings.py
```diff
@@ -257,8 +257,8 @@ class Base(StyleguideMixin, DRFMixin, RichieCoursesConfigurationMixin, Configura
         # e.g: for user.username = johndoe, /u/(username) will be /u/johndoe
         "PROFILE_URLS": values.ListValue(
             [
-                {"label": _("Profile"), "href": _("{base_url:s}/u/(username)")},
-                {"label": _("Account"), "href": _("{base_url:s}/account/settings")},
```
sampaccoud, line 261: I don't think it's useless to translate path!?

jbpenrath: In case of url rewriting ?

sampaccoud: useless for sandbox ok, but not in other sites...

sampaccoud (approved):

PR #1190, opened 2020-11-13
✨(course_glimpse) add large variant for course glimpse plugin

jbpenrath:
## Purpose
On homepage, there is a layout issue when course glimpses section includes another plugin before them. In fact for this section, custom styles are applied to the first three children but this tweak does not take in account that other plugins than course glimpse plugin can be included in this section.

## Proposal

- [x] Create a large variants of course glimpse that editor can use everywhere


src/richie/apps/courses/defaults.py
```diff
@@ -242,7 +242,10 @@
 COURSE_GLIMPSE_VARIANT_CHOICES = getattr(
     settings,
     "RICHIE_COURSE_GLIMPSE_VARIANT_CHOICES",
-    [(None, _("Inherit")), ("glimpse", _("Default")), ("small", "small")],
+    [
+        (None, _("Inherit")),
+        ("large", _("Large")),
```
sampaccoud, line 247: you need to add a migration because these choices are in database for someone who did not tweak the setting... 
I would say you also need to keep the "small" option. So the migration could do:
glimpse -> None
small -> small
large (new)

src/richie/apps/courses/templates/courses/cms/fragment_course_glimpse.html
```diff
@@ -1,13 +1,14 @@
...
-    <div class="course-{{ course_variant }}__media">
+{% with course_page=course.extended_object course_state=course.state main_org_title=course.get_main_organization.extended_object.get_title course_variant=course_variant %}
+<a class="course-glimpse  
+    {% if course_variant is not None %} course-glimpse--{{ course_variant }}{% endif %}
+    {% if course_page.publisher_is_draft is True %} course-glimpse--draft{% endif %}"
+    href="{{ course_page.get_absolute_url }}">
+    <div class="course-glimpse__media">
         {% get_page_plugins "course_cover" course_page as cover_plugins or %}
-            <p class="course-{{ course_variant }}__empty">{% trans "Cover" %}</p>
+            <p class="course-glimpse__empty">{% trans "Cover" %}</p>
```
sampaccoud, line 11: should it be --empty ? it is somewhere else in the code...

src/richie/locale/fr_FR/LC_MESSAGES/django.po
```diff
@@ -1614,8 +1618,11 @@ msgstr[1] ""
...
+#~ msgid "Fluid"
+#~ msgstr "Extensible"
+
+#~ msgid "Profile"
+#~ msgstr "Profil"
 
-msgid "Account"
-msgstr "Compte"
+#~ msgid "Account"
+#~ msgstr "Compte"
```
sampaccoud, line 1628: remove it completely if it's obsolete

src/richie/locale/fr_CA/LC_MESSAGES/django.po
```diff
@@ -1614,8 +1618,11 @@ msgstr[1] ""
...
+#~ msgid "Fluid"
+#~ msgstr "Extensible"
+
+#~ msgid "Profile"
+#~ msgstr "Profil"
 
-msgid "Account"
-msgstr "Compte"
+#~ msgid "Account"
+#~ msgstr "Compte"
```
sampaccoud, line 1628: remove it completely if it's obsolete

src/richie/apps/courses/templates/courses/cms/fragment_course_glimpse.html
jbpenrath: Good catch. Yep a wrong class name was used.

src/richie/apps/courses/defaults.py
jbpenrath: Small variant do currently nothing.

sampaccoud: then ok to remove it with a migration setting the value to "None"... because someone might have it as value in their database.

src/richie/apps/courses/migrations/0019_auto_20201113_1630.py
```diff
@@ -0,0 +1,25 @@
...
+            field=models.CharField(
+                blank=True,
+                choices=[(None, "Inherit"), ("large", "Large")],
+                help_text="Optional glimpse variant for a custom look.",
+                max_length=50,
+                null=True,
+                verbose_name="variant",
+            ),
+        ),
+    ]
```
sampaccoud, line 25: This migration is requested by Django (although useless) but you most importantly need a "data migration" to correct the field values See example here: https://github.com/openfun/richie/blob/master/src/richie/apps/courses/migrations/0013_migrate_data_translated_licence_fields.py

src/richie/apps/courses/migrations/0020_deprecate_course_glimpse_plugin_small_variant.py
```diff
@@ -0,0 +1,25 @@
...
+
+def forwards_func(apps, schema_editor):
+    """
+    Update every course plugin objects which used the deprecated "small" variant
+    to default course variant.
+    """
+    CoursePlugin = apps.get_model("courses", "coursepluginmodel")
+    CoursePlugin.objects.filter(variant="small").update(
+        variant=COURSE_GLIMPSE_VARIANT_CHOICES[0][0]
+    )
```
sampaccoud, line 16: Sorry I just realized that this may break the db for someone who has customized these choices
Maybe add an instruction to UPGRADE.md to make the necessary treatments themselves?

src/richie/apps/courses/defaults.py
jbpenrath: Finally I kept the small variant and its value is 'glimpse' to not display something broken if end user selects this value. Furthermore since I add a `large` variant, `small` variant seems logic.

jbpenrath: Finally I kept the small variant and its value is 'glimpse' to not display something broken if end user selects this value. Furthermore since I add a `large` variant, `small` variant seems logic.

sampaccoud (approved):

PR #1196, opened 2020-11-17
🐛(LMSHandler) check resource_link matches lms COURSE_REGEX

jbpenrath:
## Purpose
Enrollment widget can fail to enroll if course run resource_link is malformed.

## Proposal

- [x] Display enrollment widget only if there is a backend for the course run and a course_id can be extracted from its resource_link

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 ### Fixed
 
 - Fix an issue about PaginateSearchCourse layout
+- Prevent enrollment failure if course run's resource_link does not match the
+  course regexp
 
 ## [2.0.0-beta.20] - 2020-11-10
 
--- a/env.d/development/dev-ssl
+++ b/env.d/development/dev-ssl
@@ -4,5 +4,6 @@ AUTHENTICATION_BACKEND=richie.apps.courses.lms.edx.TokenEdXLMSBackend
 
 # LMS Backend
 EDX_BACKEND=richie.apps.courses.lms.edx.TokenEdXLMSBackend
+EDX_COURSE_REGEX=^.*/courses/(?P<course_id>.*)/info$
 EDX_JS_COURSE_REGEX=^.*/courses/(?<course_id>.*)/info$
 EDX_BASE_URL=https://edx.local.dev:8073
--- a/env.d/development/dev-ssl
+++ b/env.d/development/dev-ssl
@@ -4,6 +4,6 @@ AUTHENTICATION_BACKEND=richie.apps.courses.lms.edx.TokenEdXLMSBackend
 
 # LMS Backend
 EDX_BACKEND=richie.apps.courses.lms.edx.TokenEdXLMSBackend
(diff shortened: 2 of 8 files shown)


sandbox/settings.py
```diff
@@ -275,9 +275,14 @@ class Base(StyleguideMixin, DRFMixin, RichieCoursesConfigurationMixin, Configura
                 environ_name="EDX_BACKEND",
                 environ_prefix=None,
             ),
-            "SELECTOR_REGEX": values.Value(
+            "COURSE_REGEX": values.Value(
                 r".*", environ_name="EDX_SELECTOR_REGEX", environ_prefix=None
             ),
+            "SELECTOR_REGEX": values.Value(
+                r"^(?P<course_id>.*)$",
+                environ_name="EDX_COURSE_REGEX",
```
sampaccoud, line 283: you mixed up the environ names

sampaccoud:
I'm wondering if it's not the frontend's job to determine if the course "has_connected_lms"? Since the frontend and the backend regexes are different settings, It feels like each one should check this based on its regex!

jbpenrath:
> I'm wondering if it's not the frontend's job to determine if the course "has_connected_lms" based on its regex?
> Since the frontend and the backend regexes are different settings, It feels like each one should check this based on its regex!

@sampaccoud Delegate this check to the back allow us to prevent a useless react component generation.

> I'm wondering if it's not the frontend's job to determine if the course "has_connected_lms" based on its regex?
> Since the frontend and the backend regexes are different settings, It feels like each one should check this based on its regex!

Do this check from backend prevents rendering of a React component to just display a link.

sampaccoud:
Ok then. The only drawback I see is that the backend and frontend regexes must be the same regex translated in 2 languages. I suppose it's fine and will mention it in the documentation revamp.

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 ### Fixed
 
 - Fix an issue about PaginateSearchCourse layout
+- Prevent enrollment failure if resource_link do not match COURSE_REGEX
```
sampaccoud, line 18:
```suggestion
- Prevent enrollment failure if a course run's resource link does not match the course regex
```

env.d/development/dev-ssl
```diff
@@ -4,5 +4,6 @@ AUTHENTICATION_BACKEND=richie.apps.courses.lms.edx.TokenEdXLMSBackend
 
 # LMS Backend
 EDX_BACKEND=richie.apps.courses.lms.edx.TokenEdXLMSBackend
+EDX_COURSE_REGEX=^.*/courses/(?P<course_id>.*)/info$
```
sampaccoud, line 7: Since you said /info redirects to /course :man_shrugging: :
```suggestion
EDX_COURSE_REGEX=^.*/courses/(?P<course_id>.*)/course$
```

sampaccoud (approved):

env.d/development/dev-ssl
jbpenrath: Yes, I will also update `create_demo_site.py` !

PR #1197, opened 2020-11-18
✨(Plugins/Glimpse) add a new "Badge" variant

jbpenrath:
## Purpose

A glimpse variant was missing to format content of FUN sites as designed in mockups

## Proposal

- [x] Create the missing "Badge" variant

badge-glimpse-showcase


src/frontend/scss/settings/_fonts.scss
```diff
@@ -2,7 +2,7 @@
 // https://fonts.google.com/?query=Montserrat&selection.family=Hind:wght@400;600|Montserrat:wght@300;700;800&sidebar.open
 //
 
-@import url('https://fonts.googleapis.com/css2?family=Hind:wght@400;600&family=Montserrat:wght@300;700;800&display=swap');
+@import url('https://fonts.googleapis.com/css2?family=Hind:wght@400;600&family=Montserrat:wght@300;500;600;700;800&display=swap');
```
sampaccoud, line 5: not sure why this was necessary but may require a separate commit to explain?

sampaccoud (approved):

src/frontend/scss/settings/_fonts.scss
jbpenrath: As designed in mockups, Glimpse badge title use Montserrat Medium font.
I moved it in a separate commit and I removed the useless `600` variant.

sampaccoud: so the category plugin should also use medium so that they can be mixed?

PR #1199, opened 2020-11-23
Fix/course resource link validator

jbpenrath:
## Purpose
When a course use a known `LMS_BACKEND` but its associated `COURSE_REGEX` does not match a `courseId` an exception occured.

## Proposal

- [x] `select_lms` now checks that course run has an associated LMS_BACKEND and that COURSE_REGEX finds a `courseId` in its resource link.


jbpenrath:
Duplicate #1196

PR #1211, opened 2020-12-07
🗃(backend) add object names plurals migration

jbpenrath:
## Purpose

Object names plurals migration was missing

## Proposal

- [x] Generate missing migration file


jbpenrath:
This migration is useless. We will include it later with other migrations.

PR #1223, opened 2020-12-15
🐛(frontend) fix intlrelativeformat polyfill

jbpenrath:
## Purpose

When intl-relativetimeformat polyfill must be applied, the locale module was
not found. In fact when locale is composed of a languageCode identical
to its countryCode, intl-relativetimeformat locale filename is only languageCode

i.e: if locale is fr-FR related locale filename is fr.js
if locale is fr-CA related locale filename is fr-CA.js

## Proposal

- [x] check if locale languageCode is identical to countryCode, if yes just use languageCode as locale filename

### DIFF
--- a/src/frontend/js/index.tsx
+++ b/src/frontend/js/index.tsx
@@ -53,13 +53,17 @@ async function render() {
       // TODO: remove type assertion when typescript libs include RelativeTimeFormat
       if (!(Intl as any).RelativeTimeFormat) {
         await import('@formatjs/intl-relativetimeformat');
-        // Polyfilled locale data is keyed by 2-letter language code
-        let languageCode = locale;
-        if (languageCode.match(/^.*_.*$/)) {
-          [languageCode] = locale.split('_');
+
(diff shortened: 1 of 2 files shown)


src/frontend/js/index.tsx
```diff
@@ -25,7 +25,9 @@ async function render() {
   // Only move on with anything if there are actually components to render
   if (richieReactSpots.length) {
     // Determine the BCP47/RFC5646 locale to use
-    const locale = document.querySelector('html')!.getAttribute('lang');
+    // This standard specifies that some locale implementations may use '_' instead of '-'
```
lunika, line 28: Where did you see that ? I don't see usage of `_` in the RFC5646

jbpenrath: Whoops you are right. I search a little bit quickly, this concerns OpenID that can use locale with BCP47/RFC4646 format and some implementation may use an underscore instead of a dash.

lunika (approved):

PR #1236, opened 2021-01-07
✨(development) allow connections from openedx-docker development stack

sampaccoud:
## Purpose

For ouath2 we already allowed richie's development stack to access openedx-docker's development stack. For course runs synchronizations, it is the contrary: openedx-docker's development tack needs to be able
to call richie's course synchronization web hook on the development stack.

## Proposal

Connect nginx in the `lms_outside` network so it can be accessed from outside its docker-compose project with the `richie` host.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -44,8 +44,12 @@ services:
     ports:
       - "8070:8070"
     networks:
-      - default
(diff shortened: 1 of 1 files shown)


jbpenrath (approved):
LGTM

PR #1244, opened 2021-01-11
♻️(settings) remove JS_SELECTOR_REGEXP

jbpenrath:
## Purpose

As it has been done in backend, we only use the course regexp to both select the lms api and extract the course id. So `JS_SELECTOR_REGEXP` is deprecated.

## Proposal

- [x] Use `JS_COURSE_REGEXP` to select LMS Api to use


sampaccoud:
@jbpenrath maybe add the fix release to this PR?

sampaccoud (approved):
Up to now we didn't make new doc releases for richie fix releases. Apart from this GTM!

PR #1250, opened 2021-01-19
🐛(courses) fix displaying course runs in list on course detail page

sampaccoud:
## Purpose

The template code that was taking care of displaying the course run details with or without a title, was broken in french because the translation included a comma where it shouldn't. It turns out that the implementation is a git too "smart" and thus fragile. 

## Proposal

We refactor it to use a good old "if" clause.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 - Improve UX of course search pagination by avoiding truncation of page number
   when it is not relevant
 
+### Fixed
+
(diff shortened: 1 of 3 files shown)


jbpenrath (commented):
Hmmm, Is not the extra comma due to a mistake in crowdin translations ?

sampaccoud:
@jbpenrath yes you are right, it could be fixed by correcting the translation in Crowdin but:
- the implementation was fragile if it relies on the string being of a certain format
- when the course run has a title, it is in fact better to have the comma like before...

jbpenrath (approved):

Issue #1249, opened 2021-01-19
Display the course code on the detail page and make it frontend editable


sampaccoud:
## Feature Request

**Is your feature request related to a problem or unsupported use case? Please describe.**
We want to make it easier to edit a course's "code" field.

**Describe the solution you'd like**
It's also interesting to display it on the course syllabus, e.g. in small below the title.
Once the code is displayed, add frontend edition to it.

PR #1251, opened 2021-01-19
✨(courses) display code on course detail page with frontend editing

sampaccoud:
## Purpose

It was not straightforward to go through the "course settings..." menu item to edit a course code. 

## Proposal

Since this is an important and uniquely defining information about the course, we propose to display it on the
detail page and allow frontend editing.

Fixes https://github.com/openfun/richie/issues/1249

## Snapshots

Screenshot from 2021-01-19 15-45-16

### DIFF
--- a/src/frontend/scss/components/_subheader.scss
+++ b/src/frontend/scss/components/_subheader.scss
@@ -31,6 +31,22 @@ $r-subheader-search-title-width: 19rem !default; // aligned on computed search r
     }
   }
 
+  // Main document title
+  &__code {
+    @include make-container();
+    @include make-container-max-widths();
+    @include sv-flex(1, 0, 100%);
+
+    margin-top: -0.5rem;
+    margin-bottom: 0.5rem;
+
+    @include media-breakpoint-down(lg) {
+      padding-left: 0;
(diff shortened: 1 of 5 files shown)


src/frontend/scss/components/_subheader.scss
```diff
@@ -31,6 +31,19 @@ $r-subheader-search-title-width: 19rem !default; // aligned on computed search r
     }
   }
 
+  // Main document title
+  &__code {
+    @include make-container();
+    @include make-container-max-widths();
+    @include sv-flex(1, 0, 100%);
+
```
jbpenrath, line 39: Just a nitpick. As code is not an important information for user, I think it should be visually discreet.

```css
margin-top: -0.5rem;
margin-bottom: 0.5rem;
```

By using this vertical margin, Reference code will be visually linked to the course title. So visually these two elements appears like a one block.


discreet-ref

jbpenrath (approved):

src/frontend/scss/components/_subheader.scss
sampaccoud: Nice suggestion. I applied it.

Issue #1240, opened 2021-01-07
Allow overriding OR completing a person's bio in the person plugin


sampaccoud:
## Feature Request

**Is your feature request related to a problem or unsupported use case? Please describe.**
When we include a person via a plugin on the course detail page, we sometimes want to use a custom bio instead of the bio defined on that person's page.

**Describe the solution you'd like**
Add a "bio" field in the person's plugin model which, when filled, will replace the person's bio in the html rendered by the plugin.

PR #1252, opened 2021-01-19
✨(courses) allow overriding a person's bio on the person plugin

sampaccoud:
## Purpose

In some cases, we want to precise or customize a person's bio in the context of his/her participation to a course. 

## Proposal

We propose to add an optional field on the person plugin that, if filled, is displayed in place of the person's bio on the glimpse.

closes https://github.com/openfun/richie/issues/1240


jbpenrath (approved):
Perfect ! LGTM

Issue #1239, opened 2021-01-07
Allow glimpses in the main content of the person detail page


sampaccoud:
## Feature Request

**Is your feature request related to a problem or unsupported use case? Please describe.**
Some person pages are in fact collectives. In this case, we want to include thumbnails of the persons in this collective.

**Describe the solution you'd like**
It should be possible to include person plugins or glimpse plugins depending on whether the person in the collective has a page in richie or not.

PR #1254, opened 2021-01-20
✨(persons) allow linking person plugins on a person's page

sampaccoud:
## Purpose

Some person pages are in fact collectives. In this case, we want to include thumbnails of the persons in this collective.

## Proposal

Allow the person plugin on the person detail page.

closes https://github.com/openfun/richie/issues/1239

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Added
 
+- Allow person and glimpse plugins on the person detail page
 - Allow overriding a person's bio on the person plugin
 - Add new section on the course detail page to display related programs
 - Display "code" on the course detail page and allow frontend editing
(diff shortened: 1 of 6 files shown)


jbpenrath (commented):
I suggested you to update styles of `.person-detail__row` to accept a column to display properly person plugin or glimpse plugin. https://github.com/openfun/richie/pull/1254/commits/b9eb67a395fa7cc1fd199f47a88d0abd024c4928

Capture d’écran 2021-01-20 à 13 44 00

Capture d’écran 2021-01-20 à 13 43 33

Also, I wonder if a new variant of the preview plugin should be created to have the same layout as a block person?

sampaccoud:
@jbpenrath the way to add person plugins is to group them under a section. Did you try that?

sampaccoud:
@jbpenrath good idea to add a glimpse variant that looks like a person.

jbpenrath:
> @jbpenrath the way to add person plugins is to group them under a section. Did you try that?

You are right, SectionPlugin is far better than my fix to address this topic.

jbpenrath (approved):
LGTM!

PR #1257, opened 2021-01-22
🐛(plugins) fix HTMLSiteMap plugin when placed in a static placeholder

sampaccoud:
## Purpose

Fix HTMLSiteMap plugin when placed in a static placeholder

This is not the preferred use case but it can nevertheless be done and
should not raise a 500 error 😱

## Proposal

Get page from context instead of poking the placeholder (static placeholders are not linked to any page).


jbpenrath (approved):
LGTM

PR #1262, opened 2021-01-27
🐛(courses) delete orphan course run nodes

sampaccoud:
## Purpose

When we converted course runs to simple objects, we deleted the CMS pages related to what, at the time, were course run page extensions, but forgot to delete the corresponding MPTree nodes.

## Proposal

Add a data migration to delete the orphan nodes that were previously linked to course run pages.


src/richie/apps/courses/migrations/0028_auto_20210126_2200.py
```diff
@@ -0,0 +1,35 @@
...
+        .distinct()
+        .iterator()
+    ):
+        # When deleting a node, we must also update its parent to decrement
+        # its number of children
+        parentpath = node.path[0 : (node.depth - 1) * 4]
+        parent = TreeNode.objects.get(path=parentpath)
+        node.delete()
+        parent.numchild -= 1
+        parent.save()
```
jbpenrath, line 26: Treebeard documentation advises to consider `num_child` as read-only.
After node deletion, `fix_tree` should do the job to detect and fix inconsistency in parent node.

```suggestion
        parent = node.get_parent()
        node.delete()
        parent.fix_tree()
        parent.save()
```

https://django-treebeard.readthedocs.io/en/latest/mp_tree.html#treebeard.mp_tree.MP_Node.fix_tree

sampaccoud: I know because I went through the whole process of "why just deleting the node doesn't fix my tree" :smile: I was just trying to make it fast because running `fix_tree` takes forever and I didn't want to run it in a migration. They say it's read-only because they are keeping it up-to-date but in my case they don't because a migration is operating on a dummy model that does not have any of the model's methods... 
What do you think? Should I call the "fix_tree" command in my migration?

jbpenrath: Ok, if `fix_tree` is a heavy operation, let's keep your implementation.

jbpenrath (approved):
I did not know that models for migration are faked and only fields are available. With that in mind, your implementation immediately seems more logical.

Issue #1263, opened 2021-01-27
Remove console.error during tests


jbpenrath:
## Bug Report

**Expected behavior/code**
If we test case of failure, we would like to ensure that error handler is trigger but not polluted the stack trace with console logs.

**Actual Behavior**
If we test case of failure, the error handler is trigger then emits a `console.error` that polluted the test stack trace.

**Steps to Reproduce**
1. Run tests
2. Check stack trace

**Environment**
- Richie version: 2.1.0
- Platform: Mac OS

**Possible Solution**
Mock the error handler to check it has been called and prevents its behavior.

**Additional context/Screenshots**
https://app.circleci.com/pipelines/github/openfun/richie/1598/workflows/8b96a0b1-97d5-4d0e-a46e-d6379f52b21e/jobs/97763

PR #1264, opened 2021-01-27
🚸(frontend) hide console error in tests outputs

kernicPanel:
## Purpose

When frontend tests are run, as console.error and console.warning are displayed, developer can think that some tests are failing.

## Proposal

Capture console.error and console.warning on concerned tests, and assert they have been called.
Only test displaying errors has been modified for now, but if the approach is ok, remaining test has to be updated as well.

fixes https://github.com/openfun/richie/issues/1263

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Changed
 
+- Mock error handler on frontend tests
 - Show 403 page instead of course run list when trying to edit a course run
   with insufficient permissions
 - Improve UX of course search pagination by avoiding truncation of page number
--- a/src/frontend/js/components/CourseRunEnrollment/index.spec.tsx
+++ b/src/frontend/js/components/CourseRunEnrollment/index.spec.tsx
@@ -11,6 +11,11 @@ import { Deferred } from 'utils/test/deferred';
 import * as factories from 'utils/test/factories';
 import { SESSION_CACHE_KEY } from 'settings';
 
+import { handle } from 'utils/errors/handle';
+
+const mockHandle: jest.Mock<typeof handle> = handle as any;
+jest.mock('utils/errors/handle');
+
 describe('<CourseRunEnrollment />', () => {
   const endpoint = 'https://demo.endpoint';
 
@@ -138,6 +143,7 @@ describe('<CourseRunEnrollment />', () => {
 
     screen.getByRole('button', { name: 'Enroll now' });
     screen.getByText('Your enrollment request failed.');
+    expect(mockHandle).toHaveBeenCalledWith('500 - Internal Server Error');
   });
 
   it('shows a link to the course if the user is already enrolled', async () => {
--- a/src/frontend/js/utils/api/lms/index.spec.ts
+++ b/src/frontend/js/utils/api/lms/index.spec.ts
@@ -1,5 +1,10 @@
 import { ApiBackend } from 'types/api';
 
+import { handle } from 'utils/errors/handle';
+
+const mockHandle: jest.Mock<typeof handle> = handle as any;
+jest.mock('utils/errors/handle');
+
 describe('API LMS', () => {
   (window as any).__richie_frontend_context__ = {
     context: {
@@ -34,6 +39,9 @@ describe('API LMS', () => {
   it('throw an error if an unknown url is provided', () => {
     expect(() => LMSHandler('https://unknown.org/course/a-test-course')).toThrow(
       `No LMS Backend found for https://unknown.org/course/a-test-course.`,
+    );
+    expect(mockHandle).toHaveBeenCalledWith(
+      new Error('No LMS Backend found for https://unknown.org/course/a-test-course.'),
     );
   });
 });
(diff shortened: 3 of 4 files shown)


src/frontend/js/components/CourseRunEnrollment/index.spec.tsx
```diff
@@ -128,6 +128,9 @@ describe('<CourseRunEnrollment />', () => {
 
     const button = screen.getByRole('button', { name: 'Enroll now' });
 
+    // eslint-disable-next-line no-console
+    console.error = jest.fn();
```
lunika, line 132:
```suggestion
    spyConsole = jest.spyOn(console, 'error');
```

In this cas `spyOn` is more appropriated IMO.

jbpenrath, line 132: We use a function to handle errors all over our frontend (https://github.com/openfun/richie/blob/master/src/frontend/js/utils/errors/handle.ts). According to the environment, it is in charge to send error to sentry or just display it through a `console.error`.

So to make this test more relevant, you have to mock our `handle` method then check it has been called.

kernicPanel:
I've mixed your suggestions, @lunika @jbpenrath
What do you think ?

src/frontend/js/components/CourseRunEnrollment/index.spec.tsx
```diff
@@ -128,6 +129,8 @@ describe('<CourseRunEnrollment />', () => {
 
     const button = screen.getByRole('button', { name: 'Enroll now' });
 
+    const spyConsole = jest.spyOn(ErrorHandler, 'handle').mockImplementation(() => {});
```
jbpenrath, line 132: IMO spyOn was relevant to spy the `error` method of `console`. Since `handle` is the only method exposed by the `handle` module, we could fully mock the module with `jest.mock('utils/errors/handle')`. Then we could simply import `handle` method and test it has been called.

May be you will have to reset mock before each tests.

src/frontend/js/components/CourseRunEnrollment/index.spec.tsx
```diff
@@ -138,6 +141,7 @@ describe('<CourseRunEnrollment />', () => {
 
     screen.getByRole('button', { name: 'Enroll now' });
     screen.getByText('Your enrollment request failed.');
+    expect(spyConsole).toHaveBeenCalled();
```
jbpenrath, line 144: To go further, we could use `toHaveBeenCalledWith` to also test the value received by `handle`. In this way our tests will failed if error code change.

src/frontend/js/utils/api/lms/index.spec.ts
```diff
@@ -31,9 +32,11 @@ describe('API LMS', () => {
     expect(api).toBeDefined();
   });
 
-  it('throw an error if an unknown url is provided', () => {
+  it.only('throw an error if an unknown url is provided', () => {
```
jbpenrath, line 35: Whoops.
```suggestion
  it('throw an error if an unknown url is provided', () => {
```

kernicPanel:
I found this file https://github.com/openfun/richie/blob/master/src/frontend/js/common/searchFields/getSuggestionsSection.spec.ts
and, after searching for meaning of `const mockHandle: jest.Mock<typeof handle> = handle as any;` which I begin to understand, I've decided to use the same approach.

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Changed
 
+- Hide console.errors on frontend tests output
```
jbpenrath, line 21:
```suggestion
- Mock error handler on frontend tests
```

src/frontend/js/utils/api/lms/openedx-hawthorn.spec.ts
```diff
@@ -143,6 +152,12 @@ describe('OpenEdX Hawthorn API', () => {
         );
 
         expect(response).toBeFalsy();
+        expect(mockHandle).toHaveBeenCalledWith(
+          new Error('[GET - Enrollment] > 500 - Internal Server Error'),
+        );
```
jbpenrath, line 157: This is a false positive. This assertion is true only because the mock is not restore before each tests. So in the call stack there is still the error caught in the previous test.

In the `beforeEach` function, you have to add `mockHandle.mockRestore`.

jbpenrath (approved):
Well done ! 👌

kernicPanel:
> Well done ! ok_hand

Thanks ! :slightly_smiling_face:

kernicPanel:
Can I do something to complete this PR ?

jbpenrath:
> Can I do something to complete this PR ?

Nope thats good, the merge is blocked by a ci job failure. We have to improve our workflow to merge PR from forks. We identified a potential solution that we have to test (maybe on your PR rightly ^^)

kernicPanel:
> > Can I do something to complete this PR ?
> 
> @kernicPanel Nope thats good, the merge is blocked by a ci job failure. We have to improve our workflow to merge PR from forks. We identified a potential solution that we have to test (maybe on your PR rightly ^^)

OK, thanks :)

jbpenrath:
@kernicPanel CI issue has been fixed. Can you rebase your branch from master then push force?

kernicPanel:
> @kernicPanel CI issue has been fixed. Can you rebase your branch from master, fix merge conflict on CHANGELOD.md then push force?

Yes, it's done

PR #1266, opened 2021-01-28
🐛(cms_plugins) fix HTMLSiteMap plugin when current_page is not defined

jbpenrath:
## Purpose

HTMLSiteMap plugin use current_page context property to know if this page is a draft. If this property is not defined in the context, an exception is raised and we got a 500 error.

## Proposal

- [x] Add default value to `current_page_is_draft` (`False`) to manage case where current_page is not defined

### DIFF
--- a/tests/plugins/html_sitemap/test_cms_plugins.py
+++ b/tests/plugins/html_sitemap/test_cms_plugins.py
@@ -564,3 +564,53 @@ class HTMLSitemapPluginTestCase(CMSPluginTestCase):
             </div>
             """,
         )
+
+    def test_cms_plugins_htmlsitemap_no_current_page(self):
+        """
+        A sitemap plugin inserted on a page with no current_page in its context
+        should display all published pages from its root
+        """
+        self.create_page_tree()
+
+        root = PageFactory(
+            title__title="Root", title__language="en", should_publish=True
+        )
+        PageFactory(
+            title__title="Parent",
+            parent=root,
+            title__language="en",
+            should_publish=True,
+        )
+        PageFactory(
+            title__title="Uncle", parent=root, title__language="en", should_publish=True
+        )
+        self.assertEqual(StaticPlaceholder.objects.count(), 1)
+        placeholder = StaticPlaceholder.objects.get()
+
+        context = self.get_practical_plugin_context()
+        parent_instance = add_plugin(placeholder.draft, HTMLSitemapPlugin, "en")
+        add_plugin(
+            placeholder.draft,
+            plugin_type="HTMLSitemapPagePlugin",
+            language="en",
+            target=parent_instance,
+        )
+
+        html = context["cms_content_renderer"].render_placeholder(
+            placeholder.draft, context=context, language="en"
+        )
+        self.assertHTMLEqual(
+            html,
+            """
+            <div class="sitemap">
+              <ul>
+                <li><a href="/en/root/">Root</a>
+                  <ul>
+                    <li><a href="/en/root/parent/">Parent</a></li>
+                    <li><a href="/en/root/uncle/">Uncle</a></li>
+                  </ul>
+                </li>
+              </ul>
+            </div>
+            """,
+        )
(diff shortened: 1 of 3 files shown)


sampaccoud (approved):

src/richie/apps/core/views/error.py
```diff
@@ -42,6 +43,10 @@ def error_view_handler(request, exception, status):
             "status": status,
             "title": CONTEXT_ERRORS[status]["title"],
             "content": CONTEXT_ERRORS[status]["content"],
+            # current_page is required by several plugins,
+            # so in error view request does not have current_page
+            # we pass the homepage
+            "current_page": Page.objects.get_home()
```
sampaccoud, line 49: Add a test maybe...

tests/apps/core/test_error_view.py
```diff
@@ -71,9 +84,20 @@ def test_error_view_handler_with_unsupported_status_code(self):
...
         request.current_page = page
         request.user = AnonymousUser()
+
         with self.assertTemplateUsed("richie/error.html"):
             response = error.error_view_handler(request, Exception, 405)
             self.assertContains(response, "500 - Server bad request", status_code=500)
+
+    def test_error_view_handler_raises_error_when_no_homepage(self):
+        """
+        As error_view_handler set context['current_page'] to the homepage site,
```
sampaccoud, line 99: Or make it work when there is no home page?

jbpenrath: Yep. Can we assume to pass None if the exception `NoHomeFound` is raised ? I guess if a site has no homepage, there is also no plugins used?

sampaccoud: Yes I agree with this take. Maybe add a test on the sitemap plugin that was problematic to be sure it behaves when current_page is set to None.

tests/plugins/html_sitemap/test_cms_plugins.py
```diff
@@ -564,3 +564,53 @@ def test_cms_plugins_htmlsitemap_static_placeholder(self):
             </div>
             """,
         )
+
+    def test_cms_plugins_htmlsitemap_no_current_page(self):
+        """
+        A sitemap page with no current_page in its context should display
```
sampaccoud, line 570:
```suggestion
        A sitemap plugin inserted on a page with no current_page in its context should display
```

tests/plugins/html_sitemap/test_cms_plugins.py
```diff
@@ -564,3 +564,53 @@ def test_cms_plugins_htmlsitemap_static_placeholder(self):
...
+            title__language="en",
+            should_publish=True,
+        )
+        PageFactory(
+            title__title="Uncle", parent=root, title__language="en", should_publish=True
+        )
+        self.assertEqual(StaticPlaceholder.objects.count(), 1)
+        placeholder = StaticPlaceholder.objects.get()
+
+        context = self.get_practical_plugin_context({"current_page": None})
```
sampaccoud, line 590: actually you should go further and not include the "current_page" parameter in the context

jbpenrath: Whoops, undo / redo failed.

tests/plugins/html_sitemap/test_cms_plugins.py
```diff
@@ -564,3 +564,53 @@ def test_cms_plugins_htmlsitemap_static_placeholder(self):
...
+        context = self.get_practical_plugin_context()
+        parent_instance = add_plugin(placeholder.draft, HTMLSitemapPlugin, "en")
+        add_plugin(
+            placeholder.draft,
+            plugin_type="HTMLSitemapPagePlugin",
+            language="en",
+            target=parent_instance,
+        )
+
+        html = context["cms_content_renderer"].render_placeholder(
```
sampaccoud, line 599: I would add a line to underline that the "current_page" parameter is absent from context. For clarity to the developper who reads this test:
```
self.assertFalse("current_page" in context)
```

sampaccoud (approved):

tests/plugins/html_sitemap/test_cms_plugins.py
jbpenrath: Good idea!

PR #1272, opened 2021-02-02
🐛(frontend) replace "go to course" button with dashboard link for unstarted lms courses and add keys to profile_url settings

### LINKED ISSUE #1260: Send enrolled user to LMS only when the course run is open
## Bug Report

**Problematic behavior/code**
When a user is enrolled on a course, we show him/her a button that sends him/her to the course on the LMS. This leads to a 500 error in OpenEdX if the course is not yet open.

**Expected Behavior**
The user should not be proposed a link before the start date of the course.

**Steps to Reproduce**
1. Enroll to a course for which the start date is in the future
2. Click on the button to access the course
3. And then the bug happens!

**Environment**
- Richie version: 2.0.1
- Platform: Docker

kernicPanel:
## Purpose

In openedx, when we try to start a course which is not open yet, we might get a 500 error. The "go to course" button is not displayed in this case, and has been replaced with a dashboard link.

Also, as discussed, profile_urls in settings are stored in a list, which we need to loop on to find a particular url

## Proposal

New prop attributes have been added to CourseRunEnrollment component with a message indicating when the course will start and a dashboard link.
If the message is defined we display the button.

A key has been added to each url in profile_urls settings.

fixes #1260

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unrealeased]
 
+- Hide "Go to course" button if lms course hasn't started yet 
+
 ### Added
 
 - Add a "person" variant to the glimpse plugin for persons without a page
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,11 +8,6 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unrealeased]
 
-- Hide "Go to course" button if lms course hasn't started yet 
-- Add dashboard link to user menu
-- Use dashboard link if enrolled lms course hasn't started yet
-- Urls keys added in profile_url settings 
-
 ### Added
 
 - Add a "person" variant to the glimpse plugin for persons without a page
@@ -20,6 +15,8 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 - Allow overriding a person's bio on the person plugin
 - Add new section on the course detail page to display related programs
 - Display "code" on the course detail page and allow frontend editing
+- Add dashboard link to user menu
+- Use dashboard link if enrolled lms course hasn't started yet
 
 ### Changed
 
@@ -28,6 +25,8 @@ Versioning](https://semver.org/spec/v2.0.0.html).
   with insufficient permissions
 - Improve UX of course search pagination by avoiding truncation of page number
   when it is not relevant
+- Hide "Go to course" button if lms course hasn't started yet
+- Add Urls `key` property in profile_url settings
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,8 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 ## [Unrealeased]
 
 - Hide "Go to course" button if lms course hasn't started yet 
-- Add dashboard link to user menu 
+- Add dashboard link to user menu
+- Use dashboard link if enrolled lms course hasn't started yet
 
 ### Added
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 - Hide "Go to course" button if lms course hasn't started yet 
 - Add dashboard link to user menu
 - Use dashboard link if enrolled lms course hasn't started yet
+- Urls keys added in profile_url settings 
 
 ### Added
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 ## [Unrealeased]
 
 - Hide "Go to course" button if lms course hasn't started yet 
+- Add dashboard link to user menu 
 
 ### Added
 
--- a/src/frontend/js/components/CourseRunEnrollment/index.spec.tsx
+++ b/src/frontend/js/components/CourseRunEnrollment/index.spec.tsx
@@ -56,6 +56,7 @@ describe('<CourseRunEnrollment />', () => {
     id: courseRun.id,
     resource_link: courseRun.resource_link,
     priority: courseRun.state.priority,
+    starts_in_message: courseRun.starts_in_message,
   });
 
   afterEach(() => {
@@ -175,6 +176,40 @@ describe('<CourseRunEnrollment />', () => {
 
     screen.getByRole('link', { name: 'Go to course' });
     screen.getByText('You are enrolled in this course run');
+  });
+
+  it("shows remaining course opening time and a link to the lms dashboard if the user is already enrolled and if the course hasn't started yet", async () => {
+    const username = initializeUser();
+    const courseRun: CourseRun = factories.CourseRunFactory.generate();
+    courseRun.state.priority = 0;
+    courseRun.starts_in_message = 'The course will start in 3 days';
+
+    const courseRunDeferred = new Deferred();
+    fetchMock.get(`/api/v1.0/course-runs/${courseRun.id}/`, courseRunDeferred.promise);
+    const enrollmentsDeferred = new Deferred();
+    fetchMock.get(
+      `${endpoint}/api/enrollment/v1/enrollment/${username},${courseRun.resource_link}`,
+      enrollmentsDeferred.promise,
+    );
+
+    render(
+      <IntlProvider locale="en">
+        <SessionProvider>
+          <CourseRunEnrollment context={contextProps} courseRun={getCourseRunProp(courseRun)} />
+        </SessionProvider>
+      </IntlProvider>,
+    );
+    screen.getByRole('status', { name: 'Loading enrollment information...' });
+
+    await act(async () => {
+      courseRunDeferred.resolve(courseRun);
+      enrollmentsDeferred.resolve({ is_active: true });
+    });
+
+    expect(screen.queryByRole('link', { name: 'Go to course' })).toBeNull();
+    // screen.getByRole('link', { name: 'Go to dashboard' });
+    screen.getByText('You are enrolled in this course run');
+    screen.getByText('The course will start in 3 days');
   });
 
   it('shows a helpful message if the course run is closed', async () => {
--- a/src/frontend/js/components/CourseRunEnrollment/index.spec.tsx
+++ b/src/frontend/js/components/CourseRunEnrollment/index.spec.tsx
@@ -57,6 +57,7 @@ describe('<CourseRunEnrollment />', () => {
     resource_link: courseRun.resource_link,
     priority: courseRun.state.priority,
     starts_in_message: courseRun.starts_in_message,
+    dashboard_link: courseRun.dashboard_link,
   });
 
   afterEach(() => {
@@ -183,6 +184,7 @@ describe('<CourseRunEnrollment />', () => {
     const courseRun: CourseRun = factories.CourseRunFactory.generate();
     courseRun.state.priority = 0;
     courseRun.starts_in_message = 'The course will start in 3 days';
+    courseRun.dashboard_link = 'https://edx.local.dev:8073/dashboard';
 
     const courseRunDeferred = new Deferred();
     fetchMock.get(`/api/v1.0/course-runs/${courseRun.id}/`, courseRunDeferred.promise);
@@ -207,8 +209,10 @@ describe('<CourseRunEnrollment />', () => {
     });
 
     expect(screen.queryByRole('link', { name: 'Go to course' })).toBeNull();
-    // screen.getByRole('link', { name: 'Go to dashboard' });
-    screen.getByText('You are enrolled in this course run');
+    expect(screen.getByText('You are enrolled in this course run')).toHaveAttribute(
+      'href',
+      'https://edx.local.dev:8073/dashboard',
+    );
     screen.getByText('The course will start in 3 days');
   });
 
--- a/src/frontend/js/components/CourseRunEnrollment/index.tsx
+++ b/src/frontend/js/components/CourseRunEnrollment/index.tsx
@@ -57,6 +57,7 @@ interface CourseRunEnrollmentProps {
     resource_link: string;
     priority: number;
     starts_in_message: Nullable<string>;
+    dashboard_link: string;
   };
 }
 
@@ -236,7 +237,7 @@ export const CourseRunEnrollment: React.FC<CourseRunEnrollmentProps & CommonData
         <React.Fragment>
           {courseRun.starts_in_message ? (
             <div>
-              <a href={courseRun.resource_link} className="course-run-enrollment__cta">
+              <a href={courseRun.dashboard_link} className="course-run-enrollment__cta">
                 <FormattedMessage {...messages.enrolled} />
               </a>
               <div className="course-run-enrollment__helptext">{courseRun.starts_in_message}</div>
--- a/src/frontend/js/components/CourseRunEnrollment/index.tsx
+++ b/src/frontend/js/components/CourseRunEnrollment/index.tsx
@@ -56,6 +56,7 @@ interface CourseRunEnrollmentProps {
     id: number;
     resource_link: string;
     priority: number;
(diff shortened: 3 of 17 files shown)


src/frontend/js/components/CourseRunEnrollment/index.spec.tsx
```diff
@@ -177,6 +180,38 @@ describe('<CourseRunEnrollment />', () => {
...
+    );
+    screen.getByRole('status', { name: 'Loading enrollment information...' });
+
+    await act(async () => {
+      courseRunDeferred.resolve(courseRun);
+      enrollmentsDeferred.resolve({ is_active: true });
+    });
+
+    expect(screen.queryByRole('link', { name: 'Go to course' })).toBeNull();
+    screen.getByText('You are enrolled in this course run');
```
sampaccoud, line 212: In terms of Ux, do we need to precise that we are now waiting for the course to open or will the user understand?

src/frontend/js/components/CourseRunEnrollment/index.tsx
```diff
@@ -233,9 +234,11 @@ export const CourseRunEnrollment: React.FC<CourseRunEnrollmentProps & CommonData
     case step === Step.ENROLLED:
       return (
         <React.Fragment>
-          <a href={courseRun.resource_link} className="course-run-enrollment__cta">
-            <FormattedMessage {...messages.goToCourse} />
-          </a>
+          {new Date().toISOString() >= courseRun.start ? (
```
sampaccoud, line 237: Is this timezone aware?

src/frontend/js/components/CourseRunEnrollment/index.spec.tsx
kernicPanel: Just above the hidden button, start date is already displayed.
I can't post a screenshot right now, but I'll do it tomorrow.
I think it's understandable without additional message.

src/frontend/js/components/CourseRunEnrollment/index.tsx
kernicPanel: Hmm, good point, I'll check this.
In fact I was using timestamp, but I've seen that ISO strings were used elsewhere.
If you're OK, I can use timestamp instead.

src/richie/apps/courses/templatetags/extra_tags.py
```diff
@@ -235,6 +235,7 @@ def course_enrollment_widget_props(context):
                 "id": course_run.id,
                 "resource_link": course_run.resource_link,
                 "priority": course_run.state["priority"],
+                "start": course_run.start.isoformat(),
```
kernicPanel, line 238: @sampaccoud @jbpenrath what do you think about replacing `start` attribute with `has_started` or `starts_in` : 
both permits to replace frontend test `new Date().toISOString() >= courseRun.start` and prevent timezone issues.
`has_started` would be a boolean.
`starts_in` would be remaining days before the course start, and we could display this value in front

sampaccoud: Smart and seems legit. I'll let @jbpenrath comment on the possibility to compare this date to now in the frontend (since the isoformat returned by the backend is timezone aware, it should be feasible).

kernicPanel: In fact, I don't see any reason to compare user local time with the course start datetime

sampaccoud: If the local time is timezone aware you can compare them to know if the course is open at a given time. But I agree that it is easier on the backend...

src/frontend/js/components/CourseRunEnrollment/index.spec.tsx
kernicPanel: image
Currently, it looks likes this

src/richie/apps/courses/templatetags/extra_tags.py
jbpenrath: You are right @kernicPanel, as we only use start date to check if course started, backend could be in charge to send this information. I think it is a good idea to send `starts_in` and add a message below `You are enrolled in this course run`: `The course will start in X days`.

src/frontend/js/components/CourseRunEnrollment/index.spec.tsx
sampaccoud: oh ok! Then I think we should provide a link on this phrase that points to the user's dashboard (see profile menu) so that he/she can go see the courses to which he/she is enrolled. WDYT?

src/richie/apps/courses/templatetags/extra_tags.py
kernicPanel: ok, great ! Thanks for pointing that lib @jbpenrath

src/frontend/js/components/CourseRunEnrollment/index.spec.tsx
kernicPanel: Good for me, any thought @jbpenrath ?

jbpenrath: Yep I agree with @sampaccoud, good idea to allow user to access to its dashboard and see its enrollments.

kernicPanel:
Can you give me your toughs on this rewrite @sampaccoud @jbpenrath please ?
User's dashboard link is still missing, I'm on it.

src/frontend/js/components/CourseRunEnrollment/index.spec.tsx
kernicPanel: In the menu, I see a ink like this `https://edx.local.dev:8073/u/edx` but, you want this one `https://edx.local.dev:8073/dashboard` , right ?

src/frontend/i18n/locales/es-ES.json
```diff
@@ -35,6 +35,10 @@
     "description": "Helper text in the enroll button for non logged in users",
     "message": "Log in to enroll"
   },
+  "components.CourseRunEnrollment.courseStartsInDays": {
+    "description": "Helper text for days remaining before course starts",
+    "message": "The course will start in {days} day(s)"
```
sampaccoud: you can pluralize with ReactIntl. See https://github.com/openfun/richie/blob/1e5279fadb8e01df46f65a355567aa90720493d2/src/frontend/i18n/locales/fr-FR.json#L140

kernicPanel:
image
Now, message is generated by django, dashboard url still need to be added

src/richie/apps/courses/templatetags/extra_tags.py
```diff
@@ -229,12 +234,21 @@ def course_enrollment_widget_props(context):
     """
     course_run = context["run"]
 
+    starts_in_message = None
+    if course_run.start > timezone.now():
+        course_start = arrow.get(course_run.start)
+        humanized_course_start = course_start.humanize(
+            arrow.now(), locale=to_locale(get_language())
+        )
+        starts_in_message = f"{_('The course will start')} {humanized_course_start}"
```
sampaccoud, line 243: Having doubts whether we can presume that order will be the same in every language... If not this is safer:
```suggestion
        starts_in_message = _("The course will start {:s}'').format(humanized_course_start)
```

tests/apps/courses/test_templatetags_extra_tags_course_enrollment_widget_props.py
```diff
@@ -20,19 +24,25 @@ def test_course_enrollment_widget_props_tag(self):
...
-        self.assertEqual(
-            course_enrollment_widget_props(context),
-            json.dumps(
-                {
-                    "courseRun": {
-                        "id": course_run.id,
-                        "resource_link": course_run.resource_link,
-                        "priority": course_run.direct_course.state["priority"],
+        now = timezone.now()
+        with mock.patch.object(timezone, "now", return_value=now):
```
sampaccoud, line 28: I would say that for clarity of intention, and avoiding side effects, this mock should only wrap the call to `course_enrollment_widget_props`

src/frontend/js/components/CourseRunEnrollment/index.tsx
```diff
@@ -56,6 +56,7 @@ interface CourseRunEnrollmentProps {
     id: number;
     resource_link: string;
     priority: number;
+    starts_in_message: string;
```
jbpenrath, line 59: Since this property can be null if the course started, the correct type should be:
```suggestion
    starts_in_message: Nullable<string>;
```

jbpenrath (commented):
Looks good to me at this point.

The link to the dashboard is related to the LMS, so I think we have to add a property into the `LMS_BACKEND` and pass it to the frontend through the `context_processor`. No better idea to propose for the moment. @sampaccoud WDYT?

sampaccoud:
> The link to the dashboard is related to the LMS, so I think we have to add a property into the `LMS_BACKEND` and pass it to the frontend through the `context_processor`. No better idea to propose for the moment. @sampaccoud WDYT?

Yes I have better: it is already available to the front-end and available in the top menu right!?

jbpenrath:
> 
> 
> > The link to the dashboard is related to the LMS, so I think we have to add a property into the `LMS_BACKEND` and pass it to the frontend through the `context_processor`. No better idea to propose for the moment. @sampaccoud WDYT?
> 
> 
> 
> Yes I have better: it is already available to the front-end and available in the top menu right!? 
> 
> 

Not the dashboard (only user profile and settings) but it is  very easy to add. But I think about that and this link is related to the LMS not the Authentication even if today both are EDX.

sampaccoud:
> > > The link to the dashboard is related to the LMS, so I think we have to add a property into the `LMS_BACKEND` and pass it to the frontend through the `context_processor`. No better idea to propose for the moment. @sampaccoud WDYT?
> > 
> > 
> > Yes I have better: it is already available to the front-end and available in the top menu right!?
> 
> Not the dashboard (only user profile and settings) but it is very easy to add. But I think about that and this link is related to the LMS not the Authentication even if today both are OpenEdX.

Well it will never be related to each LMS because there can only be one dashboard for the user:
- either you have 1 LMS like now and define it in settings,
- or you have several LMSes and you need to use Joanie as a broker and the dashboard will be in Joanie and you also define it in settings.

Does it make sense?

kernicPanel:
@sampaccoud @jbpenrath I think you can review this once again.

tests/apps/courses/test_templatetags_extra_tags_course_enrollment_widget_props.py
kernicPanel: Is it ok like this, or should I extract `course_enrollment_widget_props(context),` from assertEqual to wrap it alone in `course_enrollment_widget_props` ?

CHANGELOG.md
```diff
@@ -8,6 +8,11 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unrealeased]
 
+- Hide "Go to course" button if lms course hasn't started yet 
+- Add dashboard link to user menu
+- Use dashboard link if enrolled lms course hasn't started yet
+- Urls keys added in profile_url settings
```
jbpenrath, line 14:
```suggestion
- Add Urls `key` property in profile_url settings 
```

src/frontend/js/components/CourseRunEnrollment/index.tsx
```diff
@@ -233,12 +235,23 @@ export const CourseRunEnrollment: React.FC<CourseRunEnrollmentProps & CommonData
...
-            <FormattedMessage {...messages.goToCourse} />
-          </a>
-          <div className="course-run-enrollment__helptext">
-            <FormattedMessage {...messages.enrolled} />
-          </div>
+          {courseRun.starts_in_message ? (
+            <div>
+              <a href={courseRun.dashboard_link} className="course-run-enrollment__cta">
+                <FormattedMessage {...messages.enrolled} />
+              </a>
```
jbpenrath, line 242: If `dashboard_link` is null, we could not wrap `<FormattedMessage />` into a link.

src/frontend/js/types/index.ts
```diff
@@ -7,6 +7,8 @@ export interface CourseRun {
   enrollment_end: string;
   languages: string[];
   state: CourseState;
+  starts_in_message: string;
```
jbpenrath, line 10: Both properties can be null.
```suggestion
  starts_in_message: Nullable<string>;
  dashboard_link: Nullable<string>;
```

src/frontend/js/components/CourseRunEnrollment/index.tsx
```diff
@@ -56,6 +56,8 @@ interface CourseRunEnrollmentProps {
     id: number;
     resource_link: string;
     priority: number;
+    starts_in_message: Nullable<string>;
+    dashboard_link: string;
```
jbpenrath, line 60:
```suggestion
    dashboard_link: Nullable<string>;
```

src/frontend/js/components/UserLogin/index.tsx
```diff
@@ -93,11 +93,13 @@ export const UserLogin = ({ profileUrls = [] }: UserLoginProps) => {
           user={{
             ...user,
             urls: [
-              ...profileUrls.map(({ label, action }) => ({
+              ...profileUrls.map(({ key, label, action }) => ({
```
jbpenrath, line 96: `profileUrls` is no more an array, this is an object.
To adapt the code to this new format, you could write:
```suggestion
              ...Object.entries(profileUrls).map(([key, { label, action }]) => ({
```

You also have to update Component Props type.

src/richie/apps/courses/templatetags/extra_tags.py
```diff
@@ -229,12 +234,29 @@ def course_enrollment_widget_props(context):
...
+
+    starts_in_message = None
+    if course_run.start > timezone.now():
+        course_start = arrow.get(course_run.start)
+        humanized_course_start = course_start.humanize(
+            arrow.now(), locale=to_locale(get_language())
+        )
+        starts_in_message = _("The course will start {:s}").format(
+            humanized_course_start
+        )
```
sampaccoud, line 250: I think you should name the variable and add a comment for translators explaining what it looks like because it is not obvious:
```suggestion
        # Translators: delay indicates when the course will start as a duration. In english the string will be "The course will start in 3 days"
        starts_in_message = _("The course will start {delay:s}").format(
            delay=humanized_course_start
        )
```

See https://docs.djangoproject.com/fr/2.2/topics/i18n/translation/#comments-for-translators

sampaccoud (approved):
image

src/frontend/js/components/UserLogin/index.tsx
kernicPanel: :man_facepalming:

src/frontend/js/components/CourseRunEnrollment/index.tsx
```diff
@@ -233,12 +235,29 @@ export const CourseRunEnrollment: React.FC<CourseRunEnrollmentProps & CommonData
...
-            <FormattedMessage {...messages.enrolled} />
-          </div>
+          {courseRun.starts_in_message ? (
+            <div>
+              {courseRun.dashboard_link ? (
+                <a href={courseRun.dashboard_link} className="course-run-enrollment__cta">
+                  <FormattedMessage {...messages.enrolled} />
+                </a>
+              ) : (
+                <div className="course-run-enrollment__helptext">
```
jbpenrath, line 245: More semantically correct.
```suggestion
                <p className="course-run-enrollment__helptext">
```

src/frontend/js/components/CourseRunEnrollment/index.tsx
```diff
@@ -233,12 +235,29 @@ export const CourseRunEnrollment: React.FC<CourseRunEnrollmentProps & CommonData
...
+                  <FormattedMessage {...messages.enrolled} />
+                </div>
+              )}
+              <div className="course-run-enrollment__helptext">{courseRun.starts_in_message}</div>
```
jbpenrath, line 249: More semantically correct.
```suggestion
              <p className="course-run-enrollment__helptext">{courseRun.starts_in_message}</div>
```

jbpenrath (approved):
A last one comment about semantic html tag.

Otherwise, looks good to me ! 🚀

Issue #1270, opened 2021-02-01
Add a configurable LTI plugin


sampaccoud:
## Feature Request

**Is your feature request related to a problem or unsupported use case? Please describe.**
We want to use [Marsha](https://github.com/openfun/marsha) to host our teaser videos on course detail pages.

**Describe the solution you'd like**
Add an LTI plugin to https://github.com/openfun/richie/tree/master/src/richie/plugins.

This plugin should have the same features as our [OpenEdX configurable LTI xblock](https://github.com/openfun/xblock-configurable-lti-consumer):
- configurable via Django's settings
- able to generate and propose unique launch url when creating a new plugin (see [exrex](https://github.com/openfun/xblock-configurable-lti-consumer/blob/1e63ddcfabf0f4b347f6cec4ba4aa0015c728ac4/configurable_lti_consumer/configurable_lti_consumer.py#L67))
- able to resize the consumer iframe according to its content (see [iframe resizer](https://github.com/openfun/xblock-configurable-lti-consumer/tree/1e63ddcfabf0f4b347f6cec4ba4aa0015c728ac4/configurable_lti_consumer/static/js))

The LTI role should be "instructor" in edition mode and "student" otherwise.

Open questions:
- What should the lti_id be?  
    I propose to generate a uuid in a readonly field of the plugin (it will thus be shared by the draft of public version of the plugin contrary to the `id` primary key)
- What should the context_id be?
    I propose to use the course's code. The next question being: how do we relate to a course from any page? I think we need a ForeignKey to a course on the plugin which we can then try to fill in smartly when computing the CMSPlugin context.

**Discovery, Documentation, Adoption, Migration Strategy**
Here are 2 examples of handcrafting LTI requests:
- for [Ashley](https://github.com/openfun/ashley): https://github.com/openfun/ashley/blob/v1.0.0-beta.4/tests/ashley/test_auth_backend.py
- for [Marsha](https://github.com/openfun/marsha): https://github.com/openfun/videofront-to-marsha/blob/master/transfer.py#L80

PR #1274, opened 2021-02-08
✨(lti-consumer) add LTI consumer plugin

kernicPanel:
## Purpose

Users need to use marsha as LTI video provider in course teaser placeholder

## Proposal

A configurable generic LTI consumer plugin has been added.
Configuration may allow any LTI content to be added anywhere,
but currently, only marsha videos are allowed in course teaser placeholder

#1270


src/richie/plugins/lti_consumer/cms_plugins.py
```diff
@@ -0,0 +1,59 @@
...
+
+from richie.apps.core.defaults import PLUGINS_GROUP
+
+from .models import LTIConsumer
+
+
+@plugin_pool.register_plugin
+class LTIConsumerPlugin(CMSPluginBase):
+    """
+    A plugin to add plain text.
```
sampaccoud, line 17: :raised_hand_with_fingers_splayed:

src/richie/plugins/lti_consumer/cms_plugins.py
```diff
@@ -0,0 +1,59 @@
...
+    #         form.base_fields["body"].validators.append(MaxLengthValidator(max_length))
+    #
+    #     return form
+    #
+    def render(self, context, instance, placeholder):
+        """
+        Build plugin context passed to its template to perform rendering
+        """
+        context = super().render(context, instance, placeholder)
+        context["url"] = instance.url
```
sampaccoud, line 58: or pass the whole instance?

tests/plugins/lti_consumer/test_cms_plugins.py
```diff
@@ -0,0 +1,61 @@
...
+            course=lti_consumer.course,
+        )
+        plugin_instance = model_instance.get_plugin_class_instance()
+        context = plugin_instance.render({}, model_instance, None)
+
+        # Check if "instance" is in context
+        self.assertIn("instance", context)
+
+        # Check if parameters, generated by the factory, are correctly set in "instance" of context
+        self.assertEqual(context["instance"].url, lti_consumer.url)
```
sampaccoud, line 54: weird if it works as is :wink:

tests/plugins/lti_consumer/test_cms_plugins.py
```diff
@@ -0,0 +1,61 @@
...
+            "resource_link_id": "Marsha Video",
+            "oauth_consumer_key": "InsecureOauthConsumerKey",
+            "oauth_signature_method": "InsecureSharedSecret",
+            "oauth_timestamp": True,
+            "oauth_nonce": True,
+            "oauth_version": 0.5625,
+            "oauth_signature": 0.5625,
+            "oauth_callback": 0.5625,
+        }
+    )
```
sampaccoud, line 30: all this won't come from settings. The setting should get the same parameters as the one in our [LTI configurable xblock](https://www.google.com/url?sa=t&rct=j&q=&esrc=s&source=web&cd=&cad=rja&uact=8&ved=2ahUKEwinns3b6truAhWTD-wKHSdeAbYQFjAAegQIAhAC&url=https%3A%2F%2Fgithub.com%2Fopenfun%2Fxblock-configurable-lti-consumer&usg=AOvVaw2-e1rNRw2DI97h2_HwLx0D)

About the params listed here:

- "lti_message_type": not sure but could come from the setting
- "lti_version": probably 1.1 since we don't need advanced features
- "resource_link_id": use the plugin instance id?
- "oauth_consumer_key":  comes from a secret setting (see our LTI configurable xblock)
- "oauth_signature_method", "oauth_timestamp", "oauth_nonce", "oauth_version", "oauth_signature": generated when computing the signature

src/richie/plugins/lti_consumer/templates/richie/lti_consumer/lti_consumer.html
```diff
@@ -0,0 +1,28 @@
+<form
+    id="lti_form"
+    action="{{ url }}"
+    method="POST"
+    encType="application/x-www-form-urlencoded"
+    style="display:none;"
+    target="launch_iframe"
```
jbpenrath, line 7: Should be `lti_iframe`?

jbpenrath (commented):
Nothing to add related to @sampaccoud comments at this point 🙂

src/richie/plugins/lti_consumer/cms_plugins.py
kernicPanel: Instance is already passed in CMSPluginBase.render()

```python
    def render(self, context, instance, placeholder):
        context['instance'] = instance
        context['placeholder'] = placeholder
        return context
```
In fact overriding render is useless for now.
But do you prefer to have it explicitly defined at the plugin level ?

tests/plugins/lti_consumer/test_cms_plugins.py
kernicPanel: It is, see my previous comment :)

src/richie/plugins/lti_consumer/templates/richie/lti_consumer/lti_consumer.html
kernicPanel: Oh, yes thanks :)

src/richie/plugins/lti_consumer/models.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+class LTIConsumer(CMSPlugin):
+    """
+    LTI consumer plugin model.
+    """
+
+    url = models.URLField(_("LTI url"))
+    course = models.ForeignKey(to=Course, on_delete=models.PROTECT)
+
```
sampaccoud, line 19: Note that you also need a field to point to one of the LTI configurations in settings.
We will need to force this field for some placeholders (see what we did to force CKEditor configuration on the SimpleText plugin) so that a placeholder can only receive an LTI plugin targetting Marsha and not some other LTI service that we don't want to see here...

kernicPanel:
@sampaccoud can you do a quick check on tests/plugins/lti_consumer/test_cms_plugins.py specially the settings and build parameters, please ?

tests/plugins/lti_consumer/test_cms_plugins.py
```diff
@@ -0,0 +1,172 @@
...
+            course=lti_consumer.course,
+        )
+        plugin_instance = model_instance.get_plugin_class_instance()
+        context = plugin_instance.render({}, model_instance, None)
+
+        # Check if "instance" is in context
+        self.assertIn("instance", context)
+
+        # Check if parameters, generated by the factory, are correctly set in "instance" of context
+        self.assertEqual(context["instance"].url, lti_consumer.url)
```
sampaccoud, line 57: Or straightaway do:
```suggestion
        # Check if instance is in context
        self.assertEqual(context["instance"], model_instance)
```

tests/plugins/lti_consumer/test_cms_plugins.py
```diff
@@ -0,0 +1,172 @@
...
+
+    def test_cms_plugins_lti_auth_parameters(self):
+        lti_consumer = LTIConsumerFactory()
+        expected_auth_parameters = {
+            "lti_message_type": settings.LTI_CONSUMER.get("display_name"),
+            "lti_version": "LTI-1p0",
+            "resource_link_id": str(lti_consumer.id),
+            "context_id": lti_consumer.course.code,
+            "user_id": "richie",
+            "lis_person_contact_email_primary": "fun.dev@fun-mooc.fr",
```
sampaccoud, line 77: better leave it blank :wink:

tests/plugins/lti_consumer/test_cms_plugins.py
```diff
@@ -0,0 +1,172 @@
...
+            "roles": "instructor",
```
sampaccoud, line 78: It should only be instructor when in edit mode

tests/plugins/lti_consumer/test_cms_plugins.py
```diff
@@ -0,0 +1,182 @@
...
+        return_value=(
+            "uri",
+            {
+                "Authorization": (
+                    'OAuth oauth_nonce="80966668944732164491378916897", '
+                    'oauth_timestamp="1378916897", '
+                    'oauth_version="1.0", oauth_signature_method="HMAC-SHA1", '
+                    'oauth_consumer_key="InsecureOauthConsumerKey", '
+                    'oauth_signature="frVp4JuvT1mVXlxktiAUjQ7%2F1cw%3D"'
+                )
```
sampaccoud, line 93: I don't think you should mock this but let it be computed and then check it because the entry parameters are known. Maybe take a look at this test suite: https://github.com/openfun/marsha/blob/ee6e325aa84ee1d65dc1c6b49019dcc80c9d6616/src/backend/marsha/core/tests/test_lti.py

src/richie/plugins/lti_consumer/models.py
```diff
@@ -0,0 +1,91 @@
...
+    """
+
+    url = models.URLField(_("LTI url"))
+    course = models.ForeignKey(to=Course, on_delete=models.PROTECT)
+    edit_mode = False
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        if not self.url:
+            self.url = getattr(settings, "LTI_CONSUMER", {}).get("base.url")
```
kernicPanel, line 28: I'm not sure about this approach…

src/richie/plugins/lti_consumer/models.py
kernicPanel: It might be ok now

tests/plugins/lti_consumer/test_cms_plugins.py
kernicPanel: It's fixed now

kernicPanel:
This PR should be barely OK, but maybe we can add some tests on generated url.
For now automatic iframe resizing is missing.

@sampaccoud and @jbpenrath, I think you can review this as I work on resizing.

https://media.giphy.com/media/l4FATJpd4LWgeruTK/giphy.gif

src/richie/plugins/lti_consumer/models.py
```diff
@@ -0,0 +1,91 @@
...
+        super().__init__(*args, **kwargs)
+        if not self.url:
+            self.url = getattr(settings, "LTI_CONSUMER", {}).get("base_url")
+
+    @property
+    def auth_parameters(self):
+        """
+        Builds required parameters for LTI authentication
+        """
+        role = "instructor" if self.edit_mode else "student"
```
sampaccoud, line 35: can you declare "instructor" and "student" as constants in defaults.py?

src/richie/plugins/lti_consumer/models.py
```diff
@@ -0,0 +1,91 @@
...
+        return {
+            "lti_message_type": settings.LTI_CONSUMER.get("display_name"),
```
sampaccoud, line 37: if lti_service is None, we probably need a field on the plugin for this info as well? Ideally, these fields need to disappear when an lti_service is selected from the list of predefined services ;-)

src/frontend/js/components/LTIConsumer/index.spec.tsx
```diff
@@ -0,0 +1,148 @@
...
+    //   '<!DOCTYPE html><html lang=""><body><p>It works !!</p></body></html>',
+    // );
+
+    // HTMLFormElement.prototype.submit = jest.fn();
+
+    HTMLFormElement.prototype.submit = () => {
+      const iframeDocument = document.getElementsByTagName('iframe')[0].contentDocument!;
+      iframeDocument.open();
+      iframeDocument.write(
+        '<!DOCTYPE html><html lang=""><body>' +
```
jbpenrath, line 52: You could use a template string if you want
```suggestion
        `<!DOCTYPE html><html lang=""><body>
            <p style="height: 400px;">It works !!</p>
          </body></html>`,
```

src/frontend/js/components/LTIConsumer/index.spec.tsx
```diff
@@ -0,0 +1,148 @@
...
+    const { container } = render(
+      <IntlProvider locale="en">
+        <LTIConsumer {...ltiConsumerProps} />
+      </IntlProvider>,
+    );
+
+    // check form inputs
+    Object.entries(ltiConsumerProps.contentParameters).forEach(([name, value]) => {
+      const input = container.querySelector(`input[name=${name}]`);
+      expect(input).not.toBeNull();
```
jbpenrath, line 100:
```suggestion
      expect(input).toBeInstanceOf(HTMLInputElement);
```

src/frontend/js/components/LTIConsumer/index.spec.tsx
```diff
@@ -0,0 +1,148 @@
...
+    // });
+
+    // const iframe: HTMLIFrameElement = container.getElementsByTagName('iframe')[0];
+    // await waitFor(() => {
+    //   expect(iframe.height).toEqual(400);
+    // }, {timeout: 5000});
+
+    // check if iframeresizer does its job
+    const iframe: HTMLIFrameElement = container.getElementsByTagName('iframe')[0];
+    expect(iframe.id).not.toBeNull();
```
jbpenrath, line 143:
```suggestion
  expect(iframe.id).toMatch(/^iFrameResizer[0-9]$/);
```

src/frontend/js/components/LTIConsumer/index.tsx
```diff
@@ -0,0 +1,38 @@
...
+import { LTIConsumer as LTIConsumerProps } from 'types/LTIConsumer';
+
+export const LTIConsumer = ({ url, contentParameters }: LTIConsumerProps) => {
+  const formRef = React.useRef<HTMLFormElement>(null);
+  const iframeRef = React.useRef<HTMLIFrameElement>(null);
+
+  useEffect(() => {
+    formRef.current?.submit();
+    iframeResizer({ sizeWidth: true }, iframeRef.current!);
+  });
```
jbpenrath, line 12:
```suggestion
  }, []);
```

You should specify an empty array as second argument of your `useEffect` hook elsewhere the function will be trigger at each renders. We would like to trigger it only when the component is mounted.

src/frontend/js/components/Root/index.tsx
```diff
@@ -23,6 +24,7 @@ interface ComponentLibrary {
   Search: typeof Search;
   SearchSuggestField: typeof SearchSuggestField;
   UserLogin: typeof UserLogin;
+  LTIConsumer: typeof LTIConsumer;
```
jbpenrath, line 27: Can you sort it alphabetically ?

src/frontend/js/components/Root/index.tsx
```diff
@@ -32,6 +34,7 @@ const componentLibrary: ComponentLibrary = {
   Search,
   SearchSuggestField,
   UserLogin,
+  LTIConsumer,
```
jbpenrath, line 37: Ditto

src/frontend/package.json
```diff
@@ -80,6 +80,7 @@
     "faker": "5.4.0",
     "fetch-mock": "9.11.0",
     "glob": "7.1.6",
+    "iframe-resizer": "^4.3.1",
```
jbpenrath, line 83: We always use pinned dependency, it avoids some bad suprises 😅
```suggestion
    "iframe-resizer": "4.3.1",
```

src/richie/plugins/lti_consumer/templates/richie/lti_consumer/lti_consumer.html
```diff
@@ -0,0 +1,4 @@
+<div
+  class="richie-react richie-react--l-t-i-consumer"
```
jbpenrath, line 2:
```suggestion
  class="richie-react richie-react--lti-consumer"
```

Nope?

kernicPanel:
@sampaccoud @jbpenrath I think we are finally ok for this PR. :sweat_smile:

src/richie/plugins/lti_consumer/models.py
```diff
@@ -0,0 +1,119 @@
...
+from .defaults import INSTRUCTOR, STUDENT
+
+
+class LtiConsumer(CMSPlugin):
+    """
+    LTI consumer plugin model.
+    """
+
+    url = models.URLField(_("LTI url"), null=True, blank=True)
+    context_id = models.CharField(max_length=100, null=True, blank=True)
```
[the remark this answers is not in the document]
sampaccoud: Actually, on second thought, you don't need this field since we always want to send the active site... so you should get it while building the LTI request and just send it...

src/richie/plugins/lti_consumer/models.py
```diff
@@ -0,0 +1,119 @@
...
+        """
+        return getattr(settings, "RICHIE_LTI_PROVIDERS", {}).get(self.lti_provider_id)
+
+    def auth_parameters(self, edit=False):
+        """
+        Builds required parameters for LTI authentication
+        """
+        role = INSTRUCTOR if edit else STUDENT
+        return {
+            "lti_message_type": self.lti_provider.get("display_name"),
```
sampaccoud, line 70:
```suggestion
            "lti_message_type": self.lti_provider.get("display_name", self.lti_provider_id),
```

src/richie/plugins/lti_consumer/models.py
```diff
@@ -0,0 +1,119 @@
...
+                )
+        super().save(*args, **kwargs)
+
+    @staticmethod
+    def available_providers():
+        """
+        Returns available predefined LTI providers settings
+        """
+        return [
+            (provider_key, provider.get("display_name"))
```
sampaccoud, line 51: Avoid an empty display name?
```suggestion
    def available_providers(self):
        """
        Returns available predefined LTI providers settings
        """
        return [
            (provider_key, provider.get("display_name", self.lti_provider_id))
```

src/richie/plugins/lti_consumer/forms.py
```diff
@@ -0,0 +1,59 @@
+"""SimplePicture plugin forms."""
+from django import forms
+from django.utils.translation import gettext_lazy as _
+
+from .models import LtiConsumer
+
+
+class LtiConsumerForm(forms.ModelForm):
+    """Plugin form used to add a picture from the frontend admin."""
```
sampaccoud, line 9: copy/pasta fail

src/richie/plugins/lti_consumer/forms.py
```diff
@@ -0,0 +1,59 @@
...
+                        "custom",
+                        _("Choose a predefined LTI provider or fill fields below"),
+                    ),
+                    *LtiConsumer.available_providers(),
+                ],
+                attrs={"class": "form-control"},
+            ),
+        }
+
+    def clean(self):
```
sampaccoud, line 34: missing docstring

src/richie/plugins/lti_consumer/forms.py
```diff
@@ -0,0 +1,59 @@
...
+        if self.cleaned_data.get("lti_provider_id") == "custom":
```
sampaccoud, line 35: I think the custom LTI case should be when the `lti_provider_id` field is left blank.
```suggestion
        if self.cleaned_data.get("lti_provider_id"):
```

src/richie/plugins/lti_consumer/forms.py
```diff
@@ -0,0 +1,59 @@
...
+            missing_url = not self.cleaned_data.get("url")
+            missing_oauth_consumer_key = not self.cleaned_data.get("oauth_consumer_key")
+            missing_shared_secret = not self.cleaned_data.get("shared_secret")
+
+            if missing_url and missing_oauth_consumer_key and missing_shared_secret:
```
sampaccoud, line 40: All the error messages below should be translated because targeted to users.

src/richie/plugins/lti_consumer/cms_plugins.py
```diff
@@ -0,0 +1,89 @@
...
+
+    def render(self, context, instance, placeholder):
+        """
+        Build plugin context passed to its template to perform rendering
+        and pass edit mode
+        """
+        edit = False
+        if "request" in context:
+            request = context["request"]
+            edit = request.toolbar and request.toolbar.edit_mode_active
```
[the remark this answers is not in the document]
kernicPanel: actually id would be this (after black)
```
edit = (
    "request" in context
    and context["request"].toolbar
    and context["request"].toolbar.edit_mode_active
)
Do you prefer this ?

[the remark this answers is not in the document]
Comment (sampaccoud) [reply src/richie/plugins/lti_consumer/cms_plugins.py:72]: Personally yes... Although technically it is no more a one-liner 😢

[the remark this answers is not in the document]
Comment (kernicPanel) [reply src/richie/plugins/lti_consumer/cms_plugins.py:72]: Another option :
```python
        try:
            edit = context["request"].toolbar.edit_mode_active
        except KeyError:
            edit = False
```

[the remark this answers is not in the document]
sampaccoud: Then catch (KeyError, AttributeError) no?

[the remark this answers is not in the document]
kernicPanel: I tried some fake assignations like 
```python
edit = context["request"].toolbar_fail.edit_mode_active
```
and 
```python
edit = context["request"].toolbar.edit_mode_active_fail
```
AttributeError isn't raised in those cases.
Any idea of something that will raise it ?

src/richie/plugins/lti_consumer/forms.py
kernicPanel: I thought it would be clearer to have a "custom" value added to the field select, but leaving it empty is ok for me.

src/richie/plugins/lti_consumer/models.py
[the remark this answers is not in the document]
kernicPanel: But in case of multiple site, we couldn't retrieve lti content across all sites, no ?

src/richie/plugins/lti_consumer/models.py
kernicPanel: Ok for avoiding empty name, but making this function non static will cause problems, as it is used before instantiation.

src/richie/plugins/lti_consumer/cms_plugins.py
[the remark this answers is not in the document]
sampaccoud: the case of `if context["request"] ? if the request is None their will be an AttributeError. The earlier version is better and more explicit imo.

src/richie/plugins/lti_consumer/models.py
```diff
@@ -0,0 +1,119 @@
...
+
+from .defaults import INSTRUCTOR, STUDENT
+
+
+class LtiConsumer(CMSPlugin):
+    """
+    LTI consumer plugin model.
+    """
+
+    url = models.URLField(_("LTI url"), null=True, blank=True)
```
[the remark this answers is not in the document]
kernicPanel: Because in case of predefined lti provider, this field value is generated, so not required in the form.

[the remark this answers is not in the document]
sampaccoud: then you should have `null=False, blank=True` and set it in the `save` method.

src/richie/plugins/lti_consumer/models.py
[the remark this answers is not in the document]
sampaccoud: why not? anyway I'm not sure if a given page/placeholder/plugin can be part of multiple sites... but the idea is to avoid storing in db something that is already known (and could change by the way)...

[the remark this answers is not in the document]
kernicPanel: That's the point : if the current site domain can change and so the context_id, I suppose that LTI content couldn't be retrieved…
Storing context_id ensures that the one used for adding lti content will always be used in the student view.

src/richie/plugins/lti_consumer/models.py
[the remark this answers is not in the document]
kernicPanel: :facepalm: django 101 fail…

.circleci/config.yml
```diff
@@ -229,6 +229,9 @@ jobs:
           DJANGO_SETTINGS_MODULE: settings
           DJANGO_CONFIGURATION: Test
           DJANGO_SECRET_KEY: ThisIsAnExampleKeyForTestPurposeOnly
+          LTI_1_BASE_URL: http://localhost:8060/lti/videos/[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}
+          LTI_1_OAUTH_CONSUMER_KEY: InsecureOauthConsumerKey
+          LTI_1_SHARED_SECRET: InsecureSharedSecret
```
sampaccoud, line 234: I think we should use the LTI testing tool as example in the sandbox : https://lti.tools/saltire/
```
url: https://lti.tools/saltire/tp
consumer key: your choice
shared secret: secret
```

src/richie/plugins/lti_consumer/models.py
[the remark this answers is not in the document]
sampaccoud: I would say that in this case, we should have a setting with a list of predefined contexts to which an LTI consumer plugin can be associated :man_shrugging:  I find it weird to get the site info and save it in a field without any formatting. This is opening the way to context hell?

.circleci/config.yml
sampaccoud: About the LTI_1, I thought of what you said that you wanted to convey the fact that it allow more than one... but in the case of this sandbox setting, only one is possible via environment variables. So I'm not a big fan of these names... so since we will be pointing to the LTI test servi ce, I propose to name them: 
`LTI_TEST_BASE_URL`, `LTI_TEST_OAUTH_CONSUMER_KEY` and `LTI_TEST_SHARED_SECRET`.

kernicPanel: Shall I use a regex_url for https://lti.tools/saltire/tp ?

sampaccoud: well, there is no variable uuid in it :man_shrugging: so I would say no.

src/richie/plugins/lti_consumer/models.py
kernicPanel: This function has been removed now.

env.d/development/dev
```diff
@@ -5,3 +5,6 @@ AUTHENTICATION_BACKEND=base
 # LMS Backend
 EDX_BASE_URL=http://localhost:8073
 EDX_BACKEND=richie.apps.courses.lms.base.BaseLMSBackend
+
+# LTI Backend
+LTI_TEST_BASE_URL=http://localhost:8060/lti/videos/[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}
```
sampaccoud, line 10: do you need to repeat it here as you already set it in the common env file?

sandbox/settings.py
```diff
@@ -312,6 +312,29 @@ class Base(StyleguideMixin, DRFMixin, RichieCoursesConfigurationMixin, Configura
...
+                environ_name="LTI_TEST_OAUTH_CONSUMER_KEY",
+                environ_prefix=None,
+            ),
+            "shared_secret": values.Value(
+                "InsecureSharedSecret",
+                environ_name="LTI_TEST_SHARED_SECRET",
+                environ_prefix=None,
+            ),
+            "base_url": values.Value(
+                environ_name="LTI_TEST_BASE_URL", environ_prefix=None
```
sampaccoud, line 329: Why not put the LTI test url here as default and remove your configurations in env files and in the CI? This would be more coherent with what you do for consumer and secret keys.

sandbox/settings.py
```diff
@@ -569,6 +593,7 @@ class Development(Base):
 
     DEBUG = True
     ALLOWED_HOSTS = ["*"]
+    SECURE_REFERRER_POLICY = "unsafe-url"
```
sampaccoud, line 596: Can you explain and add a comment why this is necessary?

src/richie/plugins/lti_consumer/cms_plugins.py
```diff
@@ -0,0 +1,60 @@
...
+    cache = True
+    disable_child_plugins = True
+    form = LtiConsumerForm
+    model = LtiConsumer
+    module = PLUGINS_GROUP
+    name = _("LTI consumer")
+    render_template = "richie/lti_consumer/lti_consumer.html"
+
+    class Media:
+        js = ("lti_consumer/js/change_form.js",)
```
sampaccoud, line 33: Add a docstring explaining why this js file?

src/richie/plugins/lti_consumer/cms_plugins.py
```diff
@@ -0,0 +1,60 @@
...
+        return context
+
+    @staticmethod
+    def get_lti_consumer_widget_props(instance, edit=False):
+        """
+        Return all LTI consumer properties required by LtiConsumer React widget
+        """
+        return {
+            "url": instance.url,
+            "contentParameters": instance.content_parameters(edit=edit),
```
sampaccoud, line 59: I would keep it snake_case up to the template :man_shrugging:

src/richie/plugins/lti_consumer/cms_plugins.py
```diff
@@ -0,0 +1,60 @@
...
+from cms.plugin_pool import plugin_pool
+
+from richie.apps.core.defaults import PLUGINS_GROUP
+
+from .forms import LtiConsumerForm
+from .models import LtiConsumer
+
+
+@plugin_pool.register_plugin
+class LtiConsumerPlugin(CMSPluginBase):
```
sampaccoud, line 18: I think in marsha we chose to write LTI in capital letters, so maybe keep this pattern :man_shrugging:  `LTIConsumerPlugin`

src/richie/plugins/lti_consumer/factories.py
```diff
@@ -0,0 +1,20 @@
+"""
+LtiConsumer CMS plugin factories
+"""
+import factory
+
+from .models import LtiConsumer
+
+
+class LtiConsumerFactory(factory.django.DjangoModelFactory):
```
sampaccoud, line 9: ditto

src/richie/plugins/lti_consumer/forms.py
```diff
@@ -0,0 +1,54 @@
+"""SimplePicture plugin forms."""
+from django import forms
+from django.utils.translation import gettext_lazy as _
+
+from .models import LtiConsumer
+
+
+class LtiConsumerForm(forms.ModelForm):
```
sampaccoud, line 8: ditto

src/richie/plugins/lti_consumer/forms.py
```diff
@@ -0,0 +1,54 @@
+"""SimplePicture plugin forms."""
```
sampaccoud, line 1: copy/pasta fail

src/richie/plugins/lti_consumer/models.py
```diff
@@ -0,0 +1,122 @@
...
+
+import exrex
+from cms.models.pluginmodel import CMSPlugin
+from cms.utils import get_current_site
+from oauthlib import oauth1
+
+from .defaults import INSTRUCTOR, PREDEFINED_LTI_PROVIDERS, STUDENT
+
+
+class LtiConsumer(CMSPlugin):
```
sampaccoud, line 19: ditto

src/richie/plugins/lti_consumer/models.py
```diff
@@ -0,0 +1,122 @@
...
+    """
+    LTI consumer plugin model.
+    """
+
+    url = models.URLField(
```
sampaccoud, line 24: and it seems Django is also following this pattern :wink:  `URLField` vs `UrlField`

src/richie/plugins/lti_consumer/models.py
```diff
@@ -0,0 +1,122 @@
...
+        verbose_name=_("LTI url"),
+        help_text=_(
+            "For a predefined provider, leave this field empty for uploading new content"
```
sampaccoud, line 27: Not really "uploading new content"? :thinking: 
```suggestion
            "For a predefined provider, leave this field empty."
```

src/richie/plugins/lti_consumer/models.py
```diff
@@ -0,0 +1,122 @@
...
+    def save(self, *args, **kwargs):
+        """
+        Initialize attributes :
+        - context_id from current site domain
+        - url if predefined provider is used
+        """
+        if not self.id:
+            if self.lti_provider and not self.url:
+                self.url = "{:s}".format(
+                    exrex.getone(self.lti_provider.get("base_url")),
```
sampaccoud, line 53: This will fail if base_url is not set. Add a test?
```suggestion
                    exrex.getone(self.lti_provider.get("base_url", "")),
```

tests/plugins/lti_consumer/test_cms_plugins.py
```diff
@@ -0,0 +1,88 @@
...
+                'OAuth oauth_nonce="80966668944732164491378916897", oauth_timestamp="1378916897", '
+                'oauth_version="1.0", oauth_signature_method="HMAC-SHA1", '
+                'oauth_consumer_key="InsecureOauthConsumerKey", '
+                'oauth_signature="frVp4JuvT1mVXlxktiAUjQ7%2F1cw%3D"'
+            )
+        },
+    )
+    def test_cms_plugins_get_lti_consumer_widget_props(self, _):
+        """
+        Verify that LTI content consumption parameters are correctly build trough
```
sampaccoud, line 61:
```suggestion
        Verify that LTI content consumption parameters are correctly built trough
```

tests/plugins/lti_consumer/test_factories.py
```diff
@@ -0,0 +1,15 @@
+"""
+Factory tests
+"""
+from django.test import TestCase
+
+from richie.plugins.lti_consumer.factories import LtiConsumerFactory
+
+
+class LtiConsumerFactoriesTestCase(TestCase):
```
sampaccoud, line 9:
```suggestion
class LTIConsumerFactoriesTestCase(TestCase):
```

tests/plugins/lti_consumer/test_forms.py
```diff
@@ -0,0 +1,113 @@
+"""
+Forms tests
+"""
+from django.test import TestCase
+
+from richie.plugins.lti_consumer.forms import LtiConsumerForm
+
+
+class LtiConsumerFormTestCase(TestCase):
```
sampaccoud, line 9:
```suggestion
class LTIConsumerFormTestCase(TestCase):
```

tests/plugins/lti_consumer/test_forms.py
```diff
@@ -0,0 +1,113 @@
...
+            ],
+            LtiConsumerForm().fields["lti_provider_id"].widget.choices,
+        )
+
+    def test_forms_lti_consumer_clean_errors(self):
+        """
+        Verify that LTI consumer form is displaying errors with predefined or
+        custom provider
+        """
+        for data, errors in [
```
sampaccoud, line 29: :medal_sports: :wink:

tests/plugins/lti_consumer/test_models.py
```diff
@@ -0,0 +1,146 @@
...
+
+from richie.plugins.lti_consumer.models import LtiConsumer
+
+
+class LtiConsumerModelsTestCase(TestCase):
+    """Model tests case"""
+
+    def test_models_lti_consumer_with_predefined_provider(self):
+        """
+        HTML5lib Cleaner fix unclosed element.
```
sampaccoud, line 16: copy/pasta fail?

tests/plugins/lti_consumer/test_models.py
```diff
@@ -0,0 +1,146 @@
...
+        """
+        instance = LtiConsumer(lti_provider_id="lti_provider_test")
+
+        instance.save()
+
+        self.assertIsNotNone(instance.url)
+
+    def test_models_lti_consumer_auth_parameters_no_edit_mode(self):
+        """
+        Verify that student LTI authentication parameters are correctly build
```
sampaccoud, line 26:
```suggestion
        Verify that student LTI authentication parameters are correctly built without edit mode
```

tests/plugins/lti_consumer/test_models.py
```diff
@@ -0,0 +1,146 @@
...
+            "context_id": "example.com",
+            "user_id": "richie",
+            "lis_person_contact_email_primary": "",
+            "roles": "student",
+        }
+        self.assertDictEqual(expected_auth_parameters, instance.auth_parameters())
+
+    def test_models_lti_consumer_auth_parameters_edit_mode(self):
+        """
+        Verify that instructor LTI authentication parameters are correctly build
```
sampaccoud, line 43:
```suggestion
        Verify that instructor LTI authentication parameters are correctly built with edit mode
```

tests/plugins/lti_consumer/test_models.py
```diff
@@ -0,0 +1,146 @@
...
+            "oauth_consumer_key",
+            "oauth_signature",
+        )
+        auth_headers = instance.authorize()
+        for expected_key in expected_auth_headers_keys:
+            self.assertIn(expected_key, auth_headers.get("Authorization"))
+
+    def test_models_lti_consumer_build_content_parameters_edit_mode(self):
+        """
+        Verify that LTI content consumption parameters are correctly build
```
sampaccoud, line 79:
```suggestion
        Verify that LTI content consumption parameters are correctly built
```

tests/plugins/lti_consumer/test_models.py
```diff
@@ -0,0 +1,146 @@
...
+                'OAuth oauth_nonce="80966668944732164491378916897", oauth_timestamp="1378916897", '
+                'oauth_version="1.0", oauth_signature_method="HMAC-SHA1", '
+                'oauth_consumer_key="InsecureOauthConsumerKey", '
+                'oauth_signature="frVp4JuvT1mVXlxktiAUjQ7%2F1cw%3D"'
+            )
+        },
+    )
+    def test_models_lti_consumer_content_parameters(self, _):
+        """
+        Verify that LTI content consumption parameters are correctly build trough
```
sampaccoud, line 124:
```suggestion
        Verify that LTI content consumption parameters are correctly built through the
```

env.d/development/dev
kernicPanel: I didn't look how and when env files are loaded, so I put it everywhere :sweat_smile: 
I'll check it.

sandbox/settings.py
kernicPanel: It seemed strange to me to to have localhost:port by default, but if you're ok with that, I'll put it here :)

env.d/development/dev
jbpenrath: I'm asking the same question about circleci environment variables?

src/richie/plugins/lti_consumer/cms_plugins.py
kernicPanel: The problem with LTI (uppercase) is at the call for react component :
with LTI, we have to write `class="richie-react richie-react--l-t-i-consumer"` which I found awful…

src/frontend/js/components/LtiConsumer/index.spec.tsx
```diff
@@ -0,0 +1,67 @@
...
+        expect(input.getAttribute('value')).toEqual(value);
+      }
+    });
+
+    // check if iframeresizer does its job
+    const iframe: HTMLIFrameElement = container.getElementsByTagName('iframe')[0];
+    expect(iframe.id).toMatch(/^iFrameResizer[0-9]$/);
+    expect(iframe.style).not.toBeNull();
+    // never set, probably because of faked form submit :/
+    // expect(iframe.height).toEqual(400);
```
jbpenrath, line 65: Looks like a debug trace. Should be remove or detailed?

src/richie/plugins/lti_consumer/cms_plugins.py
kernicPanel: So I choose to use LtiConsumer everywhere.

src/frontend/js/components/LtiConsumer/index.tsx
```diff
@@ -0,0 +1,38 @@
...
+        ref={formRef}
+        action={url}
+        method="POST"
+        encType="application/x-www-form-urlencoded"
+        target="lti_iframe"
+        style={{ display: 'none' }}
+      >
+        {Object.entries(contentParameters).map(([name, value]) => {
+          return <input key={name} type="hidden" name={name} value={value} />;
+        })}
```
jbpenrath, line 27: Just to show a little shortcut.
```suggestion
        {Object.entries(contentParameters).map(([name, value]) => (
          <input key={name} type="hidden" name={name} value={value} />
        ))}
```

sandbox/settings.py
sampaccoud: Aaah no sorry I didn't see. I meant to put the LTI demo site here.

src/frontend/js/types/LtiConsumer.ts
```diff
@@ -0,0 +1,21 @@
+export interface LtiConsumerContentParameters {
```
jbpenrath, line 1: Can you sort ascending properties just to improve readability?

src/richie/plugins/lti_consumer/cms_plugins.py
sampaccoud: I didn't get how the Python class name is link to the class attribute.

src/frontend/scss/components/_subheader.scss
```diff
@@ -112,7 +121,7 @@ $r-subheader-search-title-width: 19rem !default; // aligned on computed search r
   // Main subheader container alongside 'aside' container
   &__main {
     @include media-breakpoint-up(lg) {
-      @include sv-flex(1, 0, 75%);
+      @include sv-flex(1, 0, #{100% - $r-course-aside});
```
jbpenrath, line 124:
```suggestion
      @include sv-flex(1, 0, calc(100% - #{$r-course-aside}));
```

src/richie/plugins/lti_consumer/models.py
kernicPanel: Leaving url empty generates a new one for a new content, but we want to provide a way to use an existing one, no ?

tests/plugins/lti_consumer/test_forms.py
kernicPanel: Thanks :)

src/richie/plugins/lti_consumer/cms_plugins.py
kernicPanel: Oh, there's no direct link, but I thought having the same name everywhere were better for consistency.

src/frontend/scss/components/templates/courses/cms/_course_detail.scss
```diff
@@ -55,7 +55,7 @@
 
       @include media-breakpoint-up(lg) {
         @include sv-flex(1, 0, $r-course-aside);
-        padding: 3rem 0.5rem;
```
jbpenrath, line 58: I'll add a `padding-right: 3rem` to `.course-detail__content`. In my opinion, course content and course aside elements need to be spaced on desktop.

I think it is a good to add padding right to `course-detaill__content` and not on `.course-detail__aside` to reduce paragraph width which are a little bit too long.

src/richie/plugins/lti_consumer/cms_plugins.py
jbpenrath: This comment could be also applied for frontend

kernicPanel: I mean : having the same name for the django plugin/model and the react component.

tests/plugins/lti_consumer/test_cms_plugins.py
```diff
@@ -0,0 +1,88 @@
+"""Testing DjangoCMS plugin declaration for Richie's lti consumer plugin."""
```
jbpenrath, line 1:
```suggestion
"""Testing DjangoCMS plugin declaration for Richie's LTI consumer plugin."""
```

tests/plugins/lti_consumer/test_cms_plugins.py
```diff
@@ -0,0 +1,88 @@
...
+from cms.api import add_plugin
+from cms.models import Placeholder
+from cms.plugin_rendering import ContentRenderer
+
+from richie.plugins.lti_consumer.cms_plugins import LtiConsumerPlugin
+from richie.plugins.lti_consumer.factories import LtiConsumerFactory
+
+
+class LtiConsumerPluginTestCase(TestCase):
+    """Test suite for the lti consumer plugin."""
```
jbpenrath, line 16:
```suggestion
    """Test suite for the LTI consumer plugin."""
```

src/frontend/js/components/LtiConsumer/index.spec.tsx
kernicPanel: It was for leaving a trace of what a pertinent test could be, but as it isn't working, I'll remove it.

src/frontend/scss/components/templates/courses/cms/_course_detail.scss
kernicPanel: Do you mean in place of this modification, or in addition ?

jbpenrath: In addition

jbpenrath (commented):
The part to avoid iframe resize flickering looks good to me. No better idea to address this issue.

tests/plugins/lti_consumer/test_models.py
```diff
@@ -0,0 +1,146 @@
...
+                'OAuth oauth_nonce="80966668944732164491378916897", oauth_timestamp="1378916897", '
+                'oauth_version="1.0", oauth_signature_method="HMAC-SHA1", '
+                'oauth_consumer_key="InsecureOauthConsumerKey", '
+                'oauth_signature="frVp4JuvT1mVXlxktiAUjQ7%2F1cw%3D"'
+            )
+        },
+    )
+    def test_models_lti_consumer_content_parameters(self, _):
+        """
+        Verify that LTI content consumption parameters are correctly built trough the
```
jbpenrath, line 172:
```suggestion
        Verify that LTI content consumption parameters are correctly built through the
```

tests/plugins/lti_consumer/test_models.py
```diff
@@ -3,23 +3,71 @@
...
+        RICHIE_LTI_PROVIDERS={
+            "lti_provider_test": {
+                "base_url": "http://localhost:8060/lti/videos/[0-9a-f]{8}-[0-9a-f]",
+                "is_base_url_regex": True,
+            }
+        }
+    )
+    def test_models_lti_consumer_url_regex(self):
+        """
+        Verify that a regex url is generated when a predifined LTI provider is used
```
jbpenrath, line 66:
```suggestion
        Verify that a regex url is generated when a predefined LTI provider is used
```

jbpenrath (commented):
Last review before approval. 💪

src/frontend/scss/_main.scss
```diff
@@ -105,6 +105,7 @@
 @import './components/templates/courses/cms/blogpost_detail';
 @import './components/templates/courses/cms/program_list';
 @import './components/templates/courses/cms/program_detail';
+@import '../js/components/LtiConsumer/styles';
```
jbpenrath, line 108: Just move this line to line 73 (below `@import '../js/components/CourseRunEnrollment/styles';`)

jbpenrath (approved):
✅ Looks good to me!
Do not forgot to fixup or amend the second commit.

LGTM

PR #1279, opened 2021-02-10
🐛(frontend) fix autosuggest to redirect directly on course page

jbpenrath:
## Purpose

When user select a course entry in the autosuggest menu, it should be redirect to the course page and not the search view. As `hasHighlightedSuggestion` state was not set correctly, that was not work.

## Proposal

- [x] Set `hasHighlightedSuggestion`correctly on `onSuggestionHighlighted` callback

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,11 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unrealeased]
 
+### Fixed
+
+- Fix autosuggest to redirect user directly on the course page when a course
+  entry has been selected from the suggestion dropdown
+
 ## [2.1.0] - 2021-01-22
 
 ### Added
--- a/src/frontend/js/components/RootSearchSuggestField/index.tsx
+++ b/src/frontend/js/components/RootSearchSuggestField/index.tsx
@@ -132,7 +132,7 @@ export const RootSearchSuggestField = ({
       onSuggestionsFetchRequested={async ({ value: incomingValue }) =>
         onSuggestionsFetchRequested(await getFilters(), setSuggestions, incomingValue)
       }
-      onSuggestionHighlighted={({ suggestion }) => setHasHighlightedSuggestion(!suggestion)}
+      onSuggestionHighlighted={({ suggestion }) => setHasHighlightedSuggestion(!!suggestion)}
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,11 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unrealeased]
 
+### Fixed
+
+- Fix autosuggest to redirect user directly on the course page when a course
+  entry has been selected from the autocomplete drodopwn
```
sampaccoud, line 14: drodopwn? Quessidi ?

src/frontend/js/components/RootSearchSuggestField/index.tsx
```diff
@@ -132,7 +132,7 @@ export const RootSearchSuggestField = ({
       onSuggestionsFetchRequested={async ({ value: incomingValue }) =>
         onSuggestionsFetchRequested(await getFilters(), setSuggestions, incomingValue)
       }
-      onSuggestionHighlighted={({ suggestion }) => setHasHighlightedSuggestion(!suggestion)}
+      onSuggestionHighlighted={({ suggestion }) => setHasHighlightedSuggestion(!!suggestion)}
```
sampaccoud, line 135: Is there a test that we can add that should have caught it?

jbpenrath: That's a point to investigate, from my point of view, the following test should already warn us about this mistake.
https://github.com/openfun/richie/blob/master/src/richie/apps/search/indexers/courses.py#L89

jbpenrath: Ok understood, the test passed because in fact with the bug, location.assign was called two times for this use case.
First time, by the `onSuggestionSelected` callback then by `onKeyDown` input trigger.

So, the first redirect (`en/courses/42`) exists but in reality there was a second redirect at the same time which redirected on `/en/courses/?limit=13&offset=0&query=Course%20%2342`.

sampaccoud: :muscle:

sampaccoud (approved):

sampaccoud (approved):

Issue #1302, opened 2021-03-08
Fix mouse interaction on full text search input in FireFox


sampaccoud:
## Bug Report

**Problematic Behavior**
When using FireFox, it is not possible to interact with the full text search input with the mouse, only with the keyboard.

**Steps to Reproduce**
1. Go to search page with a full text search: https://demo.richie.education/fr/cours/?limit=21&offset=0&query=avantage%20rouler
2. Try selecting the text in the input or bringing the cursor between the 2 words by clicking between them

**Environment**
- Richie version: 2.2.0
- Platform: Docker
- FireFox version: 86.0

PR #1303, opened 2021-03-08
💄(frontend) fix text selection on search input on firefox

jbpenrath:
## Purpose

On Firefox, user cannot select text in search input.
A `user-select: none` rule was applied to the search input that makes text
unselectable only on Firefox.

#1302

## Proposal

- [x] Inherit `user-select` rule from `button-base` mixin for search-input element


mbenadda (approved):
GTM

PR #1305, opened 2021-03-12
⚡️(frontend) lazy load react widgets

jbpenrath:
## Purpose

As frontend widgets aims to grow, we need to optimize js chunks to load them only where they are needed.

## Proposal

- [x] Lazy load components from Root component


mbenadda (approved):
Looks good to me!! I think it was high time we did this! 👍

PR #1306, opened 2021-03-12
Add language fallback on page extension plugins

sampaccoud:
## Purpose

When a page extension plugin (CoursePlugin, OrganizationPlugin, etc.) was pointing to a page that had **never been published** in the current language, but published in another language, we were displaying broken plugins because the content could not be found in the current language.

When a page extension plugin was pointing to a page that had **been published in the current language but later unpublished**, we were displaying the unpublished content because DjangoCMS does not take into account the publication status of content when it uses a fallback language (looks like a bug to me).

## Proposal

To fix this, I implemented language fallback on the page extension plugins, getting inspiration from the [DjangoCMS main view](https://github.com/django-cms/django-cms/blob/3.8.0/cms/views.py#L37) where pages are retrieved taking into account their language fallback.

I also took this opportunity to show plugins only if the page pointed by the plugin is published. This decision follows a complaint from our support team stating that:

> In draft, we want to see the page exactly how it will look when we press the publish button

We get around the above mentioned DjangoCMS bug by not displaying the plugin at all for a language that was published and later unpublished, even if it has a fallback language that is published. This is because in this case, DjangoCMS would not display the fallback language but the very content that was voluntarily unpublished... So this could be very annoying. Hiding the plugin completely in this language is also what was the user's intention when unpublishing it.

### DIFF
--- a/tests/apps/courses/test_cms_plugins_course.py
+++ b/tests/apps/courses/test_cms_plugins_course.py
@@ -3,10 +3,13 @@
 Unit tests for the Course plugin and its model
 """
 import re
+from datetime import datetime
+from unittest import mock
 
 from django import forms
 from django.conf import settings
 from django.test import TestCase
+from django.utils import timezone
 
 from cms.api import add_plugin, create_page
 
@@ -127,9 +130,7 @@ class CoursePluginTestCase(TestCase):
             )
         )
         self.assertContains(
-            response,
-            organization.extended_object.get_title(),
-            status_code=200,
+            response, organization.extended_object.get_title(), status_code=200
         )
 
         # The course's cover should be present
@@ -188,9 +189,7 @@ class CoursePluginTestCase(TestCase):
             )
         )
         self.assertContains(
-            response,
-            organization.extended_object.get_title(),
-            status_code=200,
+            response, organization.extended_object.get_title(), status_code=200
         )
 
         # The course's cover should be present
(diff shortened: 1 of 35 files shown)


tests/apps/courses/test_cms_plugins_course.py
```diff
@@ -284,3 +285,88 @@ def test_cms_plugins_course_render_variant(self):
...
+        # Publish only the French version of the course
+        course_page.publish("fr")
+
+        # Check the page content in English
+        page.publish("en")
+        url = page.get_absolute_url(language="en")
+        response = self.client.get(url)
+
+        # The course path in french should be in the url the locale in the url
+        # should remain "en"
```
jbpenrath, line 318:
```suggestion
        # The course path in french should be in the url.
        # Furthermore the locale in the url should remain "en"
```

tests/apps/courses/test_cms_plugins_organization.py
```diff
@@ -247,3 +253,84 @@ def test_cms_plugins_organization_render_context_variant(self):
...
+        )
+        # The organization's full name should be wrapped in a h2
+        self.assertContains(
+            response,
+            '<div class="organization-glimpse__title">organisation publique</div>',
+            html=True,
+        )
+        self.assertNotContains(response, "public organization")
+
+        # Organziation's logo should be present
```
jbpenrath, line 300:
```suggestion
        # Organization's logo should be present
```

jbpenrath (approved):
Well done! Great improvement for our users

tests/apps/courses/test_cms_plugins_course.py
sampaccoud: I added "but" to match what I wrote in other places:
```suggestion
        # The course path in french should be in the url but the locale in the
        # url should remain "en"
```

PR #1313, opened 2021-03-22
⚡️(courses) cache getting course runs dict from the course models

sampaccoud:
## Purpose

On one of our sites, we want to display the first open course run in the top header of the course syllabus, and only the remaining ones in the side column.

## Proposal

This will is done via overridding template blocks in [richie site factory](https://github.com/openfun/richie-site-factory) and we will need to call the method "get_course_runs_dict" several times in the template.

This method being very costly, it is a good idea to first make sure that it is cached on the instance: make it a cached property.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Changed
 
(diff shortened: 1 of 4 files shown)


jbpenrath (approved):
LGTM ! ✅

PR #1315, opened 2021-03-23
🔖(chore) bump version to 2.3.0

sampaccoud:
### Changed

- Cache getting the course runs dictionary from the course model
- Take fallback languages into account in page extension plugins
- Load components lazily in Root component

### Fixed

- Fix LTI consumer plugin (edit mode, manual config and update after create)
- Fix an issue about text selection on search input on Firefox


jbpenrath (approved):
🚢

PR #1318, opened 2021-03-24
🐛(frontend) fix lti consumer issues

jbpenrath:
## Purpose

When iframeResizer was enabled, it enforced iframe content width and leads to layout issue on small screens.
Furthermore, option to allow fullscreen was missing.

## Proposal

- [x] Add attributes to allow use of fullscren within LTI Consumer iframe
- [x] Do not use `sizeWidth` iframeResizer options

### DIFF
--- a/src/frontend/js/components/LtiConsumer/index.tsx
+++ b/src/frontend/js/components/LtiConsumer/index.tsx
@@ -16,10 +16,7 @@ const LtiConsumer = ({
       // Retrieve and inject current component container height to prevent flickering
       // and remove aspect-ratio trick which is not compatible with iframeResizer
       const componentContainer = document.querySelector('.richie-react--lti-consumer');
-      iframeResizer(
-        { sizeWidth: true, minHeight: componentContainer?.clientHeight },
-        iframeRef.current!,
-      );
+      iframeResizer({ minHeight: componentContainer?.clientHeight }, iframeRef.current!);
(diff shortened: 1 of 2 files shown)


src/frontend/js/components/LtiConsumer/index.tsx
```diff
@@ -45,7 +42,8 @@ const LtiConsumer = ({
         name="lti_iframe"
         title={url}
         src={url}
-        allow="microphone *; camera *; midi *; geolocation *; encrypted-media *"
+        allow="microphone *; camera *; midi *; geolocation *; encrypted-media *; fullscreen *"
```
sampaccoud, line 45: Should we remove geolocation and midi?

sampaccoud (approved):

src/frontend/js/components/LtiConsumer/index.tsx
jbpenrath: I hesitated to delete them, but this component should be support all kind of LTI tool and may be one can require geolocation and midi?

kernicPanel: > I hesitated to delete them, but this component should support all kind of LTI tool and may be one of them can require geolocation and midi?

That's why I've keeped them, but I don't remember where I found them.

kernicPanel (approved):

sampaccoud:
:clap:

PR #1325, opened 2021-03-30
🐛(lti_consumer) retrieve context through an api endpoint

jbpenrath:
## Purpose

When the page is cached, LTI Consumer plugin embedded could contain stale oauth parameters that cause authentication failure to the LTI service. To keep a relevant cache duration on our CMS pages, we decided to retrieve the LTI Consumer context through an API endpoint.

## Proposal
- [x] Add an API endpoint to retrieve fresh LTI Consumer context

### DIFF
--- a/src/frontend/js/components/LtiConsumer/index.spec.tsx
+++ b/src/frontend/js/components/LtiConsumer/index.spec.tsx
@@ -1,14 +1,25 @@
-import { render } from '@testing-library/react';
+import { act, render } from '@testing-library/react';
 import React from 'react';
 import { IntlProvider } from 'react-intl';
+import fetchMock from 'fetch-mock';
 
+import { Deferred } from 'utils/test/deferred';
 import {
-  LtiConsumerContentParameters as LtiConsumerContentParametersProps,
-  LtiConsumer as LtiConsumerProps,
+  LtiConsumerContentParameters,
+  LtiConsumerContext,
+  LtiConsumerProps,
 } from 'types/LtiConsumer';
+import { handle } from 'utils/errors/handle';
 import LtiConsumer from '.';
 
+const mockHandle: jest.Mock<typeof handle> = handle as any;
+jest.mock('utils/errors/handle');
+
 describe('components/LtiConsumer', () => {
+  afterEach(() => {
+    fetchMock.restore();
+  });
+
   it('renders an auto-resized iframe with a LTI content', async () => {
     const contentParameters = {
       lti_message_type: 'Marsha Video',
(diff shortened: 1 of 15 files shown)


src/frontend/js/components/LtiConsumer/index.tsx
```diff
@@ -1,47 +1,63 @@
...
-      componentContainer?.attributes.removeNamedItem('style');
+  const checkResponseStatus = (response: Response) => {
+    if (!response.ok) {
+      throw new Error(`Failed to retrieve LTI consumer context at placeholder ${id}`);
     }
+    return response.json();
+  };
+
+  useAsyncEffect(async () => {
+    await fetch(`/api/v1.0/plugins/lti-consumer/${id}/context/`)
```
kernicPanel, line 20: I don't get how toolbar.edit_mode_active is added to the call here…
Is it managed natively by django ?

src/richie/plugins/lti_consumer/urls.py
```diff
@@ -0,0 +1,17 @@
+"""Plugins URLs configuration."""
```
kernicPanel, line 1:
```suggestion
"""LTI Consumer plugin URLs configuration."""
```

tests/plugins/lti_consumer/test_api.py
```diff
@@ -0,0 +1,59 @@
+"""
+Tests for LTIConsumerPlugin api
```
kernicPanel, line 2:
```suggestion
Tests for LTI Consumer plugin api
```

src/frontend/js/components/LtiConsumer/index.spec.tsx
```diff
@@ -118,6 +149,35 @@ describe('components/LtiConsumer', () => {
 
     // check if iframeresizer does its job
     const iframe: HTMLIFrameElement = container.getElementsByTagName('iframe')[0];
+    const form: HTMLFormElement = container.getElementsByTagName('form')[0];
     expect(iframe.id).not.toMatch(/^iFrameResizer[0-9]$/);
+    expect(form.submit).toHaveBeenCalledTimes(1);
+  });
+
+  it('renders anything and handle error if context fetching failed', async () => {
```
kernicPanel, line 157:
```suggestion
  it('renders nothing and handle error if context fetching failed', async () => {
```

tests/plugins/lti_consumer/test_api.py
```diff
@@ -0,0 +1,59 @@
...
+        self.assertEqual(response.status_code, 200)
+        self.assertIn(content["url"], lti_consumer.url)
+        self.assertEqual(content["automatic_resizing"], lti_consumer.automatic_resizing)
+        self.assertEqual(content["content_parameters"], "test_content")
+        mock_params.assert_called_once()
+
+    def test_lti_consumer_api_get_context_to_unknown_placeholder(self):
+        """
+        Make a request to its API endpoint to try to retrieve a context
+        to an unknown placeholder should return a 404 error response.
```
kernicPanel, line 54: Is it an unknown placeholder, or an unknown plugin instance (pk) ?

jbpenrath: You're this is an unknown plugin instance

src/frontend/js/components/LtiConsumer/index.tsx
jbpenrath: There is `ToolbarMiddleware` which is in charge to check if request `is_cms_request` then to bind the toolbar to the request.

https://github.com/django-cms/django-cms/blob/6e5caadc30748eb944129c950f668292aae9a348/cms/middleware/toolbar.py#L18

tests/plugins/lti_consumer/test_api.py
```diff
@@ -0,0 +1,59 @@
...
+        self.assertEqual(response.status_code, 200)
+        self.assertIn(content["url"], lti_consumer.url)
+        self.assertEqual(content["automatic_resizing"], lti_consumer.automatic_resizing)
+        self.assertEqual(content["content_parameters"], "test_content")
+        mock_params.assert_called_once()
+
+    def test_lti_consumer_api_get_context_to_unknown_placeholder(self):
+        """
+        Make a request to its API endpoint to try to retrieve a context
+        to an unknown plugin instance should return a 404 error response.
```
kernicPanel, line 54: Maybe simplify a bit, something like
```suggestion
        Making a context API request to an unknown plugin instance should return a 404 error response.
```

kernicPanel (approved):

sampaccoud (approved):

sampaccoud (approved):

PR #1327, opened 2021-04-01
🚸(programs) remove date on program glimpse

sampaccoud:
## Purpose

The date displayed on the glimpse was the "creation_date" which is useless. 
The support team requested that it be removed.

## Proposal

Remove program glimpse footer with creation date.


jbpenrath (approved):
LGTM ✅

PR #1332, opened 2021-04-07
🚚(frontend) clean built frontend files before each build

jbpenrath:
## Purpose

Webpack uses hashes to name files it generate. But currently old files stay in place. So after several builds, we can have a js folder with a very large amount of files which is not efficient and furthermore can produce issues and slow down script executions.

In order to clean these files before each build, we need to put them in a dedicated directory to not remove
other js scripts placed in `static/js` directory.

## Proposal

- [x] Generate frontend scripts into a dedicated `build` folder
- [x] Clean this new folder before each build


mbenadda (approved):
👍

PR #1336, opened 2021-04-14
🐛(frontend) fix quality issues with <SearchFilterGroupModal />

mbenadda:
## Purpose

There were several issues on `<SearchFilterGroupModal />`, which we're fixing here:

- placeholder and label for the search input were not localized;
Capture d’écran 2021-04-14 à 11 31 02

- close button created a double border as it overlapped with a list item;
Capture d’écran 2021-04-14 à 11 31 11

- positioning for the close button was broken in Safari;
Capture d’écran 2021-04-14 à 14 34 41

- document body kept scrolling even as the modal was opened.

## Proposal

For Safari I moved the scrolling container to the list of items as opposed to the whole modal. I also had to inject the Safari "hack" directly into the DOM as sass does not recognize it as valid CSS.

There's also another issue which I did not address: color contrast for the placeholder is not accessible. There's the same problem on the main search menu. I think both should be addressed together, I was not sure how we want to fix this (white input background, different colors, etc.)

### DIFF
--- a/src/frontend/js/components/SearchFilterGroupModal/_styles.scss
+++ b/src/frontend/js/components/SearchFilterGroupModal/_styles.scss
@@ -70,10 +70,12 @@
     background: r-theme-val(search-filters-group-modal, close-background);
     border: none;
     border-top: 1px solid r-theme-val(search-filters-group-modal, close-border);
+    margin-top: -1px; // Avoid having a double border when the button overlaps with a list item
   }
 
   &__overlay {
-    z-index: 10; // Make sure the overlay is above any random 0-1-2 z-indexes in content
+    // Make sure the overlay is above any random 0-1-2 z-indexes in content and 200 z-index to the top bar
+    z-index: 300;
     position: fixed;
     top: 0;
     left: 0;
@@ -93,6 +95,27 @@
 
     &:hover {
       color: r-theme-val(search-filters-group-modal, button-color);
+    }
+  }
+}
+
+body.has-search-filter-group-modal {
+  overflow: hidden;
+
+  /*
+   * Flex children in a fixed/absolute container breaks Safari.
+   * Use a specific, admittedly hacky media query to target it with a special layout
+   * that solves the issue on that browser.
+   */
+  @media not all and (min-resolution: 0.001dpcm) {
+    @media all {
+      .search-filter-group-modal {
+        overflow: hidden;
+      }
+
+      .search-filter-group-modal__form__values {
+        overflow: auto;
+      }
     }
   }
 }
--- a/src/frontend/js/components/SearchFilterGroupModal/index.tsx
+++ b/src/frontend/js/components/SearchFilterGroupModal/index.tsx
@@ -1,5 +1,5 @@
 import React, { useEffect, useState } from 'react';
-import { defineMessages, FormattedMessage, MessageDescriptor } from 'react-intl';
+import { defineMessages, FormattedMessage, MessageDescriptor, useIntl } from 'react-intl';
 import ReactModal from 'react-modal';
 
 import { fetchList } from 'data/getResourceList';
@@ -24,6 +24,16 @@ const messages = defineMessages({
     description:
       'Error message when the search for more filter value fails in the search filters modal.',
     id: 'components.SearchFilterGroupModal.error',
+  },
+  inputLabel: {
+    defaultMessage: 'Search for filters to add',
+    description: 'Accessible label for the search input in the search filter modal.',
+    id: 'components.SearchFilterGroupModal.inputLabel',
+  },
+  inputPlaceholder: {
+    defaultMessage: 'Search in { filterName }',
+    description: 'Placeholder message for the search input in the search filter modal.',
+    id: 'components.SearchFilterGroupModal.inputPlaceholder',
   },
   modalTitle: {
     defaultMessage: 'Add filters for {filterName}',
@@ -53,6 +63,8 @@ if (!isTestEnv) {
 }
 
 export const SearchFilterGroupModal = ({ filter }: SearchFilterGroupModalProps) => {
+  const intl = useIntl();
+
   const [modalIsOpen, setModalIsOpen] = useState(false);
   const [values, setValues] = useState([] as FilterValue[]);
   const [query, setQuery] = useState('');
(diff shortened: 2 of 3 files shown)


src/frontend/js/components/SearchFilterGroupModal/_styles.scss
```diff
@@ -70,6 +70,7 @@
     background: r-theme-val(search-filters-group-modal, close-background);
     border: none;
     border-top: 1px solid r-theme-val(search-filters-group-modal, close-border);
+    margin-top: -1px; // Avoid having a double border when the button overlaps with a list item
   }
 
   &__overlay {
```
jbpenrath, line 76: Can you also increase the `z-index: 9999` because currently the modal is under the `topbar` element which has a `z-index: 200`...

src/frontend/js/components/SearchFilterGroupModal/_styles.scss
```diff
@@ -70,6 +70,7 @@
     background: r-theme-val(search-filters-group-modal, close-background);
     border: none;
     border-top: 1px solid r-theme-val(search-filters-group-modal, close-border);
+    margin-top: -1px; // Avoid having a double border when the button overlaps with a list item
```
jbpenrath, line 73: What do you think to not apply a border-bottom to the latest button item?
```css
&__item {
  ....
  &:not(:last-child) > button {
    border-bottom: 1px solid r-theme-val(search-filters-group-modal, item-border);
  }
}
```

src/frontend/js/components/SearchFilterGroupModal/index.tsx
```diff
@@ -9,6 +9,25 @@ import { FacetedFilterDefinition, FilterValue } from 'types/filters';
 import { Nullable } from 'utils/types';
 import { useAsyncEffect } from 'utils/useAsyncEffect';
 
+/**
+ * Flex children in a fixed/absolute container breaks Safari.
+ * Use a specific, admittedly hacky media query to target it with a special layout
+ * that solves the issue on that browser.
+ */
+const safariFixStyles = `
```
jbpenrath, line 17: Not a fan to add a `style` tag anywhere on the page.

You could add these styles within `body.has-search-filter-group-modal` in `SearchFilterGroupModal/_styles.scss ` right?

Wow never this hacky media query to target Safari! Just to my understanding, where did you find this?

src/frontend/js/components/SearchFilterGroupModal/index.tsx
```diff
@@ -119,51 +151,61 @@ export const SearchFilterGroupModal = ({ filter }: SearchFilterGroupModalProps)
...
+                ))}
+              </ul>
+            )}
+          </fieldset>
+          <button
+            className="search-filter-group-modal__close"
+            onClick={() => setModalIsOpen(false)}
+          >
+            <FormattedMessage {...messages.closeButton} />
+          </button>
```
jbpenrath, line 207: The label is no more centered. 🤔

src/frontend/js/components/SearchFilterGroupModal/_styles.scss
```diff
@@ -70,6 +70,7 @@
     background: r-theme-val(search-filters-group-modal, close-background);
```
jbpenrath, line 70: Line 7, instead to use `bottom: 0`, we could use `max-height: calc(100vh - 6rem);` in this way we do not a large blank space when list item is short and viewport height is high.

mbenadda:
For me this was supposed to be the look for the modal if there are not so many elements that we need to scroll.

Capture d’écran 2021-04-14 à 15 49 08

I guess I have nothing against letting it shrink to content. It's an edge case anyway as it's only for facets with 11-~16-17 elements. 10 and below, the modal does not exist, above 17 it takes up the whole screen and there's scrolling.

src/frontend/js/components/SearchFilterGroupModal/index.tsx
mbenadda: It's centered for me on Latest firefox & Safari 13. In which browser are you seeing this?

src/frontend/js/components/SearchFilterGroupModal/index.tsx
mbenadda: I took this one from this blog post (https://www.ryadel.com/en/css3-media-query-target-only-ie-ie6-ie11-firefox-chrome-safari-edge/) which I've used as a reference in the past.

The issue is that `_styles.scss` does not work as this is not valid SASS (although is does work in CSS). It is in that sense I said it's a hack. I don't really have a better solution to offer.

The solution I propose for Safari is a worse experience in my opinion (as scrolling only through the list makes the UI less clear), which I'd rather not force on all other browsers. On the other hand, I wrestled with it a little bit this morning and could not find a way to make it work as intended on Safari. It seems like the sizing of flex children inside the `fixed` element is fundamentally broken.

src/frontend/js/components/SearchFilterGroupModal/index.tsx
jbpenrath: Firefox 87.0 on Mac OS.

jbpenrath:
> For me this was supposed to be the look for the modal if there are not so many elements that we need to scroll.
> 
> Capture d’écran 2021-04-14 à 15 49 08
> 
> I guess I have nothing against letting it shrink to content. It's an edge case anyway as it's only for facets with 11-~16-17 elements. 10 and below, the modal does not exist, above 17 it takes up the whole screen and there's scrolling.

From my point of view, it is visually better to shrink the modal but as you said it is an edge case.

src/frontend/js/components/SearchFilterGroupModal/_styles.scss
mbenadda: With the fixed-height modal this is not ideal as it leaves the last element without a border if the list it not long enough to reach the bottom of the modal.

src/frontend/js/components/SearchFilterGroupModal/index.tsx
mbenadda: I think everything else is done now, we just have this one left to decide on.

If you want to try your hand to a better fix than what I offer in this PR,  there might be something I missed!

jbpenrath: This snippet works on my side. Just use a valid a media query (`@media all`)
```scss
body.has-search-filter-group-modal {
  overflow: hidden;

  @media not all and (min-resolution: 0.001dpcm) {
    @media all {
      .search-filter-group-modal {
        overflow: hidden;
      }
  
      .search-filter-group-modal__form__values {
        overflow: auto;
      }
    }
  }
}
```

mbenadda: Great! I tried it on Edge and it works there too.

I changed it to your suggestion.

jbpenrath (approved):
Many thanks for this fix @mbenadda 🙏

PR #1341, opened 2021-04-16
✨(blogposts) add language fallback to get_related_blogposts

sampaccoud:
### Purpose

The `get_related_blogposts` method on the blogpost model does not take into account language fallbacks.

## Proposal

Refactor the `get_related_blogposts` method on the blogpost model to take into account language fallbacks.
Write a test to secure this behavior.


jbpenrath (approved):
LGTM ✅

PR #1343, opened 2021-04-19
Update references on documentation site and fix target blank links

sampaccoud:
## Purpose

Now that richie is in production on https://www.fun-mooc.fr, we should point to it in the documentation website.

## Proposal

- Add links to existing richie sites
- Remove CEA site that never went live
- Add missing `rel="noopener noreferer"` attribute on links with `target="_blank"` attribute


jbpenrath (approved):
Since you add `rel="noopener noreferrer"` attribute, it seems you have to update several tests.

Otherwise, good catch! ✅

PR #1345, opened 2021-04-20
Fix/frontend issues

jbpenrath:
## Purpose

1. Old browsers does not support group regexp. So enrollment widget does not work on these browsers.
https://sentry.io/organizations/gip-fun-mooc/issues/2337189525/?environment=production&project=1454000&query=is%3Aunresolved

2. There are too much request to the search api endpoint.
From our search components, autocomplete and search queries are immediately fired on each user inputs that is too much. We have to debounce these requests. (The debounce was in place on SuggestField but did not work properly).

## Proposal

- [x] Remove use of group regexp from frontend
- [x] Debounce search requests

### DIFF
--- a/sandbox/settings.py
+++ b/sandbox/settings.py
@@ -543,7 +543,9 @@ class Base(StyleguideMixin, DRFMixin, RichieCoursesConfigurationMixin, Configura
 
     # For more details about CMS_CACHE_DURATION, see :
     # http://docs.django-cms.org/en/latest/reference/configuration.html#cms-cache-durations
-    CMS_CACHE_DURATIONS = values.DictValue({"menus": 0, "content": 0, "permissions": 0})
+    CMS_CACHE_DURATIONS = values.DictValue(
+        {"menus": 3600, "content": 60, "permissions": 3600}
+    )
 
     # Sessions
     SESSION_ENGINE = values.Value("django.contrib.sessions.backends.db")
--- a/sandbox/settings.py
+++ b/sandbox/settings.py
@@ -298,7 +298,7 @@ class Base(StyleguideMixin, DRFMixin, RichieCoursesConfigurationMixin, Configura
                 "base", environ_name="EDX_JS_BACKEND", environ_prefix=None
             ),
             "JS_COURSE_REGEX": values.Value(
-                r"^.*/courses/(?<course_id>.*)/course/?$",
+                r"^.*/courses/(.*)/course/?$",
                 environ_name="EDX_JS_COURSE_REGEX",
                 environ_prefix=None,
             ),
@@ -543,9 +543,7 @@ class Base(StyleguideMixin, DRFMixin, RichieCoursesConfigurationMixin, Configura
 
     # For more details about CMS_CACHE_DURATION, see :
     # http://docs.django-cms.org/en/latest/reference/configuration.html#cms-cache-durations
-    CMS_CACHE_DURATIONS = values.DictValue(
-        {"menus": 3600, "content": 60, "permissions": 3600}
-    )
+    CMS_CACHE_DURATIONS = values.DictValue({"menus": 0, "content": 0, "permissions": 0})
 
     # Sessions
     SESSION_ENGINE = values.Value("django.contrib.sessions.backends.db")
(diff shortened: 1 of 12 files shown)


sandbox/settings.py
```diff
@@ -543,9 +543,7 @@ class Base(StyleguideMixin, DRFMixin, RichieCoursesConfigurationMixin, Configura
 
     # For more details about CMS_CACHE_DURATION, see :
     # http://docs.django-cms.org/en/latest/reference/configuration.html#cms-cache-durations
-    CMS_CACHE_DURATIONS = values.DictValue(
-        {"menus": 3600, "content": 60, "permissions": 3600}
-    )
+    CMS_CACHE_DURATIONS = values.DictValue({"menus": 0, "content": 0, "permissions": 0})
```
sampaccoud, line 546: fail?

src/frontend/js/utils/api/lms/openedx-hawthorn.ts
```diff
@@ -20,8 +20,10 @@ const API = (
   APIConf: AuthenticationBackend | LMSBackend,
   options?: ApiOptions,
 ): ApiImplementation => {
-  const extractCourseIdFromUrl = (url: string): Maybe<Nullable<string>> =>
-    url.match((APIConf as LMSBackend).course_regexp)?.groups?.course_id;
+  const extractCourseIdFromUrl = (url: string): Maybe<Nullable<string>> => {
+    const matches = url.match((APIConf as LMSBackend).course_regexp);
+    return matches && matches[1] ? matches[1] : null;
```
sampaccoud, line 25: I'm wondering, do you see anyway to lint that this is not used in future code?

src/frontend/js/components/SearchSuggestField/index.tsx
```diff
@@ -87,7 +87,10 @@ const SearchSuggestField = ({ context }: CommonDataProps) => {
       });
     }
   };
-  const updateCourseSearchParamsDebounced = debounce(searchAsTheUserTypes, 500, { maxWait: 1100 });
+  const updateCourseSearchParamsDebounced = useMemo(
+    () => debounce(searchAsTheUserTypes, 500, { maxWait: 1100 }),
+    [],
+  );
```
sampaccoud, line 93: I don't understand how it is going to work. If you pass [] as params then they are never going to change and memoization is forever so debounce is called only once?

sandbox/settings.py
jbpenrath: Oups !

src/frontend/js/components/SearchSuggestField/index.tsx
jbpenrath: Yes, the debounce method is called once, so only once timer is instantiated during the component life cycle. Previously, we called debounce at every render, so a timer was created each time and debounce cannot work.

sampaccoud: Would be nice if this can be tested ^^

jbpenrath: Yes I'm currently fighting to fix tests and remove this TODO ^^.
https://github.com/openfun/richie/blob/f2c68abe06d2e711f5bea0c9835f84de82777a81/src/frontend/js/components/SearchSuggestField/index.spec.tsx#L470

src/frontend/js/utils/api/lms/openedx-hawthorn.ts
jbpenrath: For this case, as the regex is provided by the backend it is something hard to detect from the frontend.

sampaccoud: ok... You could look for `.groups` but that may be a bit too common...

jbpenrath: The error occured when the regex is instantiate. We couldn't even test whether the `groups` attribute exists. We could try to detect if the regex course string contains `<?.*>` but from my point of view it is not a good idea to add this complexity to support old browsers.

sampaccoud: :+1: I follow you.

src/frontend/js/components/SearchSuggestField/index.tsx
```diff
@@ -66,28 +66,39 @@ const SearchSuggestField = ({ context }: CommonDataProps) => {
    * Helper to update the course search params when the user types. We needed to take it out of
    * the `onChange` handler to wrap it in a `debounce` (and therefore a `useRef` to make the
    * debouncing effective).
+   *
+   * This method should be memoized and updated only courseSearchParams changed.
```
sampaccoud, line 70:
```suggestion
   * This method should be memoized and updated only when courseSearchParams change.
```

sampaccoud (approved):

PR #1346, opened 2021-04-20
🐛(search) avoid KeyError when a course is indexed with no titles

sampaccoud:
## Purpose

A course should not get indexed with no titles, but if it does, querying the API should not raise a 500 error.

## Proposal

- Fix source fields definition to include the `title` field even if it's an empty dict.


jbpenrath (approved):
I checked on frontend, a Resource without title should not break anything. On the other hand, we could have blank lines in suggestion fields and in Course Glimpses.

LGTM ✅

PR #1351, opened 2021-04-29
🔧(docker) use a custom name for the app service on default network

jbpenrath:
## Purpose

Joanie, Richie and OpenEdX have to run over the same network. But both Richie and Joanie have a service called `app`. From Richie, `nginx` create a reverse proxy resolving urls with the domain `app`. If Joanie is run before Richie, the nginx service use the Joanie app service. To avoid this issue, we could add a network alias to the richie `app` service.

Related to https://github.com/openfun/joanie/pull/26

## Proposal

- [x] Update docker app service networks
- [x] Update nginx conf to use the network app service alias name

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -27,7 +27,9 @@ services:
       - env.d/development/${DEV_ENV_FILE:-dev}
       - env.d/development/${DB_HOST:-postgresql}
     networks:
-      - default
+      default:
+        aliases:
+          - app-service
     volumes:
       - .:/app
       - ./data/static:/data/static
--- a/docker/files/etc/nginx/conf.d/default.conf
+++ b/docker/files/etc/nginx/conf.d/default.conf
@@ -1,5 +1,5 @@
 upstream richie-backend {
-  server app:8000 fail_timeout=0;
+  server app-service:8000 fail_timeout=0;
 }
 
 server {


docker-compose.yml
```diff
@@ -27,7 +27,9 @@ services:
       - env.d/development/${DEV_ENV_FILE:-dev}
       - env.d/development/${DB_HOST:-postgresql}
     networks:
-      - default
+      default:
```
sampaccoud, line 30: Actually, you can declare different names depending on the network like we do here: https://github.com/openfun/openedx-docker/blob/master/docker-compose.yml#L180

This should allow you to keep app for the default name which is better?

jbpenrath: The problem is that Joanie django application service name is also `app` and this service is exposed on the `edx_outside` network. Of course, I expose Joanie on the `edx_outside` with an alias `joanie`, but it seems the service is accessible from `app` and `joanie` host names on this network.

From Richie, the `nginx` needs to resolve the `app` host name and the conflict appears here:
- On `default` network, it can resolve the Richie `app` service
- On `edx_outside` network, it can resolve the Joanie `app` service

So I have two solutions:
- Rename the joanie `app` service
- Use an alias to the richie `app` service on `default` network (Solution I implemented here)

sampaccoud (approved):

PR #1352, opened 2021-04-29
💄(program) make related courses optional for program page

2018D:
## Purpose

To be able to present programs that won't have or doesn't yet have courses
ready to be presented.

## Proposal

Add an if statement to check for course presence

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unrealeased]
 
+- Make related courses optional for program page
+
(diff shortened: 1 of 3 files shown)


sampaccoud (approved):
Thank you for your contribution @2018D. It would be nice to add a small test that demonstrate what can be done now that couldn't be done before (hide the "Related courses" section when it is empty I suppose).

2018D:
I've added the test to check the page when there's no course.

sampaccoud (approved):
Sweet! Thank you for this contribution. I hope you're still happy with your Richie sites 🤗

PR #1359, opened 2021-05-12
Refactor/react query

jbpenrath:
## Purpose

[React Query](https://react-query.tanstack.com/) is a powerful tool to fetch, cache and update data without to have to manage any "global state". 
With Joanie, we will have to make several requests that `react-query` will greatly ease management. Furthermore in order to homogenize codebase, it is interesting to use react-query for existing queries. For example, we will able to manage session state and cache uniquely within `react-query`, we could also minimize number of requests by finetuning `react-query`cache precisely.

## Proposal

- [x] Install `react-query`
- [x] Refactor code to use `react-query` to make requests when its useful.

### DIFF
--- a/src/frontend/js/components/Root/index.spec.tsx
+++ b/src/frontend/js/components/Root/index.spec.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { PropsWithChildren } from 'react';
 import { IntlProvider } from 'react-intl';
 
 import { findByText, render } from '@testing-library/react';
@@ -13,6 +13,11 @@ jest.mock('components/RootSearchSuggestField', () => ({
   __esModule: true,
   default: ({ exampleProp }: { exampleProp: string }) =>
     `root search suggest field component rendered with ${exampleProp}`,
+}));
+
+jest.mock('data/useSession', () => ({
+  __esModule: true,
+  SessionProvider: (props: PropsWithChildren<any>) => props.children,
 }));
 
 describe('<Root />', () => {
@@ -73,7 +78,7 @@ describe('<Root />', () => {
     // Render the root component, passing our real element and our bogus one
     render(
       <IntlProvider locale="en">
-        <Root richieReactSpots={[userLoginContainer, userFeedbackContainer]} />
+        <Root richieReactSpots={[userFeedbackContainer, userLoginContainer]} />
       </IntlProvider>,
     );
 
--- a/src/frontend/js/components/SearchFilterValueParent/index.tsx
+++ b/src/frontend/js/components/SearchFilterValueParent/index.tsx
@@ -2,14 +2,14 @@ import React, { useState } from 'react';
 import { defineMessages, useIntl } from 'react-intl';
 
 import { SearchFilterValueLeaf } from 'components/SearchFilterValueLeaf';
-import { fetchList } from 'data/getResourceList';
 import { useCourseSearchParams } from 'data/useCourseSearchParams';
 import { useFilterValue } from 'data/useFilterValue';
 import { RequestStatus } from 'types/api';
 import { FacetedFilterDefinition, FilterValue } from 'types/filters';
 import { getMPTTChildrenPathMatcher } from 'utils/mptt';
 import { Nullable } from 'types/utils';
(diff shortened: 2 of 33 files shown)


kernicPanel (approved):

src/frontend/js/utils/react-query/createQueryClient.ts
```diff
@@ -0,0 +1,50 @@
...
+
+const createQueryClient = (options?: QueryClientOptions) => {
+  const context: CommonDataProps['context'] = (window as any)?.__richie_frontend_context__?.context;
+  const environment = context?.environment;
+
+  const queryClient = new QueryClient({
+    defaultOptions: {
+      queries: {
+        cacheTime: REACT_QUERY_SETTINGS.cacheTime,
+        refetchOnWindowFocus: false,
```
mbenadda, line 22: Is there a reason we want this disabled? In general I think it's a great feature of react-query.

Is it because it might cause inconsistencies with stale server-generated content (and we'd rather have the whole page be stale)?

src/frontend/js/components/SearchFilterValueParent/index.tsx
```diff
@@ -54,24 +54,27 @@ export const SearchFilterValueParent = ({ filter, value }: SearchFilterValuePare
...
   const [children, setChildren] = useState([] as FilterValue[]);
-  useAsyncEffect(async () => {
-    if (showChildren) {
-      // Get only the filters & facet counts for the children of the current parent
-      const childrenResponse = await fetchList('courses', {
-        ...courseSearchParams,
-        [`${filter.name}_include`]: childrenPathMatch,
-        scope: 'filters',
-      });
+  const { data: childrenResponse } = useCourseSearch(
```
mbenadda, line 57: IMO the best outcome for this refactor would be to use the query state to render different things between loading/success/failure, and ditch the `[children, setChildren]` state altogether. 

I don't think it makes sense to store react-query return values in component state. Then you can also remove the whole `useEffect`call  below.

src/frontend/js/utils/react-query/createQueryClient.ts
jbpenrath: From my point of view, default react-query parameters are too agressive. Especially this option to refetch data each time user get focus on the window.

So I choose to disable it by default but I enable this option to retrieve session state. In this way, if the user logged out from OpenEdX, we will be able to destroy the user session from richie.

src/frontend/js/components/Root/index.spec.tsx
```diff
@@ -62,9 +67,9 @@ describe('<Root />', () => {
 
   it('prints a console warning and still renders everything else when it fails to find a component', async () => {
     // Create the containers for the component we're about to render
-    const userLoginContainer = document.createElement('div');
-    userLoginContainer.setAttribute('class', 'richie-react richie-react--user-login');
-    document.body.append(userLoginContainer);
+    // const userLoginContainer = document.createElement('div');
+    // userLoginContainer.setAttribute('class', 'richie-react richie-react--user-login');
+    // document.body.append(userLoginContainer);
```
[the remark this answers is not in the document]
jbpenrath: Whoops, of course not.

src/frontend/js/components/SearchFilterValueParent/index.tsx
jbpenrath: You are totally right. I removed the state and process date through a `useMemo` instead of `useEffect`

mbenadda: I don't understand why we need a `useMemo` here: as far as I understand it, it's supposed to be a performance optimisation tool, but we're not doing any heavy work.

Error reporting could probably move to `useCourseSearch` directly (maybe in an `onError` handler?), then we're basically just setting a default value `const children = childrenResponse?.content.filters[filter.name].values || []`

jbpenrath: Ok, I removed the `useMemo` and use the `onSettled` method to handle error to be keep the detailed error message from `SearchFilterValueParent`.

mbenadda (approved):
GTM 👍

PR #1362, opened 2021-05-19
Chore/reconfigure renovate

jbpenrath:
## Purpose

Lately pyup bot was out of service. So we think about to switch to another dependency update bot.
First, we though about use of dependabot but after a quick test, it risks to be too noisy as it is not able to group PR (It creates one PR per package to update...). https://github.com/dependabot/dependabot-core/issues/1190

And then we discovered that renovate is now supporting `setup.cfg` dependencies update. In this way, we could use only one bot to manage our dependencies update.

We also discovered that our renovate configuration file was outdated. So this PR aims to also update the renovate file to make renovate works again.

## Proposal

- [x] Remove pyup
- [x] Add python managers to renovate
- [x] Migrate renovate configuration file to new version
- [x] Validate configuration file through `renovate-config-validator` tool


jbpenrath:
We could also add managers to upgrade docker images? WDYT?

renovate.json
```diff
@@ -2,13 +2,37 @@
...
+      "schedule": [
+        "before 7am on monday"
+      ]
+    },
+    {
+      "matchManagers": [
+        "setup-cfg"
+      ],
+      "matchDepTypes": [
+        "devDependencies"
```
lunika, line 30: If I understand the [documentation](https://docs.renovatebot.com/configuration-options/#matchdeptypes), you limit the packages to `devDependencies`. But in `setup.cfg` this is not existing and why limiting to development ?
This part is unclear for me.

jbpenrath: I limited to development because previously pyup ignored `install_requires` dependency (except django-redis and exrex don't no why just this deps are allowed).

Then about the use of `devDependencies`, I don't know yet which value to use (`options.extras_require`?).
Or maybe I should use [`excludePackageNames`](https://docs.renovatebot.com/configuration-options/#excludepackagenames) and declare one by one each dependency to ignore in this array?

lunika: Ok I didn't notice in the PR the removal of `pyup` comments, sorry. Did you try to use the `renovate-config-validator` program ? https://docs.renovatebot.com/reconfigure-renovate/

jbpenrath: Yes, I did it.

lunika: So to know if it works we have to merge it and wait to see ?

lunika: This scenario looks good :+1:

renovate.json
```diff
@@ -2,20 +2,48 @@
...
+        "djangocms-googlemap",
+        "djangocms-link",
+        "djangocms-picture",
+        "djangocms-text-ckeditor",
+        "djangocms-video",
+        "djangorestframework",
+        "elasticsearch",
+        "social-auth-core[openidconnect]",
+        "social-auth-app-django",
+        "django-redis",
```
lunika, line 43: You decided to exclude both `django-redis` and `exrex`. Is it intended ? They should be excluded and they were not with pyup ?

jbpenrath: Oh copy/pasta failed !

setup.cfg
```diff
@@ -25,23 +25,23 @@ classifiers =
 [options]
 include_package_data = True
 install_requires =
```
lunika, line 27: Can you add a comment here saying to complete the renovate `excludePackageNames` when someone add a new requirement ?

setup.cfg
```diff
@@ -24,24 +24,27 @@ classifiers =
 
 [options]
 include_package_data = True
+# When you install a new package, if you do not want our renovate bot checks
+# updates for it, please add the exact name of the package within
+# `excludePackageNames` within renovate.json file at the root of this repository
```
jmaupetit, line 29:
```suggestion
# `excludePackageNames` (in the "python dependencies" group) within renovate.json file at the root of this repository.
```

jmaupetit (approved):

lunika (approved):

PR #1366, opened 2021-05-27
🔧(renovate) fix ignored python dependencies

jbpenrath:
## Purpose

Following renovate configuration update, instead of ignore dependencies listed in
`excludePackageNames`, Renovate creates a dedicated PR for each of these dependencies.
So to renovate really ignores update of these dependencies we have to create
a dedicated packageRule

Furthermore with this syntax, it's easier to create a generic preset for all our repositories.

## Proposal

- [x] Add `ignored python dependencies` package rule

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -23,8 +23,13 @@
       "groupName": "python dependencies",
       "matchManagers": ["setup-cfg"],
       "schedule": ["before 7am on monday"],
-      "matchPackagePatterns": ["*"],
-      "excludePackageNames": [
+      "matchPackagePatterns": ["*"]
+    },
+    {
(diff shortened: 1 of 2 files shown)


lunika:
And instead of having a list of excluded package can we create a list of allowed dependencies listing them in the `matchPackagePatterns` config ?

jbpenrath:
> And instead of having a list of excluded package can we create a list of allowed dependencies listing them in the `matchPackagePatterns` config ?

We could but I prefer to explicit packages to ignore. In this way it's easier to be aware of packages which can be obsolet in the repository.

lunika (approved):

PR #1368, opened 2021-05-28
🔍️(i18n) add "hreflang" meta to avoid duplicates in search engines

sampaccoud:
## Purpose

I noticed that our pages where appearing twice in Google's search results, once per language. The good news is our SEO as skyrocketed, the bad news is we have duplicate content.

## Proposal

This is a well known problem solved by adding `hreflang` metas to all pages.

### DIFF
--- a/src/richie/apps/core/templates/richie/hreflang.html
+++ b/src/richie/apps/core/templates/richie/hreflang.html
@@ -1,0 +1,6 @@
+{% load i18n menu_tags cms_tags %}
+{% if languages|length > 1 %}
+    {% for code, name in languages %}
+        <link rel="alternate" href="{% page_language_url code %}" hreflang="{{ code }}" />
+    {% endfor %}
+{% endif %}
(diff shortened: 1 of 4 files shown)


src/richie/apps/core/templates/richie/hreflang.html
```diff
@@ -0,0 +1,7 @@
+{% load i18n menu_tags cms_tags %}
+######################################### {{languages}}
```
jbpenrath, line 2: What is the purpose of this line?

tests/apps/core/test_pages.py
```diff
@@ -37,6 +37,32 @@ def test_pages_i18n(self):
         # ... and make sure the page menu is present in english on the page
         self.assertIn(content["en"], response.rendered_content)
 
+    @override_settings(LANGUAGES=(("fr-ca", "Canadian"), ("es", "Spanish")), LANGUAGE_CODE="fr-ca", CMS_LANGUAGES={})
```
jbpenrath, line 40: Too long for the linter ^^

jbpenrath (approved):
✅

src/richie/apps/core/templates/richie/hreflang.html
sampaccoud: image

tests/apps/core/test_pages.py
sampaccoud: :+1:

NamFra (approved):
👍

PR #1369, opened 2021-05-28
✨(frontend) send username to lti consumer if he/she login

jbpenrath:
## Purpose
We would like to monitor which watch marsha video on Richie. In order to achieve this, we pass the username of the logged in user through the LTI context.

## Proposal

- [x] Pass username to the lti consumer plugin if a user is logged in on Richie

### DIFF
--- a/src/frontend/js/components/LtiConsumer/index.spec.tsx
+++ b/src/frontend/js/components/LtiConsumer/index.spec.tsx
@@ -1,7 +1,9 @@
 import { act, render } from '@testing-library/react';
-import React from 'react';
+import React, { PropsWithChildren } from 'react';
 import { IntlProvider } from 'react-intl';
+import faker from 'faker';
 import fetchMock from 'fetch-mock';
+import { Session } from 'data/useSession';
 
 import { Deferred } from 'utils/test/deferred';
 import {
@@ -9,6 +11,8 @@ import {
   LtiConsumerContext,
   LtiConsumerProps,
 } from 'types/LtiConsumer';
+import { Nullable } from 'utils/types';
+import { User } from 'types/User';
 import { handle } from 'utils/errors/handle';
 import LtiConsumer from '.';
 
@@ -19,6 +23,17 @@ describe('components/LtiConsumer', () => {
   afterEach(() => {
     fetchMock.restore();
   });
+
+  const SessionProvider = ({ username, children }: PropsWithChildren<{ username?: string }>) => {
+    const user: Nullable<User> = username ? { username, urls: [] } : null;
+    const destroy = jest.fn();
+    const login = jest.fn();
+    const register = jest.fn();
+
+    return (
+      <Session.Provider value={{ user, login, register, destroy }}>{children}</Session.Provider>
+    );
+  };
 
   it('renders an auto-resized iframe with a LTI content', async () => {
     const contentParameters = {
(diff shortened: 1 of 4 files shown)


src/frontend/js/components/LtiConsumer/index.tsx
```diff
@@ -19,7 +21,12 @@ const LtiConsumer = ({ id }: LtiConsumerProps) => {
   useAsyncEffect(async () => {
     await fetch(`/api/v1.0/plugins/lti-consumer/${id}/context/`)
       .then(checkResponseStatus)
-      .then(setContext)
+      .then((ltiContext: LtiConsumerContext) =>
+        setContext({
+          ...ltiContext,
+          ...(user && { lis_person_sourcedid: user.username }),
```
sampaccoud, line 27: maybe add a test if easy?

jbpenrath: Yes of course, I just quickly put the code on github.

src/frontend/js/components/LtiConsumer/index.tsx
```diff
@@ -19,7 +21,15 @@ const LtiConsumer = ({ id }: LtiConsumerProps) => {
   useAsyncEffect(async () => {
     await fetch(`/api/v1.0/plugins/lti-consumer/${id}/context/`)
       .then(checkResponseStatus)
-      .then(setContext)
+      .then((ltiContext: LtiConsumerContext) =>
+        setContext({
+          ...ltiContext,
+          content_parameters: {
+            ...ltiContext.content_parameters,
+            ...(user && { lis_person_sourcedid: user.username }),
```
lunika, line 29: should we also send the email or not ?

src/frontend/js/components/LtiConsumer/index.spec.tsx
```diff
@@ -180,4 +199,76 @@ describe('components/LtiConsumer', () => {
...
+      expect(input).toBeInstanceOf(HTMLInputElement);
+      if (input) {
+        expect(input.getAttribute('value')).toEqual(value);
+      }
+    });
+
+    // Check that `lis_person_sourcedid` input exists when user is logged in
+    const input = container.querySelector(`input[name=lis_person_sourcedid]`);
+    expect(input).toBeInstanceOf(HTMLInputElement);
+    if (input) {
```
lunika, line 270: This if will never happen if previous `expect` fails. I suggest to remove this `if` and act like if `input` is not `null`.

src/frontend/js/components/LtiConsumer/index.spec.tsx
```diff
@@ -180,4 +199,76 @@ describe('components/LtiConsumer', () => {
...
+      expect(input.getAttribute('value')).toEqual(username);
```
lunika, line 271:
```suggestion
      expect(input!.getAttribute('value')).toEqual(username);
```

src/frontend/js/components/LtiConsumer/index.tsx
jbpenrath: I don't have this information right now. Without using a custom route to retrieve user information as we will do with fonzie, currently the only one solution is to make two request to OpenEdX:
- first request `/me` to retrieve the username of the logged in user
- then a second request to retrieve account information from this `username`

lunika: Ok the username is enough for now.

kernicPanel (approved):

lunika (approved):

PR #1374, opened 2021-06-03
Fix/lti consumer

jbpenrath:
## Purpose

LTIConsumerPlugin is not fit to be added several times on the same page. Moreover we have to support this use case.

## Proposal
We have to fix all problems related to LTIConsumerPlugin multi instance
- [x] Use a unique name for LTIConsumer iframe
- [x] Target properly component parent used to instantiate iframeResizer if needed
- [x] Move styles related to LTIConsumerPlugin out of `.subheader` element

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 ### Fixed
 
 - Use a unique name to LTIConsumerPlugin iframes to allow several plugins on the
+- Fix iframeResizer on LTIConsumer component when several plugins are on the
   same page
 
 ## [2.6.0] - 2021-05-03
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,11 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 - Send username to LTI Consumer if user is logged in
 - Make related courses optional for program page
 
+### Fixed
+
+- Use a unique name to LTIConsumerPlugin iframes to allow several plugins on the
+  same page
+
 ## [2.6.0] - 2021-05-03
 
 ### Added
--- a/src/frontend/js/components/LtiConsumer/_styles.scss
+++ b/src/frontend/js/components/LtiConsumer/_styles.scss
@@ -1,9 +1,28 @@
-.lti-consumer {
-  animation: fadeIn 600ms cubic-bezier(0.33, 1, 0.68, 1) forwards;
-  background-color: #ffffff;
+.richie-react--lti-consumer {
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -17,6 +17,14 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 - Send username to LTI Consumer if user is logged in
 - Make related courses optional for program page
 
+### Fixed
+
+- Use a unique name to LTIConsumerPlugin iframes
+  to avoid a bug when several plugins was on the same page
+- Change logic to retrieve LTIConsumerPlugin component parent DOM node to avoid
+  an error that occured on iframeResizer instantiation when several plugins was
+  on the same page
```
sampaccoud, line 26: Maybe too long and too vague?
```suggestion
- Fix iframeResizer on LTIConsumer plugin when several plugins are on the same page
```

CHANGELOG.md
```diff
@@ -17,6 +17,14 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 - Send username to LTI Consumer if user is logged in
 - Make related courses optional for program page
 
+### Fixed
+
+- Use a unique name to LTIConsumerPlugin iframes
+  to avoid a bug when several plugins was on the same page
```
sampaccoud, line 23:
```suggestion
- Use a unique name to LTIConsumerPlugin iframes
  to allow several plugins on the same page
```

src/frontend/js/components/LtiConsumer/_styles.scss
```diff
@@ -1,9 +1,29 @@
-.lti-consumer {
-  animation: fadeIn 600ms cubic-bezier(0.33, 1, 0.68, 1) forwards;
-  background-color: #ffffff;
+.richie-react--lti-consumer {
+  &.aspect-ratio {
+    position: relative;
+    padding-bottom: 56.25%; // Aspect ratio 16/9
```
sampaccoud, line 4: isn't the aspect ratio supposed to be passed via the template context?

jbpenrath: Yes, 16/9 is default value. Check `inline_ratio_css_padding_bottom` to see how this style is overrided.

sampaccoud: But there is a default value also in the context so maybe this is useless 🤔

jbpenrath: You are right, `padding-bottom` rule is always overridden in the LTIConsumerPlugin context.

sampaccoud (approved):

PR #1375, opened 2021-06-03
Fix LTI consumer manual configuration

sampaccoud:
## Purpose

When declaring an LTI consumer plugin manually, the "is_automatic_resizing" and "inline_ratio" parameters can not be set and have bad default values leading to broken displays that can't be fixed by the editor.

## Proposal

We propose to add model fields on the plugin instance so that the editor can set specific values for the `is_automatic_resizing` and `inline_ratio` parameters.

For preconfigured providers, the parameters defined in settings take precedence.


src/richie/plugins/lti_consumer/templates/richie/lti_consumer/lti_consumer.html
```diff
@@ -1,5 +1,5 @@
 <div
   class="richie-react richie-react--lti-consumer aspect-ratio"
-  style="{{ instance.inline_ratio_css_padding_bottom }}"
+  style="padding-bottom: {% widthratio instance.get_inline_ratio 1 100 %}%"
```
jbpenrath, line 3: There is a tag for that! I see that the value is rounded to the nearest integer. It does make sens to interpolate values in pixels but for percentage values, it bothers me a bit.
What do you think to simply multiply by 100 the `inline_ratio` value?

CHANGELOG.md
```diff
@@ -19,6 +19,8 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Fixed
 
+- Allow configuring the inline ratio and auto-resizing for manually defined
+  LTIConsumerPlugin isntances
```
jbpenrath, line 23:
```suggestion
- Allow configuring the inline ratio and auto-resizing manually for defined
  LTIConsumerPlugin instances
```

src/richie/plugins/lti_consumer/models.py
```diff
@@ -41,6 +41,8 @@ class LTIConsumer(CMSPlugin):
     )
     oauth_consumer_key = models.CharField(max_length=50, null=True, blank=True)
     shared_secret = models.CharField(max_length=50, null=True, blank=True)
+    is_automatic_resizing = models.BooleanField(default=True, blank=True)
+    inline_ratio = models.FloatField(default=0.5625, blank=True)  # default is 16/9
```
jbpenrath, line 45:
```suggestion
    inline_ratio = models.FloatField(default=0.5625, blank=True, min_value=0)  # default is 16/9
```

We could also add a `max_value` constraint but not sure about the value to use (3?)

jbpenrath (approved):
Just nitpicks!

CHANGELOG.md
sampaccoud: actually this would mean something different. I really meant "for manually-defined instances". Let me add the "-".

src/richie/plugins/lti_consumer/models.py
sampaccoud: Good catch. Let me put 0.1 as mean value because a ratio of 0 doesn't make sense anyway.

CHANGELOG.md
jbpenrath: Oh ok!

src/richie/plugins/lti_consumer/templates/richie/lti_consumer/lti_consumer.html
sampaccoud: It's the `widthratio` tag that rounds it down :thinking: If you think the precision is important, I will compute it myself on the model.

jbpenrath: I prefer to have our own computation and preserver precision.

src/richie/plugins/lti_consumer/models.py
sampaccoud: There's no `min_value` or `max_value` properties on `FloatField`. So I used validators and chose 0.1 < x < 10:
```
inline_ratio = models.FloatField(
    ...
    validators=[MinValueValidator(0.1), MaxValueValidator(10)]
)
```

sampaccoud:
@jbpenrath I'm done fixing your comments. The third one was not just a neatpick :sweat_smile: so feel free to review again.

tests/plugins/lti_consumer/test_models.py
```diff
@@ -25,10 +26,40 @@ def get_lti_settings(is_regex=True):
...
+        with self.assertRaises(ValidationError) as context:
+            instance.full_clean()
+
+        self.assertEqual(
+            str(context.exception),
+            "{'inline_ratio': ['Ensure this value is greater than or equal to 0.1.']}",
+        )
+
+    def test_lti_consumer_models_inline_ratio_max(self):
+        """The "inline_ratio" field should not accept values bigger than 1"""
```
jbpenrath, line 48:
```suggestion
        """The "inline_ratio" field should not accept values bigger than 10"""
```

jbpenrath (approved):
Great!

PR #1376, opened 2021-06-04
🔖(chore) bump version to 2.7.0

jbpenrath:
## Description

Bump richie to version 2.7.0

#### Added
- Add "hreflang" meta to all pages to avoid duplicates in search engines

#### Changed
- Send username to LTI Consumer if user is logged in
- Make related courses optional for program page

#### Fixed
- Fix permissions on the LTI consumer plugin when granting instructor role
- Allow configuring the inline ratio and auto-resizing for manually-defined
  LTIConsumerPlugin instances
- Use a unique name to LTIConsumerPlugin iframes to allow several plugins on the
- Fix iframeResizer on LTIConsumer component when several plugins are on the
  same page
- Use a unique name to LTIConsumerPlugin iframes
  to avoid a bug when several plugins are on the same page


sampaccoud (approved):
Maybe take this opportunity to fix UPGRADE.md?

PR #1377, opened 2021-06-04
🐛(lti_consumer) fix permissions granting institutor role more carefully

sampaccoud:
## Purpose

We noticed that when a Marsha LTI consumer plugin is created, all editors on the site can update it regardless of the permissions they have to update this specific page and plugin.

## Proposal

We were granting the LTI `instructor` role as soon as edit mode was on. 
We propose to check plugin update permissions before granting this role and add tests to secure it.

### DIFF
--- a/tests/plugins/lti_consumer/test_api.py
+++ b/tests/plugins/lti_consumer/test_api.py
@@ -7,20 +7,23 @@ import random
 from unittest import mock
 
 from django.contrib.auth.models import AnonymousUser
-from django.test import RequestFactory, TestCase
+from django.test import RequestFactory
 from django.test.utils import override_settings
 
 from cms.api import add_plugin
-from cms.models import Placeholder
+from cms.models import PagePermission, Placeholder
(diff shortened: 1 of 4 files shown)


jbpenrath (approved):
LGTM ✅

kernicPanel (commented):
:thinking: `s/institutor/instructor/g` ?

tests/plugins/lti_consumer/test_api.py
```diff
@@ -177,10 +284,10 @@ def test_lti_consumer_view_set_get_context_method_is_cached(self, mock_params):
         request.toolbar = CMSToolbar(request)
         view_set = LTIConsumerViewsSet()
 
-        with self.assertNumQueries(1):
+        with self.assertNumQueries(9):
```
jbpenrath, line 287: Seems the right value is still `1`

sampaccoud: :man_shrugging: not on my laptop... reverting to old code.

PR #1394, opened 2021-06-25
💥 Fixed error connect with elasticsearch

leduong:
## Purpose

Solve the problem when `make bootstrap` cannot connect to `elasticsearch`

## Proposal
- [x] 💥 Fixed error connect with elasticsearch

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -46,7 +46,7 @@ services:
   nginx:
     image: nginx
     ports:
-      - "8070:8070"
+      - "${NGINX_PORT:-8070}:8070"
     networks:
       default:
         aliases:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -13,7 +13,9 @@ services:
     command: mysqld --character-set-server=utf8 --collation-server=utf8_general_ci
 
   elasticsearch:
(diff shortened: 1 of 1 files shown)


sampaccoud:
@leduong Nice to see that richie works with Elasticsearch 7.5 :astonished: 
Could you paste the error you got with Elasticsearch 6.6 as it should be working as well?

leduong:
# Origin
image

# After
image

sampaccoud:
@leduong Sorry I was busy on something else. I think you are a victim of this configuration trick: https://github.com/openfun/richie/blob/master/docs/docker-development.md#elasticsearch-service-is-always-down

Can you try to follow the instructions in this section of the docs to see if it solves it with Elasticsearch 6.6?

leduong:
Thanks of alot @sampaccoud

sampaccoud:
> Thanks of alot @sampaccoud

So does it mean it works on your side with Elasticsearch 6.6 after this trick?

If yes, we should definitely upgrade to Elasticsearch 7 as it will avoid this bad experience for new comers. However, we will need to add an item in the CI to run tests for version 7 and still continue testing with Elasticsearch 6. 6 as well. Do you want to try to do this in your PR or should we close it and do the change on our side?

PR #1401, opened 2021-07-28
Accessibility: Define base HTML landmarks for screen readers

sandroscosta:
Define the base landmark on base.html to improve overall readability
Create Skip to Content link as the first link on page

## Purpose and Objectives

The improvement of Richies' accessibility rating is a multiple step process. This PR defines main landmarks, such as Header, Footer and Main to allow better readability from Screen Readers. It also defines a first link as a skip to content link, along with a sr-only class to display links and text off screen.

## Evaluation results

Results now show minor issues with accessibility, mainly regarding color contrasts.

Where previously, we had more structural issues, such as landmarks and skip navigation links.


src/richie/apps/core/templates/richie/base.html
```diff
@@ -66,48 +66,60 @@
 
         {# Wrap all body content in an ID'd element to easily remove it from view (using aria-hidden) #}
         {# when a modal is open. This is important for accessibility. #}
+
+        {# First link on a website should always be a skip to content #}
+        {# Access key directive allows users to use keyboard shortcuts #}
+        <a href="#site-content" class="sr-only sr-only-focus" accesskey="c">{% trans "Skip to main content" %}</a>
```
mbenadda, line 72: I think this should be inside `#modal-exclude`? `#modal-exclude` is used to remove content while showing a modal, so I'm not sure the link would even do anything while a modal is open

src/richie/apps/core/templates/richie/base.html
```diff
@@ -66,48 +66,60 @@
...
+                            {% language_chooser "menu/language_menu.html" %}
+                        </div>
                     </div>
-                </div>
-            </nav>
+                </nav>
+            </header>
             {% endblock body_header %}
+            
+            {# Main landmark helps redirect users with dissabilities directly to the content #}
```
mbenadda, line 120:
```suggestion
            {# Main landmark helps redirect users with disabilities directly to the content #}
```

sveetch (approved):
I won't argue on the accessibility improvements themselves since i don't have many skills on it.

About the changes on templates, they seems harmless, it needs to be checked in practice to ensure they are not disturbing the layout since it surround some component with new elements, and maybe Bootstrap adds some basic styles on them (like some margin or else).

I would say it's ok.

src/frontend/scss/generic/_accessibility.scss
```diff
@@ -4,3 +4,45 @@
...
+  height: 1px;
+  padding: 0;
+  overflow: hidden;
+  clip: rect(0, 0, 0, 0);
+  white-space: nowrap;
+  border: 0;
+}
+
+// Screen reader toggable function, including offscreen navigation
+.sr-only-focus:active,
```
mbenadda, line 31: I'm not sure it's appropriate to use this generic class to specifically style the "skip to main content" link.

IMO this class should just basically reset what `.sr-only` does and set the element back in the flow of the page. We can then use specific classes to create styles individually for these kinds of elements as we want to show them.

sandroscosta:
I did some testing and it didn't affect anything in the demo site. Bootstrap is pretty much class based, it only resets default blocks.
I can verify everything a bit better but I didn't notice any issues with spacing of element display.

mbenadda (changes requested):
Hey @sandroscosta and thanks for what looks like a very useful contribution 🙏

I had some remarks as to how best to make these improvements

src/frontend/scss/generic/_accessibility.scss
```diff
@@ -4,3 +4,45 @@
   position: absolute;
   left: -9999rem;
 }
+
+// Screen reader only content
+.sr-only {
```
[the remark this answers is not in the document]
sandroscosta: @mbenadda we can do something like:
```css
.offscreen,
.sr-only {
// rules here
}
```
to mitigate this issues.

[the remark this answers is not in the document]
mbenadda: I mean just replace `offscreen` with `sr-only` wherever it appears. If you're reluctant to go over the whole codebase I can do it, I think it's just a search and replace.

src/richie/apps/core/templates/richie/base.html
sandroscosta: I'll review that.

src/frontend/scss/generic/_accessibility.scss
sandroscosta: I'll refactor that. `.sr-only`classes can be available to every element, when needed, but `:focus` is only accessible through element with a `tabindex` defined.

sandroscosta:
Fixed an additional issue related with the homepage link not having a title.

src/richie/apps/core/templates/richie/base.html
```diff
@@ -66,48 +66,61 @@
...
+
+            {# Wrap header content in a header tag to improve landmarks around the site #}
+
+            <header id="site-header">
+                <nav class="topbar {% block topbar_classes %}{% endblock topbar_classes %}" id="main-menu">
+                    <div class="topbar__container">
+                        <div class="topbar__brand">
+                            <button class="topbar__hamburger"
+                                    data-target="main-menu" aria-label="menu" aria-expanded="false">&#8801;</button>
+                            <a href="/" title="{% trans "Go to homepage" %}" rel="home" accesskey="h">
```
mbenadda, line 86: I think the correct way to do this is to just add "Home" to the alt text for the logo, not add a title attribute for the link

sandroscosta: From my experience, accessibility wise, you shouldn't have elements without context. You have a img element wrapped in an anchor tag, but the screen reader will read and list the link and not the element inside it. NVDA and JAWS will internally create a link index for all links with the link itself and the text inside it (context). We can either do this or opt for a `<span class="sr-only">SITE TITLE</span>` inside the link, to give it context. I've been using https://accessmonitor.acessibilidade.gov.pt/ to do validation, along with Deque AXE and WAVE.

mbenadda: I don't get what "context" is here.

This is actually something we used to do wrong in Richie. We used to put titles on all links and when we did an accessibility audit, the auditor warned us that is was not good practice, as link content is sufficient and titles are just additional noise (except for some cases such as file downloads).

For a link that just contains an image, alt text on the image is sufficient as it is the content of the link and the link's name in the accessibility tree.

sandroscosta: You're absolutely right @mbenadda. An image with an alt tag will do the job. For this specific link I just like to add better context. At the moment, the "homepage" link says the name of the site as defined in the configurations. I like to have a bit more context in this homepage link, because of its importance. Having a direct access key and a title that describes what it does makes sense to me, due to my experience in this are. We can refactor it and leave the title out or leave it as it was previously. Sorry if I explained myself wrong and thank you for your feedback.

sampaccoud: @mbenadda we discussed this irl and it's true that, for a screen reader, the most important thing about this block is that it links back to the homepage. The "alt" as it is does not bring any information...

Maybe a solution could be to keep the "title" with the call to action "Go to home page" and set the image alt to an empty string?

Other solutions seem less good. Removing the title and setting the image "alt" to:
- "Go to home page" does not make sense because the image is a logo and has nothing to do with the action assigned to this image by the link...
- "{{ SITE.name }} logo" does not bring any value to the screen reader user

mbenadda: I think setting the alt to an empty string and using the title might be a RGAA non-conformity.

What about changing the alt of the image to "{{ SITE.name }} home"?

sandroscosta: Like I said previously, Your solution is correctly. The documentation confirms it (https://www.w3.org/WAI/WCAG21/Techniques/failures/F89.html). The only note a leave is that this is a special case. The problem with title is that, most times, it repeats the alt text, so blind users ear the same text twice. In this case, this is my proposed solution, and the one I'll implement on our end.

```html
<!-- this gives context to my action, navigation to the homepage. you have aditional access via keyboard shortcut and the rel microformat as described in http://microformats.org/wiki/existing-rel-values#HTML5_link_type_extensions -->
<a href="/" title="{% trans "Go to homepage" %}" rel="home" accesskey="h">
    <!-- this can be an image or an inline svg (in our case). it provides a bit more flexibility, because i'm not bound to the content inside the link-->
    <img src="<image source>" alt="Institution"> <!-- the alt here is still the image description, as a complement to the previous -->
</a>
```
As I said, I can regress this to a previous version or add a `home` at the end of `alt` and then make the changes on our theme. Will be waiting for you feedback. Thank you.

mbenadda: Ok, we can implement this in Richie 👍

No need to theme this on your end.

src/frontend/scss/generic/_accessibility.scss
```diff
@@ -1,6 +1,44 @@
 // Helper class to hide elements from sighted users and have accessibility devices still read them
 // to users that need it.
-.offscreen {
+
+// Screen reader only content
+.offscreen,
```
mbenadda, line 5: So do you want someone from FUN to remove the `offscreen` class and replace it with `sr-only`, or would you do it in this PR?

src/frontend/scss/generic/_accessibility.scss
```diff
@@ -1,6 +1,44 @@
...
+    display: block;
+    font-size: 1em;
+    height: auto;
+    left: 5px;
+    line-height: normal;
+    text-decoration: none;
+    top: 5px;
+    width: auto;
+    z-index: 100000; /* Above WP toolbar. */
+    background-color: r-color('midnightblue');
```
mbenadda, line 40: I think a lot of these styles should not be on this class. This should just "cancel out" `sr-only` IMO and put the element back in the flow of the page.

Then we can make a class for `.skip-to-content` to add styles that are specific to this exact link

sandroscosta: Ok. Will do these changes. So we'll stay with the offscreen, remove `sr-only` and implemente `.skip-to-content` with button specific styling? Does this sound good? If so, I'll do that.

mbenadda: We can keep the `sr-only` name, IMO it is clearer than `offscreen`, the only downside is that it needs a search-and-replace to remove existing `offscreen` uses.

Apart from that, sounds great!

sandroscosta: Exactly that @mbenadda. Will fix that ASAP. Thanks for all the help.

mbenadda (approved):
Great work! Thanks for your contribution 👍

sampaccoud:
@sandroscosta Can you rebase it and we'll merge it?

sampaccoud:
@sandroscosta Welcome back :partying_face: 
Can you rebase this PR so we can include it in the upcoming release! :pray:

PR #1402, opened 2021-07-28
🐛(SEO) fix opengraph og:image meta

igobranco:
The **og:image** head meta need to have an absolute URL.

Currently on [demo](https://demo.richie.education/en/) we have this value http://edx.richie.education//dpaz9rouqy0e2.cloudfront.net/static/richie/images/logo.bc96dea393cb.png when a social network scraps the page it won't found the image because the URL is bad formed. 

Richie is prefixing the `og:image` meta content value with the site name. But when static assets of the richie site is behind a CDN the URL of the og:image has an incorrect value.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Fixed
 
+- Fix open graph og:image meta when static files are on a CDN
 - Fix course teaser layout issues on Safari
 
 ### Added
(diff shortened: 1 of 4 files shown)


sampaccoud:
@igobranco we are finalizing PR https://github.com/openfun/richie/pull/1390 which is refactoring OpenGraph tags. Can you check if the subject of your PR is covered or make a PR directly on the [add-schema-metadata-courses](https://github.com/openfun/richie/tree/add-schema-metadata-courses) branch to improve it?

igobranco:
@sampaccoud the PR still makes sense to include on the add-schema-metadata-courses branch. So I'm going to create a new PR to that branch.

PR #1404, opened 2021-07-28
🔍️(SEO) add course cover to og image meta

igobranco:
Added the course cover image using 1200x630 size for the open graph og:image meta

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 ### Fixed
 
 - Fix course teaser layout issues on Safari
+- Added the course cover image using 1200x630 size for the open graph og:image meta
 
 ### Added
 
--- a/src/richie/apps/courses/templates/courses/cms/course_detail.html
+++ b/src/richie/apps/courses/templates/courses/cms/course_detail.html
@@ -11,6 +11,19 @@
 {% endblock meta_index_rules %}
 
 
+{% block meta_opengraph_image %}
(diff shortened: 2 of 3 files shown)


sampaccoud:
@igobranco we are finalizing PR https://github.com/openfun/richie/pull/1390 which is refactoring OpenGraph tags. Can you check if the subject of your PR is covered or make a PR directly on the [add-schema-metadata-courses](https://github.com/openfun/richie/tree/add-schema-metadata-courses) branch to improve it?

igobranco:
It's almost the same, almost. 
What you have on your branch want work in PROD, because you have a CDN.
The `og:image` requires an absolute URL of the image.

Another thing, why do you decide to use 300x170 image? 
I think the minimum recommended size are 200x200. https://developers.facebook.com/docs/sharing/webmasters/images/

Can we use the maximum permitted value of 1200 x 630 ?

sampaccoud:
> It's almost the same, almost.
> What you have on your branch want work in PROD, because you have a CDN.
> The `og:image` requires an absolute URL of the image.

Ok let's change this!

> Another thing, why do you decide to use 300x170 image?
> I think the minimum recommended size are 200x200. https://developers.facebook.com/docs/sharing/webmasters/images/
> Can we use the maximum permitted value of 1200 x 630 ?

You're right it seems to be Facebook's recommandation :+1:

igobranco:
@sampaccoud ok, so I'm going to send a new PR to fix those on the add-schema-metadata-courses branch.

igobranco:
Create a new PR https://github.com/openfun/richie/pull/1406

PR #1406, opened 2021-07-28
🔍️(SEO) fix og image meta

igobranco:
Change course cover image using 1200x630 size and if the course don't have a cover show the site logo.
Also add the og:image:width and og:image:height for sharing performance.

### DIFF
--- a/src/richie/apps/core/context_processors.py
+++ b/src/richie/apps/core/context_processors.py
@@ -46,6 +46,17 @@ def site_metas(request):
     if getattr(settings, "CDN_DOMAIN", None):
         context["CDN_DOMAIN"] = settings.CDN_DOMAIN
 
+    # Add a MEDIA_URL_PREFIX to context to prefix the media url files to have an absolute URL
+    if settings.MEDIA_URL.startswith("//"):
+        # Eg. //my-cdn-user.cdn-provider.com/media/
+        context["MEDIA_URL_PREFIX"] = f"{request.scheme:s}:"
+    elif settings.MEDIA_URL.startswith("/"):
+        # Eg. /media/
+        context["MEDIA_URL_PREFIX"] = f"{protocol:s}://{site_current.domain:s}"
+    else:
+        # Eg. https://my-cdn-user.cdn-provider.com/media/
+        context["MEDIA_URL_PREFIX"] = ""
+
     authentication_delegation = getattr(
         settings, "RICHIE_AUTHENTICATION_DELEGATION", None
     )
--- a/src/richie/apps/core/templates/richie/base.html
+++ b/src/richie/apps/core/templates/richie/base.html
@@ -1,4 +1,4 @@
-{% load i18n cms_tags menu_tags rfc_5646_locale static sekizai_tags %}
+{% load i18n cms_tags menu_tags rfc_5646_locale static sekizai_tags full_static_tags %}
 <!doctype html>
 <html lang="{{ LANGUAGE_CODE|rfc_5646_locale }}">
     <head>
@@ -39,7 +39,7 @@
             <meta property="og:url" content="{{ SITE.web_url }}{{ current_page.get_absolute_url }}">
             {% block meta_rdfa_context %}
             <meta property="og:type" content="website" />
-            <meta property="og:image" content="{{ SITE.web_url }}{% static "richie/images/logo.png" %}" />
+            <meta property="og:image" content="{% static_absolute "richie/images/logo.png" %}" />
             {% endblock meta_rdfa_context %}
             {% endblock meta_rdfa %}
 
--- a/src/richie/apps/courses/templates/courses/cms/course_detail.html
+++ b/src/richie/apps/courses/templates/courses/cms/course_detail.html
@@ -11,9 +11,16 @@
 {% endblock meta_index_rules %}
 
 {% block meta_rdfa_context %}
-    <meta property="og:type" content="website" />
-    {% get_placeholder_plugins "course_cover" as plugins %}
-    <meta property="og:image" content="{% thumbnail plugins.0.picture 300x170 crop upscale subject_location=plugins.0.picture.subject_location %}">
+    {% get_placeholder_plugins "course_cover" as plugins or %}
+        {{ block.super }}
+    {% endget_placeholder_plugins %}
+    {% blockplugin plugins.0 %}
+        <meta property="og:type" content="website" />
+        {% thumbnail instance.picture 1200x630 crop upscale subject_location=instance.picture.subject_location as thumb %}
+        <meta property="og:image" content="{{ MEDIA_URL_PREFIX }}{{ thumb.url }}">
+        <meta property="og:image:width" content="{{ thumb.width }}">
+        <meta property="og:image:height" content="{{ thumb.height }}">
+    {% endblockplugin %}
 {% endblock meta_rdfa_context %}
 
 {% block body_rdfa %} vocab="https://schema.org/" typeof="Course"{% endblock body_rdfa %}
--- a/tests/apps/core/test_cache.py
+++ b/tests/apps/core/test_cache.py
@@ -74,7 +74,7 @@ class TestSentinelClient(TestCase):
         self.assertEqual(res, 9223372036854775805)
 
     def test_delete(self):
-        """Test the delete cache operation """
+        """Test the delete cache operation"""
         cache.set_many({"a": 1, "b": 2, "c": 3})
         res = cache.delete("a")
         self.assertTrue(bool(res))
@@ -136,7 +136,7 @@ class TestSentinelClient(TestCase):
         self.assertGreater(ttl + 0.5, 20)
 
     def test_get_default(self):
-        """Test get cache operation with a default value """
+        """Test get cache operation with a default value"""
         res = cache.get("test_key")
         self.assertIsNone(res)
 
@@ -240,7 +240,7 @@ class TestSentinelClient(TestCase):
         self.assertEqual(res, "heló")
 
     def test_incr(self):
-        """"Test the incr cache operation"""
+        """ "Test the incr cache operation"""
         cache.set("num", 1)
 
         cache.incr("num")
--- a/tests/apps/courses/test_templates_course_detail_opengraph.py
+++ b/tests/apps/courses/test_templates_course_detail_opengraph.py
@@ -1,0 +1,156 @@
+"""
+Test suite for the Open Graph for course pages
+"""
+import re
+
+from django.test.utils import override_settings
+
+from cms.test_utils.testcases import CMSTestCase
+
+from richie.apps.courses.factories import CourseFactory
+
+
+class TemplatesCourseDetailOpengraphCMSTestCase(CMSTestCase):
+    """Testing a course page open graph meta content values"""
+
+    def test_open_graph_course_meta_og_image(self):
+        """
+        Test if the og:image of a course has the correct value image url, width and height on the
+        metas og:image og:image:width and og:image:heigth
+        """
+        cover_file_name = "cover.jpg"
+
+        course = CourseFactory(
+            page_title="Introduction to Programming",
+            code="IntroProg",
+            page_languages=["en", "fr"],
+            fill_cover={
+                "original_filename": cover_file_name,
+            },
+        )
+        course_page = course.extended_object
+        course_page.publish("en")
+
+        url = course_page.get_absolute_url(language="en")
+        response = self.client.get(url)
+        response_content = response.content.decode("UTF-8")
+
+        match = re.search("<meta[^>]+og:image[^>]+>", response_content)
+        html_meta_og_image = match.group(0)
+        self.assertIn('content="http://example.com/media', html_meta_og_image)
+        self.assertIn(cover_file_name, html_meta_og_image)
+
+        match = re.search("<meta[^>]+og:image:width[^>]+>", response_content)
+        html_meta_og_image = match.group(0)
+        self.assertIn(
+            "1200",
+            html_meta_og_image,
+            msg="og:image:width should have the maximum permitted value of 1200",
+        )
+
+        match = re.search("<meta[^>]+og:image:height[^>]+>", response_content)
+        html_meta_og_image = match.group(0)
+        self.assertIn(
+            "630",
+            html_meta_og_image,
+            msg="og:image:height should have the maximum permitted value of 630",
+        )
+
+    @override_settings(MEDIA_URL="https://xyz.cloudfront.net/media/")
+    def test_open_graph_course_meta_og_image_cdn_domain(self):
+        """
+        Test if the og:image of a course has the correct value when the MEDIA_URL setting is
+        configured with a CDN.
+        """
+        cover_file_name = "cover.jpg"
+
+        course = CourseFactory(
+            page_title="Introduction to Programming",
+            code="IntroProg",
+            page_languages=["en", "fr"],
+            fill_cover={
+                "original_filename": cover_file_name,
+            },
+        )
+        course_page = course.extended_object
+        course_page.publish("en")
+
+        url = course_page.get_absolute_url(language="en")
+        response = self.client.get(url)
+        response_content = response.content.decode("UTF-8")
+
+        match = re.search("<meta[^>]+og:image[^>]+>", response_content)
+        html_meta_og_image = match.group(0)
+        self.assertIn('content="https://xyz.cloudfront.net/media/', html_meta_og_image)
+        self.assertIn(cover_file_name, html_meta_og_image)
+
+    @override_settings(MEDIA_URL="//xyz.cloudfront.net/media/")
+    def test_open_graph_course_meta_og_image_cdn_double_slash(self):
+        """
+        Test if the og:image of a course has the correct value when the MEDIA_URL setting is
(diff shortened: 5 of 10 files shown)


src/richie/apps/courses/templates/courses/cms/course_detail.html
```diff
@@ -11,9 +11,16 @@
...
+    {% get_placeholder_plugins "course_cover" as plugins or %}
+        {{ block.super }}
+    {% endget_placeholder_plugins %}
+    {% blockplugin plugins.0 %}
+        <meta property="og:type" content="website" />
+        {% thumbnail instance.picture 1200x630 crop upscale subject_location=instance.picture.subject_location as thumb %}
+        <meta property="og:image" content="{{ request.scheme }}:{% if not CDN_DOMAIN %}//{{ request.get_host }}{% endif %}{{ thumb.url }}">
+        <meta property="og:image:width" content="{{ thumb.width }}">
+        <meta property="og:image:height" content="{{ thumb.height }}">
+    {% endblockplugin %}
```
sampaccoud, line 23: Are you sure that this won't be printed as well when there is no cover?

src/richie/apps/courses/templates/courses/cms/course_detail.html
```diff
@@ -11,9 +11,16 @@
...
-    <meta property="og:type" content="website" />
-    {% get_placeholder_plugins "course_cover" as plugins %}
-    <meta property="og:image" content="{% thumbnail plugins.0.picture 300x170 crop upscale subject_location=plugins.0.picture.subject_location %}">
+    {% get_placeholder_plugins "course_cover" as plugins or %}
+        {{ block.super }}
+    {% endget_placeholder_plugins %}
+    {% blockplugin plugins.0 %}
+        <meta property="og:type" content="website" />
+        {% thumbnail instance.picture 1200x630 crop upscale subject_location=instance.picture.subject_location as thumb %}
+        <meta property="og:image" content="{{ request.scheme }}:{% if not CDN_DOMAIN %}//{{ request.get_host }}{% endif %}{{ thumb.url }}">
```
sampaccoud, line 20: I'm surprised that you need `http:` when `CDN_DOMAIN` is set?

tests/apps/courses/test_open_graph_course.py
```diff
@@ -0,0 +1,74 @@
+"""
```
sampaccoud, line 1: [The way we name our test files](https://handbook.openfun.fr/python#test-files), this file could be called something like:
`test_templates_course_detail_opengraph.py`

src/richie/apps/courses/templates/courses/cms/course_detail.html
igobranco: Yes, I think the `og:image` requires the full URL. 
On reference documentation https://ogp.me/ and on forums, all the examples start with `http:` or `https:`.

igobranco: On other PR i've added the [fullstatic templatetag](https://github.com/igobranco/richie/commit/34c107ae9bd1d053d360f53d6389fb007a48316c) but the image thumbnail is completely different, and I think this simple if resolves the problem.
There is an alternative one is to put it to the richie context processor.

tests/apps/courses/test_open_graph_course.py
igobranco: done

tests/apps/courses/test_templates_course_detail_opengraph.py
```diff
@@ -0,0 +1,74 @@
...
+        html_meta_og_image = match.group(0)
+        self.assertIn(
+            "630",
+            html_meta_og_image,
+            msg="og:image:height should have the maximum permitted value of 630",
+        )
+
+    def test_open_graph_course_no_cover_image(self):
+        """
+        Test if a course withthout a cover image has on og:image meta the default logo
```
sampaccoud, line 58:
```suggestion
        Test if a course without a cover image has the default logo as og:image meta
```

src/richie/apps/courses/templates/courses/cms/course_detail.html
sampaccoud: I wasn't sure so I pulled the branch. In fact, when `CDN_DOMAIN` is set, you will also use a custom storage backend that will return the full path for media files as you can see here: https://github.com/openfun/richie-site-factory/blob/master/sites/funmooc/src/backend/base/storage.py#L50

Here is the diff of the tests I made locally:
```diff
diff --git a/src/richie/apps/courses/templates/courses/cms/course_detail.html b/src/richie/apps/courses/templates/courses/cms/course_detail.html
index f7799b01..06b584ff 100644
--- a/src/richie/apps/courses/templates/courses/cms/course_detail.html
+++ b/src/richie/apps/courses/templates/courses/cms/course_detail.html
@@ -17,7 +17,7 @@
     {% blockplugin plugins.0 %}
         <meta property="og:type" content="website" />
         {% thumbnail instance.picture 1200x630 crop upscale subject_location=instance.picture.subject_location as thumb %}
-        <meta property="og:image" content="{{ request.scheme }}:{% if not CDN_DOMAIN %}//{{ request.get_host }}{% endif %}{{ thumb.url }}">
+        <meta property="og:image" content="{% if not CDN_DOMAIN %}{{ SITE.web_url }}{% endif %}{{ thumb.url }}">
         <meta property="og:image:width" content="{{ thumb.width }}">
         <meta property="og:image:height" content="{{ thumb.height }}">
     {% endblockplugin %}
diff --git a/tests/apps/courses/test_templates_course_detail_opengraph.py b/tests/apps/courses/test_templates_course_detail_opengraph.py
index faa31b79..65769492 100644
--- a/tests/apps/courses/test_templates_course_detail_opengraph.py
+++ b/tests/apps/courses/test_templates_course_detail_opengraph.py
@@ -3,6 +3,9 @@ Test suite for the Open Graph for course pages
 """
 import re
 
+from django.test.utils import override_settings
+from django.utils.translation import override
+
 from cms.test_utils.testcases import CMSTestCase
 
 from richie.apps.courses.factories import CourseFactory
@@ -35,6 +38,7 @@ class TemplatesCourseDetailOpengraphCMSTestCase(CMSTestCase):
 
         match = re.search("<meta[^>]+og:image[^>]+>", response_content)
         html_meta_og_image = match.group(0)
+        self.assertIn('content="http://example.com/media', html_meta_og_image)
         self.assertIn(cover_file_name, html_meta_og_image)
 
         match = re.search("<meta[^>]+og:image:width[^>]+>", response_content)
@@ -53,6 +57,34 @@ class TemplatesCourseDetailOpengraphCMSTestCase(CMSTestCase):
             msg="og:image:height should have the maximum permitted value of 630",
         )
 
+    @override_settings(CDN_DOMAIN='xyz.cloudfront.net')
+    def test_open_graph_course_meta_og_image_cdn_domain(self):
+        """
+        Test if the og:image of a course has the correct value image url, width and height on the
+        metas og:image og:image:width and og:image:heigth
+        """
+        cover_file_name = "cover.jpg"
+
+        course = CourseFactory(
+            page_title="Introduction to Programming",
+            code="IntroProg",
+            page_languages=["en", "fr"],
+            fill_cover={
+                "original_filename": cover_file_name,
+            },
+        )
+        course_page = course.extended_object
+        course_page.publish("en")
+
+        url = course_page.get_absolute_url(language="en")
+        response = self.client.get(url)
+        response_content = response.content.decode("UTF-8")
+
+        match = re.search("<meta[^>]+og:image[^>]+>", response_content)
+        html_meta_og_image = match.group(0)
+        self.assertIn('content="/media', html_meta_og_image)
+        self.assertIn(cover_file_name, html_meta_og_image)
+
     def test_open_graph_course_no_cover_image(self):
         """
         Test if a course withthout a cover image has on og:image meta the default logo
@@ -71,4 +103,5 @@ class TemplatesCourseDetailOpengraphCMSTestCase(CMSTestCase):
 
         match = re.search("<meta[^>]+og:image[^>]+>", response_content)
         html_meta_og_image = match.group(0)
+        self.assertIn('content="http://example.com/static', html_meta_og_image)
         self.assertIn("richie/images/logo.png", html_meta_og_image)
(END)
```

igobranco: > In fact, when CDN_DOMAIN is set, you will also use a custom storage backend that will return the full path for media files as you can see here: https://github.com/openfun/richie-site-factory/blob/master/sites/funmooc/src/backend/base/storage.py#L50

It will append the `//<cdn_domain>/media` to the static files, but they want start with `https://<cdn_domain>/media` that the specs requires. I'm going to add a test with the `CDN_DOMAIN` setting similar to the one you sent.

igobranco: I think this code is more obvious:
```html
<meta property="og:image" content="{% if CDN_DOMAIN %}{{ request.scheme }}:{% else %}{{ SITE.web_url }}{% endif %}{{ thumb.url }}">
```

The test is kind of strange, because we need to test with `content="http:/media` because we can't test with the setting `STATICFILES_STORAGE` with the class CDNManifestStaticFilesStorage

igobranco: @sampaccoud why do you have the CDNManifestStaticFilesStorage class ?
https://github.com/openfun/richie-site-factory/blob/master/sites/funmooc/src/backend/base/storage.py#L13-L52

I think it would be better to remove that class from your code and derived if from `STATIC_URL` on the context processor. And use the `STATIC_URL` and `MEDIA_URL` on our Richie apps (Demo, Fun-mooc, NAU, etc.).
Current settings snippet:
```python
    STATIC_URL = "/static/"
    MEDIA_URL = "/media/"
    MEDIA_ROOT = os.path.join(DATA_DIR, "media")
    STATIC_ROOT = os.path.join(DATA_DIR, "static")
```

Proposed settings snippet:
```python
    STATIC_URL = "https://" + os.environ.get("CDN_DOMAIN") + "/static/"
    MEDIA_URL = "https://" + os.environ.get("CDN_DOMAIN") + "/media/"
    MEDIA_ROOT = os.path.join(DATA_DIR, "media")
    STATIC_ROOT = os.path.join(DATA_DIR, "static")
```

Note, the STATIC_URL and MEDIA_URL need to include the protocol.

With this proposed change the code is much more simpler:
```
<meta property="og:image" content="{{ thumb.url }}">
```

What do you think ?

igobranco:
@sampaccoud 
I've updated this PR. I've included the fix for the default og:image meta value.

To make it optional the use of a CDN, I needed to configure the course og:image like this:
```python
        <meta property="og:image" content="{% if MEDIA_URL|first in '/' %}{{ SITE.web_url }}{% endif %}{{ thumb.url }}">
```
https://github.com/igobranco/richie/blob/igobranco/opengraph_seo_add_course_cover_to_og_image_2/src/richie/apps/courses/templates/courses/cms/course_detail.html#L20

I think it would be better if we separate the `og:type` and `og:image`on the `meta_rdfa_context` block. Eg. if a person don't have any image, it would be better to fall back to the site logo.

src/richie/apps/core/context_processors.py
```diff
@@ -46,6 +46,17 @@ def site_metas(request):
     if getattr(settings, "CDN_DOMAIN", None):
         context["CDN_DOMAIN"] = settings.CDN_DOMAIN
 
+    # Add a MEDIA_URL_PREFIX to context to prefix the media url files to have an absolute URL
+    if settings.MEDIA_URL.startswith('//'):
+        # Eg. //my-cdn-user.cdn-provider.com/media/
+        context["MEDIA_URL_PREFIX"] = request.scheme + ":"
```
sampaccoud, line 52: Make it an f string as well?
```suggestion
        context["MEDIA_URL_PREFIX"] = f"{request.scheme:s}:"
```

src/richie/apps/core/templates/richie/base.html
```diff
@@ -39,7 +39,7 @@
             <meta property="og:url" content="{{ SITE.web_url }}{{ current_page.get_absolute_url }}">
             {% block meta_rdfa_context %}
             <meta property="og:type" content="website" />
-            <meta property="og:image" content="{{ SITE.web_url }}{% static "richie/images/logo.png" %}" />
+            <meta property="og:image" content="{% fullstatic "richie/images/logo.png" %}" />
```
sampaccoud, line 42: Is `static_absolute`  a better name?

src/richie/apps/courses/templates/courses/cms/fragment_organization_glimpse.html
```diff
@@ -2,6 +2,7 @@
 {% comment %}Obviously, the context template variable "organization" is required and must be an Organization page extension{% endcomment %}
 
 {% with organization_page=organization.extended_object organization_variant=organization_variant|default:"glimpse" %}
+<div>
```
sampaccoud, line 5: this change does not seem related to this PR?

tests/apps/core/test_templates_base.py
```diff
@@ -0,0 +1,52 @@
...
+        match = re.search("<meta[^>]+og:image[^>]+(/)?>", response_content)
+        html_meta_og_image = match.group(0)
+        self.assertIn(
+            'content="http://testserver/static/richie/images/logo.png',
+            html_meta_og_image,
+        )
+        self.assertIn("richie/images/logo.png", html_meta_og_image)
+
+    @override_settings(STATIC_URL="https://xyz.cloudfront.net/static/")
+    def test_templates_base_meta_og_image_with_cdn(self):
```
sampaccoud, line 35: I think this test should be in the courses app and in the existing file `courses/test_templates_course_detail.py`

tests/apps/core/test_templates_base.py
```diff
@@ -0,0 +1,52 @@
...
+
+from cms.test_utils.testcases import CMSTestCase
+
+from richie.apps.core.helpers import create_i18n_page
+
+
+class TemplatesBaseCMSTestCase(CMSTestCase):
+    """Testing the base.html template"""
+
+    def test_templates_base_meta_og_image(self):
```
sampaccoud, line 16: I think this test should be `test_templates_richie_homepage_meta_og_image` in a file named `test_templates_richie_homepage.py`

tests/apps/core/test_templates_base.py
sampaccoud: Actually it seems redundant with your tests below: https://github.com/openfun/richie/pull/1406/files#diff-5bef639c5eb529ca347f700081792cea0d3530d86d1b385b235f25af1f0cf35bR115

If I did not miss anything it could then be removed?

tests/apps/courses/test_templates_course_detail_rdfa.py
```diff
@@ -523,6 +523,7 @@ def test_templates_course_detail_rdfa(self):
             self.assertTrue(subject in course_run_subjects)
 
         for end_date in ["2030-07-10", "2030-08-01"]:
+            subject, = graph.subjects(
```
sampaccoud, line 526: seems to break the build... keep it like it is?

src/richie/apps/core/context_processors.py
igobranco: Ok

src/richie/apps/core/templates/richie/base.html
igobranco: Renamed.

src/richie/apps/courses/templates/courses/cms/fragment_organization_glimpse.html
igobranco: Probably not. Removing it.

tests/apps/core/test_templates_base.py
igobranco: The test documentation was incorrect, but the code was ok. We need to test the og:image value for the base templates specifically the homepage. The tests are similar on test_templates_course_detail_opengraph.py and test_templates_base.py but the first tests on course pages and the second currently is testing on homepage, both have different implementations, so both need different tests.

tests/apps/core/test_templates_base.py
igobranco: Renamed

tests/apps/courses/test_templates_course_detail_rdfa.py
igobranco: I didn't change this file... merge problem?

igobranco: It has take me some time, but I fixed. The branch add-schema-metadata-courses had different hashed between the upstream openfun and my fork. So at least now I don't have any conflicts.

igobranco:
@sampaccoud I think the current problem is that the base branch is failing tests.

tests/apps/core/test_cache.py
```diff
@@ -240,7 +240,7 @@ def test_save_unicode(self):
         self.assertEqual(res, "heló")
 
     def test_incr(self):
-        """"Test the incr cache operation"""
+        """ "Test the incr cache operation"""
```
sampaccoud, line 243:
```suggestion
        """Test the incr cache operation"""
```

sampaccoud (approved):
@igobranco you're right. So let's merge it like this and I will finalize the work on schema.org to include it in the next release!

Issue #1347, opened 2021-04-22
Remove accents from search queries


sampaccoud:
## Bug Report

**Problematic behavior/code**
If a course has "électricité" in its title, searching "electricite" does not bring any results whereas "électricité" does.

**Expected Behavior**
Accents in search queries should not be taken into account.

**Steps to Reproduce**

https://user-images.githubusercontent.com/1427165/115736071-49456b80-a38b-11eb-959a-cd9fcee97839.mp4

**Environment**
- Richie version: 2.5.0
- Platform: docker

sampaccoud:
So we just need to apply the same solution to the search query as what was done for autosuggest?

mbenadda:
From a cursory glance, removing the accents is handled by the `asciifolding` filter, which is already present in our analyzers for search. As far as I understand it, accents should already be ignored. I have to work on this to find out.

mbenadda:
I spent more time digging into this. It seems to me like it's not a general problem around accents in words.

There's a specific problem with "électricité" as a word. If you try accented/accentless variations for any other word (any I could find in our course titles anyway), the search results are the same.

I tried looking into the ES side with `explain` and `analyze` queries but it did not yield any results for me. I think we need a pair-programming session to find out what's going on with "électricité" that breaks search.

PR #1412, opened 2021-08-05
🐛(search) fix search bug due to wrong ordering of char filters

sampaccoud:
### Purpose

We noticed that courses with the word "`électricité`" could not be found with a search query lacking accents: "`electricite`".

### Proposal

The problem was due to a wrong ordering of filters. We were stemming before doing the ascii folding. As a result:

- "`électricité`" was indexed as "`electr`"
- searching "`électricité`" was analyzed as "`electr`" => result found :smiley: 
- searching "`electricite`" was analylzed as "`electricit`" => result not found :thinking: 

I changed the order of the char filters in our custom analyzers and added tests to secure the fix.

Fixes https://github.com/openfun/richie/issues/1347

### DIFF
--- a/tests/apps/search/test_partial_mappings.py
+++ b/tests/apps/search/test_partial_mappings.py
@@ -130,3 +130,35 @@ class PartialMappingsTestCase(TestCase):
                 },
             },
         )
+
+    def test_partial_mappings_french_diacritics(self):
+        """
+        Make sure words ending in "icité" are analyzed the same way whether or not there
+        is an accent.
+        """
+        document_type = "stub"
+        index_name = "stub_index"
+        mapping = {"dynamic_templates": MULTILINGUAL_TEXT}
+
+        # Create the index and set a mapping that includes the pattern we want to test
+        self.indices_client.create(index=index_name)
+        self.indices_client.put_mapping(
+            index=index_name, doc_type=document_type, body=mapping
+        )
+        # The index needs to be closed before we set an analyzer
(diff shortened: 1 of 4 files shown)


NamFra (approved):
👏😎

tests/apps/search/test_partial_mappings.py
```diff
@@ -130,3 +130,32 @@ def test_partial_mappings_multilingual_text(self):
                 },
             },
         )
+
+    def test_partial_mappings_french_diacritics(self):
+        """Make sure words ending in "icité" are analyzed as expected"""
```
mbenadda, line 135:
```suggestion
        """Make sure words ending in "icité" are analyzed the same way whether or not there is an accent."""
```

tests/apps/search/test_query_courses_edge_cases.py
```diff
@@ -316,3 +316,44 @@ def test_query_courses_no_title(self, *_):
         content = json.loads(response.content)
         self.assertEqual(len(content["objects"]), 1)
         self.assertEqual(content["objects"][0]["id"], "xyz")
+
+    def test_query_courses_french_ending_with_diacritic(self, *_):
+        """
+        Check that words ending in "ité" are well analyzed.
+        This test follows are bug in ordering of char filters.
```
mbenadda, line 323:
```suggestion
        This test follows a bug in ordering of char filters.
```

mbenadda (approved):
👍👍

PR #1413, opened 2021-08-05
Frontend/extend window interface

jbpenrath:
## Purpose

Within window objects we add two custom properties:
- `__richie_frontend_context__`
- `__RICHIE__`

As Typescript do not know that window object contains those properties, currently we are using `(window as any)` syntax which is a little ugly.

## Proposal

- [x] Declare custom properties added to window through global interface.

### DIFF
--- a/src/frontend/js/index.tsx
+++ b/src/frontend/js/index.tsx
@@ -106,4 +106,4 @@ async function render() {
 document.addEventListener('DOMContentLoaded', render);
 
 // In some case, you would like to render Richie components manually
-(window as any).__RICHIE__ = Object.create({ render });
+window.__RICHIE__ = Object.create({ render });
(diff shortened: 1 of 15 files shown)


src/frontend/js/index.tsx
```diff
@@ -106,4 +106,4 @@ async function render() {
 document.addEventListener('DOMContentLoaded', render);
 
 // In some case, you would like to render Richie components manually
-(window as any).__RICHIE__ = Object.create({ render });
+window.__RICHIE__ = Object.create({ render });
```
kernicPanel, line 109: `__RICHIE__` attribute name does'nt reflects what it contains, maybe something like `__PRERENDERED_COMPONENT__` would be more meaningful ?

kernicPanel (approved):

src/frontend/js/index.tsx
jbpenrath: Agree with you that `__RICHIE__` is unclear about it's content.
But I think this name should be kept as generic as possible to put all other stuff related to richie frontend  in the future. Furthermore, I think it's a good practice to keep `RICHIE` in the name as this property is mix with all other window properties, in this way we are able to distinguish richie's properties easily.

kernicPanel: I see your point :+1:

PR #1414, opened 2021-08-05
✨(blogposts) sort related blog posts by their publication date

sampaccoud:
## Purpose

When displaying related blogposts on a blogpost's detail page, we were sorting by the related blogposts' "path" (same order as in  the pages tree).

We were waiting for user feedback which requested to sort them by their publication date (the latest coming first).

## Proposal

Sort related blogposts by their publication date and add a test to secure this behavior.


jbpenrath (approved):
LGTM ✅

PR #1416, opened 2021-08-06
✨(search) add introduction text to search index and API

sampaccoud:
## Purpose

The course introduction is an important piece of text on which we should search.

Also, we want to share the course introduction on the API to be able to build glimpses with more information than the
present card.

## Proposal

Add course introduction text to the course search index and to the displayed fields on the search API.


jbpenrath (approved):
Looks good to me ✅ 
I think #1418 is warmly welcome after this improvement !

PR #1417, opened 2021-08-06
Add course code to search index and api

sampaccoud:
## Purpose

We want to be able to search a course by its code. 

## Proposal

- [x] add course code to the search index and API
- [x] use specific analyzers that don't alter the code but normalize it and allow finding it with a partial match (via ngrams)
- [x] display code on the course glimpse (both DjangoCMS and ReactJS versions)
- [x] write frontend and backend tests to secure this work
- [x] change the "organization" icon on the course glimpse to the icon of a "public building" to better symbolize an institution providing courses (the previous pin icon carried the wrong idea of a geographical location)

## Screenshots

This is how the course glimpse now looks with the code:


tests/apps/search/test_partial_mappings.py
```diff
@@ -162,3 +162,51 @@ def test_partial_mappings_french_diacritics(self):
...
+
+        # Create the index and set a mapping that includes the pattern we want to test
+        self.indices_client.create(index=index_name)
+        # The index needs to be closed before we set an analyzer
+        self.indices_client.close(index=index_name)
+        self.indices_client.put_settings(body=ANALYSIS_SETTINGS, index=index_name)
+        self.indices_client.open(index=index_name)
+
+        self.assertEqual(
+            list(
```
sveetch, line 178: Why are you enforcing a list() on a list you've just constructed inside ?

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
```
jbpenrath, line 10: This commit (3538b2ef30096f81428ea8497eca40f149635044) should not be embedded in this PR as it has a dedicated one (#1416)

src/richie/apps/search/forms.py
```diff
@@ -167,18 +167,45 @@ def get_queries(self):
...
+                                                f"organizations_names.*^{related_content_boost}",
+                                                f"persons_names.*^{related_content_boost}",
+                                            ],
+                                            "query": full_text,
+                                            "type": "best_fields",
+                                        }
+                                    },
+                                    # Boost exact code matches so that they jump our state buckets
+                                    # A closed course with a perfect code match should be
+                                    # displayed bedore an open course with no perfect code match
```
jbpenrath, line 189:
```suggestion
                                    # displayed before an open course with no perfect code match
```

tests/apps/search/test_partial_mappings.py
sampaccoud: actually it's a list comprehension, so technically a generator so I need to make it a list to compare it. There may be other ways to compare list?

src/richie/apps/search/forms.py
```diff
@@ -167,18 +167,45 @@ def get_queries(self):
...
+                                    {
+                                        "match": {
+                                            "code": {
+                                                "query": full_text,
+                                                "analyzer": "code",
+                                                "boost": 20,
```
jbpenrath, line 195: I tested it locally to search a closed course through its exact code and open courses with no perfect match are still displayed before.

Capture d’écran 2021-08-09 à 10 23 13
Capture d’écran 2021-08-09 à 10 23 04

CHANGELOG.md
sampaccoud: Yes I was waiting for it to be merged to rebase :pray:

src/frontend/js/components/CourseGlimpse/index.spec.tsx
```diff
@@ -92,4 +94,16 @@ describe('components/CourseGlimpse', () => {
     screen.getByText('Course 42');
     screen.getByText('Cover');
   });
+
+
+  it('shows does not include the Ref. label when the course code is not set', () => {
+    render(
+      <IntlProvider locale="en">
+        <CourseGlimpse context={contextProps} course={{ ...course, code: null }} />
```
jbpenrath, line 102: According to the `Course` type, `code` property cannot be null.

src/frontend/js/components/CourseGlimpse/index.spec.tsx
```diff
@@ -92,4 +94,16 @@ describe('components/CourseGlimpse', () => {
     screen.getByText('Course 42');
     screen.getByText('Cover');
   });
+
+
+  it('shows does not include the Ref. label when the course code is not set', () => {
```
mbenadda, line 99:
```suggestion
  it('does not include the Ref. label when the course code is not set', () => {
```

src/frontend/js/components/CourseGlimpse/index.tsx
```diff
@@ -43,6 +48,16 @@ const CourseGlimpseBase = ({ context, course }: CourseGlimpseProps & CommonDataP
...
       <div className="course-glimpse__wrapper">
+        <p className="course-glimpse__code">
+          {course.code ? (
+            <FormattedMessage
+              {...messages.ref}
+              values={{
+                code: course.code,
+              }}
+            />
+          ) : null}
```
jbpenrath, line 59: Just a shorcut which improves readability from my point of view.
```suggestion
          {course.code &&
            <FormattedMessage
              {...messages.ref}
              values={{
                code: course.code,
              }}
            />
          }
```

src/frontend/scss/objects/_course_glimpses.scss
```diff
@@ -150,6 +150,14 @@ $course-glimpse-content-padding-sides: 0.7rem !default;
     position: relative;
   }
 
+  &__code {
+    font-size: 0.7rem;
+    color: r-theme-val(course-glimpse, content-color);
+    line-height: 1rem;
+    min-height: 1rem;
+    margin-bottom: 0;
```
jbpenrath, line 158: Sort line ascending.

I don't understand the use of `min-height`, `line-height` should be sufficient here?

mbenadda (approved):
LGTM

src/frontend/scss/objects/_course_glimpses.scss
```diff
@@ -111,11 +111,11 @@ $course-glimpse-content-padding-sides: 0.7rem !default;
...
     position: absolute;
     top: -1.1rem;
-    left: 0;
+    right: 0;
     flex-direction: row;
     align-items: center;
-    border-top-left-radius: 0;
-    border-bottom-left-radius: 0;
+    border-top-right-radius: 0;
+    border-bottom-right-radius: 0;
```
jbpenrath, line 118: There could be quickly an overlap.

Capture d’écran 2021-08-09 à 10 39 44

src/frontend/js/components/CourseGlimpse/index.tsx
mbenadda: IMO the ternary is clearer and almost as short as the `&&` to have conditionals in JSX. [Kent C. Dodds](https://kentcdodds.com/blog/use-ternaries-rather-than-and-and-in-jsx) agrees with me, although I'll admit it's more about readability and the error cases he mentions are a rare occurence.

src/frontend/js/components/CourseGlimpse/index.spec.tsx
sampaccoud: Yep. The linter saw it too so I changed it.

src/frontend/js/components/CourseGlimpse/index.tsx
sampaccoud: +1 to keep the ternary in this PR as it is what is used in the code base. If we want to switch to `&&` there can be a PR for this on the whole code base :wink:

src/frontend/scss/objects/_course_glimpses.scss
sampaccoud: Apparently `line-height` doesn't keep the line height when the code is null that's why I had to add a `min-height` property as I don't want the title to go up in this case.

src/frontend/scss/objects/_course_glimpses.scss
sampaccoud: True... and the code may be longer too... should I lower the code or truncate the icon label?

src/richie/apps/search/forms.py
sampaccoud: True it seems to be because of fuzziness which I added later. Let me write a test for this.

src/frontend/js/components/CourseGlimpse/index.tsx
jbpenrath: Ok let's keep this syntax 👍
I note to myself to use it elsewhere in the codebase.

src/frontend/scss/objects/_course_glimpses.scss
jbpenrath: Ah ok! From my point of view, if `course.code` is null, the DOM should not contain the node `course-glimpse__code`.

So may be we could add a `margin-top: 1rem` to `course-glimpse*__title`
Then a `margin-bottom: -1rem` to `course-glimpse*__code. In this title will have a margin-top to 1rem only if there is no `course.code` above.

sampaccoud: I asked myself this question and in view of schema.org I thought the right way to do it is to keep the markup and give the information that the code is null.

src/frontend/scss/objects/_course_glimpses.scss
jbpenrath: As a quick solution, I think truncate the icon label is the best choice.

src/frontend/scss/objects/_course_glimpses.scss
jbpenrath: Oh ok, I see your point. Let's this node empty ! 👍

tests/apps/search/test_partial_mappings.py
sampaccoud: @sveetch apparently it works without converting to a list :man_shrugging:

src/richie/apps/search/forms.py
sampaccoud: @jbpenrath I added a test and adjusted the field boost. Can you test again?

jbpenrath: 👍  Great ! It works as expected. Search `00046` returns the course with code `00046` first.
Search `004` returns all opened course containing the trigram `004` first then other course containing this trigram and finally other courses.

sampaccoud: cool

src/frontend/scss/objects/_course_glimpses.scss
sampaccoud: I proposed something...

sampaccoud: @sveetch you might want to take a look at the css :blush:

sveetch: This is more about layout than CSS itself. You putted the reference id just to the left of the category, if category is longer than in your preview, it will go on two lines or more and will go over title to mask it.

You should have put the reference elsewhere like under the title.

jbpenrath: You edited `category_glimpse.scss` so course glimpses rendered through Django are well displayed but glimpses rendered through React are not well rendered because they didn't use category glimpse.


Capture d’écran 2021-08-09 à 16 28 18

sampaccoud: @jbpenrath @sveetch So your comments have discouraged me from pursuing this option. I have pushed a new version to display the code below the organization with a "barcode" icon to illustrate it. WDYT?

sveetch: @sampaccoud i thought about a different approach, making it absolute in the top left glimpse corner like this:

reference_sample

Here is the sample element CSS (you will have to use the right variable for color, radius, etc..):

```
.course-glimpse__reference {
    position: absolute;
    top: 0;
    left: 0;
    z-index: 4;
    padding: 0.2rem 0.4rem;
    background: white;
    border-top-left-radius: .5rem;
    border-bottom-right-radius: .5rem;
    color: black;
}
```

jbpenrath: From my point of view , barcode is fine.
I'll just use red color and same font-size like the organization to group these information visually. (like a metadata section).

sampaccoud: After brainstorming and testing several options with @jbpenrath we decided on:

jbpenrath (approved):
LGTM

PR #1418, opened 2021-08-07
🧐(search) boost title field for a better search experience

sampaccoud:
## Purpose

When the query we type matches the title of a course that is not open, it may appear after some open courses with titles that don't match as good. 

## Proposal

Users expect to see the titles that match their query first.

### DIFF
--- a/tests/apps/search/test_forms_search_courses.py
+++ b/tests/apps/search/test_forms_search_courses.py
@@ -227,7 +227,7 @@ class CourseSearchFormTestCase(TestCase):
                                 "fields": [
                                     "description.*",
                                     "introduction.*",
-                                    "title.*",
+                                    "title.*^20",
                                     "categories_names.*^0.05",
                                     "organizations_names.*^0.05",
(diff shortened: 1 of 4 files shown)


mbenadda (approved):

sampaccoud:
I added a test and found that empirically that a boost factor of 20 is necessary to jump buckets and a good compromise.

tests/apps/search/test_forms_search_courses.py
```diff
@@ -226,7 +226,7 @@ def test_forms_courses_build_es_query_search_by_match_text(self, *_):
                                 "analyzer": "english",
                                 "fields": [
                                     "description.*",
-                                    "title.*",
+                                    "title.*^2",
```
mbenadda, line 229:
```suggestion
                                    "title.*^20",
```

mbenadda (approved):
I was not sure this was practical to test but it turns out that it is. 👍

jbpenrath (approved):

PR #1420, opened 2021-08-10
🚸(blogposts) increase excerpt size limit to 240 characters

sampaccoud:
## Purpose

The default size limit for a blogpost excerpt was a bit short.

## Proposal

Bump the default limit to 240 characters.


jbpenrath (approved):
LGTM ✅

PR #1422, opened 2021-08-11
♻️(courses) improve handling of course and organization codes

sampaccoud:
## Purpose

We were experiencing 500 errors when trying to create a course with a code that already existed. It turned out that organization and course codes were not handled properly, each one doing something better than the other. 

## Proposal

I took both advantages and normalized everything about code handling, then added tests to secure it.

It also fixes normalization on course run synchronization.


src/richie/apps/courses/cms_wizards.py
```diff
@@ -300,10 +314,20 @@ def save(self):
...
+            extended_object=page, code=self.cleaned_data["code"]
+        )
         organization.create_page_role()
         return page
 
+    def clean_code(self):
+        """Ensure that the code field is unique among pages."""
+        code = utils.normalize_code(self.cleaned_data["code"])
+        if code and models.Organization.objects.filter(code=code).exists():
+            raise forms.ValidationError("A course already exists with this code.")
```
jbpenrath, line 327:
```suggestion
            raise forms.ValidationError("An organization already exists with this code.")
```

src/richie/apps/courses/api.py
```diff
@@ -182,7 +183,7 @@ def course_runs_sync(request, version):
         )
 
     # Look for the course targeted by the resource link
-    course_number = lms.extract_course_number(request.data)
+    course_number = normalize_code(lms.extract_course_number(request.data))
```
jbpenrath, line 186: Maybe write a migration to normalize existing codes ?
Because if some codes are not well formatted, `course_run_sync` will not found the course.

sampaccoud: Good point! I thought that in openedx we only use digits but you are right that technically, we should do this data migration. 
I just added the migration file to this PR.

jbpenrath (approved):
👏

PR #1423, opened 2021-08-11
🐛(courses) hide unpublished pages from related objects on public page

sampaccoud:
## Purpose 

When getting related objects on a public page, we excluded related objects that were not published but forgot to exclude objects that had been published and later unpublished.

## Proposal

- [x] Add filtering clause to filter out unpublished objects when getting related objects for a public page
- [x] Add tests to secure behavior

### DIFF
--- a/tests/apps/courses/test_models_person.py
+++ b/tests/apps/courses/test_models_person.py
@@ -30,6 +30,8 @@ class PersonModelsTestCase(TestCase):
         with self.assertNumQueries(2):
             self.assertEqual(str(person), "Person: Page of Lady Louise Dupont")
 
+    # get_courses
+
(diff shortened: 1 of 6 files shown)


tests/apps/courses/test_models_person.py
```diff
@@ -583,6 +593,12 @@ def test_models_person_get_blogposts_public_person_page(self):
             list(person.public_extension.get_blogposts()), [blog_post.public_extension]
         )
 
+        # If the course is unpublished, it should not be displayed on the public
```
jbpenrath, line 596:
```suggestion
        # If the blog post is unpublished, it should not be displayed on the public
```

jbpenrath (approved):
LGTM ✅

PR #1425, opened 2021-08-12
🐛(search) delete object from search indices when a page is unpublished

sampaccoud:
## Purpose

When a page was published in a new version, we were updating it in our search indices, but we forgot to do the same upon unpublishing a page in a language

As a result, some hidden pages were remaining in the search engine until the next complete reindexation.

## Proposal

- [x] Add a signal "post_unpublish"
- [x] On unpublishing a page, check if it is still published in a language:
    * [x] If yes, reindex it to take into account the fact that a language was newly hidden
    * [x] if no, delete it from the index
- [x] Add missing signal and update methods for `person` pages

### DIFF
--- a/tests/apps/search/test_signals.py
+++ b/tests/apps/search/test_signals.py
@@ -10,6 +10,7 @@ from richie.apps.courses.factories import (
     CategoryFactory,
     CourseFactory,
     OrganizationFactory,
+    PersonFactory,
 )
 from richie.apps.search.indexers.courses import CoursesIndexer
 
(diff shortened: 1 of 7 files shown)


tests/apps/search/test_signals.py
```diff
@@ -185,6 +298,163 @@ def test_signals_categories_no_parent(self, mock_bulk, *_):
...
+        self.assertEqual(actions[0]["_type"], "course")
+        self.assertEqual(
+            actions[1]["_id"], str(person.public_extension.extended_object_id)
+        )
+        self.assertEqual(actions[1]["_op_type"], "index")
+        self.assertEqual(actions[1]["_type"], "person")
+
+    def test_signals_persons_unpublish(self, mock_bulk, *_):
+        """
+        Unpublishing an person in a language should update its document in the Elasticsearch
```
jbpenrath, line 402:
```suggestion
        Unpublishing a person in a language should update its document in the Elasticsearch
```

jbpenrath (approved):
Good catch !
Really enjoyable to read.

mbenadda (approved):
LGTM

PR #1426, opened 2021-08-13
✅(search) clarify languages used on course unpublished in a language

sampaccoud:
## Purpose

When a course is unpublished in one language, but it has related objects that are still published in this language, should or shouldn't we use this language's content for the related objects when building the course glimpse? For the moment it does.

## Proposal

This choice is controversial and I had to write these tests to fully understand it. So I propose to commit them to keep this memory.

Note: We could argue in the future that a course glimpse should be built with content only in the same language and not with "as few fallback languages as possible, using available content in each language".

### DIFF
--- a/tests/apps/search/test_indexers_courses.py
+++ b/tests/apps/search/test_indexers_courses.py
@@ -601,6 +601,136 @@ class CoursesIndexersTestCase(TestCase):
             },
         )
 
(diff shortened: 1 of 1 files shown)


jbpenrath (approved):
👍

mbenadda (approved):

PR #1427, opened 2021-08-13
🐛(blogposts) fix section on blogpost detail page

sampaccoud:
### Purpose

On blogpost pages, an apparently useless override on the section styles was breaking the flow.

As a result, adding several glimpses in a section was displaying them on several lines.

### Proposal

- [x] Remove css style overrides for sections on the blogpost detail page


sveetch (approved):
This is the way, it seems like something which is not used anymore, it can be removed safely

PR #1431, opened 2021-08-16
Add a topbar branding block to easily switch logos and images

sandroscosta:
Adding a `topbar_branding` block allows us to easily switch and change properties on the site logo ``. In specific cases, it allows users to place inline svg logos.


sampaccoud (approved):
good one :+1:

PR #1434, opened 2021-08-23
Make Richie compatible with both ES6 and ES7

### LINKED ISSUE #1428: Elasticsearch 7.0 compatibility
Hello Richie team,

I have seen that a elasticsearch 6 is needed to run richie.
Is there any plan to move to ES 7 or is it already compatible ?

Thanks a lot for your work !

mbenadda:
## Purpose

Richie was built with Elasticsearch (ES) 6. We want to upgrade to ES7 while still supporting ES6.

To do this, we need to be able to run tests and run the development project for both versions of ES, and run a CI with both versions to we can ensure we do not break compatibility going forward.

## Proposal

- [x] Update the docker config to enable development & tests with both versions
- [x] Create a new CI job that runs version 6, update existing CI jobs to run version 7
- [x] Ensure compatibility of Richie with both ES6/ES7 with a compatibility layer

Closes #1428.


mbenadda:
NB: disregard the failing test, it's a "sorting results with an empty query" situation. I'll fix it by adding alphabetical default sorting on non-course searches, which is something we want anyway.

sampaccoud:
@mbenadda I just pushed a proposal commit to allow switching the Elasticsearch image but keeping it simple. We are only going to do it if the CI fails with ES6 and we want to investigate. It can be done with:
```bash
ELASTICSEARCH_IMAGE=fundocker/openshift-elasticsearch:6.6.2 bin/pytest
ELASTICSEARCH_IMAGE=fundocker/openshift-elasticsearch:6.6.2 make run
```

src/richie/apps/search/index_manager.py
```diff
@@ -127,8 +128,13 @@ def perform_aliases_update():
             # attempting to create. In Richie, this is not supposed to happen and is usually the
             # result of a broken ES state.
             if exception.error == "invalid_alias_name_exception":
-                # Identify the broken index
-                broken_index = exception.info["error"]["index"]
+                try:
+                    # ES6 provides the name of the broken index directly
+                    broken_index = exception.info["error"]["index"]
```
sampaccoud, line 126: I'm wondering if we could instead get the index name from the alias?

src/richie/apps/search/viewsets/categories.py
```diff
@@ -60,7 +59,7 @@ def list(self, request, version, kind):
             "meta": {
                 "count": len(query_response["hits"]["hits"]),
                 "offset": offset,
-                "total_count": query_response["hits"]["total"],
+                "total_count": query_response["hits"]["total"]["value"],
```
sampaccoud, line 66: this does not seem related to the doc type commit?

tests/apps/search/test_viewsets_courses.py
```diff
@@ -77,7 +77,10 @@ def test_viewsets_courses_search(self, mock_search, *_):
         def mock_search_implementation(index, **_):
             if index == "richie_courses":
                 return {
-                    "hits": {"hits": [{"_id": 523}, {"_id": 861}], "total": 35},
+                    "hits": {
+                        "hits": [{"_id": 523}, {"_id": 861}],
+                        "total": {"value": 35, "relation": "eq"},
```
sampaccoud, line 82: ditto and several other occurences in between.

tests/apps/search/test_viewsets_organizations.py
```diff
@@ -76,7 +76,7 @@ def test_viewsets_organizations_search(self, mock_search):
                         },
                     },
                 ],
-                "total": 32,
+                "total": {"relation": "eq", "value": 32},
```
sampaccoud, line 79: ditto

src/richie/apps/search/elasticsearch.py
```diff
@@ -0,0 +1,96 @@
...
+            for index_name in mappings.keys():
+                mappings[index_name]["mappings"] = mappings[index_name]["mappings"][
+                    DOC_TYPE
+                ]
+
+        return mappings
+
+    def put_mapping(self, body, index=None, params={}):
+        """
+        Add a "_doc" type in kwargs of the put_mapping call to satisfy the requirement in ES.
```
sampaccoud, line 72: explain why the dummy "_doc" maybe?

src/richie/apps/search/elasticsearch.py
```diff
@@ -0,0 +1,96 @@
...
+        return search_response
+
+
+class ElasticsearchIndicesClientCompat7to6(IndicesClient):
+    """
+    Compatibility wrapper around the ES IndicesClient from elasticsearch-py that
+    handles incompatibilities to let Richie run ES6 and ES7.
+    """
+
+    def get_mapping(self, index=None, params={}):
```
sampaccoud, line 54: declaring {} as default param is a Python anti-pattern. Not sure where it comes from and why you need it? It is also causing E1120 if I understand correctly so maybe there is a better way?

src/richie/apps/search/indexers/persons.py
```diff
@@ -96,6 +101,7 @@ def get_es_document_for_person(cls, person, index=None, action="index"):
             },
             "portrait": portrait_images,
             "title": titles,
+            "title_raw": titles,
```
sampaccoud, line 104: Can you explain why we can't sort on "titles"?

src/richie/apps/search/index_manager.py
mbenadda: This information is unavailable in the exception. Here's what it looks like in ES7:
```
{'error': {'root_cause': [{'type': 'invalid_alias_name_exception', 'reason': 'Invalid alias name [richie_test_courses]: an index or data stream exists with the same name as the alias'}], 'type': 'invalid_alias_name_exception', 'reason': 'Invalid alias name [richie_test_courses]: an index or data stream exists with the same name as the alias'}, 'status': 400}
```

Also we're performing a bunch of updates all at once so the list of actions does not tell us which index is broken either.

I don't see another way than using a regex on the error here.

src/richie/apps/search/viewsets/categories.py
mbenadda: I put this in a separate commit

src/richie/apps/search/indexers/persons.py
mbenadda: I added a comment in the relevant mappings. The analysis alters the values and prevents us from using them for sorting (ES raises an exception, but I think the ordering wouldn't make any sense even if it did not).

src/richie/apps/search/text_indexing.py
```diff
@@ -165,5 +165,6 @@
                 "token_chars": ["letter", "digit"],
             }
         },
-    }
+    },
+    "max_ngram_diff": "20",
```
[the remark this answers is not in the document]
mbenadda: Basically what I did here is keep the existing behavior, which needed this change.

We could try different n-grams lengths with our actual data set to try and find if it fits better than 3-20 but IMO it's out of the scope of this PR.

src/richie/apps/search/elasticsearch.py
mbenadda: I tried another way that does not cause E1120. The decorator used by the elasticsearch-py lib on their methods makes the class painful to subclass, hence these shenanigans...

src/richie/apps/search/indexers/persons.py
sampaccoud: IC :pray:

src/richie/apps/search/viewsets/categories.py
sampaccoud: :+1:

src/richie/apps/search/text_indexing.py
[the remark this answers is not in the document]
sampaccoud: Agreed.

src/richie/apps/search/index_manager.py
sampaccoud: ok then. Let's keep it like this and see how it works.

src/richie/apps/search/elasticsearch.py
sampaccoud: Your modification solves the anti-pattern. So can you remove E1120 comments now or it still raises issues with pylint?

mbenadda: I did remove it. The `W0221` is a real need as it expected that my methods and the ones from the lib do not have the same number of arguments (as they do not have the doc type argument)

tests/apps/search/test_query_organizations.py
```diff
@@ -4,16 +4,14 @@
...
+from richie.apps.search import ES_CLIENT, ES_INDICES_CLIENT
+from richie.apps.search.elasticsearch import bulk_compat
 from richie.apps.search.indexers.organizations import OrganizationsIndexer
 from richie.apps.search.text_indexing import ANALYSIS_SETTINGS
 
 ORGANIZATIONS = [
-    {"id": "4214", "title": {"en": "Ünìversity of Paris 14"}},
-    {"id": "4209", "title": {"en": "ÙnĪversity of Lyon VI"}},
+    {"id": "4214", "title": {"en": "University of Pârïs 14"}},
+    {"id": "4209", "title": {"en": "University of Lÿõn VI"}},
```
sampaccoud, line 14: Just curious why you changed the accents in the test.

sampaccoud (approved):

PR #1439, opened 2021-09-06
Fix facet localization bug after ES7 migration

mbenadda:
## Purpose

During our migration to ES7, we introduced a bug that was not directly caused by ES7, but by code changes we made during the migration. We naively used the result of a dict's `.keys()` method like it was a list, however it is not a list nor can it be serialized to one. Instead we needed to just spread the dict to get a list of its keys.

## Proposal

This was not caught by our tests as the place where the bug occurs was mocked out in all of our tests. We added a new test suite that does not mock out this part so we can avoid regressions with this code in the future.


jbpenrath (approved):
~Maybe update the CHANGELOG by adding an entry under Fixed.~

LGTM

PR #1441, opened 2021-09-08
🐛(courses) reverse 401 and 403 errors in course run sync webhook

sampaccoud:
## Purpose

We mixed up both error codes:
- if there is no token, the user is anonymous and is not allowed to access the resource => 403
- if the token is invalid, the user can't be authenticated => 401

## Proposal

Reverse error codes and update tests.


jbpenrath (approved):
LGTM ✅

Issue #1440, opened 2021-09-08
Can not configure JS_COURSE_REGEX with backslash on regex


igobranco:
## Bug Report

**Expected behavior/code**
When configuring the `JS_COURSE_REGEX` Django setting with a value like described in the documentation:
```
^https://lms\.example\.com/courses/(.*)/course/?$
```

The _Courses_ search page don't work because the JavaScript that starts the richie frontend has a syntax error.

JS console log message: `Uncaught SyntaxError: Unexpected token . in JSON at position 381`

The code generated by the Richie back end to be assigned to the `window.__richie_frontend_context__` JS variable raises a parsing error on the brower.

Generated JS that raised the error during JSON parse.
```
"course_regexp": "^https://lms\\.example\\.com/courses/(.*)/course/?$"
```

**Steps to Reproduce**
From a running local demo-site, change the file `env.d/development/dev` content with:
```
# Authentication Backend
AUTHENTICATION_BASE_URL=http://localhost:8073
AUTHENTICATION_BACKEND=base

# LMS Backend
EDX_BASE_URL=http://localhost:8073
EDX_BACKEND=richie.apps.courses.lms.edx.EdXLMSBackend
EDX_JS_BACKEND=openedx-hawthorn
EDX_COURSE_REGEX=^https://lms\.example\.com/courses/(?P<course_id>.*)/course/?$
EDX_JS_COURSE_REGEX=^https://lms\.example\.com/courses/(.*)/course/?$
```

Start the richie
```
make run
```

The page http://localhost:8070/en/courses/ won't start the react frontend that searches courses.

This only occurs when we include a backslash on the `JS_COURSE_REGEX`.

jbpenrath:
Wow big issue, thanks 🙏 

To inject `__richie_frontend_context__` we are using `JSON.parse` but the regexp you pass contains unescaped characters (`\`). I found an easy fix (just use `escapejs` template tag in place of `safe`), I just have to update our tests. I keep you up to date when I will open a PR to fix that !

PR #1442, opened 2021-09-09
🐛(backend) escape frontend_context characters for use in javascript s…

jbpenrath:
## Purpose

`FRONTEND_CONTEXT` could contains some characters that's need to be escaped for use in javascript string. For example, if this variable contains some regexp string, they can contain unescaped back slash characters that raise an error when JSON.parse try to parse this string.

Related issue: #1440

## Proposal

- [x] Escape `FRONTEND_CONTEXT` characters for use in javascript string

### DIFF
--- a/tests/apps/core/test_authentication_delegation.py
+++ b/tests/apps/core/test_authentication_delegation.py
@@ -28,8 +28,9 @@ class UserMenuTests(CMSTestCase):
     )
     def test_user_menu_with_profile_urls(self):
         """
-        user menu should render correctly
-        and includes formatted profile_urls
+        User menu should render correctly and includes formatted profile_urls.
+        Characters should be escaped with the `escapejs` template tag to prevent
+        an issue when this variable contains unescaped characters for javascript (e.g: '\').
         """
         page = create_page(
             title="Home",
@@ -47,7 +48,10 @@ class UserMenuTests(CMSTestCase):
         self.assertIsNotNone(re.search(pattern, str(response.content)))
         self.assertContains(
             response,
-            r'"authentication": {{"endpoint": "{}", "backend": "{}"}}'.format(
+            (
+                r"\u0022authentication\u0022: "
+                r"{{\u0022endpoint\u0022: \u0022{}\u0022, \u0022backend\u0022: \u0022{}\u0022}}"
+            ).format(
                 "https://richie.education:9999",
                 "richie.apps.courses.lms.base.BaseLMSBackend",
             ),
--- a/tests/apps/core/test_pages.py
+++ b/tests/apps/core/test_pages.py
@@ -70,16 +70,46 @@ class PagesTests(CMSTestCase):
     @override_settings(SENTRY_DSN="https://example.com/sentry/dsn")
     @override_settings(RELEASE="9.8.7")
     @override_settings(ENVIRONMENT="test_pages")
+    @override_settings(
+        RICHIE_LMS_BACKENDS=[
+            {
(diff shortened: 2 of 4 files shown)


tests/apps/core/test_authentication_delegation.py
```diff
@@ -47,9 +48,11 @@ def test_user_menu_with_profile_urls(self):
         self.assertIsNotNone(re.search(pattern, str(response.content)))
         self.assertContains(
             response,
-            r'"authentication": {{"endpoint": "{}", "backend": "{}"}}'.format(
-                "https://richie.education:9999",
-                "richie.apps.courses.lms.base.BaseLMSBackend",
+            html.escapejs(
```
sampaccoud, line 51: are we able to foresee what the exact string should be? Because here you are applying the same method as in the template but that does not confirm that the escaping is done as expected and accepting back slashes?

tests/apps/core/test_pages.py
```diff
@@ -80,6 +81,8 @@ def test_page_includes_frontend_context(self):
         """
         page = PageFactory(should_publish=True, template="richie/single_column.html")
         response = self.client.get(page.get_public_url())
-        self.assertContains(response, '"environment": "test_pages"')
-        self.assertContains(response, '"release": "9.8.7"')
-        self.assertContains(response, '"sentry_dsn": "https://example.com/sentry/dsn"')
+        self.assertContains(response, html.escapejs('"environment": "test_pages"'))
+        self.assertContains(response, html.escapejs('"release": "9.8.7"'))
+        self.assertContains(
+            response, html.escapejs('"sentry_dsn": "https://example.com/sentry/dsn"')
```
sampaccoud, line 87: ditto

tests/apps/core/test_authentication_delegation.py
jbpenrath: Ok I'll explicity write expected output.

tests/apps/core/test_authentication_delegation.py
```diff
@@ -28,8 +28,7 @@ class UserMenuTests(CMSTestCase):
     )
     def test_user_menu_with_profile_urls(self):
         """
-        user menu should render correctly
-        and includes formatted profile_urls
+        User menu should render correctly and includes formatted profile_urls
```
sampaccoud, line 31: Add the explanation so that we are not surprised by the way you test these strings :wink: 
```suggestion
        User menu should render correctly and includes formatted profile_urls.
        Characters should be escaped with the `escapejs` template tag to prevent
        an issue when this variable contains unescaped characters for javascript (e.g: '\').
```

tests/apps/core/test_pages.py
```diff
@@ -70,16 +70,45 @@ def test_pages_i18n_hreflang(self):
...
+                "JS_COURSE_REGEX": r"^https://lms\.example\.com/courses/(.*)/course/?$",
+            }
+        ]
+    )
     def test_page_includes_frontend_context(self):
         """
         Create a page and make sure it includes the frontend context as included
-        in `base.html`.
+        in `base.html`. All characters for use in javascript string should be
+        escaped. (i.e: '"' should be escaped with \u0022 and '\' with \u005C).
```
sampaccoud, line 88: Same here
```suggestion
        escaped. (i.e: '"' should be escaped with \u0022 and '\' with \u005C).
        to prevent an issue when this variable contains unescaped characters for javascript (e.g: '\').
```

sampaccoud (approved):

PR #1453, opened 2021-09-20
🐛(website) use falsy IS_PULL_REQUEST values to force docs to deploy

mbenadda:
## Purpose

We previously used false as a value to force deployments of our docs regardless of the (incorrect) values provided by the
Circle CI environment. 

## Proposal

As environment variables are always strings, "false" is actually a truthy value. Use an empty string instead to we do provide a falsy value.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -748,11 +748,13 @@ jobs:
       - checkout
       - run:
           name: Deploying to GitHub Pages
+          # NB: pass falsy values for the PULL REQUEST environment variables to make sure the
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -748,11 +748,13 @@ jobs:
       - checkout
       - run:
           name: Deploying to GitHub Pages
+          # NB: pass falsy values for the PULL REQUEST environment variables to make sure the
+          # deployment is not cancelled by Docusaurus based on incorrect values provide by Circle CI.
```
sampaccoud, line 752: * Provided

sampaccoud (approved):

PR #1456, opened 2021-09-21
✨(search) improve aggregations ordering on filters

mbenadda:
## Purpose

Facet counts (aggregations) in the filters on course searches had different logics applied to them for ordering depending on the kind of filter definition.

## Proposal

We rationalized the way facets are ordered to be either "conf", "count" then "alphabetical" or "alphabetical" then "count" so we always have a consistent expected ordering of results. It also makes it way easier to write tests as there is only one correct ordering.

### DIFF
--- a/src/richie/apps/search/filter_definitions/base.py
+++ b/src/richie/apps/search/filter_definitions/base.py
@@ -252,8 +252,7 @@ class BaseChoicesFilterDefinition(BaseFilterDefinition):
         Add the counts to the values from the initial definition to make them complete
         and sorted as the frontend expects them.
         """
-        # Get filter values to derive human names
-        values = self.get_values()
+        human_names = self.get_values()
 
         # for each facet, we derive the value and the count:
         # eg. for facet key `availability@coming_soon`, the value is `coming_soon`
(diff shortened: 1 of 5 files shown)


src/richie/apps/search/filter_definitions/base.py
```diff
@@ -252,8 +252,7 @@ def get_faceted_definitions(self, facets, data, *args, **kwargs):
         Add the counts to the values from the initial definition to make them complete
         and sorted as the frontend expects them.
         """
-        # Get filter values to derive human names
-        values = self.get_values()
+        static_values = self.get_values()
```
sampaccoud, line 255: should be called get_human_names since you later do 
human_name = static_values[name]

src/richie/apps/search/filter_definitions/base.py
```diff
@@ -281,25 +280,34 @@ def get_faceted_definitions(self, facets, data, *args, **kwargs):
         # Check if there are more available values than we're about to return on the API
         has_more_values = len(facet_counts) > len(facet_counts_dict)
 
+        # Add human names to keys and counts before sorting as some of our sortings
+        # use alphabetical ordering.
+        values = [
```
sampaccoud, line 285: I found it hard cognitively that static values are a couple machine name/human name and that values are a dictionary with these 3 keys... 
I propose that `static_values` is renamed to `human_names` in this method.

src/richie/apps/search/filter_definitions/base.py
```diff
@@ -309,14 +317,15 @@ def get_faceted_definitions(self, facets, data, *args, **kwargs):
                 # If values are removed due to `min_doc_count`, we are not returning them, and
                 # therefore by definition our filter `has_more_values`.
                 "has_more_values": has_more_values
-                or any(count < self.min_doc_count for name, count in facet_counts),
+                or any(value["count"] < self.min_doc_count for value in values),
                 "values": [
-                    {"count": count, "human_name": values[name], "key": name}
-                    for name, count in facet_counts
+                    value
```
sampaccoud, line 322: maybe a good use cas for `filter`

src/richie/apps/search/filter_definitions/courses.py
```diff
@@ -289,22 +290,54 @@ def get_faceted_definitions(self, facets, data, *args, **kwargs):
...
+            values = sorted(
+                values,
+                key=lambda value: (value["count"] * -1, value["human_name"]),
+            )
+        else:
+            # NB: self.SORTING_CONF is not appropriate for Indexables as they are not defined
+            # in the static filter definition but generated from indices.
+            raise ImproperlyConfigured(
+                f'Facet sorting "{self.sorting}" is invalid for filter {self.name}.'
+            )
```
sampaccoud, line 323: Same code a above so could be factorized?

src/richie/apps/search/filter_definitions/courses.py
```diff
@@ -289,22 +290,54 @@ def get_faceted_definitions(self, facets, data, *args, **kwargs):
...
                 # therefore by definition our filter `has_more_values`.
                 "has_more_values": has_more_values
-                or any(count < self.min_doc_count for count in key_count_map.values()),
+                or any(value["count"] < self.min_doc_count for value in values),
                 "values": [
-                    # Aggregate the information from right above to build the values
-                    {"count": count, "human_name": key_i18n_name_map[key], "key": key}
-                    for key, count in key_count_map.items()
-                    # We filter out values that do not meet `min_doc_count` manually instead of
+                    value
```
sampaccoud, line 334: good use cas for `filter`?

src/richie/apps/search/filter_definitions/courses.py
mbenadda: It's not exactly the same as in one case we have SORTING_CONF in our cases, whereas in the other it does not exist. Not sure it's worth it to factorize with only 2 slightly differing instances of this code

sampaccoud: ok ok. I missed the difference.

PR #1459, opened 2021-09-22
🐛(search) lazily retrieve es version

jbpenrath:
## Purpose

Retrieve the elasticsearch server version on initialization could creates bug in contexts where elasticsearch host is not accessible. e.g: during a docker build if we run a manage command, an elasticsearch client is initiliazed and build
failed due to a ConnectionError. So a workaround is to lazily retrieve the elasticsearch version the first time we need to know the version then reuse the value.

## Proposal

- [x] Retrieve `ElasticsearchClientCompat7to6.__es_version__` only when it is needed and cache the result

### DIFF
--- a/sandbox/settings.py
+++ b/sandbox/settings.py
@@ -139,7 +139,7 @@ class Base(StyleguideMixin, DRFMixin, RichieCoursesConfigurationMixin, Configura
 
     # Security
     ALLOWED_HOSTS = []
-    SECRET_KEY = values.Value(None)
+    SECRET_KEY = values.Value("ThisIsAnExampleKeyForDevPurposeOnly")
     # System check reference:
     # https://docs.djangoproject.com/en/2.2/ref/checks/#security
     SILENCED_SYSTEM_CHECKS = values.ListValue(
@@ -648,13 +648,15 @@ class ContinuousIntegration(Test):
 class Production(Base):
     """Production environment settings
 
-    You must define the DJANGO_ALLOWED_HOSTS environment variable in Production
-    configuration (and derived configurations):
+    You must define the DJANGO_ALLOWED_HOSTS and DJANGO_SECRET_KEY environment
+    variables in Production configuration (and derived configurations):
 
     DJANGO_ALLOWED_HOSTS="foo.com,foo.fr"
+    DJANGO_SECRET_KEY="your-secret-key"
     """
 
     # Security
+    SECRET_KEY = values.SecretValue()
     ALLOWED_HOSTS = values.ListValue(None)
     CSRF_COOKIE_SECURE = True
     SECURE_BROWSER_XSS_FILTER = True
--- a/src/richie/apps/search/elasticsearch.py
+++ b/src/richie/apps/search/elasticsearch.py
@@ -3,6 +3,8 @@ Build an interoperability layer that allows Richie to work with ElasticSearch 6
 just like it does with ElasticSearch 7.
 """
 # pragma pylint: disable=W0221
+from django.utils.functional import cached_property
+
 from elasticsearch import Elasticsearch, Transport
 from elasticsearch.client import IndicesClient
 from elasticsearch.helpers import bulk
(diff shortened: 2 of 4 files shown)


sandbox/settings.py
```diff
@@ -139,7 +139,7 @@ class Base(StyleguideMixin, DRFMixin, RichieCoursesConfigurationMixin, Configura
 
     # Security
     ALLOWED_HOSTS = []
-    SECRET_KEY = values.Value(None)
+    SECRET_KEY = values.Value("ThisIsAnExampleKeyForDevPurposeOnly")
```
sampaccoud, line 142: Should be another commit?

src/richie/apps/search/elasticsearch.py
```diff
@@ -24,9 +26,13 @@ def __init__(self, hosts=None, transport_class=Transport, **kwargs):
         of Elasticsearch we're working with.
         """
         super().__init__(hosts=hosts, transport_class=transport_class, **kwargs)
-        # Store the ES version information in memory so we don't have to re-make the
-        # info request constantly.
-        self.__es_version__ = self.info()["version"]["number"][:1]
+
+    @cached_property
+    def __es_version__(self):
```
sampaccoud, line 31: Cached_property is thread safe so I'm pretty sure this would make the query for each incoming request. You should set it yourself in the client class. Would be nice if a test can evidence that 😏

jbpenrath: I think you would say `@cached_property` is not thead safe. I'll try to test this point and update the code accordingly.

jbpenrath: I'll make some successive autocomplete request that use `self.__es_version__` and info is not trigger on each request.

tests/apps/search/test_elasticsearch_compat_layer.py
```diff
@@ -22,48 +47,77 @@ class ElasticSearchCompatLayerTestCase(TestCase):
...
+
+                    self.client.get("/api/v1.0/subjets/")
+                    self.client.get("/api/v1.0/levels/")
+                    self.client.get("/api/v1.0/subjets/8312")
+
+                    mock_es_info.assert_called_once()
+                    mock_es_info.reset_mock()
+
+                    # - Autocomplete checks the es_version,
+                    #   info() should not be trigger again
```
sampaccoud, line 117:
```suggestion
                    #   info() should not be triggered again
```

tests/apps/search/test_elasticsearch_compat_layer.py
```diff
@@ -22,48 +47,77 @@ class ElasticSearchCompatLayerTestCase(TestCase):
...
-        ]
-        bulk_compat(actions=actions, chunk_size=500, client=ES_CLIENT)
-        ES_INDICES_CLIENT.refresh()
+        with mock.patch(
+            "elasticsearch.Elasticsearch.info", wraps=es_client.info
+        ) as mock_es_info:
+            with mock.patch("richie.apps.search.ES_CLIENT", es_client):
+                with mock.patch(
+                    "richie.apps.search.ES_INDICES_CLIENT", es_indices_client
+                ):
```
sampaccoud, line 75: You can avoid indentation:
```python
with mock.patch("elasticsearch.Elasticsearch.info") as mock_info, mock.patch(
    "richie.apps.search.ES_CLIENT"
) as mock_es_info, mock.patch(
    "richie.apps.search.ES_INDICES_CLIENT"
) as mock_es_indices_client:
    # Perform a bunch of ES actions, from index management all the way to searches
```

Actually you only need to grab `mock_es_info` so:
```python
with mock.patch("elasticsearch.Elasticsearch.info"), mock.patch(
    "richie.apps.search.ES_INDICES_CLIENT"
), mock.patch("richie.apps.search.ES_CLIENT") as mock_es_info:
    # Perform a bunch of ES actions, from index management all the way to searches
```

Not why you mock the two other things but I will trust you on this :stuck_out_tongue:

sampaccoud (approved):

tests/apps/search/test_elasticsearch_compat_layer.py
jbpenrath: Because If I do not use a mock here, `ES_CLIENT` and `ES_INDICES_CLIENT` can already be instanciated before and assertion `mock_es_info.assert_called_once()` fails.

PR #1465, opened 2021-09-29
Add section id to the section

sandroscosta:
## Purpose

To allow for easier navigation inside a page, a propose the introduction of a new field in the section plugin, called `section_id`.
This field will allow us to create same page navigation through the use of anchors.


sampaccoud:
@sandroscosta Nice feature. Maybe what we need is the same as what is on the link plugin? 
https://github.com/django-cms/djangocms-link/blob/master/djangocms_link/models.py#L119

sandroscosta:
> @sandroscosta Nice feature. Maybe what we need is the same as what is on the link plugin? https://github.com/django-cms/djangocms-link/blob/master/djangocms_link/models.py#L119

Thank you @sampaccoud. Makes much more sense to have a general attributes field where we can put id and class attributes, as well as data-* fields if we need to. Now to see how can I implement this.

sampaccoud:
> Now to see how can I implement this.

You should be able to use the same field as they use in the link plugin :man_shrugging:

sandroscosta:
@sampaccoud Does it make sense to do a new PR with the attributes field?

sampaccoud:
@sandroscosta can you squash the commits and fix the changelog then we will merge it?

sandroscosta:
@sampaccoud rebase as well?

sampaccoud:
@sandroscosta yes please!

Issue #1415, opened 2021-08-06
Stop using MPTT Paths in ElasticSearch index


mbenadda:
## Feature Request

**Is your feature request related to a problem or unsupported use case? Please describe.**
In Richie's search engine, categories and organization IDs are replaced with MPTT paths from the page tree in Django CMS. This is useful in that it provides a simple way to access the page tree for categories and organizations, and make requests against it, in the context of courses.

However, the page tree frequently changes, which makes the paths very unstable. As a result, there are a lot of ways the ElasticSearch index can become out-of-sync with the actual state of the database, causing bugs throughout Richie which are not easily mapped to this issue.

As a result, we decided the best course of action was to remove MPTT paths from the ElasticSearch index, using real IDs instead. This makes it easier to know what needs to be re-indexed when changes happen and greatly reduces the number of changes that need to be made.

**Describe the solution you'd like**
On the frontend side, there are 3 uses of MPTT paths:
- In `<SearchFilterGroupModal />`, it is used through the regex `x_includes` query string parameter to get facet counts for the filter values in the results of the search. It's just a bunch of OR in the regex, which can easily be replaced by a list of IDs for which to request facets through the API.
- In `<SearchFilterValueParent />`, it is used through the regex `x_includes` query string parameter to get child filter values of the current filter. This can be replaced by a new query string parameter to get the children of a given filter value.
- In `<SearchFilterValueParent />`, it is used to find out if a child or grandchild (or further) of the current filter value is active in the query string. This usage can be kept if MPTT paths are kept as a separate field on the relevant filter values. This use case does not rely on the paths being kept in the ES index.

On the backend side, there is one main use case:
- In `IndexableMPTTFilterDefinition`, it limits the default set of filter values we facet to the direct children of the base page for that filter. We can cut out the middleman and directly use children here. We should end up merging it with `IndexableFilterDefinition`.

Our proposed solution would entirely remove MPTT paths from the ES index, and replace them with real, stable IDs. We can then use Postgres requests and a cache layer to provide MPTT paths on API responses, and a simple Postgres request to get the children of a given filter value.

We can keep using the actual `include` parameter for aggregations in our ElasticSearch requests, abstracting it in our API methods as its use is limited to providing lists of IDs, relying on the server to provide the correct regex from query string parameters which are more structured.

**Discovery, Documentation, Adoption, Migration Strategy**
This needs to mostly be done at the same time as we're changing something that vertically goes through a lot of Richie's search.

[...]

PR #1466, opened 2021-09-29
♻️(search) stop relying on MPTT paths to define search filters

mbenadda:
## Purpose

Task n°2 for #1415 

Filters based on the IndexableMPTTFilterDefinition, which are filters defined by a base page and for which a hierarchy of pages indexed in ES is the list of available filters, were using a regex filter that required the MPTT path of said filters to be used as IDs and indexed on courses.

## Proposal

We replaced the regexp with a (cached) database call to directly provide a list of children instead of an MPTT based regex. This will enable us to link filters with courses through proper IDs as filters do not rely on MPTT paths in their definition.

### DIFF
--- a/src/richie/apps/search/filter_definitions/courses.py
+++ b/src/richie/apps/search/filter_definitions/courses.py
@@ -3,6 +3,8 @@ import re
 from operator import itemgetter
 
 from django import forms
+from django.core.cache import caches
+from django.core.cache.backends.base import InvalidCacheBackendError
 from django.core.exceptions import ImproperlyConfigured
 from django.utils import timezone, translation
 from django.utils.translation import gettext as _
@@ -122,6 +124,21 @@ class IndexableFilterDefinition(TermsQueryMixin, BaseFilterDefinition):
             }
         }
 
+        def value_included(value, include):
+            """
+            Do not build aggregations for values from query string if they do not match
+            the current include filter.
+            Use a function to perform the check as the include filter can be either a regex
+            or a list.
+            """
+            try:
+                # The Elasticsearch include regex matches exact values so we must do the same
+                # by adding ^ (resp. $) at the beginning (resp. at the end) of the pattern.
+                return re.match(f"^{include:s}$", value)
+            # Trying to build the regex with a list will trigger a TypeError
+            except TypeError:
+                return value in include
+
         # Filters aggregation for values that were selected in the querystring (we must force
         # them because they may not be in the n top facet counts but we must make sure we keep
         # them so that they remain available as options that the user can see and unselect)
@@ -142,9 +159,7 @@ class IndexableFilterDefinition(TermsQueryMixin, BaseFilterDefinition):
                     }
                 }
                 for value in data.get(self.name, [])
-                # The Elasticsearch include regex matches exact values so we must do the same
-                # by adding ^ (resp. $) at the beginning (resp. at the end) of the pattern.
-                if re.match(f"^{include:s}$", value)
+                if value_included(value, include)
             }
         )
 
@@ -344,7 +359,7 @@ class IndexableFilterDefinition(TermsQueryMixin, BaseFilterDefinition):
         }
 
 
-class IndexableMPTTFilterDefinition(IndexableFilterDefinition):
+class IndexableHierarchicalFilterDefinition(IndexableFilterDefinition):
     """
     Some of our filters are a special case of terms-based filter as they use MPTT paths
     as IDs and enable limiting faceting by path.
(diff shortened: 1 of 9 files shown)


src/richie/apps/search/filter_definitions/courses.py
```diff
@@ -368,8 +385,16 @@ def aggs_include(self):
         """
         if self.reverse_id:
             if self.base_page:
-                node = self.base_page.node
-                return f".*-{node.path:s}.{{{node.steplen:d}}}"
+                if not self._aggs_include:
+                    self._aggs_include = []
+                    for page in self.base_page.get_child_pages().select_related("node"):
```
sampaccoud, line 390: Wondering how it scales if the number of children is high? Should we limit to a max number of children? For performance but above all for Ux reasons because we can't propose more than a few and should show the "see more" button if there are more results?

src/richie/apps/search/filter_definitions/courses.py
```diff
@@ -368,8 +385,16 @@ def aggs_include(self):
...
+                        path = page.node.path
```
sampaccoud, line 391: Since what you are looking for is on the node, not on the page, it's probably better to work on the node directly?
Maybe something like this:
```
paths = self.base_page.node.get_children().values_list("path", flat=True)
return [f"{prefix}-{path}" for path in paths for prefix in ["P", "L"]]
```

src/richie/apps/search/filter_definitions/courses.py
```diff
@@ -368,8 +385,16 @@ def aggs_include(self):
         """
         if self.reverse_id:
             if self.base_page:
-                node = self.base_page.node
-                return f".*-{node.path:s}.{{{node.steplen:d}}}"
+                if not self._aggs_include:
```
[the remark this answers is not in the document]
mbenadda: I thought `@cached_property` was not shared between requests but what is set on `self` was. Should we use django Local Memory Caching then? Maybe the same thing is appropriate for `base_page` too?

src/richie/apps/search/filter_definitions/courses.py
mbenadda: But at this point the issue is we have not made the ES search request yet, so we do not know which of the children will be in the top 10 values returned by ES.

If there are 10000 children we need to allowlist them all in the `include` param, or look into a way to filter out facets after the ES request is done

src/richie/apps/search/filter_definitions/courses.py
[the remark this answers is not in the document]
sampaccoud: I'm not sure actually it depends when the class is instantiated. The problem is that if it is cached then the problem is worse because we would have to restart all pods for a change to be taken into account.
So caching in local memory with a timeout seems better?

src/richie/apps/search/filter_definitions/courses.py
sampaccoud: oh yes! Maybe it scales enough then as long as we cache it good.

src/richie/apps/search/filter_definitions/courses.py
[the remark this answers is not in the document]
mbenadda: Ok then, I'll use LocMemCache 👍

tests/apps/search/test_query_courses.py
```diff
@@ -272,18 +120,288 @@ def get_expected_courses(courses_definition, course_run_ids):
...
+        - create relevant objects that will be linked with courses,
         - generate a set of courses randomly associated to our "interesting" course runs,
-        - prepare the Elasticsearch index,
-        - execute the query.
+        - prepare the Elasticsearch index.
         """
+        filter_pages = self.create_filter_pages()
+
+        top_subjects = [
+            # 00010001 througL 00010004
```
sampaccoud, line 134:
```suggestion
            # 00010001 through 00010004
```

tests/apps/search/test_query_courses.py
```diff
@@ -272,18 +120,288 @@ def get_expected_courses(courses_definition, course_run_ids):
...
+        self.persons = [
+            PersonFactory(
+                page_title="Mikhaïl Boulgakov",
+                should_publish=True,
+            )
+        ]
+
+        # NB: Indexed human names do not match the actual titles for the pages.
+        # This makes it simpler to run full-text searches and check their results.
+        courses = [
```
sampaccoud, line 212: I didn't get why this is hardcoded and not generated via the indexer's method. Since we are testing the queries and not the indexer's document formatting it would be more robust and simpler?

tests/apps/search/test_query_courses_edge_cases.py
```diff
@@ -181,39 +226,53 @@ def test_query_courses_rare_facet_force(self, *_):
...
+                {"count": 2, "human_name": "Organization #1", "key": "L-00030002"},
+                {"count": 2, "human_name": "Organization #2", "key": "L-00030003"},
+                {"count": 2, "human_name": "Organization #3", "key": "L-00030004"},
+                {"count": 2, "human_name": "Organization #4", "key": "L-00030005"},
+                {"count": 2, "human_name": "Organization #5", "key": "L-00030006"},
+                {"count": 2, "human_name": "Organization #6", "key": "L-00030007"},
+                {"count": 2, "human_name": "Organization #7", "key": "L-00030008"},
+                {"count": 2, "human_name": "Organization #8", "key": "L-00030009"},
+                {"count": 2, "human_name": "Organization #9", "key": "L-0003000A"},
+                {"count": 1, "human_name": "Organization #10", "key": "L-0003000B"},
```
sampaccoud, line 247: I'm curious what made it change to hexadecimal?

mbenadda: Node paths are always hexadecimal. It's just when we wrote them manually we wrote them in decimal 😅

tests/apps/search/test_query_courses.py
mbenadda: I tried to make the fewest changes I could get away with, and decided not to touch this.

I also think it makes the tests less brittle: if we're testing search queries, we have to hardcode every bit of searchable data, otherwise the tests could have different results between runs.

sampaccoud: ok ok

tests/apps/search/test_query_courses_edge_cases.py
sampaccoud: ah

sampaccoud (approved):
Don't forget the CHANGELOG!

PR #1469, opened 2021-10-04
📝(UPGRADE) add missing instruction from 2.7.x to 2.8.x

sampaccoud:
## Purpose

We forgot to add an instruction about the new `robots.txt` path.

## Proposal

Add instruction:

```python
    path(
        "robots.txt",
        TemplateView.as_view(
            template_name="richie/robots.html", content_type="text/plain"
        ),
    )
```


jbpenrath (approved):
Ah !
AH!

PR #1470, opened 2021-10-04
✨(course) add enrollment count

igobranco:
Add enrollment count to course run and to course page.
Allow its update from openedx course run api.
The course page show the sum of all enrollment count from every course run.
Half the courses on the demo site has enrollment count defined on its runs.

Proposed presentation of the enrolled students.
image


sandbox/settings.py
```diff
@@ -498,6 +498,12 @@ class Base(StyleguideMixin, DRFMixin, RichieCoursesConfigurationMixin, Configura
     PAGINATION_DEFAULT_WINDOW = 2
     PAGINATION_DEFAULT_MARGIN = 1
 
+
+    # Django - core
+    # A boolean that specifies whether to display numbers using a thousand separator.
+    # Visible at least on the enrollment count of the course pages
+    USE_THOUSAND_SEPARATOR = True
```
sampaccoud, line 505: This is normally already taken care of by Django: https://github.com/django/django/blob/main/django/conf/locale/pt/formats.py#L34

src/richie/apps/courses/models/course.py
```diff
@@ -683,6 +701,12 @@ class CourseRun(TranslatableModel):
         choices=lazy(lambda: ALL_LANGUAGES, tuple)(),
         help_text=_("The list of languages in which the course content is available."),
     )
+    enrollment_count = models.PositiveIntegerField(
+        _("enrollment count"),
+        blank=True,
+        null=True,
```
sampaccoud, line 707: Maybe it should be required and default to 0 so we never display a null value?

src/richie/apps/courses/templates/courses/cms/course_detail.html
```diff
@@ -111,6 +111,16 @@ <h1 class="subheader__title" property="name">{% render_model current_page "title
                         {% endif %}
                     </div>
 
+                    <div class="subheader__enrollment_count">
+                        {% block enrollment_count %}
+                            {% with enrollment_count=course.course_runs_enrollment_count_sum %}
+                                {% if enrollment_count is not None and enrollment_count > 0 %}
```
sampaccoud, line 117: see above comment about null values

sampaccoud:
Nicely done @igobranco :astonished: You make my day :hugs:

sandbox/settings.py
igobranco: Yes, you are right. The solution was to use the [intcomma](https://docs.djangoproject.com/en/3.2/ref/contrib/humanize/#intcomma), to only format that specific number. But it requires to add `django.contrib.humanize` to the `INSTALLED_APPS`.

src/richie/apps/courses/models/course.py
igobranco: Ok.
For sure we aren't going to show zero enrollments to the user, so for the users the null or zero is the same, it wan't display any enrollment information.

igobranco:
@sampaccoud now you can merge the PR.

PR #1471, opened 2021-10-05
💄(subheader) use a dedicated variable to set aside column width

jbpenrath:
## Purpose

Previously, we used `$r-course-aside` variable to define both the course content aside column width and  the subheader aside column width. But in some use case, we want to define this variable to null to make `course__content` in full width. But we do not want this behavior for the `subheader` block. So we decided to use a dedicated variable to set the subheader aside column width.

## Proposal

- [x] Use `$r-course-subheader-aside` variable to set subheader aside column width

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unrealeased]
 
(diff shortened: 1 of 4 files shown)


sampaccoud (approved):
Well spotted my dear Watson!

PR #1472, opened 2021-10-05
🐛(courses) fix "pt_effort" string computation when effort is not set

sampaccoud:
## Purpose

While computing the "pt_effort" string for schema.org's [timeRequired field](https://schema.org/timeRequired) on the course, we forgot to handle the case when the effort field is not set.

## Proposal

- [x] Return an empty string if the effort is not set on the course.
- [x] Add a test to secure this case


jbpenrath (approved):
✅ LGTM !

PR #1473, opened 2021-10-05
🔖(chore) bump release to version 2.8.2

sampaccoud:
### Fixed

- Fix schema.org related "pt_effort" string computation when effort is not set
- Use `$r-course-subheader-aside` to define subheader aside column width


jbpenrath (commented):
You have embedded `fa_IR.mo`file.

jbpenrath (approved):
LGTM

PR #1476, opened 2021-10-06
🩹(tests) remove error message when a test fails

igobranco:
Remove error message related with cache when any test fails.
When any test fails locally, any test shows this annoying error. 
This pull request removes that message.

```
-------------------------------------------------------------------------------------------- Captured log call --------------------------------------------------------------------------------------------
ERROR    richie.apps.core.cache:cache.py:163 int() argument must be a string, a bytes-like object or a number, not 'NoneType'
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 ### Fixed
 
 - Prevent logout action to be called twice
+- Remove error message related with cache when any test fails
 
 ## [2.9.1] - 2021-11-03
 
--- a/src/richie/apps/core/cache.py
+++ b/src/richie/apps/core/cache.py
@@ -155,7 +155,7 @@ class LimitBrowserCacheTTLHeaders(MiddlewareMixin):
         Rewrite the "Cache-control" and "Expires headers" in the response
         if needed.
         """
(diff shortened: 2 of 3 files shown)


igobranco:
> Knowing this, do you think we should have a different default for this setting and use a default of 1h for example?

@sampaccoud I think we should need to have a reasonable default value for every configuration. The current default production value is `600` seconds. I think that 10 minutes is long enough to the browser cache.

sampaccoud:
@igobranco Following our discussions, I propose to add this commit on top of your PR: https://github.com/openfun/richie/commit/f423c79f0159c24ab443be67581e7f82c93f6b23

igobranco:
> @igobranco Following our discussions, I propose to add this commit on top of your PR: [f423c79](https://github.com/openfun/richie/commit/f423c79f0159c24ab443be67581e7f82c93f6b23)

@sampaccoud  Ok. 
How do you want to proceed? Do you approve this PR and then merge this https://github.com/openfun/richie/compare/igobranco/fix_test_fail_message to master branch ?

sampaccoud:
@igobranco I would say: cheery pick this commit to add it to your PR?

igobranco:
> @igobranco I would say: cheery pick this commit to add it to your PR?

Done.
@sampaccoud  approve and merge ?

sampaccoud (approved):

PR #1511, opened 2021-11-05
✨(search) update ES indices when a page is moved

mbenadda:
## Purpose

When a page is moved in the MPTT tree, it can change the paths of an unbounded number of other pages linked to the same kind of objects. Right now, paths are only used for categories.

## Proposal

Therefore, when a category is moved in the page tree, we need to re-index all categories, as we do not know which pages may be impacted by the move.

Note: although there still is a `page_moved` signal (https://github.com/django-cms/django-cms/blob/f622008b471e664249b72db6dec528d4d8bd4255/cms/signals/__init__.py#L16), it is not called anywhere as it was intentionally removed (https://github.com/django-cms/django-cms/issues/6356).

The approach we propose relies on an internal tool, but thanks to our unit test, we should be warned if/when it is broken by Django CMS.

### DIFF
--- a/src/richie/apps/search/signals.py
+++ b/src/richie/apps/search/signals.py
@@ -2,12 +2,16 @@
 from django.conf import settings
 from django.core.exceptions import ObjectDoesNotExist
 from django.db import transaction
+from django.dispatch import receiver
 
+from cms import operations
 from cms.models import Title
+from cms.signals import post_obj_operation
 
 from richie.apps.courses.models import Category, Course, Organization, Person
 from richie.apps.search.index_manager import richie_bulk
 from richie.apps.search.indexers import ES_INDICES
+from richie.apps.search.indexers.categories import CategoriesIndexer
 
 
 def apply_es_action_to_course(instance, action, _language):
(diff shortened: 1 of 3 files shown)


src/richie/apps/search/signals.py
```diff
@@ -165,3 +169,19 @@ def on_page_unpublished(sender, instance, language, **kwargs):
...
             lambda: apply_es_action_to_page(instance, action, language)
         )
+
+
+# pylint: disable=unused-argument
+@receiver(post_obj_operation)
+def on_page_moved(sender, **kwargs):
+    """
+    When a page is moved, we may need to re-index all pages linked to objects of
+    the same kind. This applies to all pages linked to categories as they have the
```
sampaccoud, line 179: I misunderstood what you meant by "pages linked to":
```suggestion
    the same kind. This applies to all category pages as they have the
```

src/richie/apps/search/signals.py
```diff
@@ -165,3 +169,19 @@ def on_page_unpublished(sender, instance, language, **kwargs):
...
+    *path* of the page in the ES index.
```
sampaccoud, line 180: It's not clear to me anymore why we need the path for categories and not for organizations or persons?
Do we really still need the path for categories?

mbenadda: We need the paths to display the categories in the frontend with their parent-children hierarchy

sampaccoud: Ok. Yes it's true that it's useless for orgs and persons although theoretically we could have it.
Good thanks for the clarifications. I don't think we can avoid this without making ugly postgresql queries. So let's go.

sampaccoud (approved):

PR #1515, opened 2021-11-08
🐛(frontend) do not call logout twice

jbpenrath:
## Purpose
In DesktopUserMenu, when user clicked on logout button, the action was trigger twice. First through the `onClick` handler then second from the downshift `onSelectedItemChange` handler used for accessibility purpose.

## Proposal

- [x] Trigger actions only through `onSelectedItemChange` handler.


mbenadda (approved):
LGTM

PR #1518, opened 2021-11-09
Style/stuff for joanie

jbpenrath:
## Purpose

In order to prepare integration of Joanie, we need to create a reusable modal component and update some styles mixin.

## Proposal

- [x] Create a reusable modal component
- [x] Update styles (add timing-function variable, button outline variant, shadowed-box mixin)


mbenadda (approved):
LTGM 👍

PR #1519, opened 2021-11-09
♻️(search) fix autocomplete filters in Search

mbenadda:
## Purpose

As a result of an oversight in our latest Search refactor, we broke the addition of autocompleted filters to the current course search params.

This was due to not having the correct "kind" on categories, them having "categories" instead of eg. "subjects" or "levels". There was a fallback relying on MPTT paths as IDs, which we recently removed.

## Proposal

We updated autocompleted categories to have the correct kind and removed the now unnecessary fallback.

NB: no changelog as the commits that broke this are not yet part of a release.


jbpenrath (approved):
LGTM ✅

PR #1526, opened 2021-11-23
🔧(frontend) use new jsx transform

jbpenrath:
## Purpose

Since React 17, a new JSX transformer is available.
It does not require anymore to load React package from jsx files and can in some case optimize bundle size.

## Proposal

- [x] Update babel config to use new JSX transform
- [x] Run `npx react-codemod update-react-imports` to update codebase


src/frontend/js/components/CourseRunEnrollment/index.tsx
```diff
@@ -1,4 +1,5 @@
-import React, { useCallback, useEffect, useReducer } from 'react';
+import { useCallback, useEffect, useReducer } from 'react';
+import * as React from 'react';
```
mbenadda, line 2: This seems worse than before: 2 lines and a wildcard import. Can we just leave it like it was?

jbpenrath: Oh yes ! I trusted `react-codemod`...

mbenadda: IMO we can leave `import React from 'react'` wherever it inserted `import * as React`. What do you think?

jbpenrath: Yes I also find the syntax better. But I took the opportunity to import only stuff used when there is only one thing used from React.

mbenadda: No more React Football Club!

mbenadda (approved):

PR #1527, opened 2021-11-24
🐛(opengraph) fix og:image

### LINKED ISSUE #1510: og:image is not set propertly
## Bug Report

**Expected behavior/code**
For SEO purpose, we add open graph tags into CMS pages. These tags should be properly configured.

**Actual Behavior**
On CMS Page, `og:image` url is malformed.

**Steps to Reproduce**
1. Go to a CMS Page
2. Use [facebook debug tool](https://developers.facebook.com/tools/debug/) to check link preview for the CMS Page
3.  Note that `og:image` tag is not valid

**Environment**
- Richie version: 2.8.2
- Platform: Mac OS 12.0.1 - Firefox 93.0

**Possible Solution**
- Fix the `og:image` attribute

- https://developers.facebook.com/tools/debug/?q=https%3A%2F%2Fwww.fun-mooc.fr%2Ffr%2Fcours%2Fmooc-elles-font-lart%2F
- https://developers.facebook.com/tools/debug/?q=https%3A%2F%2Fwww.fun-mooc.fr%2Ffr%2Factualites%2Fdeveloppement-des-territoires-comment-agir-pour-construire-laven%2F

igobranco:
Fix og:image when using S3Boto3Storage.

Fixes #1510

### DIFF
--- a/tests/apps/courses/test_templates_course_detail_opengraph.py
+++ b/tests/apps/courses/test_templates_course_detail_opengraph.py
@@ -2,7 +2,9 @@
 Test suite for the Open Graph for course pages
 """
 import re
+from unittest import mock
 
+from django.core.files.storage import FileSystemStorage
 from django.test.utils import override_settings
 
 from cms.api import add_plugin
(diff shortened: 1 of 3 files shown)


tests/apps/courses/test_templates_course_detail_opengraph.py
```diff
@@ -185,3 +187,35 @@ def test_open_graph_course_description_no_introduction(self):
         response = self.client.get(url)
 
         self.assertNotContains(response, "og:description")
+
+    # pylint: disable=no-self-use
+    def mock_storage_url(self, name):
```
jbpenrath, line 192: I'll move this function above test methods in order to split utils and tests.

tests/apps/courses/test_templates_course_detail_opengraph.py
```diff
@@ -185,3 +187,35 @@ def test_open_graph_course_description_no_introduction(self):
...
+        """
+        Method that mocks the `url` method of the django file storage.
+        """
+        return f"https://test-bucket.s3.amazonaws.com/{name:s}".format(name=name)
```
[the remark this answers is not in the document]
igobranco: Done.

tests/apps/courses/test_templates_course_detail_opengraph.py
igobranco: I think that the mock is too specific for the test case, so I refactor the function to be lambda inline. 

```python
@mock.patch.object(
        FileSystemStorage,
        "url",
        lambda storage_instance, name: f"https://test-bucket.s3.amazonaws.com/{name:s}",
    )
    def test_open_graph_course_meta_og_image_media_storage_s3(self):
```

@jbpenrath  are you ok with this solution ?

tests/apps/courses/test_templates_course_detail_opengraph.py
```diff
@@ -185,3 +187,32 @@ def test_open_graph_course_description_no_introduction(self):
         response = self.client.get(url)
 
         self.assertNotContains(response, "og:description")
+
+    @mock.patch.object(
+        FileSystemStorage,
+        "url",
+        lambda storage_instance, name: f"https://test-bucket.s3.amazonaws.com/{name:s}",
+    )
+    def test_open_graph_course_meta_og_image_media_storage_s3(self):
```
sampaccoud, line 196: I would say you are not testing the problem you had which was that it was breaking when MEDIA_URL was not set?

sampaccoud: If you want to be sure that it works with Boto3, the only way is to install it in dev dependencies and test with it and forcing the MEDIA_URL setting to None or something like this?

igobranco: @sampaccoud  the problem is that s3boto3 django-storages implementation don't use the `MEDIA_URL` setting. 
The `MEDIA_URL` setting isn't used by [s3boto3.py](https://github.com/jschneier/django-storages/blob/master/storages/backends/s3boto3.py), that is why our tests didn't detect the problem.

To make the richie implementation and its tests more generic, to run on any kind of storage, we shouldn't use or run tests for a specific storage solution, but instead use the `django.core.files.storage.Storage` base class /interface. That is why I mocked the `url` method. We could replace `s3.amazonaws.com` on the tests with any think of domain.

sampaccoud: ok... then let's derive it in 3 tests for each kind of domain?

igobranco: Yes.
* One for full URL
* another for an URL that starts with a double slash
* and a last one for an URL start with a relative URL.

sampaccoud (approved):

tests/apps/courses/test_templates_course_detail_opengraph.py
sampaccoud: :+1:

jbpenrath (approved):

PR #1528, opened 2021-11-29
✨(search) Scroll search filters more options

mbenadda:
## Purpose

It is often the case that facet categories in course search have a lot of possible filter values.

We initially built the `<SearchFilterGroupModal />` to enable users to search through those values. However, we did not set up pagination.

As it turns out, users want to be able to move down the list to get to the value they want.

## Proposal

Implement infinite scrolling in the modal using react-query and IntersectionObserver to make it as effortless as possible for users.

### DIFF
--- a/src/frontend/js/components/Modal/_styles.scss
+++ b/src/frontend/js/components/Modal/_styles.scss
@@ -1,3 +1,8 @@
+// Make sure random elements with any z-index don't show on top of the modals
+.ReactModal__Overlay {
+  z-index: 1;
+}
+
 .modal {
   background-color: r-theme-val(modal, base-background);
   border-radius: $border-radius;
--- a/src/frontend/js/components/Modal/_styles.scss
+++ b/src/frontend/js/components/Modal/_styles.scss
@@ -1,8 +1,3 @@
-// Make sure random elements with any z-index don't show on top of the modals
-.ReactModal__Overlay {
-  z-index: 1;
-}
-
 .modal {
   background-color: r-theme-val(modal, base-background);
   border-radius: $border-radius;
--- a/src/frontend/js/components/SearchFilterGroupModal/_styles.scss
+++ b/src/frontend/js/components/SearchFilterGroupModal/_styles.scss
@@ -5,6 +5,12 @@
   justify-content: space-between;
   max-width: 30rem;
 
+  &__close {
+    position: absolute;
+    right: 0;
+    top: 0.25rem;
+  }
+
   &__form {
     display: flex;
     flex-direction: column;
@@ -14,7 +20,6 @@
       display: block;
       margin: 0;
       padding: 1rem;
-      text-align: center;
       font-size: 1.1rem;
       font-weight: $font-weight-boldest;
     }
@@ -50,7 +55,7 @@
     }
   }
 
-  &__close {
+  &__more-results {
     padding: 0.5rem 1rem;
     flex-shrink: 0;
     flex-grow: 0;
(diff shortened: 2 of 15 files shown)


sampaccoud:
@mbenadda do you think it's possible and a good idea to set a focus on the search field when the modal opens?

sampaccoud:
I generated a site with 200 organizations.
When I search for an organization in the "more choices" modal box, I type a few letters that I see in an organization, then results all contain the few letters. But if I scroll down and trigger a load for more results, the new results do not contain those 3 letters :thinking:

igobranco:
> @mbenadda do you think it's possible and a good idea to set a focus on the search field when the modal opens?

@sampaccoud  only on desktop mode, if on mobile it opens the keyboard that occupies half the screen.

src/frontend/js/components/SearchFilterGroupModal/index.tsx
```diff
@@ -54,118 +63,191 @@ const messages = defineMessages({
...
+  const loadMoreButtonRef = useRef<HTMLButtonElement>(null);
+  useIntersectionObserver({
+    target: loadMoreButtonRef,
+    onIntersect: fetchNextPage,
+    enabled: !!hasNextPage,
+  });
+
+  return (
+    <Fragment>
+      <button className="search-filter-group-modal__close" onClick={() => setModalIsOpen(false)}>
```
jbpenrath, line 144: modal has an element `closeButton`.

So you could add class `.modal__closeButton`.

src/frontend/js/components/SearchFilterGroupModal/index.tsx
```diff
@@ -54,118 +63,191 @@ const messages = defineMessages({
...
+        <span className="offscreen">
+          <FormattedMessage {...messages.closeButton} />
+        </span>
+        <svg aria-hidden={true} role="img" className="icon search-filter-group-modal__close__icon">
```
jbpenrath, line 148: If you use `.modal_closeButton`, you have to use `.modal__closeButton__icon` class for the icon.
```suggestion
        <svg aria-hidden={true} role="img" className="modal__closeButton__icon">

```

src/frontend/js/components/SearchFilterGroupModal/_styles.scss
```diff
@@ -5,6 +5,23 @@
...
+    background: transparent;
+    border: none;
+    padding: 0;
+
+    &__icon {
+      width: 1rem;
+      height: 1rem;
+      fill: r-theme-val(search-filters-group-modal, base-color);
+    }
+  }
```
jbpenrath, line 23: If you use `.modal_closeButton` this code could be simplified :

```suggestion
 &__close {
    position: absolute;
    top: 0;
    right: 0;
  }
```

src/frontend/js/components/Modal/_styles.scss
```diff
@@ -1,3 +1,8 @@
+// Make sure random elements with any z-index don't show on top of the modals
+.ReactModal__Overlay {
+  z-index: 1;
+}
+
```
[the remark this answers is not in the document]
mbenadda: Yes, but I think it's more of a rebasing clash: I was already working on this before you merged the modal refactoring.

mbenadda:
@sampaccoud I think it's just an artifact of search happening in different languages all at once, as scrolling pagination does not affect search results at all (query here: https://github.com/openfun/richie/blob/master/src/richie/apps/search/forms.py#L329). And I checked "unexpected" results locally and found they contained the given string in a different language.

jbpenrath (commented):
There is an issue on Brave. The IntersectionObserver callback is never triggered due to the threshold value.

src/frontend/js/utils/useIntersectionObserver.tsx
```diff
@@ -0,0 +1,48 @@
...
+  threshold?: number;
+  rootMargin?: string;
+  enabled?: boolean;
+}
+
+export const useIntersectionObserver = ({
+  root,
+  target,
+  onIntersect,
+  threshold = 1.0,
```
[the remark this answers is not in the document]
mbenadda: 0.99 also works on Firefox, Chrome and Safari so I think we can just use that value and move on.

jbpenrath (approved):
LGTM ✅

PR #1533, opened 2021-12-03
⚡️(templates) remove image transparency when not necessary so that thumbnails are converted to jpeg

sampaccoud:
## Purpose

Easy thumbnails converts images to jpg when there is no transparency, but keeps png otherwise. We can't prevent users from uploading images with transparency in places where it is not desired but we can remove
it when generating thumbnails so that the image can be converted to jpg and thus be much smaller.

## Proposal

Force the `replace_alpha` in all thumbnails template tags to get rid of transparency and make sure images are converted to jpg and thus compressed to a much smaller size.

### DIFF
--- a/src/richie/apps/courses/templates/courses/cms/fragment_course_glimpse.html
+++ b/src/richie/apps/courses/templates/courses/cms/fragment_course_glimpse.html
@@ -8,11 +8,11 @@
             <p class="course-{{ course_variant }}--empty">{% trans "Cover" %}</p>
         {% endget_page_plugins %}
         {% blockplugin cover_plugins.0 %}
-            <img src="{% thumbnail instance.picture 300x170 crop upscale subject_location=instance.picture.subject_location %}"
+            <img src="{% thumbnail instance.picture 300x170 replace_alpha='#FFFFFF' crop upscale subject_location=instance.picture.subject_location %}"
                 srcset="
-                    {% thumbnail instance.picture 800x457 crop upscale subject_location=instance.picture.subject_location %} 300w
-                    {% if instance.picture.width >= 1600 %},{% thumbnail instance.picture 1600x914 crop upscale subject_location=instance.picture.subject_location %} 600w{% endif %}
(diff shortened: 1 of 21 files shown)


jbpenrath (commented):
I succeeded to reproduce the misbehavior of organization logo within `fragment_course_glimpse.html`.

If the organization logo is a png, this image will be a png even with `replace_alpha` set.
But in `course_detail` and `organization_detail`, the logo is a jpg...
But I noticed that only the first `srcset` source is a png other sources and `src` are jpg 🤔 

Capture d’écran 2021-12-08 à 09 48 20

sampaccoud:
> But I noticed that only the first `srcset` source is a png other sources and `src` are jpg thinking

And the only difference between the `thumbnail` tag between the `src` and `srcset` is the size of the image right? :exploding_head:

src/richie/apps/courses/templates/courses/cms/fragment_course_glimpse.html
```diff
@@ -37,12 +37,12 @@
                     {% endget_page_plugins %}
                     {% blockplugin plugins.0 %}
                         <div class="course-glimpse__organization-logo">
-                            <img src="{% thumbnail instance.picture 200x113 upscale subject_location=instance.picture.subject_location %}"
+                            <img src="{% thumbnail instance.picture 200x113 replace_alpha='#FFFFFF' upscale subject_location=instance.picture.subject_location %}"
                                 srcset="
                                     {% thumbnail instance.picture 200x113 upscale subject_location=instance.picture.subject_location %} 200w
```
jbpenrath, line 42: Missed one
```suggestion
                                    {% thumbnail instance.picture 200x113 replace_alpha='#FFFFFF' upscale subject_location=instance.picture.subject_location %} 200w
```

sampaccoud: Good catch :muscle:!
I corrected it so I hope now everything works :pray:

jbpenrath (approved):
All seems good right now ! 🎉
You just have to fix a test.

PR #1539, opened 2021-12-07
🚸(courses) include new pages in navigation by default for breadcrumbs

sampaccoud:
## Purpose

When creating a new page from the DjangoCMS page wizard, we did not set the `in_navigation` flag to `True` by default. This resulted in the new page not being present in the breadcrumb. Users had to systematically
tick the flag to fix that.

## Proposal

- [x] add tests to confirm the issue
- [x] set the `in_navigation` flag to `True` when creating a new page via the DjangoCMS wizard

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Changed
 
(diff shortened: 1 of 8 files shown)


jbpenrath (approved):
✅ LGTM !

PR #1545, opened 2021-12-15
🐛(search) fix negative scores causing errors with far future courses

sampaccoud:
## Purpose

The score computation we made to sort courses within a "search bucket" could lead to a negative score. This causes a 500 error in Elasticsearch and should not happen. When a course has dates far in the future, taking
into account the relative dates is not really relevant anyway.

## Proposal

- [x] Add a test that surfaces the problem
- [x] Make sure scores don't get negative


mbenadda (approved):
LGTM

PR #1555, opened 2021-12-28
🐛(frontend) LMS authentication is optional

mbenadda:
## Purpose

As we deployed the latest version of Richie, all frontend code failed to run because authentication via an LMS connection was assumed to be present.

## Proposal

However, LMS authentication is optional in Richie, and some deployed sites do not use it and break if an authentication backend is not provided.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 - Refactor scss color variables to be more easily customizable.
 
+### Fixed
+
+- Make LMS authentication optional.
+
 ## [2.10.0] - 2021-12-27
 
 ### Added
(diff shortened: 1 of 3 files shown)


sampaccoud:
@mbenadda Can you point to the code that broke it for my reference?

mbenadda:
Right there: https://github.com/openfun/richie/blob/master/src/frontend/js/data/useSession/index.tsx#L49

The `!` in `AuthenticationApi!.me` is wrong as it tells the compiler the authentication is available when it really can be null. We have to add a check (what I did in this PR) before we can make that assertion

sampaccoud:
Oh ok that's when we added ReactQuery! Is it possible to surface the bug with a test first?

sampaccoud (approved):

PR #1561, opened 2022-01-07
🐛(lti_consumer)  Add missing language parameter in LTI requests

sampaccoud:
## Purpose

The LTI consumer was not transmitting the language when making an LTI launch request. As a result, marsha was serving the english language on our french sites.

## Proposal

Add missing language parameter in LTI requests and set it with the current language.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unrealeased]
 
+### Fixed
+
+- Add missing language parameter in LTI requests issued by LTI consumer plugin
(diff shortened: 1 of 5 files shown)


kernicPanel (commented):
Shouldn't we have a test with another language than the default one ?

jbpenrath (commented):
Agree with @kernicPanel, maybe write a test to check that cache is language sensitive.

sampaccoud:
> Shouldn't we have a test with another language than the default one ?

Good idea! I'll do it later. Gotta move.

jbpenrath (approved):
LGTM ✅

PR #1575, opened 2022-01-26
🏷️(helix) add types for @helpscout/helix package

jbpenrath:
## Purpose

We are using `@helpscout/helix` to create factories for our frontend tests. Currently this package has no type so we have no guarantee that our factories stick to real objects implementation. As a little challenge to level up in typescript and in order to improve our factories frontend, I decided to try to type this package.

If this type is reliant from our side, I'm thinking about submit a PR to add these types within definitelyTypes.

## Proposal

- [x] Type `@helpscout/helix` package

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Added
 
+- Type @helpscout/helix package
 - Add catalog visibility field to course run so it's possible to hide a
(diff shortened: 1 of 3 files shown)


lunika (commented):
dog-golf

I'm a little bit like this in front of this PR. I have no idea what is helix internally and still after reading the code I don't see problem with this type.

jbpenrath:
Closed by #1965

PR #1576, opened 2022-01-27
📌(backend) remove use of forked djangocms-admin-style

jbpenrath:
## Purpose
[djangocms-admin-style 3.0.0](https://pypi.org/project/djangocms-admin-style/) has been released and includes fixes we need. So we are now able to remove the use of our forked version of this
package.

## Proposal

- [x] Remove the use of our forked version of djangocms-admin-style


sampaccoud (approved):
:muscle:

PR #1580, opened 2022-02-02
🐛(courses) hide unpublished objects from list views

sampaccoud:
## Purpose

On the category, program and person list views, unpublished objects are listed on the public site.

## Proposal

The template code was a bit too naïve. Exclude unpublished objects (as it was already done for blog posts and organizations) and add tests to secure it.

PR #1582, opened 2022-02-03
♿️(pagination) improve a11y labels when using multiple paginations on same page

### LINKED ISSUE #1581: Pagination a11y labels are ambigous when there are multiple paginations on the same page
## Bug Report

:wave: Starting my work on improving general accessibility on richie :)

**Expected behavior/code**

With a screen reader, when navigating on a page that has multiple lists that can be paginated, for example, the courses category details page, I must be able to differentiate each pagination _on its own_.

**Actual Behavior**

I can't differentiate each pagination easily because they all have the same label: "Pagination".

**Steps to Reproduce**

1. Go to a category detail page
2. With a screen reader, use a keyboard shortcut to jump between each region (or check the code: "Pagination" is the label for every pagination component)
3. "Pagination" is announced multiple times

**Possible Solution**

Make sure every pagination used has a specific label by changing the `aria-labelledby` attribute of each pagination component used on the same page (PR coming).

Leimi:
## Purpose

Fix #1581

## Proposal

The idea here is to easily allow custom labels when using multiple pagination components on the same page. We do this by using a new pagination-specific django tag, currently named `richie_pagination`, **but I guess the naming is not that great**.

Allowing custom labels without this new tag was cumbersome because, the dj-pagination `{% paginate %}` tag doesn't accept any custom context (like an `{% include %}` tag). So it would require to create a specific template file for each label change. I figured having a new tag was better?

Because I'm still not sure how it is usually dealt with in this project, **translation strings are missing** in this PR.

### DIFF
--- a/src/richie/apps/core/templates/richie/pagination.html
+++ b/src/richie/apps/core/templates/richie/pagination.html
@@ -1,56 +1,12 @@
-{% load i18n %}{% spaceless %}
+{% comment %}
+this is meant to be included in a template after using the "autopaginate" tag
+{% endcomment %}
+{% load i18n pagination_tags %}
 
-{% if is_paginated %}
-<div class="pagination">
-  <nav aria-label="{% trans "Pagination" %}">
-    <ul class="pagination__list">
-    {% for page in pages %}
-      {% if page %}
-        {% ifequal page page_obj.number %}
-          <li class="pagination__item pagination__item--current">
-            <span class="pagination__page-number">
-              <span class="offscreen">
-              {% if forloop.last %}
-                {% blocktrans %}Currently reading last page {{ page }}{% endblocktrans %}
-              {% else %}
-                {% blocktrans %}Currently reading page {{ page }}{% endblocktrans %}
-              {% endif %}
-              </span>
-              <span aria-hidden="true">{% if page == 1 %}Page {{ page }}{% else %}{{ page }}{% endif %}</span>
-            </span>
-          </li>
-        {% else %}
-          {% if page == 1 %}
-            <li class="pagination__item">
-              <a href="{{ request.path }}{% if getvars %}?{{ getvars|slice:"1:" }}{% endif %}#page{{ page_suffix }}" class="pagination__page-number">
-                {% blocktrans %}Page {{ page }}{% endblocktrans %}
-              </a>
-            </li>
-          {% else %}
-            <li class="pagination__item">
-              <a href="?page{{ page_suffix }}={{ page }}{{ getvars }}#page{{ page_suffix }}" class="pagination__page-number">
-                <span class="offscreen">
-                  {% if page == page_obj.previous_page_number %}
-                    {% blocktrans %}Previous page {{ page }}{% endblocktrans %}
-                  {% elif forloop.last %}
-                    {% blocktrans %}Last page {{ page }}{% endblocktrans %}
-                  {% elif page == page_obj.next_page_number %}
-                    {% blocktrans %}Next page {{ page }}{% endblocktrans %}
-                  {% else %}
-                    {% blocktrans %}Page {{ page }}{% endblocktrans %}
-                  {% endif %}
-                </span>
-                <span aria-hidden="true">{{ page }}</span>
-              </a>
-            </li>
-          {% endif %}
-        {% endifequal %}
-      {% else %}
-        <li class="pagination__item pagination__item--placeholder">...</li>
-      {% endif %}
-    {% endfor %}
-    </ul>
-  </nav>
-</div>
+{% if paginator.count > paginator.per_page %}
+  <div class="pagination">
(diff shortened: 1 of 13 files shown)


src/richie/apps/courses/templatetags/extra_tags.py
```diff
@@ -205,6 +208,61 @@ def render_tag(self, context, plugin, nodelist):
...
+
+    def render(self, context):
+        template_context = paginate(context)
+        for name, value in self.additional_context.items():
+            template_context[name] = value.resolve(context)
+        return template.loader.render_to_string(self.template, template_context)
+
+
+@register.tag()
+def richie_paginate(parser, token):
```
Leimi, line 231: Maybe registering this tag as `paginate` to replace the dj-pagination one would be better (if possible)? This would prevent future usages of the dj-pagination tag that kinda becomes obsolete with this PR.

Leimi:
I have a failing test because of the addition of the custom template tag but I don't understand why…

jbpenrath:
> I have a failing test because of the addition of the custom template tag but I don't understand why…

You are using `richie_paginate` tag which is part of `extra_tags`. So you have to load `extra_tags` in templates where you are using `richie_paginate`.

Leimi:
> > I have a failing test because of the addition of the custom template tag but I don't understand why…
> 
> You are using `richie_paginate` tag which is part of `extra_tags`. So you have to load `extra_tags` in templates where you are using `richie_paginate`.

Oh OK I missed that you needed to specifically load those. Thanks, it's fixed.

src/richie/apps/core/templates/richie/pagination.html
```diff
@@ -2,9 +2,7 @@
 
 {% if is_paginated %}
 <div class="pagination">
-  <div id="pagination-label" class="offscreen">{% trans "Pagination" %}</div>
```
sampaccoud, line 5: I'm thinking maybe it would be simpler and easier to understand if we take the wrapper markup out of this template and keep it in the calling template where it can be customized?

sampaccoud: @Leimi And you can still share the code by introducing a new template. In the page where you want the pagination you put:
```
{% include richie/pagination_wrapper_new_template with label="Custom pagination" %}
```

and in the wrapper template:

sampaccoud: :sweat_smile:

Leimi: Well… Yeah, what was I thinking? This solution is way simpler haha. Will update, thanks.

sampaccoud: @Leimi be careful I just merged this PR which adds new pagination and should also benefit from your refactoring : https://github.com/openfun/richie/pull/1580

sampaccoud:
@sandroscosta it would be nice to have your review on all a11y Pull Requests.
@Leimi would be nice to add Sandro as a reviewer on a11y PRs.

sandroscosta (approved):

Leimi:
Should be all good now @sampaccoud!

jbpenrath (approved):
LGTM

src/richie/apps/courses/templates/courses/cms/category_detail.html
```diff
@@ -152,7 +152,7 @@ <h2 class="category-detail__title">{% trans "Related organizations" %}</h2>
                                 {% include "courses/cms/fragment_organization_glimpse.html" %}
                             {% endfor %}
                             {% if paginator.num_pages > 1 %}
-                                {% paginate using "richie/pagination.html" %}
+                                {% include "richie/pagination.html" with label="Related organizations pagination" %}
```
sampaccoud, line 155: If you do like this, it will not get translated. This should make the job:
```suggestion
                                {% trans "Custom pagination" as pagination_label %}
                                {% include "richie/pagination.html" with label=pagination_label %}
```

Leimi: Oh yeah sorry. Would it be okay to use the `_()` syntax to more concise?

```diff
- {% trans "Custom pagination" as pagination_label %}
- {% include "richie/pagination.html" with label=pagination_label %}
+ {% include "richie/pagination.html" with label=_("Custom pagination") %}
```

sampaccoud: If it works OK! Never seen it though

Leimi: Tested it locally, it does work. Saw this [in the django docs](https://docs.djangoproject.com/fr/4.0/topics/i18n/translation/#string-literals-passed-to-tags-and-filters) :)

sampaccoud: Cool 🤩

PR #1583, opened 2022-02-03
✨(frontend) Create a Banner React component

jbpenrath:
## Purpose

Previously we created a styles for a banner block. In the react app context, it could be interesting to have a reusable `Banner` component base on these styles to easily display messages.

## Proposal

- [x] Create a reusable react Banner component

### DIFF
--- a/src/frontend/js/components/Banner/index.tsx
+++ b/src/frontend/js/components/Banner/index.tsx
@@ -1,0 +1,29 @@
+import { useMemo } from 'react';
+
+export enum BannerType {
+  ERROR = 'error',
+  INFO = 'info',
+  SUCCESS = 'success',
+  WARNING = 'warning',
+}
+
+interface BannerProps {
+  message: string;
+  type?: BannerType;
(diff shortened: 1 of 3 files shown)


src/frontend/js/components/Banner/index.tsx
```diff
@@ -0,0 +1,29 @@
...
+}
+
+interface BannerProps {
+  message: string;
+  type: BannerType;
+  rounded?: boolean;
+}
+
+const Banner = ({ message, rounded, type }: BannerProps) => {
+  const className = useMemo(
```
sampaccoud, line 17: Maybe have a default of info and not rounded?

sampaccoud (approved):

sampaccoud (approved):

src/frontend/js/components/Banner/index.tsx
jbpenrath: Yes good idea to use `info` as default type prop.

PR #1586, opened 2022-02-04
Joanie/setup

jbpenrath:
## Purpose

First, Joanie is an optional module of Richie. It must be activable through settings.

Then as a first step, this pull request adds api interface and hooks to retrieve/create/update/delete resources from Joanie (Course, Orders, Addresses, Credit cards).

## Proposal

- [x] Add Joanie object within settings to configure it
- [x] Add joanie frontend api interface and hooks 
- [x] Add JoanieSessionProvider to request api with jwt


docs/joanie-connection.md
```diff
@@ -0,0 +1,32 @@
+# Joanie Connection
+
+## Settings
+
+All settings related to Joanie have to be declared in the `JOANIE` dictionary
+within `settings.py`.
+To enable Joanie, the minimal configuration requires one property:
+
+- `BASE_URL` : the endpoint which Joanie is accessible
```
sampaccoud, line 9:
```suggestion
- `BASE_URL` : the endpoint at which Joanie is accessible
```

jbpenrath:
@lunika can you please review the front end part of this PR ?

docs/joanie-connection.md
```diff
@@ -0,0 +1,32 @@
...
+
+Add to your `settings.py`:
+
+```python
+...
+JOANIE = {
+  "BASE_URL": values.Value(environ_name="JOANIE_BASE_URL", environ_prefix=None)
```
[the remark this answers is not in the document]
jbpenrath: As discussed, even if in a first time course run enrollment will be still manage from Richie, in the future when Joanie will be enabled we should prevent the use of a lms backend so these two settings do not aims to live together.

src/frontend/js/utils/api/joanie.ts
```diff
@@ -0,0 +1,273 @@
...
+  }
+}
+
+/**
+ * Build Joanie API Routes interface.
+ */
+const getRoutes = (joanieEndpoint: string) => ({
+  user: {
+    creditCards: {
+      get: joanieEndpoint.concat('/api/credit-cards/:id/'),
```
lunika, line 95: Why not using string interpolation here ? `${joanieEndpoint}/api/credit-cards/:id/` ?

src/frontend/js/data/SessionProvider/BaseSessionProvider.tsx
```diff
@@ -0,0 +1,94 @@
...
+  const { data: user } = useQuery<Nullable<User>>('user', AuthenticationApi!.me, {
+    refetchOnWindowFocus: true,
+    staleTime: REACT_QUERY_SETTINGS.staleTimes.session,
+  });
+  const previousUserState = usePrevious(user);
+
+  const queryClient = useQueryClient();
+
+  const login = useCallback(() => {
+    if (!AuthenticationApi) return handle(new Error('No AuthenticationApi configured!'));
```
lunika, line 38: Line 29 you are sure `AuthenticationApi` is defined, probably you don't need this check

src/frontend/js/data/SessionProvider/JoanieSessionProvider.tsx
```diff
@@ -0,0 +1,140 @@
...
+  });
+  const previousUserState = usePrevious(user);
+
+  const queryClient = useQueryClient();
+  const addresses = useAddresses();
+  const creditCards = useCreditCards();
+  const orders = useOrders();
+
+  const login = useCallback(() => {
+    if (!AuthenticationApi) return handle(new Error('No AuthenticationApi configured!'));
```
lunika, line 60: Same as `BaseSessionProvider`, you probably don't need to do this check anymore

src/frontend/js/data/SessionProvider/index.spec.tsx
```diff
@@ -0,0 +1,193 @@
...
+    }));
+    jest.doMock('./BaseSessionProvider', () => ({
+      __esModule: true,
+      default: () => 'BaseSessionProvider',
+    }));
+    jest.doMock('./JoanieSessionProvider', () => ({
+      __esModule: true,
+      default: () => 'JoanieSessionProvider',
+    }));
+    const { SessionProvider: Provider } = await import('.');
```
lunika, line 66: As this is node code, you can also use `require` instead of `await import` here.

src/frontend/js/data/SessionProvider/index.spec.tsx
```diff
@@ -0,0 +1,193 @@
...
+      const { result } = renderHook(useSession, { wrapper });
+
+      await act(async () => {
+        userDeferred.resolve({ username });
+        jest.runAllTimers();
+      });
+
+      const { user, destroy } = result.current;
+      expect(user).toStrictEqual({ username });
+      expect(sessionStorage.getItem(REACT_QUERY_SETTINGS.cacheStorage.key) || '').toContain(
```
lunika, line 152: Dose using `|| ''` provdes better information when it fails ?

src/frontend/js/data/SessionProvider/BaseSessionProvider.tsx
jbpenrath: Yes good catch !

src/frontend/js/data/SessionProvider/index.spec.tsx
jbpenrath: Nope I could remove this fallback.

In fact when we use fallback we get this error message when the test fails : 
```
Error: expect(received).toMatch(expected)

Expected pattern: /"data":null,.*"queryKey":"user"/
Received string:  ""
```

Otherwise we get :
```
Error: expect(received).toMatch(expected)

Matcher error: received value must be a string

Received has value: null
```

src/frontend/js/utils/api/joanie.ts
jbpenrath: Because this day, I remembered the `concat` method 😅. I'll use a template string.

Issue #1594, opened 2022-02-11
Can't run tests on MySQL DB


igobranco:
## Bug Report

The CI is running fine the tests on MySQL DB. 

But there isn't a way to run them locally. When I run this command:
```bash
DB_HOST=mysql make test-back
```

It sais that the test_richie DB don't exits.

igobranco:
Good!

> Another way would be to grant database creation rights to the `fun` user but I'm not sure it's worth the hassle?
I like your solution! This is for development, so an easier and simpler solution would be better.

Can you send a PR with this?

PR #1595, opened 2022-02-11
👷(tests) fix tests for mysql

sampaccoud:
### Purpose

When Django runs tests, it creates a temporary database with the name "test_{database_name}". The problem is that our `fun` user does not have the right to create a new database.

### Proposal

We can safely run our tests with the mysql `root` user.

Fixes https://github.com/openfun/richie/issues/1594

### DIFF
--- a/env.d/development/mysql
+++ b/env.d/development/mysql
@@ -2,13 +2,11 @@
 MYSQL_ROOT_PASSWORD=
 MYSQL_ALLOW_EMPTY_PASSWORD=yes
 MYSQL_DATABASE=richie
-MYSQL_USER=fun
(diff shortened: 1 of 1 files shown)


igobranco (commented):
@sampaccoud  you forgot to add a line on the changelog.
But in my machine the tests are running file with:
```bash
DB_HOST=mysql make test-back 
```

sampaccoud:
> @sampaccoud you forgot to add a line on the changelog. 

I don't think we need an entry in the changelo as this is only scaffholding... not part of the richie product that we ship...

jmaupetit (approved):

igobranco (approved):

PR #1599, opened 2022-02-16
🚸(frontend) wrap React widgets in ErrorBoundary

jbpenrath:
## Purpose

In order to prevent that all component rendering crash when one of them raises
an error we wrap each component within an [ErrorBoundary](https://reactjs.org/docs/error-boundaries.html)

## Proposal

- [x] Create an `ErrorBoundary` component
- [x] Wrap widgets in this new component


src/frontend/js/components/Root/index.tsx
```diff
@@ -78,10 +80,15 @@ export const Root = ({ richieReactSpots }: RootProps) => {
 
       // Add context to props if they do not already include it
       if (!props.context) {
-        props.context = window.__richie_frontend_context__.context;
+        props.context = context;
```
sampaccoud, line 83: If I understand correctly, this should this be in another commit?

sampaccoud (approved):

PR #1600, opened 2022-02-16
♻️(frontend) use createWebStoragePersistor

jbpenrath:
## Purpose

Previously we created a sessionStoragePersistor to store react query state within sessionStorage. From now, react-query offers a `createWebStoragePersistor` which do the same thing, so we can use this implementation and remove your custom one.

## Proposal

- [x] Replace our `sessionStoragePersistor` by the new `createWebStoragePersistor`


sampaccoud (approved):
negative PRs are the best :star_struck:

PR #1590, opened 2022-02-08
🚸(course) add admin autocomplete feature on Page select in Person plugin

sveetch:
## Purpose

Some sites can have a huge list of objects where it is hard to search for a specific
object. The page selector from Person plugin is one of this case.

## Proposal

We implement usage of an autocomplete feature with Select2.js by the way of new
dependancy 'django-autocomplete-light' enabled and configured on the 'page' field
from Person plugin form.

PR #1601, opened 2022-02-17
🚸(courses) add admin autocomplete for remaining page related plugins

sampaccoud:
## Purpose

In PR https://github.com/openfun/richie/pull/1590, @sveetch proposed to activate an autocomplete select field for page related plugins in the frontend admin view. This feature was urgently needed for the person object as we have a lot of instances of this type of page in https://www.fun-mooc.fr, but it is also very interesting for all other types of page related plugins: blog post, category, course, organization, organization by category and program.

## Proposal

This is a generalization of @sveetch's  work to all other page related plugins.

We generate tests programmatically, for each type of page related object, in order not to repeat the exact same code 6 times.

### DIFF
--- a/tests/apps/courses/test_cms_views.py
+++ b/tests/apps/courses/test_cms_views.py
@@ -9,7 +9,7 @@ from django.urls import reverse_lazy
 from cms.test_utils.testcases import CMSTestCase
 
 from richie.apps.core.factories import UserFactory
-from richie.apps.courses.factories import PersonFactory
+from richie.apps.courses import factories
 
 
 @override_settings(LANGUAGES=(("en", "En"), ("fr", "Fr"), ("de", "De")))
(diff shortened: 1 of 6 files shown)


tests/apps/courses/test_cms_views.py
```diff
@@ -23,111 +23,120 @@
...
-        admin_1 = UserFactory(permissions=[can_view_person])
+        admin_1 = UserFactory(permissions=[f"can_view_{model_name}"])
         self.client.login(username=admin_1.username, password="password")
-        payload = self.client.get(self._VIEW_URL, follow=True).json()
+        payload = self.client.get(url, follow=True).json()
         self.assertEqual(len(payload["results"]), 0)
 
         # Staff user with view permission is allowed to have results
-        admin_2 = UserFactory(is_staff=True, permissions=[can_view_person])
+        admin_2 = UserFactory(is_staff=True, permissions=["can_view_{model_name}"])
```
jbpenrath, line 63:
```suggestion
        admin_2 = UserFactory(is_staff=True, permissions=[f"can_view_{model_name}"])
```
```suggestion
        admin_2 = UserFactory(is_staff=True, permissions=["can_view_{model_name}"])
```

tests/apps/courses/test_cms_views.py
```diff
@@ -224,11 +227,29 @@ def test_cms_views_autocomplete_search(self):
...
+    for method in (
+        f for f in dir(PluginAutocompleteTestCase) if f.startswith(TEST_PREFIX)
+    ):
+        setattr(
+            PluginAutocompleteTestCase,
+            str(method).replace(TEST_PREFIX, f"{TEST_PREFIX[1:]}_{name}"),
+            lambda self, m=method, n=name, rf=related_factory: getattr(
+                PluginAutocompleteTestCase, str(m)
+            )(self, n, rf),
         )
```
jbpenrath, line 255: Good candidate to use `subTest` nope ?

jbpenrath (approved):
It just misses a `f-string` in tests.

tests/apps/courses/test_cms_views.py
sampaccoud: :+1: :pray:

tests/apps/courses/test_cms_views.py
sampaccoud: So you think I should rewrite it with a loop in each test method? If you think it's clearer I can do it :cry:

jbpenrath: I wrote a quick test and I don't know why exactly but using `subTest` with pytest do not display the context of the subTest so it is more complicated to debug the failing test than your solution. So let's keep your implementation !

sampaccoud: Yes. I like the fact that this code runs like real tests.

PR #1611, opened 2022-02-24
♿️(templates) improve template rendering

sandroscosta:
## The following PR:
- Adds titles to all `svg` files with roles as `img`.
- Removes duplicated information from blocks to avoid SR repetition.

## Purpose
I conducted some tests on https://accessmonitor.acessibilidade.gov.pt/ and proceeded to fix some of the most glaring issues. Most of the issues highlighted were due to title repetition on Screen Readers and SVG, with a role of `img`not having an alt tag.

This PR does not fix issues related with color contrast or issues related with white spacing.

## Recent results
a11y_after


src/richie/apps/core/templates/richie/base.html
```diff
@@ -147,7 +147,7 @@
                         </div>
                         <div class="body-footer__brand">
                             {% block body_footer_brand %}
-                                <a href="/">
+                                <a href="/" title="{{ SITE.name }}">
```
Leimi, line 150: Hey,

Instead of the `title` attribute here, I'd rather fill the img `alt` attribute with the site name. Filling the alt is the primary way to add alt text on a `img`. The `title`, accessibility-wise, is more of a fallback. Here it would have the additional bonus of showing a tooltip to help mouse users understand if needed, but it's not really necessary since the logo is self-explanatory.

src/richie/apps/core/templates/social-networks/blogpost-badges.html
```diff
@@ -10,6 +10,7 @@
             title="{% trans "Share on Facebook" %}"
         >
             <svg role="img">
+                <title>{% trans "Facebook" %}</title>
```
Leimi, line 13: By doing this, screen readers might announce both the anchor title and the svg title, or just one or the other.

I think the best, for svg icon links like this one and those below, would be to:

- remove the `title` attribute from the wrapping `<a>`
- have the svg `title` tag convey the meaning of the anchor, not the image by itself. So, "Share on Facebook" here.
- add an `aria-label` on the svg tag with the exact same content as the svg `title` tag. This is mostly for people using old screen readers that do not understand correctly the `title` tag.

This should make screen readers announce stuff only one time, plus keep the mouse tooltip. What do you think?

src/richie/apps/courses/templates/courses/cms/course_detail.html
```diff
@@ -83,6 +83,7 @@ <h1 class="subheader__title" property="name">{% render_model current_page "title
                                 {% if instance.duration or current_page.publisher_is_draft %}
                                 <li class="characteristics__item">
                                     <svg role="img" class="characteristics__icon" aria-hidden="true">
+                                        <title>{% trans "Duration" %}</title>
```
Leimi, line 86: I'm not sure the 3 `<svg>` changed on this page should still have an `aria-hidden` attribute? It seems they convey meaning.

Leimi (changes requested):

src/richie/apps/courses/templates/courses/cms/course_detail.html
sandroscosta: Indeed. This is probably a mistake. If they have a `title`, they should be read by the screen reader. Will fix this.

src/richie/apps/core/templates/social-networks/blogpost-badges.html
sandroscosta: Sounds great. Will do that.

src/richie/apps/core/templates/richie/base.html
sandroscosta: The only issue here is that, in our case (as an example) we use css inline SVG's to display our logo. But we've changed the  template. That was the reason for the `title` tag, but yeah, the `alt` is a much better overall solution, specially for the older browser. I'll fix this as well. Thanks.

Leimi:
Hey @sandroscosta, don't know if you still had stuff to push on this, but if you want I can take over this and continue your work :)

There are a few tiny bits we can continue to clean here and there before merging.

sandroscosta:
> 

I don't have time ATM to solve the issues with the unit test. If you can, please continue this work so I don't lose the time that I already invested. Thank you so much @Leimi .

Leimi:
Ok perfect, will check this tomorrow.

src/richie/apps/courses/templates/courses/cms/course_detail.html
Leimi: Sorry about the wrong advice, now that i'm working on it and watched the code around this, the `aria-hidden` were actually the right choice. Because Duration/Effort/Pace are all written in plain text just after, so it's actually a bad idea to repeat it in the icon.

src/richie/apps/courses/templates/courses/cms/fragment_organization_glimpse.html
```diff
@@ -57,9 +57,6 @@
                 />
             {% endblockplugin %}
         </div>
-        <div class="organization-{{ organization_variant }}__content">
-            <h{{ header_level|default:3 }} class="organization-{{ organization_variant }}__title" property="name">{{ organization_page.get_title }}</h{{ header_level|default:3 }}>
-        </div>
```
Leimi, line 62: I'm not quite sure why this was removed? Maybe unrelated stuff that was commited alongside the icons/title attributes but that should be in another PR?

Anyway, I'm reverting back those changes to only have the icons/title attribute stuff in this PR.

tests/apps/courses/test_templates_course_detail_rendering.py
```diff
@@ -110,6 +110,12 @@ def test_templates_course_detail_cms_published_content(self):
         )
 
         response = self.client.get(url)
+        content = htmlmin.minify(
+            response.content.decode("UTF-8"),
+            reduce_empty_attributes=False,
+            remove_optional_attribute_quotes=False,
+        )
+
```
Leimi, line 118: Kinda like my comment on the removal of the category glimpse title, I'm not sure why this and a few other test changes are added in this PR. It seems unrelated to the other things. I reverted those changes too to only add the icons/title attribute changes. I'll let you handle those in another PR if it was really intended, I'm not sure what this is about :)

Leimi:
Closing this in favor of https://github.com/openfun/richie/pull/1648

PR #1614, opened 2022-02-24
✨(frontend) new Icon component

Leimi:
## Purpose

Make using icons in react templates easier for developers, without them having to ask themselves about what attributes the SVG element needs depending on use cases.

## Proposal

Add an `Icon` react component that handles common use cases:

- rendering decorative icons
- rendering informative icons
- allowing custom css classes

### DIFF
--- a/src/frontend/js/components/CourseGlimpse/index.tsx
+++ b/src/frontend/js/components/CourseGlimpse/index.tsx
@@ -3,6 +3,7 @@ import { defineMessages, FormattedMessage } from 'react-intl';
 
 import { CommonDataProps } from 'types/commonDataProps';
 import { Course } from 'types/Course';
+import { Icon } from 'components/Icon';
 import { CourseGlimpseFooter } from './CourseGlimpseFooter';
 
 export interface CourseGlimpseProps {
@@ -66,17 +67,13 @@ const CourseGlimpseBase = ({ context, course }: CourseGlimpseProps & CommonDataP
           </div>
         ) : null}
         <div className="course-glimpse__metadata course-glimpse__metadata--organization">
-          <svg aria-hidden={true} role="img" className="icon">
-            <use xlinkHref="#icon-org" />
-          </svg>
+          <Icon name="icon-org" />
           <span className="title" title={course.organization_highlighted}>
             {course.organization_highlighted}
           </span>
         </div>
         <div className="course-glimpse__metadata course-glimpse__metadata--code">
-          <svg aria-hidden={true} role="img" className="icon">
-            <use xlinkHref="#icon-barcode" />
-          </svg>
+          <Icon name="icon-barcode" />
           <span>{course.code || '-'}</span>
         </div>
       </div>
--- a/src/frontend/js/components/Icon/index.tsx
+++ b/src/frontend/js/components/Icon/index.tsx
@@ -1,0 +1,53 @@
+interface IconProps {
+  name: IconType;
+  /**
+   * Optional title, will be announced by screen readers
+   * and set as `title` attribute for mouse users
+   *
+   * If omitted, the icon is considered decorative and
+   * ignored by screen readers
+   */
+  title?: string;
+  className?: string;
+}
+
+// icons from src/richie/apps/core/templates/richie/icons.html
+type IconType =
+  | 'icon-calendar'
+  | 'icon-barcode'
+  | 'icon-chevron-down'
+  | 'icon-clock'
+  | 'icon-pace'
+  | 'icon-duration'
+  | 'icon-cross'
+  | 'icon-envelope'
+  | 'icon-facebook'
+  | 'icon-filter'
+  | 'icon-linkedin'
+  | 'icon-login'
+  | 'icon-magnifying-glass'
+  | 'icon-org'
+  | 'icon-quote'
+  | 'icon-search-fail'
+  | 'icon-stopwatch'
+  | 'icon-twitter'
+  | 'icon-arrow-right'
+  | 'icon-check'
+  | 'icon-info-rounded'
+  | 'icon-warning';
+
+export const Icon = ({ name, title, className = '' }: IconProps) => {
+  return (
+    <svg
+      className={`icon ${className}`}
+      aria-hidden={title ? undefined : true}
+      {...(title && {
+        role: 'img',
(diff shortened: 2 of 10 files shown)


src/frontend/js/components/Icon/index.tsx
```diff
@@ -0,0 +1,58 @@
...
+            }),
+      }}
+    >
+      <use href={`#${name}`} />
+    </svg>
+  );
+};
+
+// icons from src/richie/apps/core/templates/richie/icons.html
+type Icon =
```
Leimi, line 36: The idea is to restrict the `name` attribute to known available icons. Here it's just a manual copy of all icons in the `icons.html` template. So, with this merged, one should be careful in the future when adding an icon in `icons.html`: adding the icon id in the type here will also be necessary.

Didn't bother trying to automate the generation of this type as I figured it was rather OK to require this tiny new thing to do when adding an icon, I'll you decide.

src/frontend/js/components/Icon/index.tsx
```diff
@@ -0,0 +1,58 @@
...
+   */
+  title?: string;
+  className?: string;
+}
+
+export const Icon = ({ name, title, className = '' }: IconProps) => {
+  return (
+    <svg
+      {...{
+        className: `icon ${className}`,
```
jbpenrath, line 18:
```suggestion
        className: ['icon', ...className].join(' '),
```

src/frontend/js/components/CourseGlimpse/index.tsx
```diff
@@ -66,17 +67,13 @@ const CourseGlimpseBase = ({ context, course }: CourseGlimpseProps & CommonDataP
           </div>
         ) : null}
         <div className="course-glimpse__metadata course-glimpse__metadata--organization">
-          <svg aria-hidden={true} role="img" className="icon">
-            <use xlinkHref="#icon-org" />
-          </svg>
+          <Icon name="icon-org" title="prout" />
```
jbpenrath, line 70: There is some gas in the code 😅
```suggestion
          <Icon name="icon-org" />
```

src/frontend/js/components/LanguageSelector/index.tsx
```diff
@@ -62,9 +63,7 @@ const LanguageSelector: FC<LanguageSelectorProps> = ({ currentLanguage, language
       </label>
       <button {...getToggleButtonProps()} className="selector__button">
         {selectedItem?.name || intl.formatMessage(messages.languages)}
-        <svg role="img" className="selector__button__icon" aria-hidden="true">
-          <use xlinkHref="#icon-chevron-down" />
-        </svg>
+        <Icon name="icon-chevron-down" className="selector__button__icon" />
```
jbpenrath, line 66: This icon has now the `.icon` className applied but it introduces a visual regression.

image

src/frontend/js/components/UserMenu/DesktopUserMenu.tsx
```diff
@@ -42,9 +43,7 @@ export const DesktopUserMenu: FC<UserMenuProps> = ({ user }) => {
       </label>
       <button {...getToggleButtonProps()} className="selector__button">
         {user.username}
-        <svg role="img" className="selector__button__icon" aria-hidden="true">
-          <use xlinkHref="#icon-chevron-down" />
-        </svg>
+        <Icon name="icon-chevron-down" className="selector__button__icon" />
```
jbpenrath, line 46: This icon has now the `.icon` className applied but it introduces a visual regression.

image

src/frontend/js/components/Icon/index.tsx
```diff
@@ -0,0 +1,58 @@
...
+          ? {
+              role: 'img',
+              'aria-label': title,
+              title,
+            }
+          : {
+              'aria-hidden': true,
+            }),
+      }}
+    >
```
jbpenrath, line 29: What do you think to rewrite this like that in order to improve readability?
```suggestion
    <svg
      className={`icon ${className}`}
      aria-hidden={!title}
      {...(title && {
        role: 'img',
        'aria-label': title,
        title,
      })}
    >
```

src/frontend/js/components/Icon/index.tsx
```diff
@@ -0,0 +1,58 @@
...
+      <use href={`#${name}`} />
+    </svg>
+  );
+};
+
+// icons from src/richie/apps/core/templates/richie/icons.html
+type IconType =
```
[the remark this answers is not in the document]
jbpenrath: I think it is a good idea as it ease the use of icon during development because before we have to open the `icons.html` file to see the list of available icons.

I'll just move this type above the component declaration.

src/frontend/js/components/CourseGlimpse/index.tsx
Leimi: Well, my huge testing skills are revealed to the public by mistake

src/frontend/js/components/LanguageSelector/index.tsx
Leimi: Oh, I was definitely not awake yesterday when open this PR then... There were other few spots where the "icon" class wasn't used and I didn't use the new component there. Felt weird to have the choice to "disable" internal behavior via a new prop so I figured not using the component in those cases was better. Will update

src/frontend/js/components/Icon/index.tsx
Leimi: Yeah, it's better!

src/frontend/js/components/Icon/index.tsx
```diff
@@ -0,0 +1,53 @@
...
+  | 'icon-arrow-right'
+  | 'icon-check'
+  | 'icon-info-rounded'
+  | 'icon-warning';
+
+export const Icon = ({ name, title, className = '' }: IconProps) => {
+  return (
+    <svg
+      className={`icon ${className}`}
+      aria-hidden={title ? undefined : true}
```
Leimi, line 43: Went with this for the aria-hidden attribute in the end, to prevent adding unnecessary `aria-hidden="false"` on most cases

Leimi:
Should be better now @jbpenrath, thanks for the review

jbpenrath (approved):
Looks good to me !

PR #1615, opened 2022-02-24
🔧(search) add elastic search client custom settings

igobranco:
Add settings that permit to configure elastic search client with custom configuration like custom timeouts.

### DIFF
--- a/src/richie/apps/search/__init__.py
+++ b/src/richie/apps/search/__init__.py
@@ -9,8 +9,17 @@ from .elasticsearch import (
 # pylint: disable=invalid-name
 default_app_config = "richie.apps.search.apps.SearchConfig"
 
-ES_CLIENT = ElasticsearchClientCompat7to6(
-    getattr(settings, "RICHIE_ES_HOST", ["elasticsearch"])
-)
+
+def new_elasticsearch_client():
+    """
+    Utility function to allow to test the `RICHIE_ES_CLIENT_KWARGS` setting.
+    """
+    return ElasticsearchClientCompat7to6(
+        getattr(settings, "RICHIE_ES_HOST", ["elasticsearch"]),
+        **getattr(settings, "RICHIE_ES_CLIENT_KWARGS", {}),
+    )
+
+
+ES_CLIENT = new_elasticsearch_client()
 
 ES_INDICES_CLIENT = ElasticsearchIndicesClientCompat7to6(ES_CLIENT)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,11 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unrealeased]
 
+## Added
+
+- Add settings that permit to configure elastic search client with custom
+  configuration like custom timeouts.
+
 ##  [2.15.0] - 2022-06-22
 
 ### Added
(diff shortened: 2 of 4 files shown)


src/richie/apps/search/__init__.py
```diff
@@ -10,7 +10,8 @@
 default_app_config = "richie.apps.search.apps.SearchConfig"
 
 ES_CLIENT = ElasticsearchClientCompat7to6(
-    [getattr(settings, "RICHIE_ES_HOST", "elasticsearch")]
+    [getattr(settings, "RICHIE_ES_HOST", "elasticsearch")],
+    **getattr(settings, "RICHIE_ES_KWARGS", {}),
```
sampaccoud, line 14: should we use names that are coherent with the related variable names? ie `RICHIE_ES_CLIENT_KWARGS` and `RICHIE_ES_INDICES_CLIENT_KWARGS`. Maybe these variables should be renamed if their name is not good?

src/richie/apps/search/index_manager.py
```diff
@@ -46,7 +46,10 @@ def perform_create_index(indexable, logger=None):
     # Create the new index
     if logger:
         logger.info(f'Creating a new Elasticsearch index "{new_index:s}"...')
-    ES_INDICES_CLIENT.create(index=new_index)
+    ES_INDICES_CLIENT.create(
+        index=new_index,
+        **getattr(settings, "RICHIE_ES_CREATE_INDEX_KWARGS", {}),
+    )
```
sampaccoud, line 52: I would add a test for each setting to confirm that it has an effect on the underlying lib.

src/richie/apps/search/__init__.py
igobranco: Yes, I've renamed them. Also included a new setting for each elastic search lib call so it was possible to tune each call.

src/richie/apps/search/index_manager.py
igobranco: I don't know how to do that.

igobranco: @sampaccoud  I've addeed a new function `new_elasticsearch_client` so it is possible to test the `ES_CLIENT` initialization. Nevertheless, I've removed the custom initialization of all the connections to the elasticsearch (all the calls to the `ES_INDICES_CLIENT` with custom kwargs). So this PR is more specific to resolve the problem.
So each installation can increase the overall timeout to the elasticsearch.

igobranco:
@sampaccoud  I've addeed a new function `new_elasticsearch_client` so it is possible to test the `ES_CLIENT` initialization. Nevertheless, I've removed the custom initialization of all the connections to the elasticsearch (all the calls to the `ES_INDICES_CLIENT` with custom kwargs). So this PR is more specific to resolve the problem.
So each installation can increase the overall timeout to the elasticsearch.

igobranco:
During meeting with @sampaccoud we have decided that I'm going to use the Django AppConfig https://github.com/openfun/richie/blob/master/src/richie/apps/search/apps.py , with the ready method to load other ES_CLIENT object. The `ES_CLIENT` that it's defined in the `apps/search/__init__.py` should be initialized within a function that can be overwritten on subsequent richie-site-factories.

igobranco:
Replaced with: https://github.com/openfun/richie/pull/2620

PR #1618, opened 2022-02-25
♿️(search) add a heading for the search results

Leimi:
## Purpose

Make the navigation between the different sections of the search page less confusing for screen reader users.

## Proposal

In the headings hierarchy, it looked like course glimpse headings were
descendants of "Filter courses". This was pretty misleading for screen
reader users that can rely a lot on headings hierarchy to navigate a
page.

This adds a heading for the results block that is the same level as the
filters to prevent this confusion. We can argue the heading is maybe not
that necessary in the UI though, so we can just have it for screen
readers.

To better understand the problem, here are screenshots of NVDA before/after the change:

<details><summary>Screenshots</summary>

**Avant**

<a target="_blank" rel="noopener noreferrer" href="https://user-images.githubusercontent.com/221253/155706324-070f8417-2a93-4b3d-a084-75dcb258d165.jpg">avant</a>

**Après**

<a target="_blank" rel="noopener noreferrer" href="https://user-images.githubusercontent.com/221253/155706350-7a9aee93-f93d-4dce-ac4f-558b2956a02b.jpg">apres</a>

</details>

### DIFF
--- a/src/frontend/js/components/Search/index.tsx
+++ b/src/frontend/js/components/Search/index.tsx
@@ -31,6 +31,12 @@ const messages = defineMessages({
       'Accessibility text for the button/icon that toggles *on* the filters pane on mobile',
     id: 'components.Search.showFiltersPane',
   },
+  resultsTitle: {
+    defaultMessage: 'Search results',
+    description:
+      'Title for the search results pane in course search (not shown, made for screen reader users).',
+    id: 'components.Search.resultsTitle',
+  },
   spinnerText: {
     defaultMessage: 'Loading search results...',
     description: 'Accessibility text for the spinner while search results are being loaded',
@@ -114,6 +120,9 @@ const Search = ({ context }: CommonDataProps) => {
         )}
       </div>
       <div className="search__results">
+        <h2 className="offscreen">
+          <FormattedMessage {...messages.resultsTitle} />
+        </h2>
         {courseSearchResponse && courseSearchResponse.status === RequestStatus.SUCCESS ? (
           <Fragment>
             {query && query.length < 3 ? (
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 - Create a React `Icon` component that can optionally take alternative text
   for screen reader users.
+- Add a heading for screen reader users on the search results page to have a
+  more understandable navigation.
 
 ### Changed
 


src/frontend/js/components/Search/index.tsx
```diff
@@ -31,6 +31,12 @@ const messages = defineMessages({
       'Accessibility text for the button/icon that toggles *on* the filters pane on mobile',
     id: 'components.Search.showFiltersPane',
   },
+  resultsTitle: {
+    defaultMessage: 'Search results',
+    description:
+      'Title for the search filters pane in course search (not shown, made for screen reader users).',
```
sampaccoud, line 37:
```suggestion
      'Title for the search results pane in course search (not shown, made for screen reader users).',
```

sampaccoud (approved):

src/frontend/js/components/Search/index.tsx
```diff
@@ -114,6 +120,9 @@ const Search = ({ context }: CommonDataProps) => {
         )}
       </div>
       <div className="search__results">
+        <h2 className="offscreen">
```
sampaccoud, line 123: Actually I would like to question the idea of a hidden h2. Isn't it probematic for SEO?

Leimi: I must say I'm not sure if search engines conclude that it's shaddy behavior or not if they detect a hidden _heading_.

That being said, this h2 in particular has no real impact on the SEO. If it's ignored by search engines, it's ignoring a "search results" text.

One thing that some people do is use aria attributes to have the same accessibility meaning as heading tags, hopefully being ignored by search engines ("hopefully" because I actually never seen actual research on "are aria attributes understood by google or not"):

```diff
- <h2>test</h2>
+ <div role="heading" aria-level="2">test</div>
```

sampaccoud: @jbpenrath WDYT? I remember you handcrafted the header levels on the search page

jbpenrath: It's fine for me to use a `h2` here. From my side, I replaced the `p` element by a `h3` to display course title within CourseGlimpse components so about semantic hierarchy all is fine.

jbpenrath (approved):

src/frontend/js/components/Search/index.tsx
sampaccoud: ok then :+1:

sandroscosta (approved):

PR #1620, opened 2022-02-25
♿️(course glimpse) improve global accessibility of a course glimpse

Leimi:
## Purpose

Make a course glimpse more understandable when using a screen reader.

## Proposal

A few stuff can be done:

- [x] add alt text to meaningful icons
- [x] reorder the DOM
- [x] change the structure with the one `<a>` wrapping everything

Note: some improvements are already added via https://github.com/openfun/richie/pull/1611 (most notably, the removal of redundant `title` attributes)

### DIFF
--- a/src/frontend/js/components/CourseGlimpse/CourseGlimpseFooter.tsx
+++ b/src/frontend/js/components/CourseGlimpse/CourseGlimpseFooter.tsx
@@ -1,8 +1,16 @@
-import { useIntl } from 'react-intl';
+import { defineMessages, useIntl } from 'react-intl';
 
 import { Icon } from 'components/Icon';
 import { CommonDataProps } from 'types/commonDataProps';
 import { Course } from 'types/Course';
+
+const messages = defineMessages({
+  dateIconAlt: {
+    defaultMessage: 'Course date',
+    description: 'Course date logo alternative text for screen reader users',
+    id: 'components.CourseGlimpseFooter.dateIconAlt',
+  },
+});
 
 /**
  * <CourseGlimpseFooter />.
@@ -13,7 +21,7 @@ export const CourseGlimpseFooter: React.FC<{ course: Course } & CommonDataProps>
   return (
     <div className="course-glimpse-footer">
       <div className="course-glimpse-footer__date">
-        <Icon name="icon-calendar" />
+        <Icon name="icon-calendar" title={intl.formatMessage(messages.dateIconAlt)} />
         {course.state.text.charAt(0).toUpperCase() +
           course.state.text.substr(1) +
           (course.state.datetime
--- a/src/frontend/js/components/CourseGlimpse/index.tsx
+++ b/src/frontend/js/components/CourseGlimpse/index.tsx
@@ -1,4 +1,4 @@
-import { memo } from 'react';
+import React, { memo } from 'react';
 import { defineMessages, FormattedMessage, useIntl } from 'react-intl';
 
 import { CommonDataProps } from 'types/commonDataProps';
@@ -36,25 +36,33 @@ const messages = defineMessages({
 const CourseGlimpseBase = ({ context, course }: CourseGlimpseProps & CommonDataProps) => {
   const intl = useIntl();
   return (
-    <a className="course-glimpse course-glimpse--link" href={course.absolute_url}>
-      <div className="course-glimpse__media">
-        {/* alt forced to empty string because it's a decorative image */}
-        {course.cover_image ? (
-          <img
-            alt=""
-            sizes={course.cover_image.sizes}
-            src={course.cover_image.src}
-            srcSet={course.cover_image.srcset}
-          />
-        ) : (
-          <div className="course-glimpse__media__empty">
-            <FormattedMessage {...messages.cover} />
-          </div>
-        )}
+    <div className="course-glimpse">
+      {/* the media link is only here for mouse users, so hide it for keyboard/screen reader users.
+      Keyboard/sr will focus the link on the title */}
+      <div aria-hidden="true" className="course-glimpse__media">
+        <a tabIndex={-1} href={course.absolute_url}>
+          {/* alt forced to empty string because it's a decorative image */}
+          {course.cover_image ? (
+            <img
+              alt=""
+              sizes={course.cover_image.sizes}
+              src={course.cover_image.src}
+              srcSet={course.cover_image.srcset}
+            />
+          ) : (
+            <div className="course-glimpse__media__empty">
+              <FormattedMessage {...messages.cover} />
+            </div>
+          )}
+        </a>
       </div>
       <div className="course-glimpse__content">
         <div className="course-glimpse__wrapper">
-          <h3 className="course-glimpse__title">{course.title}</h3>
+          <h3 className="course-glimpse__title">
+            <a className="course-glimpse__link" href={course.absolute_url}>
+              <span className="course-glimpse__title-text">{course.title}</span>
+            </a>
+          </h3>
           {course.organization_highlighted_cover_image ? (
             <div className="course-glimpse__organization-logo">
               {/* alt forced to empty string because the organization name is rendered after */}
@@ -95,7 +103,7 @@ const CourseGlimpseBase = ({ context, course }: CourseGlimpseProps & CommonDataP
         ) : null}
         <CourseGlimpseFooter context={context} course={course} />
       </div>
-    </a>
+    </div>
   );
 };
 
--- a/src/frontend/js/components/CourseGlimpse/index.tsx
+++ b/src/frontend/js/components/CourseGlimpse/index.tsx
@@ -1,5 +1,5 @@
 import { memo } from 'react';
-import { defineMessages, FormattedMessage } from 'react-intl';
+import { defineMessages, FormattedMessage, useIntl } from 'react-intl';
 
 import { CommonDataProps } from 'types/commonDataProps';
 import { Course } from 'types/Course';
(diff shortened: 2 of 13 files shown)


Leimi:
Creating this PR early to talk about one thing that would require some notable changes @jbpenrath @sampaccoud 

The fact that each course glimpse is a big `<a>` wrapping everything is a bit of a problem.

With a screen reader, when navigating in the course glimpse, each thing is announced as a "link". It's not obvious at all that each thing is a link to the course itself, and not a link to the thing it describes.

For example, when hearing "link, organization, [organization name]", we expect to be redirected to the organization, not the course.

And with a mouse, one big anchor wrapping everything also breaks the selection of text, preventing copy/pasting stuff.

Proposals to fix this:

- wrap everything with a classic `<div>`, then put the `<a>` only around the course name. But this would mean now only the title is clickable.
- wrap everything with a classic `<div>`, then re-organize the internal wrappers inside, to have one `<a>` wrapping both the cover and the title. This would mean we can click most of the view, as usually intended for such blocks.

I wouldn't personally recommend keeping the current behavior ("being able to click the whole block wherever to activate the link"), as it would require some dirty JS tricks to make it _kind of_ accessible.

jbpenrath:
> Creating this PR early to talk about one thing that would require some notable changes @jbpenrath @sampaccoud
> 
> The fact that each course glimpse is a big `<a>` wrapping everything is a bit of a problem.
> 
> With a screen reader, when navigating in the course glimpse, each thing is announced as a "link". It's not obvious at all that each thing is a link to the course itself, and not a link to the thing it describes.

Hmm.. I believed that screen reader cannot navigate within a course glimpse. From my side I used Voice Over to test it and I cannot navigate within the glimpse by using tab key. So I'm not sure to take your point.
Capture d’écran 2022-02-28 à 09 40 56


> For example, when hearing "link, organization, [organization name]", we expect to be redirected to the organization, not the course.
> 
> And with a mouse, one big anchor wrapping everything also breaks the selection of text, preventing copy/pasting stuff.
> 
> Proposals to fix this:
> 
> * wrap everything with a classic `<div>`, then put the `<a>` only around the course name. But this would mean now only the title is clickable.

The click area will be too much reduced, not a fan of this suggestion.

> * wrap everything with a classic `<div>`, then re-organize the internal wrappers inside, to have one `<a>` wrapping both the cover and the title. This would mean we can click most of the view, as usually intended for such blocks.

Pros: We could add link to Organization or Categories on glimpses
Cons: This change will modify the behavior of the course glimpse that will be something very confusing for our user.

> I wouldn't personally recommend trying to keep the exact same current behavior ("being able to click the whole block wherever to activate the link"), as it would require some dirty JS tricks to make it _kind of_ accessible.

Yes use custom JS here is not the right way.
However I wonder if could not use some css to expand the click area of a link to the whole glimpse.
e.g : Wrap only the course title within a link but use the pseudo-element `:before` positioned absolutly.

Leimi:
> Hmm.. I believed that screen reader cannot navigate within a course glimpse. From my side I used Voice Over to test it and I cannot navigate within the glimpse by using tab key. So I'm not sure to take your point.

With a screen reader, the usual way of navigating in the content is more to use arrow keys than tab keys. Because the tab key, like without a screen reader, only makes focusing "focusable" things possible (links, buttons, inputs, etc.), contrary to the arrow keys that lets you focus any content. So you could check by using the VO key+arrows instead of tab.

> However I wonder if we could not use some css to expand the click area of a link to the whole glimpse.
e.g : Wrap only the course title within a link but use the pseudo-element :before positioned absolutly.

Yes, I think we could have an invisible, absolutely positioned `a` on top of the whole glimpse, that is only clickable (not focusable). And have a `a` in the `h3` for keyboard users. This would be OK for screen readers and we would keep current behavior for mouse users. Didn't talk about it though because this behavior keeps the problem of "not being able to select any text in the glimpse", which I think is kind of a bummer. It's not an RGAA rule to actually have any text selectable, but it's usually a good thing to do to keep it that way. Some people have way less trouble copy/pasting stuff than retyping it.

And this would prevent the bonus of adding links on organization/categories ;)

Leimi:
My proposal to handle the big wrapping `<a>`: 

- replace the wrapping `<a>` containing everything by a `<div>`
- have a `<a>` around the media. This one is unfocusable and hidden from screen readers. It's here only to have a big clickable zone for mouse users
- have a `<a>` inside the title `<h3>`. This one is focusable normally.
- when you hover or focus the media or title, we toggle a class on the wrapper to keep the behavior with the card shadow.
- the title anchor has padding/margin trick to make it go outside it's container in order to have a bigger clickable zone.

The result is:

- while the clickable zone obviously is smaller than before, it's still pretty decent
- the organization/code/date text is selectable normally, and we'll be able to add a link on the organization if we want later
- screen reader users have stuff announced in a clearer way

https://user-images.githubusercontent.com/221253/158842152-9c25d2cb-486e-4893-965f-15fefef78dd3.mp4

What do you think? If you think it's OK I'll reproduce this in the django templates.

sveetch:
Just to notice than since Bootstrap 5.0 there is the "stretched link" feature which is somewhat to implement the "link wrapper around container" behavior without to use a link wrapper, but a simple link instead. However it may not work in some context and i don't know about its accessibility.

src/frontend/scss/objects/_course_glimpses.scss
```diff
@@ -68,8 +68,7 @@ $course-glimpse-content-padding-sides: 0.7rem !default;
     @include sv-flex(1, 0, calc(25% - #{$r-course-glimpse-gutter * 2}));
   }
 
-  &:hover,
-  &:focus {
+  &--focused {
```
jbpenrath, line 71: Instead to toggle a class focused through javascript event listeners, what do you think to keep the `:hover` then set `pointer-events` to none for the whole course-glimpse then set `pointer-events` to `fill` for interative children (`a` and `button`?)

```scss
pointer-events: none;

a, button {
  pointer-events: fill;
}

&:hover, &:focus { ... }
```

src/frontend/js/components/CourseGlimpse/CourseGlimpseFooter.tsx
```diff
@@ -1,9 +1,16 @@
...
+import { defineMessages, useIntl } from 'react-intl';
 
 import { Icon } from 'components/Icon';
 import { CommonDataProps } from 'types/commonDataProps';
 import { Course } from 'types/Course';
 
+const messages = defineMessages({
+  dateIconAlt: {
+    defaultMessage: 'Course date',
+    description: 'Course date logo alternative text for screen reader users',
```
jbpenrath, line 10:
```suggestion
    defaultMessage: 'Course date',
    description: 'Course date logo alternative text for screen reader users',
    id: 'components.CourseGlimpseFooter.dateIconAlt'
```
You forget to add an `id` which is used to generate translations json files.

src/frontend/js/components/CourseGlimpse/index.tsx
```diff
@@ -16,71 +16,115 @@ const messages = defineMessages({
...
+    description: 'Organization logo alternative text for screen reader users',
+  },
+  codeIconAlt: {
+    defaultMessage: 'Course code',
+    description: 'Course code logo alternative text for screen reader users',
+  },
+  categoryLabel: {
+    defaultMessage: 'Category',
+    description: 'Category label text for screen reader users',
+  },
```
jbpenrath, line 30: Ditto, do not forget to add an `id` property for each message.
By convention we use the following pattern :
`components.<COMPONENT_NAME>.<MESSAGE_PROPERTY_KEY>`

src/frontend/js/components/CourseGlimpse/index.tsx
```diff
@@ -16,71 +16,115 @@ const messages = defineMessages({
...
+          ) : null}
+          <div className="course-glimpse__metadata course-glimpse__metadata--organization">
+            <Icon name="icon-org" title={intl.formatMessage(messages.organizationIconAlt)} />
+            <span className="title" title={course.organization_highlighted}>
+              {course.organization_highlighted}
+            </span>
+          </div>
+          <div className="course-glimpse__metadata course-glimpse__metadata--code">
+            <Icon name="icon-barcode" title={intl.formatMessage(messages.codeIconAlt)} />
+            <span>{course.code || '-'}</span>
```
jbpenrath, line 102: Is it meaning full to add a title to the `Icon` here? Is not it better to set an aria-label on the span ? Like `Course code 0000` ?

src/frontend/js/components/CourseGlimpse/index.tsx
Leimi: woops! thanks

src/frontend/scss/objects/_course_glimpses.scss
Leimi: Hey, that's pretty clever :) great idea.

src/frontend/js/components/CourseGlimpse/index.tsx
Leimi: _Just as a side-note, setting an aria-label on a span usually does not do anything, screen readers don't announce it. We would need to change the role of the span to trigger screen readers to see the aria-label._

I'm sure I'm overthinking this but it's a great question in general: is it better to label an image, or add hidden text?

We can decide to have a decorative icon, and add the label in a text wrapped in the offscreen class.

In that case, the actual text announced by screen readers will be, as you said, "Course code 0000", in the same sentence as one block.

Or we can label the icon like in the PR.

In that case, the screen reader will announce "Course code, Image", then "0000". So, a little more verbose. But the bonus of setting the icon title is that mouse users can hover the icon in order to show its title.

Both techniques are totally valid. I usually prefer the latter, because it helps mouse users too. But I know this is a questionable bonus, one could say: if we feel the need to have a text explanation of an icon, the solution should be to change the icon, or remove it and replace it by a text for everyone. Even more so because this "tooltip" solution is not available for mobile phone users.

Here I find it great to have tooltips for the icons. It's rather difficult to have a 100% obvious icon for an "organization", and the bar code icon is pretty small, it's easy to not get what it is directly. So a little help is welcomed.

Or we could do kind of both: have an Icon that has the "Course code" title _but_ also marked aria-hidden ; and "Course code" as offscreen text in the span. This adds a tooltip for mouse users, that is not announced by screen readers, because the offscreen text is announced. But now we want to be sure we understand in the code the distinction between an icon with a screen reader label or a mouse-tooltip label...

jbpenrath: Many thanks for this explanation !

With that in mind, it seems that your solution is the best compromise to offer a good experience for screen readers and mouse users while keeping code simple. So I would say, let's keep it !

src/frontend/scss/objects/_course_glimpses.scss
Leimi: So I'm using your technique, it's great!

One small detail that can feel weird is that in order to be able to mouse hover the icons having titles, I need to activate pointer-events on them too, so that makes the card activate its shadow in this case too.

For keyboard, the idea was to activate the card shadow when focusing anything in the card *and* outline as usual the link that is actually focused inside it. Hence the usage of "focus-within" on the card wrapper, and "focus-visible" instead of "focus" on the title link, to prevent triggering the focus styles while just clicking with the mouse.

Leimi:
Hey @jbpenrath if you have time to review this today I'd appreciate it :)

src/frontend/scss/objects/_course_glimpses.scss
```diff
@@ -147,17 +159,56 @@ $course-glimpse-content-padding-sides: 0.7rem !default;
...
+    -webkit-box-orient: vertical;
+    -webkit-line-clamp: 3;
+    display: block;
+    display: -webkit-box;
     overflow: hidden;
+
+    // we extend the clickable zone of the link,
+    // so that the clickable zone actually sticks to the edges of the card
+    padding: 1.7rem $course-glimpse-content-padding-sides 1rem;
+    margin: -1.7rem #{$course-glimpse-content-padding-sides * -1} -1rem;
```
jbpenrath, line 181: The padding bottom could be increase to cover all the heading bounding box
```suggestion
    padding: 1.7rem $course-glimpse-content-padding-sides 1.7rem;
    margin: -1.7rem #{$course-glimpse-content-padding-sides * -1} -1.7rem;
```

jbpenrath (approved):
Great works 👏

src/frontend/scss/objects/_course_glimpses.scss
Leimi: If we do this and the title takes up 3 lines, the link goes above the organization text below so I'm not sure we should do this

Leimi: But your comment helped me notice we can do better, now the link takes the whole title height as it should.

Leimi: Ok, now should be good after moving the line-clamp stuff :pray: thanks

PR #1624, opened 2022-02-28
📝(upgrade) add missing instruction

jbpenrath:
## Purpose

Some instructions are missing to upgrade to Richie 2.13.0.

## Proposal

- [x] Add missing instruction into `UPGRADE.md`


UPGRADE.md
```diff
@@ -35,6 +35,20 @@ $ make migrate
       'dal_select2',
   )
   ```
+- `WEB_ANALYTICS_ID` is now a mandatory settings. You can set it to `None` if you do not
```
sampaccoud, line 38: It sucks no? Why not make None a default?

jbpenrath: Yes, finally it seems to be a bug.
https://github.com/openfun/richie/pull/1624/commits/e13853b0294b2dd0d71a182e056d4b1242f3252e

sampaccoud (approved):

PR #1625, opened 2022-02-28
♿️(search) better navigation in search page

Leimi:
## Purpose

Improve global navigation in the search results page.

## Proposal

We can do a few stuff to improve global navigation in the search page:

- [ ] add region attributes to the filters/search pane
- [ ] make sure we let the user know when search results change after toggling filters
- [ ] wrap the results in a `ul`
- [ ] improve the `title` tag. This will certainly require opening another issue/PR to discuss

This is work-in-progress and totally unfinished, will rework the commits later. Opening early for the curious.


src/frontend/js/components/SearchFiltersPane/index.tsx
```diff
@@ -33,19 +33,40 @@ export const SearchFiltersPane = ({
...
+            return prev;
+          }
+
+          const filterValues = Array.isArray(filter[1]) ? filter[1] : [filter[1]];
+          const names = current.values
+            .filter((value) => filterValues.includes(value.key))
+            .map((value) => `"${value.human_name}"`);
+          return [...prev, ...names];
+        }, [])
+      : []
```
jbpenrath, line 67: Hard to read / understand at first.

```suggestion
// Get of list of active filters human names, in the order of the shown filter list
  function getFilterNames(prev: string[], current: FacetedFilterDefinition) {
    const filter = relevantFilters.find((entry) => entry[0] === current.name);
    let names: string [] = [];

    if (filter) {
      const filterValues = Array.isArray(filter[1]) ? filter[1] : [filter[1]];
      names = current.values
        .filter((value) => filterValues.includes(value.key))
        .map((value) => `"${value.human_name}"`);
    }

    return [...prev, ...names];
  }
  const activeFilterNames = (filterList || []).reduce(getFilterNames, []).join(', ');
```

Leimi: I missed making a comment to warn you about this but yeah, I was sure you were gonna find a way to make it clearer 😅 thanks

sandroscosta (approved):

src/frontend/js/components/CourseGlimpseList/index.tsx
```diff
@@ -27,7 +33,19 @@ export const CourseGlimpseList = ({
 }: CourseGlimpseListProps & CommonDataProps) => {
   return (
     <div className="course-glimpse-list">
-      <div className="course-glimpse-list__count">
+      <div
+        className="offscreen course-glimpse-list__screen-reader-count"
```
jbpenrath, line 37: It seems the class `course-glimpse-list__screen-reader-count` is for test purpose only. So if I'm right, you should prefer to use `data-testid`.

https://testing-library.com/docs/queries/bytestid/

src/frontend/js/components/SearchFilterGroupModal/index.spec.tsx
```diff
@@ -96,6 +96,8 @@ describe('<SearchFilterGroupModal />', () => {
 
       // The modal is rendered
       const openButton = screen.getByText('More options');
+      // Make sure there is a more explicit text for screen readers
+      expect(openButton.querySelector('.offscreen')?.textContent).toEqual('(Universities)');
```
jbpenrath, line 100: Did you try to query this element with `getByRole` ?
```
screen.getByRole('button', { name: 'More options (Universities)' });
```

src/frontend/js/components/SearchFiltersPane/index.spec.tsx
```diff
@@ -118,6 +130,11 @@ describe('components/SearchFiltersPane', () => {
       'search-filters-pane__clear--hidden',
     );
 
+    // we want additional hidden text for screen reader users
+    expect(clearButton).toHaveTextContent(
+      'Clear 2 active filters ("Dummy category", "Dummy organization")',
+    );
+
```
jbpenrath, line 137: Ditto, did you try to query button through `getByRole` ?

src/frontend/js/components/SearchFilterGroupModal/index.spec.tsx
Leimi: I think I tried but will check again

src/frontend/js/components/CourseGlimpseList/index.tsx
Leimi: Oh yes!

src/frontend/js/components/SearchFilterGroupModal/index.spec.tsx
Leimi: I guess I didn't try correctly because it works :') thanks

PR #1626, opened 2022-02-28
🐛(course) fix cms toolbar bug when editing a plugin from a detail page

sveetch:
## Purpose

It has been reported that, when editing a plugin, a bug often appears which turns the CMS toolbar in mobile mode.

It was due to usage of `get_placeholder_plugins` and `blockplugin` tags in page metas. This is because in edit mode these tags append some markup required for the CMS interface to edit in place a plugin and so it appends some HTML elements like `<div>` and `<template>` which are totally invalid in `<head>` elements and also some Javascript code which invoke some object/function that are not yet loaded.

This is more probably the invalid HTML which cause the error since with Firefox 90.0 there is no bug but it always reproduce with Chrome. Firefox and Chrome quirk mode are different and Firefox have commonly a better one which just ignore the invalid HTML.

## Proposal

This commit adds a new keyword argument `silent` to previously cited template tags. This argument when set to 'True' disable default behavior of edition mode and so does not inject any markup related to placeholder and plugin edition. And finally, all detail template now use this argument on tags inside '<head>'.

Tags tests have been updated to cover this new argument behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,10 @@ Versioning](https://semver.org/spec/v2.0.0.html).
   more understandable navigation.
 - Add new page templates for various multiple columns and their own CSS to
   adjust every possible plugins to fit correctly;
+- Add new keyword argument `silent` to template tags `get_placeholder_plugins`
+  and `blockplugin`. It have to be set to `True` on these tags when they are
+  used from `<head>` so they never append any invalid markup related to
+  edition mode.
 
 ### Changed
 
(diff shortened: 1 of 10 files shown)


sveetch:
Ah my changes on templates make some tests related to metas to fails, so finally this can't even be a quick fix.

sveetch:
@sampaccoud @jbpenrath Ok i've updated my pull request with a clean solution with a new keyword argument on template tags to avoid invalid markup injection. Everything should be fine now.

PR #1629, opened 2022-03-08
🐛(layout) fix section tile title with richtext

sveetch:
## Purpose

It is required to disable automatic escaping from Django template since title is a
richtext now and CKEditor already transforms special characters which can leads to
artificial double escaping.

## Proposal

Section tile template has been changed to add '|safe' filter on section title to fix
this issue.


sampaccoud (approved):

sampaccoud:
@sveetch can you add a test to surface the problem?

sveetch:
@sampaccoud Ok i've update the PR with tests to cover the title behavior from section templates.

PR #1630, opened 2022-03-09
Feat/step manager

jbpenrath:
## Purpose

Create a frontend component to manage processes split in step.
To achieve that we created a hook `useStepManager` which takes as argument a manifest in charge to describe each steps (name, icon, callback onEnter & onExit) and their relationships.

Furthermore, we created a `StepBreadcrumb` component which is in charge to display steps of manifest and highlight the active step.

## Proposal

- [x] Create a `useStepManager` hook
- [x] Create a `StepBreadcrumb` component

### DIFF
--- a/src/frontend/js/components/StepBreadcrumb/_styles.scss
+++ b/src/frontend/js/components/StepBreadcrumb/_styles.scss
@@ -1,0 +1,96 @@
+.StepBreadcrumb {
+  $margin-y: 1.25rem;
+  $icon-size: 3.375rem;
+  display: flex;
+  flex-direction: row;
+  list-style: none;
+  margin: 0 19% 2.375rem 19%;
+  padding: 0;
+
+  &__step {
+    align-items: center;
+    display: flex;
+    flex-direction: column;
+    flex: 1;
+    margin: $margin-y 1.5rem;
+    position: relative;
+    text-align: center;
+    z-index: 1;
+
+    &__icon {
+      align-items: center;
+      background: r-theme-val(steps-breadcrumb, icon-background);
+      border-radius: 50%;
+      border: 3px solid r-theme-val(steps-breadcrumb, icon-border);
+      box-sizing: border-box;
+      color: r-theme-val(steps-breadcrumb, icon-fill);
+      display: flex;
+      font-family: $r-font-family-montserrat;
+      font-size: 1.25rem;
+      height: $icon-size;
+      justify-content: center;
+      margin-bottom: 0.3125rem;
+      width: $icon-size;
+
+      svg {
+        fill: currentColor;
+        max-height: $icon-size;
+        width: 55%;
+      }
+    }
+
+    &__label {
+      -moz-box-orient: vertical;
+      -moz-line-clamp: 3;
+      -webkit-box-orient: vertical;
+      -webkit-line-clamp: 3;
+      bottom: 0;
+      color: r-theme-val(steps-breadcrumb, label-color);
+      display: -moz-box;
+      display: -webkit-box;
+      font-weight: 600;
+      letter-spacing: 1px;
+      overflow: hidden;
+      padding-top: 0.625rem;
+      position: absolute;
+      text-transform: uppercase;
+      transform: translateY(100%);
+    }
+
+    &--active {
+      .StepBreadcrumb__step__icon {
+        background-color: r-theme-val(steps-breadcrumb, icon-background-active);
+        border-color: r-theme-val(steps-breadcrumb, icon-border-active);
+        box-shadow: 0 0 0 3px r-theme-val(steps-breadcrumb, icon-outline);
+        color: r-theme-val(steps-breadcrumb, icon-fill-active);
+      }
+
+      .StepBreadcrumb__step__label {
+        color: r-theme-val(steps-breadcrumb, label-color-active);
+      }
+    }
+  }
+
+  &__separator {
+    $separator-height: 0.1875rem;
+    background-color: r-theme-val(steps-breadcrumb, separator-background);
+    border-radius: 100vw;
+    height: $separator-height;
+    position: relative;
+    transform: translateY($margin-y + ($icon-size - $separator-height) * 0.5);
+    width: 100%;
+
+    &--active {
+      background-color: r-theme-val(steps-breadcrumb, separator-background-active);
+    }
+  }
+
+  @include media-breakpoint-down(xs) {
+    .StepBreadcrumb__separator {
+      display: none;
+    }
+    & > .StepBreadcrumb__step:not(.StepBreadcrumb__step--current) {
+      display: none;
+    }
+  }
+}
--- a/src/frontend/js/components/StepBreadcrumb/index.spec.tsx
+++ b/src/frontend/js/components/StepBreadcrumb/index.spec.tsx
@@ -1,0 +1,218 @@
+import {
+  act,
+  getAllByRole,
+  queryAllByRole,
+  queryAllByTestId,
+  queryByRole,
+  render,
+} from '@testing-library/react';
+import { renderHook } from '@testing-library/react-hooks';
+import { type Manifest, useStepManager } from 'hooks/useStepManager';
+import { StepBreadcrumb } from '.';
+
+describe('StepBreadcrumb', () => {
+  it('renders visually a minimal manifest', () => {
+    // If manifest's steps does not have `label` and `icon` property,
+    // only a breadcrumb with the step index is displayed.
+    type LastStep = 'step1';
+    type Steps = 'step0' | LastStep;
+    const manifest: Manifest<Steps, LastStep> = {
+      start: 'step0',
+      steps: {
+        step0: {
+          next: 'step1',
+        },
+        step1: {
+          next: null,
+        },
+      },
+    };
+
+    const { result } = renderHook(() => useStepManager(manifest));
+    const { container, rerender } = render(
+      <StepBreadcrumb manifest={manifest} step={result.current.step} />,
+    );
+
+    expect(getAllByRole(container, 'listitem')).toHaveLength(2);
+
+    const labels = queryAllByRole(container, 'heading', { level: 6 });
+    const stepsIcons = queryAllByTestId(container, 'StepBreadcrumb__step__icon');
+    let activeSteps = container.querySelectorAll(
+      'li.StepBreadcrumb__step.StepBreadcrumb__step--active',
+    );
+    let currentStep = container.querySelector(
+      'li.StepBreadcrumb__step.StepBreadcrumb__step--current',
+    );
+    expect(labels).toHaveLength(0);
+    expect(stepsIcons).toHaveLength(2);
+    stepsIcons.forEach((icon, index) => {
+      expect(icon.firstChild).toBeInstanceOf(HTMLSpanElement);
+      expect(icon.firstChild).toHaveTextContent((index + 1).toString());
+    });
+    expect(activeSteps).toHaveLength(1);
+    expect(currentStep).toHaveTextContent('1');
+
+    act(() => result.current.next());
+    rerender(<StepBreadcrumb manifest={manifest} step={result.current.step} />);
+
+    activeSteps = container.querySelectorAll(
+      'li.StepBreadcrumb__step.StepBreadcrumb__step--active',
+    );
+    currentStep = container.querySelector('li.StepBreadcrumb__step.StepBreadcrumb__step--current');
+    expect(activeSteps).toHaveLength(2);
+    expect(currentStep).toHaveTextContent('2');
+  });
+
+  it('renders visually a complete manifest', () => {
+    // If manifest's steps has `label` and `icon` property,
+    // this information was shown.
+    type LastStep = 'step1';
+    type Steps = 'step0' | LastStep;
+    const manifest: Manifest<Steps, LastStep> = {
+      start: 'step0',
+      steps: {
+        step0: {
+          label: '0. Step',
+          icon: 'icon-arrow-right',
+          next: 'step1',
+        },
+        step1: {
+          label: '1. Step',
+          icon: 'icon-arrow-right',
+          next: null,
+        },
+      },
+    };
+
+    const { result } = renderHook(() => useStepManager(manifest));
+    const { container, rerender } = render(
+      <StepBreadcrumb manifest={manifest} step={result.current.step} />,
+    );
+
+    expect(getAllByRole(container, 'listitem')).toHaveLength(2);
+
+    const labels = queryAllByRole(container, 'heading', { level: 6 });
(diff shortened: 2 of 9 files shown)


src/frontend/js/components/StepBreadcrumb/index.spec.tsx
```diff
@@ -0,0 +1,170 @@
...
+    let activeSteps = container.querySelectorAll(
+      'li.StepBreadcrumb__step.StepBreadcrumb__step--active',
+    );
+    let currentStep = container.querySelector(
+      'li.StepBreadcrumb__step.StepBreadcrumb__step--current',
+    );
+    expect(labels).toHaveLength(2);
+    labels.forEach((label, index) => {
+      expect(label).toHaveTextContent(`${index}. Step`);
+    });
```
carofun, line 94:
```suggestion
    expect(labels).toHaveLength(2);
    screen.getByRole('heading', { name: `0. Step` })
    screen.getByRole('heading', { name: `1. Step` })
```
As you defined labels just here, we can directly target each label with the defined name without using variables

src/frontend/js/components/StepBreadcrumb/index.spec.tsx
```diff
@@ -0,0 +1,170 @@
...
+    const activeSteps = container.querySelectorAll(
+      'li.StepBreadcrumb__step.StepBreadcrumb__step--active',
+    );
+    const currentStep = container.querySelector(
+      'li.StepBreadcrumb__step.StepBreadcrumb__step--current',
+    );
+    expect(activeSteps).toHaveLength(2);
+    expect(currentStep).toHaveTextContent('2');
+  });
+});
```
carofun, line 170: we could add an extra test to validate the number of separator items

src/frontend/js/components/StepBreadcrumb/index.tsx
```diff
@@ -0,0 +1,101 @@
...
+ * When step is null, we know that process is terminated so we have to return
+ * the last step index.
+ * @params step[]
+ * @returns {number} active step index
+ */
+function getActiveStepIndex(steps: FlattenStep[], step: Props['step']) {
+  if (step === null) {
+    return steps.length;
+  }
+  return steps.findIndex((s) => s[0] === step);
```
carofun, line 26:
```suggestion
return step === null ? steps.length : steps.findIndex((s) => s[0] === step);
```
a subjective preferred choice, I let it up to you

src/frontend/js/components/StepBreadcrumb/index.spec.tsx
```diff
@@ -0,0 +1,170 @@
...
+      'li.StepBreadcrumb__step.StepBreadcrumb__step--active',
+    );
+    currentStep = container.querySelector('li.StepBreadcrumb__step.StepBreadcrumb__step--current');
+    expect(activeSteps).toHaveLength(2);
+    expect(currentStep).toHaveTextContent('2');
+  });
+
+  it('renders visually a complete manifest', () => {
+    // If manifest's steps has `label` and `icon` property,
+    // these information are shown.
```
carofun, line 59:
```suggestion
    // If manifest's steps has `label` and `icon` property,
    // these information are shown.
```
comments of two tests, with icon and label or without icons and without labels, made me think it was either both together or without both, maybe it can be clarified and in one test use a step that has icon but not label and the other way round

src/frontend/js/hooks/useStepManager/index.spec.ts
```diff
@@ -0,0 +1,129 @@
...
+
+    // - Reset: go back to step0
+    act(() => result.current.reset());
+
+    expect(manifest.steps.step0.onEnter).toHaveBeenCalledTimes(3);
+    expect(manifest.steps.step0.onExit).toHaveBeenCalledTimes(2);
+    expect(manifest.steps.step1.onEnter).toHaveBeenCalledTimes(2);
+    expect(manifest.steps.step1.onExit).toHaveBeenCalledTimes(2);
+  });
+});
```
carofun, line 129:
```suggestion
});
//two suggested tests so that we know component still works and we are aware of this two behaviour
//i'm wondering if Manifest shouldn't be more controlled for the last test to not be possible
it('can contain a single step', () => {
    const manifest: Manifest = {
      start: 'step0',
      steps: {
        step0: {
          next: null,
        }
      },
    };
    const { result } = renderHook(() => useStepManager(manifest));

    // - step 0
    expect(result.current.step).toEqual('step0');

    // - terminated state
    act(() => result.current.next());
    expect(result.current.step).toEqual(null);

    // - Reset: should go back to step0
    act(() => result.current.reset());
    expect(result.current.step).toEqual('step0');
    
  });
  
  it("loads manifest if first step doesn't exist but steps aren't linked anymore", () => {
    const manifest: Manifest = {
      start: 'step0',
      steps: {
        step1: {
          next: 'step2',
          onEnter: jest.fn(),
          onExit: jest.fn(),
        },
        step2: {
          next: null,
          onEnter: jest.fn(),
          onExit: jest.fn(),
        },
      },
    };
    const { result } = renderHook(() => useStepManager(manifest));

    // - step 0
    expect(result.current.step).toEqual('step0');
    expect(manifest.steps.step0).toEqual(null);
    
    act(() => result.current.next());
    expect(result.current.step).toEqual(null);

    act(() => result.current.reset());
    expect(result.current.step).toEqual('step0');
  });
```

src/frontend/js/hooks/useStepManager/index.ts
```diff
@@ -0,0 +1,52 @@
...
+  label?: string;
+  next: Nullable<string>;
+  onEnter?: Function;
+  onExit?: Function;
+}
+
+export interface Manifest {
+  start: string;
+  steps: {
+    [key: string]: Step;
```
carofun, line 15:
```suggestion
    [key: string]: Step;
```
start key is not necessarily included in step, is there's something we could do about this?

src/frontend/js/components/StepBreadcrumb/index.tsx
```diff
@@ -0,0 +1,101 @@
...
+}
+
+/**
+ * The first step of the manifest could be defined as the only step which is
+ * never present in `next` property of another step.
+ * So this method searches which step is not used by another step.
+ * @param {Manifest} manifest
+ * @returns {Manifest.step} step
+ */
+function findFirstStep(manifest: Manifest) {
```
carofun, line 36:
```suggestion
function findFirstStep(manifest: Manifest) {
```
I find it confusing not to use directly the starting step declared In the manifest instead of searching for the step that doesn't contain the next step. As we have our first step declared, why not just use it ?

src/frontend/js/components/StepBreadcrumb/index.tsx
```diff
@@ -0,0 +1,101 @@
...
+      {orderedSteps.map(([key, entry], index) => (
+        <Fragment key={`step-breadcrumb-key-${key}`}>
+          <li
+            aria-current={index === activeIndex ? 'step' : 'false'}
+            className={`StepBreadcrumb__step
+              ${index <= activeIndex ? 'StepBreadcrumb__step--active' : ''}
+              ${index === activeIndex ? 'StepBreadcrumb__step--current' : ''}
+            `}
+          >
+            <div className="StepBreadcrumb__step__icon">
```
carofun, line 78:
```suggestion
            <div className="StepBreadcrumb__step__icon" data-testid="testIdIcon">
```
to target with a data-testid as suggested in the test

src/frontend/js/components/StepBreadcrumb/index.spec.tsx
```diff
@@ -0,0 +1,170 @@
...
+
+    const { result } = renderHook(() => useStepManager(manifest));
+    const { container, rerender } = render(
+      <StepBreadcrumb manifest={manifest} step={result.current.step} />,
+    );
+
+    expect(getAllByRole(container, 'listitem')).toHaveLength(2);
+
+    const labels = container.querySelectorAll('h6.StepBreadcrumb__step__label');
+    const stepsIcons = container.querySelectorAll('div.StepBreadcrumb__step__icon > *');
```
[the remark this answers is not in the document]
jbpenrath: I think, I could target icons by using using queryAllByRole('img'...

src/frontend/js/components/StepBreadcrumb/index.tsx
jbpenrath: Because the `start` property could be any step not necessary the first step of the manifest. In this way, we are able to instantiate the step manager at a given step.

src/frontend/js/hooks/useStepManager/index.ts
jbpenrath: Good idea! I should be able to do that.

src/frontend/js/hooks/useStepManager/index.spec.ts
jbpenrath: Ok for the first test you suggest. About the second, in fact we should not be able to do that. If I can type `Manifest` as discussed in your comment below, we should no longer be able to write it.

carofun: That would be great and it was the point of writing it!

src/frontend/js/components/StepBreadcrumb/index.tsx
carofun: ok start step doesn't mean first, I see. It wasn't clear to me but it makes sense. 
Simple question, shouldn't we delegate these properties such as .first like we have current to the manager instead of the component so that the component has a ready-to-use and ordered breadcrumb?

jbpenrath: Hmmm... I don't think so as step order is only important for the display not for the manager.

src/frontend/js/components/StepBreadcrumb/_styles.scss
```diff
@@ -0,0 +1,96 @@
+.StepBreadcrumb {
+  $margin-y: 20px;
+  $icon-size: 54px;
+  display: flex;
+  flex-direction: row;
+  list-style: none;
+  margin: 0 19% 2.375rem 19%;
```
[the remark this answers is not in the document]
jbpenrath: I have no problem with that. `Percentage` is relative to the width of the container while `rem` is relative to the root font size. In the first case, margin will adapt to the width of the container while with `rem`, until the root font size changes, the horizontal margin will be always the same.

src/frontend/js/hooks/useStepManager/index.ts
jbpenrath: I found a workaround by providing a generic type to `Manifest` and pass down this one to `Step` type.

src/frontend/js/components/StepBreadcrumb/index.tsx
carofun: sorry to insist :-), in the manager, we have the current, the next, the reset, are you sure to get the first() wouldn't have his place here? I find it weird to have to calculate in a component outside the manager using the method next()
I would even use the manager to get a specific step, anything to deal with and ease the use of the manifest could be plugged in the manager.

src/frontend/js/hooks/useStepManager/index.spec.ts
```diff
@@ -0,0 +1,158 @@
...
+    type Steps = 'step0' | 'step1';
+    const manifest: Manifest<Steps> = {
+      start: 'step1',
+      steps: {
+        step0: {
+          next: 'step1',
+        },
+        step1: {
+          next: 'step1',
+        },
```
carofun, line 16: next of step0 and step1 is step1, is there something wrong no?

jbpenrath: Oh yes 😅

src/frontend/js/hooks/useStepManager/index.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+
+    // - Reset: go back to step0
+    act(() => result.current.reset());
+
+    expect(manifest.steps.step0.onEnter).toHaveBeenCalledTimes(3);
+    expect(manifest.steps.step0.onExit).toHaveBeenCalledTimes(2);
+    expect(manifest.steps.step1.onEnter).toHaveBeenCalledTimes(2);
+    expect(manifest.steps.step1.onExit).toHaveBeenCalledTimes(2);
+  });
+});
```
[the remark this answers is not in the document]
jbpenrath: Ok, to show that no error is thrown. But in fact, I wonder if we do not prefer to throw an error when manifest is not well configured ?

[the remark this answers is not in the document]
jbpenrath: Ok, to show that no error is thrown. But in fact, I wonder if we do not prefer to throw an error when manifest is not well configured ?

src/frontend/js/components/StepBreadcrumb/index.spec.tsx
```diff
@@ -0,0 +1,210 @@
...
+    const activeSteps = container.querySelectorAll(
+      'li.StepBreadcrumb__step.StepBreadcrumb__step--active',
+    );
+    const currentStep = container.querySelector(
+      'li.StepBreadcrumb__step.StepBreadcrumb__step--current',
+    );
+    expect(activeSteps).toHaveLength(1);
+    expect(currentStep).toHaveTextContent('1');
+  });
+
```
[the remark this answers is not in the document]
jbpenrath: As this point is now managed by the step manager, we have already a test that check this.

src/frontend/js/hooks/useStepManager/index.spec.ts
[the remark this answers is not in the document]
carofun: I would, it disturbs me that it's not control

src/frontend/js/components/StepBreadcrumb/index.spec.tsx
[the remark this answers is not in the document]
carofun: It's tricky, true ...-)As it is the component itself that is ordering and he can order in two cases with a starting step and with none, IMO both scenarios must be tested. I think it's the result of the order that matters not the"how". "how" could change in the future. This is only my point of view and in general, the extra test doesn't arm, not doing one could be missing in the future. But I let it up to you

src/frontend/js/hooks/useStepManager/index.ts
```diff
@@ -0,0 +1,69 @@
...
+  onEnter?: Function;
+  onExit?: Function;
+}
+
+export interface Manifest<S extends PropertyKey = PropertyKey> {
+  steps: {
+    [key in S]: Step<S>;
+  };
+  start?: S;
+}
```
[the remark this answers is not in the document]
carofun: waouh my godness, Manifest just have to watch out now!

carofun (approved):
it lgtm

PR #1631, opened 2022-03-10
🐛(layout) display author name in blogpost detail in published mode also

sveetch:
## Purpose

The author person name was hidden in blogpost detail when viewed in published mode and only showed in draft mode. But alike categories, the author is required to be displayed in published mode also except if its placeholder is empty.

## Proposal

This commit fix template to use the right condition to display author and add some tests around this behavior.


sampaccoud (approved):
:clap:

PR #1632, opened 2022-03-11
✨(courses) add 'excerpt' placeholder to Organization detail template

sveetch:
## Purpose

Organization plugin 'row' variant has been reported to have broken layout in some situation.

This is due to two things, firstly the CSS was not made for the link wrapper around content, secondly Organization description can include links which break HTML since it is already included in a link wrapper, this is not valid and leads to CSS issues.

Note than the reported bug was when the Org description included a link but in fact there was other issues without any included links, this is because of invalid HTML that leads to erroneous behavior that depended from browser resolution with invalid HTML.

Below you can see a sample of render with the two possibilities (without or with included links in description):

org_row_sample_bug

Note than when row is hovered, everything adopt the "hover link" CSS behaviors with underlined and colorized text. Finally the expected layout was better when there is included links in description, since the invalid HTML break the DOM and so the row adopt the CSS parent selector which have some CSS properties to make image aside content.

Below you can see a sample with my CSS fix:

org_row_sample_fixed

Here the first row without any included link in description have the expected layout (image and text aside, correct font, correct color, etc..). And the second row which include a link in description is almost correctly but since HTML is broken, the text layout is broken.

## Proposal

This leaded to decision to add a new placeholder 'excerpt' alike with Blogpost, so Organization 'row' variant can still be in a link container since its text content is always plain text. The excerpt is used if not empty, else it will fallback to use the description but with every markup stripped out.

And now it looks like this:

org_row_sample_expected

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 - Add new page templates for various multiple columns and their own CSS to
   adjust every possible plugins to fit correctly;
 - Add theme variable to change checkmark color.
+- Add new placeholder 'excerpt' to Organization detail template
 
 ### Changed
 
@@ -24,6 +25,8 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 - Improve React search suggestion field labels for screen readers.
 - Removed usage of deprecated Sass division '/' operator in favor of
   'math.div'.
+- Organization plugin variant 'row' now use excerpt if not empty else the
+  description with every markup removed
 
 ### Fixed
 
@@ -36,6 +39,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 - Fix section tile title with richtext and special character escaping.
 - Change blogpost detail template to display author even in published mode
   except if its placeholder is empty.
+- Fix missing styles for Organization plugin 'row' variant link wrapper
 
 ## [2.13.0] - 2022-02-18
 
--- a/src/frontend/scss/objects/_organization_glimpses.scss
+++ b/src/frontend/scss/objects/_organization_glimpses.scss
@@ -186,6 +186,25 @@
     flex-basis: 100%;
   }
 
+  & > a {
+    align-content: flex-start;
+    align-items: center;
+    color: inherit;
+    display: block;
+    flex-direction: row;
+    flex-wrap: nowrap;
+
+    &:hover,
+    &:focus {
+      color: inherit;
+      text-decoration: none;
+    }
+
+    @include media-breakpoint-up(lg) {
+      display: flex;
+    }
+  }
+
   &:hover,
   &:focus {
     color: inherit;
--- a/src/richie/apps/courses/templates/courses/cms/fragment_organization_glimpse.html
+++ b/src/richie/apps/courses/templates/courses/cms/fragment_organization_glimpse.html
@@ -30,8 +30,22 @@
             {% endblockplugin %}
         </div>
         <div class="organization-{{ organization_variant }}__content">
-            <h{{ header_level|default:3 }} class="organization-{{ organization_variant }}__title" property="name">{{ organization_page.get_title }}</<h{{ header_level|default:3 }}>
-            <div class="organization-{{ organization_variant }}__excerpt" property="description">{% get_placeholder_plugins "description" organization_page as plugins or %}{% trans "Description" %}{% endget_placeholder_plugins %}{% blockplugin plugins.0 %}{{ instance.body|truncatewords_html:40|safe }}{% endblockplugin %}</div>
+            <h{{ header_level|default:3 }} class="organization-{{ organization_variant }}__title" property="name">{{ organization_page.get_title }}</h{{ header_level|default:3 }}>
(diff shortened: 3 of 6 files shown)


sveetch:
I've update the PR with new strategy with new placeholder 'excerpt'.

It still needs some tests to cover the new placeholder usage.

CHANGELOG.md
```diff
@@ -17,13 +17,16 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 - Add new page templates for various multiple columns and their own CSS to
   adjust every possible plugins to fit correctly;
 - Add theme variable to change checkmark color.
+- Add new placeholder 'excerpt' to Organization detail template;
```
jbpenrath, line 20: Too much CSS ^^
```suggestion
- Add new placeholder 'excerpt' to Organization detail template
```

CHANGELOG.md
```diff
@@ -17,13 +17,16 @@ Versioning](https://semver.org/spec/v2.0.0.html).
...
 
 ### Changed
 
 - Update frontend overriding system to allow to override any frontend module.
 - Improve React search suggestion field labels for screen readers.
 - Removed usage of deprecated Sass division '/' operator in favor of
   'math.div'.
+- Organization plugin variant 'row' now use excerpt if not empty else the
+  description with every markup removed;
```
jbpenrath, line 29:
```suggestion
  description with every markup removed
```

CHANGELOG.md
```diff
@@ -36,6 +39,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 - Fix section tile title with richtext and special character escaping.
 - Change blogpost detail template to display author even in published mode
   except if its placeholder is empty.
+- Fix missing styles for Organization plugin 'row' variant link wrapper;
```
jbpenrath, line 42:
```suggestion
- Fix missing styles for Organization plugin 'row' variant link wrapper
```

src/frontend/scss/objects/_organization_glimpses.scss
```diff
@@ -186,6 +186,25 @@
     flex-basis: 100%;
   }
 
+  & > a {
+    display: block;
+    flex-direction: row;
+    flex-wrap: nowrap;
+    align-content: flex-start;
+    align-items: center;
+    color: inherit;
```
jbpenrath, line 195: Sort line asc.

src/richie/apps/courses/templates/courses/cms/fragment_organization_glimpse.html
```diff
@@ -30,8 +30,23 @@
             {% endblockplugin %}
         </div>
         <div class="organization-{{ organization_variant }}__content">
-            <h{{ header_level|default:3 }} class="organization-{{ organization_variant }}__title" property="name">{{ organization_page.get_title }}</<h{{ header_level|default:3 }}>
-            <div class="organization-{{ organization_variant }}__excerpt" property="description">{% get_placeholder_plugins "description" organization_page as plugins or %}{% trans "Description" %}{% endget_placeholder_plugins %}{% blockplugin plugins.0 %}{{ instance.body|truncatewords_html:40|safe }}{% endblockplugin %}</div>
+            <h{{ header_level|default:3 }} class="organization-{{ organization_variant }}__title" property="name">{{ organization_page.get_title }}</h{{ header_level|default:3 }}>
+            <div class="organization-{{ organization_variant }}__excerpt" property="description">
+                {% get_placeholder_plugins "excerpt" organization_page as plugins or %}
+                    <p>{% trans "Excerpt" %}</p>
+                {% endget_placeholder_plugins %}
```
jbpenrath, line 37: When there is no plugin in this placeholder, `Excerpt` is displayed.

image

jbpenrath (commented):
I wonder if we do not encount the same accessiblity issue here than @Leimi [catches for course glimpse](https://github.com/openfun/richie/pull/1620).

src/richie/apps/courses/templates/courses/cms/fragment_organization_glimpse.html
sveetch: Yeah i reproduced for excerpt the same behavior than with description in this template variant but i can remove it if it is not wanted

jbpenrath:
> @jbpenrath
> 
> > I wonder if we do not encount the same accessiblity issue here than @Leimi #1620.
> 
> It surely seems the same issue. However for my PR i proposed first the "right" change to remove the link wrapper in profit on links on content but the current behavior have been required.

Yes, I do not expect you address this issue here. The purpose of this comment is mainly link this PR to the @Leimi to warn him when he will come back.

src/richie/apps/courses/templates/courses/cms/fragment_organization_glimpse.html
sveetch: Ok i've done it as you said.

jbpenrath (approved):
✅ LGTM

Leimi:
Hey, thanks for letting me know :)

Yeah the `<a>` wrapping everything is not ideal, we could improve that to prevent some confusion.

It's a bit less annoying than on the course glimpse though, because here there is just one paragraph in addition to the title.

PR #1633, opened 2022-03-11
💄(theme) add checkmark color variable

igobranco:
Add theme variable to change checkmark color.

On NAU we don't use the `indianred3` color, so we have detected that is not easy to change the check mark color when using a list on the field _What you will learn_ on the course page.

With this change, when we change the new theme variable `checkmark-list-decoration-color` to `blue`:
image

### DIFF
--- a/src/frontend/scss/colors/_theme.scss
+++ b/src/frontend/scss/colors/_theme.scss
@@ -348,6 +348,7 @@ $r-theme: (
     plan-title-color: r-color('firebrick6'),
     license-label-color: r-color('denim'),
     checkmark-list-decoration: url('../../richie/images/components/checkmark.svg'),
+    checkmark-list-decoration-color: r-color('indianred3'),
     view-more-runs-color: r-color('firebrick6'),
     run-catalog-visibility-hidden-logo: url('../../richie/images/catalog_visibility/hidden.svg'),
     run-catalog-visibility-course-only-logo:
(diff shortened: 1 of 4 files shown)


src/frontend/scss/components/templates/courses/cms/_course_detail.scss
```diff
@@ -273,10 +273,12 @@
...
             height: 0.8rem;
-            background-image: r-theme-val(course-detail, checkmark-list-decoration);
             background-repeat: no-repeat;
-            background-position: top left;
-            background-size: 100% 100%;
+            background-color: r-theme-val(course-detail, checkmark-list-decoration-color);
+            -webkit-mask: r-theme-val(course-detail, checkmark-list-decoration);
+            mask: r-theme-val(course-detail, checkmark-list-decoration);
+            -webkit-mask-size: cover;
+            mask-size: cover;
```
jbpenrath, line 281: The support of mask seems to be really poor currently, I don't think this is something we should use in a production environment right now.

src/frontend/scss/colors/_theme.scss
```diff
@@ -348,6 +348,7 @@ $r-theme: (
     plan-title-color: r-color('firebrick6'),
     license-label-color: r-color('denim'),
     checkmark-list-decoration: url('../../richie/images/components/checkmark.svg'),
+    checkmark-list-decoration-color: r-color('indianred3'),
```
jbpenrath, line 351: Do not forget to update the `UPGRADE.md` to mention to add this variable for those who overridden the `course-detail` theme.

jbpenrath (approved):

PR #1634, opened 2022-03-11
🐛(template_tags) fix placeholder template tags injecting in header

sampaccoud:
### Purpose

This is another approach for fixing the bug evidence in PR https://github.com/openfun/richie/pull/1626/files

Our "get_placeholder_plugins" template tag injects markup for frontend editing when rendering in edit mode. This causes a bug since we recently used them in the header for metas.

### Proposal

Refactor template tags to differentiate a template tag that only gets plugins (to simplify, I managed to merge the `get_placeholder_plugins` tag and the `get_page_plugins` tag under one tag) and a template tag that materializes a placeholder and injects markup for frontend editing.

A new tag `placeholder_as_plugins` is introduced for use cases in where the tag is the actual placeholder and should render editing tags.


sveetch (approved):
Seems you test cover every case against the markup, it should be ok for me

PR #1639, opened 2022-03-17
✨(frontend) add a PaymentButton

jbpenrath:
## Purpose

Add a payment button component which is in charge to create an order from Joanie then launch the right payment interface according to the payment provider used. Currently it supports Payplug and a dummy payment provider for development  purpose only.

## Proposal

- [x] Add a PaymentInterface in charge to lazy load the right payment component according to the payment information
- [x] Implement Payplug lightbox component
- [x] Implement a dummy payment component for dev purpose

### Classic payment with Payplug
Classic Payment

### One click payment with Payplug
One click

### DIFF
--- a/src/frontend/js/components/PaymentButton/index.spec.tsx
+++ b/src/frontend/js/components/PaymentButton/index.spec.tsx
@@ -1,0 +1,485 @@
+import { act, fireEvent, render, screen, waitFor } from '@testing-library/react';
+import fetchMock from 'fetch-mock';
+import { PropsWithChildren } from 'react';
+import { IntlProvider } from 'react-intl';
+import { hydrate, QueryClient, QueryClientProvider } from 'react-query';
+import {
+  AddressFactory,
+  ContextFactory as mockContextFactory,
+  CreditCardFactory,
+  OrderWithOneClickPaymentFactory,
+  OrderWithPaymentFactory,
+  PersistedClientFactory,
+  ProductFactory,
+  QueryStateFactory,
+} from 'utils/test/factories';
+import JoanieApiProvider from 'data/JoanieApiProvider';
+import { CourseCodeProvider } from 'data/CourseCodeProvider';
+import { PAYMENT_SETTINGS } from 'settings';
+import type * as Joanie from 'types/Joanie';
+import { OrderState } from 'types/Joanie';
+import createQueryClient from 'utils/react-query/createQueryClient';
+import PaymentButton from '.';
+
+jest.mock('utils/context', () => ({
+  __esModule: true,
+  default: mockContextFactory({
+    authentication: {
+      backend: 'fonzie',
+      endpoint: 'https://authentication.test',
+    },
+    joanie_backend: {
+      endpoint: 'https://joanie.test',
+    },
+  }).generate(),
+}));
+
+jest.mock('components/PaymentInterfaces');
+
+describe('PaymentButton', () => {
+  const formatPrice = (price: number, currency: string) =>
+    new Intl.NumberFormat('en', {
+      currency,
+      style: 'currency',
+    }).format(price);
+
+  const Wrapper = ({
+    client = new QueryClient(),
+    children,
+  }: PropsWithChildren<{ client?: QueryClient }>) => (
+    <IntlProvider locale="en">
+      <JoanieApiProvider>
+        <CourseCodeProvider code="00000">
+          <QueryClientProvider client={client}>{children}</QueryClientProvider>
+        </CourseCodeProvider>
+      </JoanieApiProvider>
+    </IntlProvider>
+  );
+
+  beforeEach(() => {
+    jest.useFakeTimers();
+  });
+
+  afterEach(() => {
+    fetchMock.restore();
+    jest.clearAllTimers();
+    jest.resetAllMocks();
+  });
+
+  it('should render a payment button', () => {
+    const product: Joanie.Product = ProductFactory.generate();
+
+    render(
+      <Wrapper>
+        <PaymentButton product={product} onSuccess={jest.fn()} />
+      </Wrapper>,
+    );
+
+    const $button = screen.getByRole('button', {
+      name: `Pay ${formatPrice(product.price, product.price_currency)}`,
+    }) as HTMLButtonElement;
+
+    // As a billing address is missing, button should be disabled
+    expect($button.disabled).toBe(true);
+  });
+
+  it('should render a payment button with a specific label when a credit card is provided', () => {
+    /*
+      If a credit card is provided, it seems that the payment should be a one click,
+      so the payment button label should mention this information.
+    */
+    const product: Joanie.Product = ProductFactory.generate();
+    const creditCard: Joanie.CreditCard = CreditCardFactory.generate();
+
+    const { rerender } = render(
+      <Wrapper>
+        <PaymentButton product={product} creditCard={creditCard.id} onSuccess={jest.fn()} />
+      </Wrapper>,
+    );
+
+    const $button = screen.getByRole('button', {
+      name: `Pay in one click ${formatPrice(product.price, product.price_currency)}`,
+    }) as HTMLButtonElement;
+
+    // As a billing address is missing, button should be disabled
+    expect($button.disabled).toBe(true);
+
+    const billingAddress: Joanie.Address = AddressFactory.generate();
+
+    rerender(
+      <Wrapper>
+        <PaymentButton
+          billingAddress={billingAddress}
+          product={product}
+          creditCard={creditCard.id}
+          onSuccess={jest.fn()}
+        />
+      </Wrapper>,
+    );
+
+    // As a billing address is given, the button should be active
+    expect($button.disabled).toBe(false);
+  });
+
+  it('should render an enabled payment button if a billing address is provided', () => {
+    const product: Joanie.Product = ProductFactory.generate();
+    const billingAddress: Joanie.Address = AddressFactory.generate();
+
+    render(
+      <Wrapper>
+        <PaymentButton billingAddress={billingAddress} product={product} onSuccess={jest.fn()} />
+      </Wrapper>,
+    );
+
+    const $button = screen.getByRole('button', {
+      name: `Pay ${formatPrice(product.price, product.price_currency)}`,
+    }) as HTMLButtonElement;
+
+    // As a billing address is given, the button should be active
+    expect($button.disabled).toBe(false);
+  });
+
+  it('should create an order then display the payment interface when user clicks on payment button', async () => {
+    const product: Joanie.Product = ProductFactory.generate();
+    const billingAddress: Joanie.Address = AddressFactory.generate();
+    const order: Joanie.OrderWithPaymentInfo = OrderWithPaymentFactory.generate();
+    const handleSuccess = jest.fn();
+
+    fetchMock
+      .post('https://joanie.test/api/orders/', order)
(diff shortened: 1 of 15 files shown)


src/frontend/js/components/PaymentButton/index.tsx
```diff
@@ -0,0 +1,226 @@
...
+}
+
+type PaymentInfo = Joanie.Payment & { order_id: string };
+type OneClickPaymentInfo = Joanie.PaymentOneClick & { order_id: string };
+
+enum ComponentStates {
+  IDLE = 'idle',
+  LOADING = 'loading',
+  ERROR = 'error',
+}
```
carofun, line 65: shouldn't we use a file outside like type/Joanie.ts or a one specific for the payment?

src/frontend/js/components/PaymentButton/index.tsx
```diff
@@ -0,0 +1,226 @@
...
+
+      setPayment(paymentInfos);
+    }
+  };
+
+  const handleSuccess = () => {
+    let round = 0;
+
+    const checkOrderValidity = async () => {
+      if (round >= 30) {
```
carofun, line 132:
```suggestion
      if (round >= 30) {
```
can we add const for this so it's easily configurable (to be used in the test as well)

kernicPanel (commented):
Some minor comments, otherwise LGTM :+1:

src/frontend/js/components/PaymentButton/index.spec.tsx
```diff
@@ -0,0 +1,451 @@
...
+    expect($button.disabled).toBe(false);
+
+    // - User clicks on pay button
+    await act(async () => {
+      fireEvent.click($button);
+    });
+
+    // - Route to create order should have been called
+    expect(fetchMock.calls()).toHaveLength(1);
+    expect(fetchMock.lastUrl()).toBe('https://joanie.test/api/orders/');
```
carofun, line 168: In general, shouldn't we check it has been called with expected params and use toHaveBeenCalledWith

src/frontend/js/components/PaymentInterfaces/index.spec.tsx
```diff
@@ -0,0 +1,67 @@
...
+    expect(container.childElementCount).toEqual(0);
+  });
+
+  it('should render the payplug lightbox when provider is Payplug', async () => {
+    const payment = PaymentFactory.afterGenerate((p: Joanie.Payment) => ({
+      ...p,
+      provider: PaymentProviders.PAYPLUG,
+    })).generate();
+    render(<PaymentProvider onSuccess={noop} onError={noop} {...payment} />);
+
```
carofun, line 54: just to be certain I would make sure on both tests that onError hasn't been called

carofun (commented):
Commit message suggestion => purpose => purposes (and spaces are weirdo :-))
Add a payment button component **that** is in charge to create an order from Joanie
 then launch the right payment interface according to the payment provider used. **Currently,** it supports Payplug and a dummy payment provider for development **purposes** only.
 
 I would advise to split in three commits as you suggest three steps in your proposal
Add a PaymentInterface in charge to lazy load the right payment component according to the payment information
 Implement Payplug lightbox component
 Implement a dummy payment component for dev purpose

src/frontend/js/components/PaymentButton/index.tsx
jbpenrath: Oh yes ! Good catch, this is something I forgot to do.

src/frontend/js/components/PaymentButton/index.tsx
```diff
@@ -0,0 +1,226 @@
+import { useEffect, useMemo, useRef, useState } from 'react';
+import { defineMessages, FormattedMessage, useIntl } from 'react-intl';
+import type * as Joanie from 'types/Joanie';
+import { OrderState } from 'types/Joanie';
+import type { Nullable } from 'types/utils';
+import { useCourseCode } from 'data/CourseCodeProvider';
+import { Spinner } from 'components/Spinner';
+import PaymentInterface from 'components/PaymentInterfaces';
+import { useOrders } from 'hooks/useOrders';
+import { useJoanieApi } from 'data/JoanieApiProvider';
```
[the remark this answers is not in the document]
jbpenrath: I do it manually right now, but I note that it could be interesting to use the `sort-imports` eslint rule.

src/frontend/js/components/PaymentButton/index.tsx
```diff
@@ -0,0 +1,226 @@
...
+
+const messages = defineMessages({
+  pay: {
```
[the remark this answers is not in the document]
jbpenrath: We could use `sort-keys` to warn developer to sort objects keys alphabetically.

jbpenrath:
> Commit message suggestion => purpose => purposes (and spaces are weirdo :-)) Add a payment button component **that** is in charge to create an order from Joanie then launch the right payment interface according to the payment provider used. **Currently,** it supports Payplug and a dummy payment provider for development **purposes** only.
> 
> I would advise to split in three commits as you suggest three steps in your proposal Add a PaymentInterface in charge to lazy load the right payment component according to the payment information Implement Payplug lightbox component Implement a dummy payment component for dev purpose

Good idea, I will split this commit in 3 once review will be finished

src/frontend/js/components/PaymentButton/index.spec.tsx
```diff
@@ -0,0 +1,451 @@
...
+  );
+
+  beforeEach(() => {
+    jest.useFakeTimers();
+  });
+
+  afterEach(() => {
+    fetchMock.restore();
+    jest.clearAllTimers();
+    jest.useRealTimers();
```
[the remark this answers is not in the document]
jbpenrath: A residue

src/frontend/js/components/PaymentButton/index.spec.tsx
```diff
@@ -0,0 +1,451 @@
...
+  it('should render a payment button', () => {
+    const product = ProductFactory.generate();
+
+    render(
+      <Wrapper>
+        <PaymentButton product={product} onSuccess={jest.fn()} />
+      </Wrapper>,
+    );
+
+    const $button = screen.getByRole('button', {
```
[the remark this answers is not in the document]
jbpenrath: This is my own convention 😅 I prefix variables with `$` ~in glory of jQuery~ when I manipulate a DOM Element.

src/frontend/js/components/PaymentButton/index.tsx
jbpenrath: These states are related to the payment button that's why I decided to let this declaration here.

src/frontend/js/components/PaymentButton/index.spec.tsx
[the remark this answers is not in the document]
carofun: great! got it thanks for the feedback

src/frontend/js/components/PaymentButton/index.spec.tsx
jbpenrath: This is true for mock using `jest.fn()` but in the case of `fetchMock` it gives access to a method to retrieve all calls executed and some methods to retrieve informations about the last call.

carofun: I mean, to do a full test, I would test the post sent to the API are correct, I don't see this kind of check-in the test.

jbpenrath: Oh ok ! Yes I should be able to check the body structure of post requests.

jbpenrath: @carofun I add assertion to check body content of each post request.

src/frontend/js/components/PaymentButton/index.spec.tsx
```diff
@@ -0,0 +1,472 @@
...
+      name: `Pay ${formatPrice(product.price, product.price_currency)}`,
+    }) as HTMLButtonElement;
+
+    // As a billing address is missing, button should be disabled
+    expect($button.disabled).toBe(true);
+  });
+
+  it('should render a payment button with a specific label when a credit card is provided', () => {
```
carofun, line 86:
```suggestion
  it('should render a payment button with a specific label when a credit card is provided', () => {
```
you are testing two things the label and the fact the button is not enabled because the address is not provided.
For this test, I would like to be sure this specific button "Pay in one click" can be enabled, to make sure we can pay :-)

If you want to test the disabled/enabled maybe it could be dissociated

src/frontend/js/components/PaymentButton/index.spec.tsx
```diff
@@ -0,0 +1,472 @@
...
+      <Wrapper>
+        <PaymentButton billingAddress={billingAddress} product={product} onSuccess={jest.fn()} />
+      </Wrapper>,
+    );
+
+    const $button = screen.getByRole('button', {
+      name: `Pay ${formatPrice(product.price, product.price_currency)}`,
+    }) as HTMLButtonElement;
+
+    // As a billing address is missing, button should be disabled
```
carofun, line 122:
```suggestion
    // As a billing address is missing, the button should be active
```

src/frontend/js/components/PaymentButton/index.spec.tsx
carofun: I like it!

src/frontend/js/components/PaymentButton/index.spec.tsx
```diff
@@ -0,0 +1,472 @@
...
+    expect(handleSuccess).toHaveBeenCalledTimes(1);
+
+    // - And poller should be stopped
+    await act(async () => {
+      jest.runOnlyPendingTimers();
+    });
+    expect(fetchMock.calls()).toHaveLength(3);
+  });
+
+  it('should abort the order if payment does not succeeded after a given delay', async () => {
```
carofun, line 309:
```suggestion
  it('should abort the order if payment does not succeed after a given delay', async () => {
```

src/frontend/js/components/PaymentButton/index.spec.tsx
```diff
@@ -0,0 +1,472 @@
...
+    expect(fetchMock.calls()).toHaveLength(PAYMENT_SETTINGS.pollLimit);
+    expect(fetchMock.lastUrl()).toBe(`https://joanie.test/api/orders/${order.id}/abort/`);
+    expect(JSON.parse(fetchMock.lastOptions()!.body!.toString())).toEqual({
+      payment_id: order.payment_info.payment_id,
+    });
+
+    // - An error message should be displayed
+    screen.getByText('An error occurred during payment. Please retry later.');
+    await act(async () => {
+      jest.runAllTimers();
```
carofun, line 396: Please can you explain why you are doing this at the end of the test?
```suggestion
    await act(async () => {
      jest.runAllTimers();
```

jbpenrath: This is a residue of a vicious bug

src/frontend/js/components/PaymentButton/index.spec.tsx
jbpenrath: I completed the test to check that button is enabled when an address is provided.

carofun (commented):
I think it's ok but I would like to validate with the rebase and the 3 commits if you don't mind

jbpenrath:
@carofun I think it's ready for the final review 🤞

carofun (approved):
Just minors returns, strange spaces in commits that you could erase and order of imports files.

PR #1643, opened 2022-03-21
Feat/sale tunnel

jbpenrath:
## Purpose

Create a sale tunnel to allow a user to purchase a product relying on a course. As user must provide a billing address for the payment, a form to manage this address has been created.

## Address Management Form
Addresses

## A three steps sale tunnel
Sale tunnel

## Proposal

- [x] Create a form to manage user billing addresses
- [x] Create a sale tunnel in three steps (validation, payment, resume)


src/frontend/js/components/AddressesManagement/index.tsx
```diff
@@ -0,0 +1,519 @@
...
+      </button>
+      {error && (
+        <Banner
+          message={intl.formatMessage(messages.error, { action: error })}
+          type={BannerType.ERROR}
+          rounded
+        />
+      )}
+      {addresses.items.length > 0 ? (
+        <section className="address-registered">
```
carofun, line 336: component is a bit long, wouldn't it be possible to have one with registered addresses and another one to edit address

carofun (commented):
Very first reading, few comments in general import orders and keys in the different files :-)

src/frontend/js/components/AddressesManagement/index.tsx
```diff
@@ -0,0 +1,519 @@
...
+import countries from 'i18n-iso-countries';
+import { Fragment, useEffect, useMemo, useState } from 'react';
+import { useForm } from 'react-hook-form';
+import { defineMessages, FormattedMessage, useIntl } from 'react-intl';
+import Banner, { BannerType } from 'components/Banner';
+import { useAddresses } from 'hooks/useAddresses';
+import type * as Joanie from 'types/Joanie';
+import { Maybe } from 'types/utils';
+import { confirm } from 'utils/indirection/window';
+import { CheckboxField, SelectField, TextField } from '../Form';
```
kernicPanel, line 11:
```suggestion
import { CheckboxField, SelectField, TextField } from 'components/Form';
```

src/frontend/js/utils/api/joanie.ts
```diff
@@ -76,14 +76,14 @@ function fetchWithJWT(routes: RequestInfo, options: RequestInit = {}) {
...
 
     options.headers = headers;
-
-    return fetch(routes, options);
   } catch (error) {
     throw new Error(
       `Joanie requires JWT Token to fetch data, but the configured authentication
       api does not contains a method \`accessToken\` to retrieve this information.`,
     );
   }
```
kernicPanel, line 84: Maybe only try catch this line ?
```const accessToken = AuthenticationApi!.accessToken!();```

**Review by kernicPanel — APPROVED** [review]
Only minor comments, as always… :clap:

File: src/frontend/js/utils/api/joanie.ts (hunk 3 above)
Comment (jbpenrath) [reply src/frontend/js/utils/api/joanie.ts:84]: Oh yes ! Good catch !

File: src/frontend/js/components/AddressesManagement/index.tsx (hunk 1 above)
Comment (jbpenrath) [reply src/frontend/js/components/AddressesManagement/index.tsx:266]: @carofun I moved the edition form into an `AddressForm` component then the registered address into a `RegisteredAddress` component. I keep the logic and methods to update address within the main `AddressesManagement`.

**Review by kernicPanel — APPROVED** [review]

File: src/frontend/js/components/AddressesManagement/index.tsx [hunk 4]
```diff
@@ -0,0 +1,309 @@
...
+  };
+
+  useEffect(() => {
+    setError(undefined);
+  }, [editedAddress]);
+
+  return (
+    <div className="AddressesManagement">
+      <button className="button button--tertiary" onClick={handleClose}>
+        <svg className="button__icon" role="img">
```
Leimi, line 253: You could add an `aria-hidden` attribute here to make sure no screen reader announce anything.

src/frontend/js/components/RegisteredAddress/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+        data-testid={`address-${address.id}-details`}
+        className="registered-addresses-item__address"
+      >
+        {address.first_name}&nbsp;{address.last_name}
+        <br />
+        {address.address} {address.postcode} {address.city}, {address.country}
+      </address>
+      <p className="registered-addresses-item__actions">
+        <button
+          title={intl.formatMessage(messages.selectTitleButton, { title: address.title })}
```
Leimi, line 82: I guess this is here mostly for screen readers in order to have a button with more context?

That's a great idea but watch out when using a `title` attribute. For screen reader it's used more as a fallback than anything. So it will not necessarily be announced if the button has content. And if it's announced, it can be announced _in addition_ to the button content. Or not. Behavior is not really consistent between screen readers.

I'd advise to rather use `aria-label` here. `aria-label` will be used as the accessible name of the element, we are sure this will be announced.

If you want to still get a mouse tooltip, I'd leave `title` as is, and add an `aria-label` with the exact same content as the title. For example, NVDA will announce the title in addition to the label, but will announce it only once if it detects it's the same content.

src/frontend/js/components/SaleTunnelStepValidation/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+        </h4>
+      </header>
+      <ol className="SaleTunnelStepValidation__product-detail-list">
+        {product.target_courses.map((course) => (
+          <li
+            key={`SaleTunnelStepValidation__product-detail-row--${course.code}`}
+            className="SaleTunnelStepValidation__product-detail-row product-detail-row product-detail-row--course"
+          >
+            <span className="product-detail-row__icon">
+              <svg role="img">
```
Leimi, line 49: As a general advice for svg icons, I'd either:

- add an `aria-hidden="true"` attribute if it's decorative _or_ if the meaning is already conveyed by visible text nearby
- add an `aria-label` attribute and a `<title>` child if the icon has meaning and only the icon carries it.

src/frontend/js/components/RegisteredCreditCard/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+  const inputId = useMemo(() => `registered-credit-card-${creditCard.id}`, [creditCard.id]);
+
+  return (
+    <label className="registered-credit-card">
+      <p className="form-field">
+        <input
+          aria-label={intl.formatMessage(messages.inputAriaLabel, {
+            title: creditCard.title,
+            selected,
+          })}
```
Leimi, line 43: Maybe adding an `id` on the `registered-credit-card__infos` div and setting it as `aria-describedby` might be good here?

With a screen reader, when in a form, it's easy to miss info surrounding the form inputs if they are not actually tied to inputs in the HTML. Because in a form, you use a "form navigation" mode, mostly using the tab key, to move from input to input, contrary to the usual arrow key to move from any content to any content.

So here, all the credit card info can be easily missed. Adding the info in the `aria-describedby` could help. It would be announced as additional info, after the label.

src/frontend/js/components/SaleTunnelStepPayment/_styles.scss
```diff
@@ -0,0 +1,180 @@
...
+    display: flex;
+    justify-content: space-between;
+    padding: 0 3rem;
+    margin-bottom: 3rem;
+    width: 100%;
+
+    @include media-breakpoint-down(md) {
+      $block-gutter: 4rem;
+      background: r-color('azure2');
+      border-radius: 6px;
```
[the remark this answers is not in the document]
jbpenrath: Read my comment above

src/frontend/js/components/SaleTunnelStepPayment/index.tsx
```diff
@@ -0,0 +1,275 @@
...
+
+  return (
+    <section className="SaleTunnelStepPayment">
+      <section className="SaleTunnelStepPayment__block">
+        <header className="SaleTunnelStepPayment__block__header">
+          <h5 className="SaleTunnelStepPayment__block__title">
+            <FormattedMessage {...messages.resumeTile} />
+          </h5>
+        </header>
+        <div className="SaleTunnelStepPayment__block--product">
```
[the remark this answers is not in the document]
jbpenrath: In Richie, we are using [BEM](http://getbem.com/introduction/) naming methodology. By convention we use `__` for element and `--` for modifier.

src/frontend/js/components/SaleTunnel/index.tsx
```diff
@@ -0,0 +1,123 @@
...
+        next: null,
+        onExit: () => {
+          course.methods.refetch();
+          orders.methods.refetch();
+          handleModalClose();
+        },
+      },
+    },
+  };
+  const { step, next, reset } = useStepManager(manifest);
```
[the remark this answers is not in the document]
jbpenrath: 🙂

src/frontend/js/components/RegisteredAddress/index.tsx
jbpenrath: Many thanks for this explanation! I will use an aria-label in this case, for mouse user, a tooltip is not necessary in this case.

src/frontend/js/components/SaleTunnelStepValidation/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+export const SaleTunnelStepValidation = ({ product, next }: SaleTunnelStepValidationProps) => {
+  const isOpenedCourseRun = (courseRun: Joanie.CourseRun) =>
+    courseRun.state.priority <= Priority.FUTURE_NOT_YET_OPEN;
+
+  return (
+    <section className="SaleTunnelStepValidation">
+      <header className="SaleTunnelStepValidation__header">
+        <h3 className="SaleTunnelStepValidation__title">{product.title}</h3>
+        <h4 className="SaleTunnelStepValidation__price">
```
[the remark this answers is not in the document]
Leimi: After seeing it in the UI I'd insist we could do better here by replacing this h4 by another tag (whatever: div, p, strong..), as it's more of a visual title than a title sectioning content.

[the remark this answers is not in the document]
Leimi: I think we could improve headings a little further, but we can do this after merging I guess. It's an issue that the modal begins with `h6` with the steps, then bumps back to `h3`.

Since we are in a modal that has its own isolated context, the step title would be better as an `h1` (or `h2`, a using multiple `h1` is always a big debate, but in the end the important part is to have a coherent headings hierarchy), followed then by the `SaleTunnelStepValidation__title` as h2, etc.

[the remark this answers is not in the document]
jbpenrath: Right, I will replace this by a paragraph.

src/frontend/js/components/SaleTunnelStepValidation/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+          <li
+            key={`SaleTunnelStepValidation__product-detail-row--${course.code}`}
+            className="SaleTunnelStepValidation__product-detail-row product-detail-row product-detail-row--course"
+          >
+            <span className="product-detail-row__icon">
+              <svg role="img">
+                <use href="#icon-check" />
+              </svg>
+            </span>
+            <h5 className="product-detail-row__title">{course.title}</h5>
```
Leimi, line 53: After the price heading change, this should then be an `h4`

src/frontend/js/components/SaleTunnelStepPayment/index.tsx
```diff
@@ -0,0 +1,275 @@
...
+          <h6 className="SaleTunnelStepPayment__block--product__title">{product.title}</h6>
```
Leimi, line 160: Seems weird to me that this is a heading. It doesn't really section content, it's content itself.

src/frontend/js/components/SaleTunnelStepPayment/index.tsx
```diff
@@ -0,0 +1,275 @@
...
+          {user.email ? (
+            <p className="SaleTunnelStepPayment__block--buyer__email">{user.email}</p>
+          ) : null}
+          <header className="SaleTunnelStepPayment__block--buyer__address-header">
+            <h6>
+              <FormattedMessage {...messages.userBillingAddressFieldset} />
+            </h6>
+            {addressesItems.length > 0 && (
+              <button
+                aria-hidden="true"
```
Leimi, line 189: Watch out for this, I guess the aria-hidden was meant on the icon inside the button?

src/frontend/js/components/SaleTunnelStepPayment/index.tsx
```diff
@@ -0,0 +1,275 @@
...
+        </div>
+      </section>
+      <section className="SaleTunnelStepPayment__block">
+        <header className="SaleTunnelStepPayment__block__header">
+          <h5 className="SaleTunnelStepPayment__block__title">
+            <FormattedMessage {...messages.userTile} />
+          </h5>
+        </header>
+        <div className="SaleTunnelStepPayment__block--buyer">
+          <h6 className="SaleTunnelStepPayment__block--buyer__name">
```
Leimi, line 177: I'd also argue this is more content that something sectioning content and remove the heading tag here.

src/frontend/js/components/RegisteredCreditCard/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+        <label htmlFor={inputId} className="form-field__label">
+          <span className="form-field__checkbox-control">
+            <svg role="img" aria-hidden="true">
+              <use href="#icon-check" />
+            </svg>
+          </span>
+        </label>
+      </p>
+      <div id={`credit-card-${creditCard.id}-infos`} className="registered-credit-card__infos">
+        <h6 className="registered-credit-card__name">{creditCard.title || creditCard.brand}</h6>
```
Leimi, line 60: As the other things, I'd also say here that maybe using a heading is overkill, since we already have the "Your registered credit card" just before. I'd say a `strong` tag is more fitting.

src/frontend/js/components/SaleTunnelStepPayment/index.tsx
jbpenrath: Oh yes 🤦‍♂️

src/frontend/js/components/SaleTunnelStepValidation/index.tsx
Leimi: Allowing myself to bump this, don't know if you still have this in mind ;)

jbpenrath:
https://github.com/openfun/richie/pull/1643#discussion_r835154976

@Leimi About this comment, I don't think it is a good idea to use `h1` or `h2` within `StepBreadcrumb` as this component aims to be reusable anywhere. Instead, I would remove the use of heading elements and replace it by an `strong` element. What do you think?

Leimi:
> [#1643 (comment)](https://github.com/openfun/richie/pull/1643#discussion_r835154976)
> 
> @Leimi About this comment, I don't think it is a good idea to use `h1` or `h2` within `StepBreadcrumb` as this component aims to be reusable anywhere. Instead, I would remove the use of heading elements and replace it by an `strong` element. What do you think?

Yeah good idea to remove the heading. Not sure I'd use _strong_ though, but it's not that impactful for a11y. Most screen readers don't change how they read normal and strong text. The important thing is that the screen reader user know what is the active step.

Leimi (approved):
Great :) thanks

PR #1644, opened 2022-03-22
✨(core) add site name in <title> everywhere

Leimi:
## Purpose

Improve a11y (and seo) by always having the website's name in the page `<title>`.

## Proposal

By default for now, the `<title>` only tells us about the current page; it does
not tell us about what website we are on.

Adding the site name after the page name mostly improves two things:

- accessibility. Knowing what is the current site is a great
improvement. Without it, if I have multiple tabs opened, all related to
learning stuff, and a few tabs don't have their website name in the
title, it can be cumbersome to navigate with a screen-reader and
remember what tab is what website.
- SEO

Stuff done to implement this:

- [x] add a new `site_title` block that contains the site name by default, just after the existing `head_title` block. This allows us to change/disable this behavior on a case by case basis on any page we want.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,9 @@ Versioning](https://semver.org/spec/v2.0.0.html).
   adjust every possible plugins to fit correctly;
 - Add theme variable to change checkmark color.
 - Add new placeholder 'excerpt' to Organization detail template
+- Add the website's name by default in every page title, that can be changed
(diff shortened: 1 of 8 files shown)


Leimi:
If you agree on merging this, I'd love to have some guidance on where to add a test for this if you think it's necessary.

jbpenrath (approved):
I did not notice this practice before but after some check I see it everywhere ^^

About test, it's ok for me to just update existing test.

PR #1645, opened 2022-03-22
♿️(styles) improve the offscreen class

Leimi:
Change the implementation of the offscreen class to match the one from bootstrap that is battle-tested.

The main idea is to stop using `left: -100000px` that can cause scrolling "flicker" or weird focus styles when using a screen reader, as that can be annoying for sighted people that use a screen reader.

### DIFF
--- a/src/frontend/scss/generic/_accessibility.scss
+++ b/src/frontend/scss/generic/_accessibility.scss
@@ -2,23 +2,20 @@
 // to users that need it.
 
 // Screen reader only content
-.offscreen {
+// code taken from orange's fork of bootstrap
+// https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/blob/v5.1.3/scss/mixins/_visually-hidden.scss#L8
+.offscreen,
+.offscreen-focusable:not(:focus) {
   position: absolute;
-  left: -10000px;
-  top: auto;
   width: 1px;
   height: 1px;
+  padding: 0;
+  margin: -1px;
   overflow: hidden;
-
-  // Screen reader toggleable function, including offscreen navigation
-  &:active,
-  &:focus {
-    left: 5px;
-    top: 5px;
-    width: auto;
-    height: auto;
-    overflow: auto;
-  }
+  clip: rect(0, 0, 0, 0);
+  clip-path: inset(50%);
+  white-space: nowrap;
+  border: 0;
 }
 
 // Skip to content button styles for it to be rendered on link focus
@@ -28,6 +25,9 @@
   background-color: r-color('midnightblue');
   padding: 1rem;
   z-index: 100000;
+  position: absolute;
+  left: 5px;
+  top: 5px;
 
   &:active,
   &:hover,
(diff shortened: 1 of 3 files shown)


src/frontend/scss/generic/_accessibility.scss
```diff
@@ -2,23 +2,19 @@
...
+.offscreen-focusable:not(:focus) {
+  position: absolute !important;
+  width: 1px !important;
+  height: 1px !important;
+  padding: 0 !important;
+  margin: -1px !important;
+  overflow: hidden !important;
+  clip: rect(0, 0, 0, 0) !important;
+  white-space: nowrap !important;
+  border: 0 !important;
```
jbpenrath, line 17: https://css-tricks.com/comparing-various-ways-to-hide-things-in-css/#aa-method-5-the-visually-hidden-class

From this article, I think you should add `clip-rect` below `clip` as this last one has been deprecated in favor of the first.

Furthermore, I think you should remove `!important`

src/frontend/scss/generic/_accessibility.scss
```diff
@@ -28,6 +24,9 @@
   background-color: r-color('midnightblue');
   padding: 1rem;
   z-index: 100000;
+  position: absolute;
```
jbpenrath, line 27: Is `fixed` better in this case ?

sandroscosta: Because it should only appear if keyboard foccused, `absolute` should be the way to go. (IMHO)

src/frontend/scss/generic/_accessibility.scss
sandroscosta: I'm with @jbpenrath on the `!important` part. It makes things more complex to debug if we ever override something in the future.

Leimi: You meant `clip-path` I guess. There were a few discussions on this in the bootstrap project, and not using it is intentional, even when knowing it's deprecated, i'll let you check https://github.com/twbs/bootstrap/pull/25197

Leimi: Will remove the `!important` bits ;)

src/frontend/scss/generic/_accessibility.scss
Leimi: It doesn't change much actually. Right now `absolute` is kinda better because it doesn't mess with the django toolbar when editing a page. But for the actual use-case without the toolbar, since the link is the first on the page, shown at the very top, absolute or fixed, whatever.

src/frontend/scss/generic/_accessibility.scss
Leimi: So what do you think @jbpenrath about the clip-path issue? Seems to have noticeable perfs issues because of a Chrome bug. My choice would be to leave it as is, but as you wish, let me know.

jbpenrath: I didn't know about that. Since february 2018, the problem seems to be fixed. I tried an example from my chromium based browser and there is no more painting issue. https://bugs.chromium.org/p/chromium/issues/detail?id=611257

PR #1646, opened 2022-03-22
🚸(search) transform pagination buttons in pagination anchors

Leimi:
Pagination buttons behave like links: they update the url and make a new request. So we could use actual links instead of buttons.

This now allows users to open search pages in new tabs from the pagination links.

### DIFF
--- a/src/frontend/js/components/PaginateCourseSearch/index.tsx
+++ b/src/frontend/js/components/PaginateCourseSearch/index.tsx
@@ -1,5 +1,7 @@
 import { Fragment } from 'react';
 import { defineMessages, FormattedMessage, useIntl } from 'react-intl';
+import { parse, stringify } from 'query-string';
+import { location } from 'utils/indirection/window';
 
 import { CourseSearchParamsAction, useCourseSearchParams } from 'data/useCourseSearchParams';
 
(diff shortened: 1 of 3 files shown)


src/frontend/js/components/PaginateCourseSearch/index.tsx
```diff
@@ -123,15 +125,20 @@ export const PaginateCourseSearch = ({ courseSearchTotalCount }: PaginateCourseS
...
                     className="pagination__page-number"
-                    onClick={() =>
+                    onClick={(event) => {
+                      event.preventDefault();
                       dispatchCourseSearchParamsUpdate({
                         // Pages are 1-indexed, we need to 0-index them to calculate the correct offset
                         offset: String((page - 1) * limit),
                         type: CourseSearchParamsAction.pageChange,
-                      })
-                    }
```
jbpenrath, line 134: I wonder if it's not overkill but we could keep the link default behavior when user presses shortcut keys.

We could do nothing if user press ctrl, shift or meta key to open link in a new tab or window.

Ctrl + Click : Open a new tab on Window (On Mac OS, it behaves like a right click so click event is not dispatched)
Meta + Click : Open a new tab on Mac OS
Shift + Click : Open a new window on Mac OS and Window 

I don't know the behaviour on Linux, so may be my suggestion becomes too complicated to support every cases.

```
onClick={(event) => {
            if(!event.metaKey && !event.ctrlKey && !event.shiftKey) {
                      event.preventDefault();
                      dispatchCourseSearchParamsUpdate({
                        // Pages are 1-indexed, we need to 0-index them to calculate the correct offset
                        offset: String((page - 1) * limit),
                        type: CourseSearchParamsAction.pageChange,
                      })
                    }
            }
```

Leimi: Oh sure :)

src/frontend/js/components/PaginateCourseSearch/index.tsx
```diff
@@ -123,15 +124,20 @@ export const PaginateCourseSearch = ({ courseSearchTotalCount }: PaginateCourseS
                 </li>
               ) : (
                 <li className="pagination__item">
-                  <button
+                  <a
```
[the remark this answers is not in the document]
jbpenrath: It's fine for me !

jbpenrath (approved):
✅ LGTM

PR #1653, opened 2022-03-25
♿️ Improve category detail page accessibility

Leimi:
## Purpose

Make sure the category page is understandable with a screen reader.

## Proposal

Globally the page is already great in terms of accessibility. There are a still a few things that we can improve. They are of the same kind as on the course page: headings hierarchy, big anchors wrapping glimpses as a whole, a few stuff to reorder in the DOM…

__________

This is still work-in-progress, opening this for the curious.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ Versioning](https://semver.org/spec/v2.0.0.html).
   to succeed
 - Improve Sale Tunnel UX
 - Improve sale tunnel theme overriding
+- Improve overall accessibility on category page, especially for screen reader
+  users
 
 ### Fixed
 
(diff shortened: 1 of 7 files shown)


Leimi:
This is composed of a few tiny changes. The most notable one is this one that may impact the visible UI: https://github.com/openfun/richie/pull/1653/commits/b7bcacf8e6cff8f9c445ca28b9ee0a74bc2ee050. 

This idea is the same as on a course glimpse, to stop having a big `<a>` wrapping a whole blogpost glimpse, to prevent confusion when using a screen reader.

For now I kept the visual behavior that consisted on changing colors of the whole glimpse when hovering with the mouse. It may bit a bit weird though, because now, if you hover the blogpost glimpse body text, it still changes color as before, but it's not clickable anymore:


https://user-images.githubusercontent.com/221253/165546549-ec3bd177-ec0e-434a-ae83-c8cc6b8f8d92.mp4

Leimi:
Actually I noticed there were a few common stuff that I found on multiple pages, so I'd rather open a PR concerning all the pages, will be easier.

PR #1655, opened 2022-03-25
💄(course_glimpse) fix style with narrow org logo

### LINKED ISSUE #1654: Course glimpse with incorrect style when Organization logo is too narrow
## Bug Report
When the organization logo is too narrow, the course glimpse is having the organization logo cropped.

**Screenshot**
For with an organization with an image that has bigger height and width. 
image

**Possible Solution**
Add a `max-height: 100%` on the organization logo on course glimpse.

I would send a PR with this fix suggestion.

igobranco:
Fix course glimpse layout when organization logo is too narrow
Closes #1654

### DIFF
--- a/src/frontend/scss/objects/_course_glimpses.scss
+++ b/src/frontend/scss/objects/_course_glimpses.scss
@@ -264,6 +264,13 @@ $course-glimpse-content-padding-sides: 0.7rem !default;
     top: -3.53rem;
     width: 4.3rem;
     z-index: 0;
+
+    img {
+      object-fit: contain;
+      object-position: center;
+      height: 100%;
+      width: 100%;
+    }
   }
 
   &__media {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -72,6 +72,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 - Change blogpost detail template to display author even in published mode
   except if its placeholder is empty.
 - Fix missing styles for Organization plugin 'row' variant link wrapper
+- Fix course glimpse layout when organization logo is too narrow
 - Fix course run deletion when translation title is empty
 - Reordered course glimpse text order in the DOM for better screen reader
   support.


igobranco:
@jbpenrath can you review this PR ?

src/frontend/scss/objects/_course_glimpses.scss
```diff
@@ -264,6 +264,10 @@ $course-glimpse-content-padding-sides: 0.7rem !default;
     top: -3.53rem;
     width: 4.3rem;
     z-index: 0;
+
+    img {
+      max-height: 100%;
+    }
```
jbpenrath, line 270: In order to cover all layout issues here, I think `object-fit` is a better workaround. What do you think?

```suggestion
    img {
      object-fit: contain;
      object-position: center;
    }
```

igobranco: I didn't know that property...

To make it work, the `img` tag also needs an explicit `height` to have the same effect.
```css
    img {
      object-fit: contain;
      object-position: center;
      height: 4rem;
    }
```

To cover other layout issues, I think it is better to include the width.
```css
    img {
      object-fit: contain;
      object-position: center;
      height: 4rem;
      width: 4rem;
    }
```

src/frontend/scss/objects/_course_glimpses.scss
```diff
@@ -264,6 +264,13 @@ $course-glimpse-content-padding-sides: 0.7rem !default;
     top: -3.53rem;
     width: 4.3rem;
     z-index: 0;
+
+    img {
+      object-fit: contain;
+      object-position: center;
+      height: 4rem;
+      width: 4rem;
```
jbpenrath, line 272: You could use percentage value instead to force image to fill its container. Just more maintainable.
```suggestion 
      height: 100%;
      width: 100%;
```

jbpenrath (approved):
Just change width and height values.
Otherwise looks good to me ✅

Issue #1661, opened 2022-03-30
Language switch while on the search page


sampaccoud:
## Bug Report

**Problematic behavior/code**
Language switch does not work good on the search engine

**Expected Behavior**
Switching language while on the search page should change the language in which results are shown

**Steps to Reproduce**
1. Go to the search page in french for example
2. See that results are in french
3. Switch to english
4. See that results are still in french

**Environment**
- Richie version: 2.11.0
- Platform: Docker

PR #1662, opened 2022-03-30
🚑(frontend) make useCourseSearch hook locale sensible

jbpenrath:
## Purpose

The `useCourseSearch` is using `useQuery` under the hood but its query key was
not localized so if user changes the language, search results was not refreshed
until current data are stale.

## Proposal

- [x] Localized the query key of `useCourseSearch` 

Fixes https://github.com/openfun/richie/issues/1661


sampaccoud (approved):
:muscle:

PR #1664, opened 2022-03-31
🍱(fonts) self-host webfonts hind and montserrat

jbpenrath:
## Purpose

Recently, Google Fonts has been considered violating GDPR. Furthermore,
since modern browsers disable cross domain caching, Google Fonts has lost some
of its interest. So today, it appears that self-host our webfonts is a
good choice.

## Proposal

- [x] Self host our fonts (Hind & Montserrat)


src/frontend/scss/settings/_fonts.scss
```diff
@@ -1,8 +1,48 @@
-//
-// https://fonts.google.com/?query=Montserrat&selection.family=Hind:wght@400;600|Montserrat:wght@300;700;800&sidebar.open
```
sampaccoud, line 2: You. Hanged the weight? Same question for hind?

jbpenrath: Yes, I do not import all font variants only those we were using previously.

**Hind**:
400: regular
600: semibold

**Montserrat**:
300: light
500: regular
700: bold
800: extrabold

sampaccoud: 👍

sampaccoud (approved):

PR #1665, opened 2022-04-01
🐛(cache) fix course page cache

igobranco:
Added a middleware `LimitCacheByCourseDates` to prevent cache of old
information on course pages.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,6 +62,8 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Fixed
 
+- Added a middleware `LimitCacheByCourseDates` to prevent cache of old
+  information on course pages.
 - Make useCourseSearch hook locale sensitive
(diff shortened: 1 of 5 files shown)


igobranco:
Fixed linting problem.

sampaccoud:
I understand the problem but I don't hink this can be a solution as it generates queries on each request so it defeats the purpose of cache.. Maybe the cache expiration can be limited when setting it?

igobranco:
> I understand the problem but I don't hink this can be a solution as it generates queries on each request so it defeats the purpose of cache.. Maybe the cache expiration can be limited when setting it?

You are right. It would be better to implement the same ideia but only when adding the page to the cache.
Do you have an ideia on how and where we could add this feature?

igobranco:
@sampaccoud  I've found the method `get_cache_expiration` on https://docs.django-cms.org/en/support-3.1.x/reference/plugins.html but I think it won't work because we need to change the ttl of the course template.

igobranco:
Replaced by #1766

PR #1671, opened 2022-04-05
💄(frontend) prevent course glimpse title color override

jbpenrath:
## Purpose

When a course glimpse is used within a section using a variant, the course
glimpse title color is set with the section title color scheme property. This
commit fix this issue.

### Before
image

### After
image


sampaccoud (approved):
Make a fix release in this PR?

PR #1674, opened 2022-04-06
🚑(frontend) fix search query string formatting

### LINKED ISSUE #1673: Search crash when several languages are selected
## Bug Report

**Expected behavior/code**
A user should be able to search courses available in several languages

**Actual Behavior**
If user wants to filter with several languages, the search crash.

**Steps to Reproduce**
1. Go to the search page
2. In the sidebar, in the languages filter pane, select several languages
3. Boom, the search crash

**Environment**
- Richie version: 2.14.0
- Platform: Mac OS

**Possible Solution**
Investigate why elasticsearch returns a 400 Bad request in this case.

**Additional context/screenshots**
```json
{"errors":{"languages":["Select a valid choice. en,ar is not one of the available choices."]}}
```

jbpenrath:
## Purpose

The search filter query string was not formatted properly so when a user wanted to filter results with several languages the search request returned a 400 Bad Request error.

Resolve #1673

## Proposal

- [x] Remove `arrayFormat: 'comma'` option to the method to serialize search params into query string

### DIFF
--- a/src/frontend/js/data/getResourceList/index.ts
+++ b/src/frontend/js/data/getResourceList/index.ts
@@ -17,14 +17,11 @@ export async function fetchList(
   params: APIListRequestParams = API_LIST_DEFAULT_PARAMS,
 ): Promise<FetchListResponse> {
   try {
-    const response = await fetch(
-      `/api/v1.0/${kind}/?${stringify(params, { arrayFormat: 'comma' })}`,
-      {
-        headers: {
-          'Content-Type': 'application/json',
-        },
+    const response = await fetch(`/api/v1.0/${kind}/?${stringify(params)}`, {
+      headers: {
+        'Content-Type': 'application/json',
       },
-    );
+    });
 
     if (!response.ok) {
       // Push remote errors to the error channel for consistency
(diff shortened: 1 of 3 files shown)


src/frontend/js/data/getResourceList/index.ts
```diff
@@ -17,14 +17,11 @@ export async function fetchList(
...
-      {
-        headers: {
-          'Content-Type': 'application/json',
-        },
+    const response = await fetch(`/api/v1.0/${kind}/?${stringify(params)}`, {
+      headers: {
+        'Content-Type': 'application/json',
       },
-    );
+    });
```
sampaccoud, line 24: I'm curious, why didn't the linter catch it?

sampaccoud (approved):

src/frontend/js/data/getResourceList/index.ts
jbpenrath: Because, there was a comma before.
It acts like black linter with tuple in python
`('item1',)` is not formatted but `('item1')` is formatted into `'item1'`

PR #1676, opened 2022-04-07
🔖(chore) bump to version 2.14.1

jbpenrath:
### Fixed

- Preserve query string on language switch
- Fix search query string formatting
- Fix a course glimpse title color issue when used within a section with variant


sampaccoud (approved):
:muscle:

PR #1682, opened 2022-04-12
Add cookiecutter project

sampaccoud:
## Purpose

We want to allow anyone to spawn their own richie site factory with one command line. 

## Proposal

I propose to integrate a Cookiecutter template inside the Richie repository. Starting your own richie site factory will be done in one command:

```bash
$ cookiecutter https://github.com/openfun/richie --directory=cookiecutter
``` 

Replaying cookiecutter on an existing repository will allow updating one's richie site factory scaffholding while respecting the `sites` directory:

```bash
$ cookiecutter -f https://github.com/openfun/richie --directory=cookiecutter
```

PR #1688, opened 2022-04-15
📝(website) fix and improve website documentation

sampaccoud:
### Purpose

The website was not building anymore following our refactoring of the documentation in PR https://github.com/openfun/richie/pull/1682 to include cookiecutter instructions.

### Proposal 

There seem to be a bug on docusaurus versioning that prevents us from renaming an existing page... so I fixed it by renaming it also in the history. This does not impact what users see and the old versions of the website are still available.

I took the opportunity to also improve the documentation to make it clearer. In particular I repeated parts of the README in the first page of the documentation to introduce the project a minimum.


jbpenrath (approved):
JPG format is more adapted to your screenshot. 
demo-screenshot-100-min

sampaccoud:
> JPG format is more adapted to your screenshot.

I changed it to jpg :pray:

PR #1687, opened 2022-04-14
✨(cookiecutter) add empty files with explanations

igobranco:
Add empty files with explanations for UI customization

PR #1695, opened 2022-04-19
♻️(scss) refactor scss code

igobranco:
Refactor frontend scss code. 
Add `_all.scss` on each folder so the `_main.scss` could be more compact.
Add `!default` to homegrown Richie shared scss settings/variables.

#1687


jbpenrath (commented):
This modification introduces deep changes about the manner to override stylesheets for our site factory. You should add entry about that in the UPGRADE.md.

igobranco:
> This modification introduces deep changes about the manner to override stylesheets for our site factory. You should add entry about that in the UPGRADE.md.

Done.

src/frontend/scss/_main.scss
```diff
@@ -20,116 +20,21 @@
...
-@import './components/templates/courses/cms/category_list';
-@import './components/templates/courses/cms/category_detail';
-@import './components/templates/courses/cms/course_detail';
-@import './components/templates/courses/cms/person_list';
-@import './components/templates/courses/cms/person_detail';
-@import './components/templates/courses/cms/blogpost_list';
-@import './components/templates/courses/cms/blogpost_detail';
-@import './components/templates/courses/cms/program_list';
-@import './components/templates/courses/cms/program_detail';
+@import './components/all';
```
jbpenrath, line 37: Not a fan to have this file at the root of js/components/ folder.

Furthermore, within scss/ folder, we have a components/ directory so, from my point of view, it seems logic that those styles are imported through components/_all.scss.

What do you think? Do you see drawbacks ?

jbpenrath:
One last little topic that I have in mind : 
In order to homegeneize naming accross frontend. What do you think to rename `_all.scss` files into `_index.scss`?

src/frontend/scss/_main.scss
igobranco: Ok. Done.

jbpenrath (approved):
Many thanks @igobranco 🚀 
It will grealty ease our upgrades through our site-factory ! 🤩

PR #1699, opened 2022-04-26
:sparkles: dashboard

gegeturambar:
## Purpose

Création d'un dashboard permettant de visualiser les informations d'un utilisateur

### DIFF
--- a/src/richie/apps/core/templates/richie/dashboard.html
+++ b/src/richie/apps/core/templates/richie/dashboard.html
@@ -1,0 +1,26 @@
+{% extends "richie/single_column.html" %}
+{% load cms_tags i18n %}
+
+{% block head_title %}
+    {% trans "dashboard" %}
+{% endblock head_title %}
+
+{% block meta_opengraph_contextuals %}
+    {% trans "dashboard" as page_title %}
+    <meta property="og:title" content="{{ page_title |truncatechars:65 }}">
+    <meta property="og:url" content="{{ SITE.web_url }}-{{ page_title }}">
+{% endblock meta_opengraph_contextuals %}
+
+
+{% block subheader_content %}
+    {% trans "dashboard" as page_title %}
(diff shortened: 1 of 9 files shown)


src/richie/apps/core/templates/richie/dashboard.html
```diff
@@ -0,0 +1,26 @@
+{% extends "richie/single_column.html" %}
+{% load cms_tags i18n %}
+
+{% block head_title %}
+    {% trans "dashboard" %}
```
jbpenrath, line 5:
```suggestion
    {% trans "Dashboard" %}
```

src/richie/apps/core/templates/richie/dashboard.html
```diff
@@ -0,0 +1,26 @@
...
+{% endblock head_title %}
+
+{% block meta_opengraph_contextuals %}
+    {% trans "dashboard" as page_title %}
```
jbpenrath, line 9:
```suggestion
    {% trans "Dashboard" as page_title %}
```

src/richie/apps/core/templates/richie/dashboard.html
```diff
@@ -0,0 +1,26 @@
...
+
+
+{% block subheader_content %}
+    {% trans "dashboard" as page_title %}
+<div class="subheader__container">
+    <h1 class="subheader__title">
+        {{ page_title|title }}
+    </h1>
+</div>
+{% endblock subheader_content %}
```
jbpenrath, line 22: Not sure, this relevant to override the `subheader_content` to display the Dashboard name.
I think we should not displayed anything here in order to display all we want about the Dashboard (breadcrumbs, current page title) through the react app.

So instead to append content within `subheader_content` block, I would override `subheader` block to display nothing at all.

```suggestion
{% block subheader %}{% endblock %}
```

PR #1706, opened 2022-05-10
Improve accessibility of header and footer

Leimi:
## Purpose

Make sure screen reader users and keyboard users get a good experience when navigating in the header, nav and footer of richie.

## Proposal

- fix how the `<header>` and `<nav>` tags are used. For best screen reader user experience, the `<nav>` shouldn't be in the `<header>` because some screen readers (NVDA, for example) will not recognize region tags that are wrapped in other region tags
- fix tabulation order, especially on mobile with the hamburger menu, but also on the footer

### DIFF
--- a/src/frontend/scss/components/_header.scss
+++ b/src/frontend/scss/components/_header.scss
@@ -35,6 +35,12 @@
     }
   }
 
+  &__header {
+    display: flex;
+    flex-wrap: wrap;
+    width: 100%;
+  }
+
   // Will show only on mobile breakpoints
   &__hamburger {
     @include sv-flex(0, 0, auto);
--- a/src/richie/apps/core/templates/richie/base.html
+++ b/src/richie/apps/core/templates/richie/base.html
@@ -204,6 +204,7 @@
                             const target = document.getElementById(el.dataset.target);
                             // Toggle the "is-active" class on both the burger and container
                             el.classList.toggle("is-active");
+                            el.setAttribute("aria-expanded", el.classList.contains("is-active"));
                             target.classList.toggle("is-open");
                         });
                     });
--- a/src/richie/apps/core/templates/richie/base.html
+++ b/src/richie/apps/core/templates/richie/base.html
@@ -81,47 +81,54 @@
 
             {# Wrap header content in a header tag to improve landmarks around the site #}
 
-            <header id="site-header">
-                <nav class="topbar {% block topbar_classes %}{% endblock topbar_classes %}" id="main-menu">
+            <div id="site-header">
+                <div class="topbar {% block topbar_classes %}{% endblock topbar_classes %}" id="main-menu">
                     <div class="topbar__container">
-                        <div class="topbar__brand">
-                            <button class="topbar__hamburger"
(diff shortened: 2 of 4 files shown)


src/frontend/scss/components/_header.scss
```diff
@@ -7,7 +7,7 @@
   background: r-theme-val(topbar, base-background);
 
   @include media-breakpoint-up($r-topbar-breakpoint) {
-    height: $r-topbar-height;
+    padding-bottom: 0;
```
Leimi, line 10: I stumbled upon this and quickly fixed it in this PR but it's kind of unrelated https://github.com/openfun/richie/pull/1706/commits/a2354f74e998ad45287222f6370a958317ef2289. *And* I'm actually not sure I didn't break some stuff, since the header is filled with templates blocks. I'm not sure that I can say to myself "this is okay" by just testing this with the demo site locally?

src/richie/apps/core/templates/richie/base.html
```diff
@@ -81,47 +81,54 @@
...
+                        <nav class="topbar__menu">
+                            <ul class="topbar__list">
+                                {% show_menu 0 100 0 0 "menu/header_menu.html" %}
+                            </ul>
+                        </nav>
                     </div>
-                </nav>
-            </header>
+                </div>
+            </div>
```
Leimi, line 131: most of the header/nav changes here are just blocks that have been reordered, don't hesitate to check this PR with hidden whitespace changes.

Leimi:
(actually I noticed a bug just after creating the PR, will let you know)

Leimi:
should be good now!

src/richie/apps/core/templates/richie/base.html
sandroscosta: Had no idea about this NVDA issue. I always encapsulate <nav> semantically in the element it should be displayed. Thanks for this note.

src/frontend/scss/components/_header.scss
sandroscosta: Those blocks are usually user to change the header logo. We might need a `min-height` just to make sure it doesn't break. But we have to test this out.

jbpenrath: I think this property has been set to fix the size of the header but yeah in fact currently its content overflows.
But from my understanding, with your fix header height has no limit. According to the logo format (especially portrait) the header height could be very large. 

So I think we should set `max-height` to the logo then use `object-fit` and `object-position` to display it well.

Leimi: Ok oh well, I actually removed my commit from this branch. Since it's not really related to a11y, since I'm not sure what max height you'd want, and since I don't have any real other actually used logos from other instances of richie to test, I'd rather leave that out for now and spend time on other a11y stuff :)

jbpenrath (approved):
I would just add an entry to the `UPDATE.md` to warn developers who have overriden header blocks to check that all is well render with the new structure.

Otherwise, looks good to me !

Leimi:
> I would just add an entry to the `UPDATE.md` to warn developers who have overriden header blocks to check that all is well render with the new structure.
> 
> Otherwise, looks good to me !

Thanks, added a note in the upgrade file.

PR #1709, opened 2022-05-10
Improve accessibility on filters pane when using a small screen

Leimi:
## Purpose

Make sure navigating inside the search page is OK when using a mobile phone/small screen + screen reader/keyboard only navigation.

## Proposal

Everything is precisely described in the commit.

*Note: instead of doing multiple very small commits, I did one big commit concerning everything, let me know if you'd rather have everything split up.*

Todo: 

- [x] add tests

### DIFF
--- a/src/frontend/js/components/Search/index.tsx
+++ b/src/frontend/js/components/Search/index.tsx
@@ -1,35 +1,23 @@
-import { Fragment, useState, useEffect } from 'react';
+import { Fragment, useState, useEffect, useRef } from 'react';
 import { defineMessages, FormattedMessage } from 'react-intl';
 
 import { CourseGlimpseList } from 'components/CourseGlimpseList';
 import { PaginateCourseSearch } from 'components/PaginateCourseSearch';
 import { SearchFiltersPane } from 'components/SearchFiltersPane';
 import { Spinner } from 'components/Spinner';
-import { Icon } from 'components/Icon';
 import { useCourseSearch } from 'data/useCourseSearch';
 import { useCourseSearchParams, CourseSearchParamsAction } from 'data/useCourseSearchParams';
 import useMatchMedia from 'utils/useMatchMedia';
 import { RequestStatus } from 'types/api';
 import { CommonDataProps } from 'types/commonDataProps';
 import { scroll } from 'utils/indirection/window';
+import FiltersPaneCloseButton from './FiltersPaneCloseButton';
 
 const messages = defineMessages({
   errorMessage: {
     defaultMessage: `Something's wrong! Courses could not be loaded.`,
     description: 'Error message for Search view when the request to load courses fails',
     id: 'components.Search.errorMessage',
-  },
-  hideFiltersPane: {
-    defaultMessage: 'Hide filters pane',
-    description:
-      'Accessibility text for the button/icon that toggles *off* the filters pane on mobile',
-    id: 'components.Search.hideFiltersPane',
-  },
-  showFiltersPane: {
-    defaultMessage: 'Show filters pane',
-    description:
-      'Accessibility text for the button/icon that toggles *on* the filters pane on mobile',
-    id: 'components.Search.showFiltersPane',
   },
   resultsTitle: {
     defaultMessage: 'Search results',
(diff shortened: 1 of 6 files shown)


src/frontend/js/components/Search/index.tsx
```diff
@@ -78,49 +78,69 @@ const Search = ({ context }: CommonDataProps) => {
...
+    }
+  }, [filtersPaneClosing]);
+
+  const onFiltersPaneCloseButtonClick = () => {
+    if (showFilters) {
+      setFiltersPaneClosing(showFilters);
+    }
+    setShowFilters(!showFilters);
+  };
+
```
Leimi, line 98: I must confess I'm not a huge fan of all this boilerplaty, sensitive code (the `cssFiltersPaneTransitionDuration` must match the value in the stylesheet). It's done to prevent having a UI jump when hiding the mobile filters pane.

This shows my goal, don't wanna start potentially spending time on something more elaborate without your opinion first @jbpenrath

src/frontend/js/components/Search/index.tsx
```diff
@@ -172,4 +192,41 @@ const Search = ({ context }: CommonDataProps) => {
...
+      <Fragment>
+        <Icon name="icon-filter" className="search__filters__toggle__icon" />
+        <span className="offscreen">
+          <FormattedMessage {...messages.showFiltersPane} />
+        </span>
+      </Fragment>
+    )}
+  </button>
+);
+
```
Leimi, line 231: Should I create a new file for this button that is now a component by itself (since I use it twice now)?

src/frontend/js/components/Search/index.tsx
jbpenrath: Instead of a timeout, you could use a event listener.

```ts
filterPaneRef.current?.addEventListener(
        'animationend',
        () => {
          setFiltersPaneClosing(false);
        },
        { once: true },
      );
```

Leimi:
Hey, any other thing to add in this review before I go back on it + add tests? Thanks.

src/frontend/js/components/Search/index.tsx
```diff
@@ -172,4 +192,41 @@ const Search = ({ context }: CommonDataProps) => {
...
+          <FormattedMessage {...messages.hideFiltersPane} />
+        </span>
+      </Fragment>
+    ) : (
+      <Fragment>
+        <Icon name="icon-filter" className="search__filters__toggle__icon" />
+        <span className="offscreen">
+          <FormattedMessage {...messages.showFiltersPane} />
+        </span>
+      </Fragment>
```
jbpenrath, line 227: I think we could factorize this part, what do you think ?
```suggestion
      const iconName = expanded ? 'icon-cross' : 'icon-filter';
      const message = expanded ? messages.hideFiltersPane : messages.showFiltersPane;
      ...
      <Fragment>
        <Icon name={iconName} className="search__filters__toggle__icon" />{' '}
        <span className="offscreen">
          <FormattedMessage {...message} />
        </span>
      </Fragment>
```

jbpenrath:
> Hey, any other thing to add in this review before I go back on it + add tests? Thanks @jbpenrath

I just add one little comment to have DRY code.

Leimi:
Thanks!

Leimi:
I'm done on this :+1:

src/frontend/js/components/Search/index.tsx
```diff
@@ -1,36 +1,24 @@
-import { Fragment, useState, useEffect } from 'react';
```
jbpenrath, line 1: You should pass `onFiltersPaneCloseButtonClick` to the `onClick` of overlay (L179) because currently when user touch the overlay the filter pane is not close smoothly.

Leimi: Good catch, thanks!

src/frontend/js/components/Search/index.tsx
```diff
@@ -78,49 +66,73 @@ const Search = ({ context }: CommonDataProps) => {
...
+  const [filtersPaneClosing, setFiltersPaneClosing] = useState(false);
+  useEffect(() => {
+    if (filtersPaneClosing) {
+      filtersPaneRef.current?.addEventListener(
+        'transitionend',
+        () => setFiltersPaneClosing(false),
+        { once: true },
+      );
+    }
+  }, [filtersPaneClosing]);
```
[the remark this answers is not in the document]
Leimi: Alright :)

src/frontend/js/components/Search/index.tsx
```diff
@@ -78,49 +74,64 @@ const Search = ({ context }: CommonDataProps) => {
...
+            filters={
+              courseSearchResponse?.status === RequestStatus.SUCCESS
+                ? courseSearchResponse.content.filters
+                : null
+            }
+            id={`${referenceId}-filters`}
+          />
+        </div>
+        {/* we repeat the close button at the end because there can be quite a lot of filters,
+				and going back to the close button at the top is cumbersome, especially for screen reader users */}
```
jbpenrath, line 119: There are tabs here instead spaces
```suggestion
        {/* we repeat the close button at the end because there can be quite a lot of filters,
        and going back to the close button at the top is cumbersome, especially for screen reader users */}
```

jbpenrath (approved):
A last one little fix then LGTM ✅

src/frontend/js/components/Search/index.tsx
Leimi: woops!

PR #1725, opened 2022-06-06
🐛(elasticsearch) multiple elasticsearch hosts

igobranco:
Allow to configure multiple elasticsearch hosts.

On sandbox we can try it like:
```
RICHIE_ES_HOST=elasticsearch_unexistence:9200,elasticsearch:9200
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 - Fix `components.SaleTunnelStepPayment.userBillingAddressNoEntry` misspell
 - Remove z-index of course glimpse icon to fix an overlay issue
 - Fix erratic frontend test failure
+- Allow to configure multiple elasticsearch hosts
 
 
 ##  [2.14.1] - 2022-04-07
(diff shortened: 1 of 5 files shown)


mbenadda (approved):
Looks great to me! This is how we should have done it to begin with 👍

sampaccoud:
But ! This is a breaking change! We must add instructions in UPGRADE.md at least.

igobranco:
> But ! This is a breaking change for people who overrode the value.... like... us: https://github.com/openfun/arnold/search?q=RICHIE_ES_HOST
> 
> We must add instructions in UPGRADE.md at least.

It isn't a breaking change, because if you pass the current value of `elasticsearch` it will continue to run has normal.

If we add the next line on the file `env.d/development/dev`, it works has expected:
```
RICHIE_ES_HOST=elasticsearch
```

This won't work:
```
RICHIE_ES_HOST=elasticsearch_unexistence
```

But with this pull request it is possible to use an elastic search high availability cluster, using configuring like:
```
RICHIE_ES_HOST=elasticsearch_1,elasticsearch_2,elasticsearch_3
```
or
```
RICHIE_ES_HOST=elasticsearch_1:9200,elasticsearch_2:9200,elasticsearch_3:9200
```

sampaccoud:
@igobranco @mbenadda perfect 💪

PR #1726, opened 2022-06-10
✨(LTIConsumer) make lti context uniquer per user 

jbpenrath:
## Purpose
In order to use properly the LTI Marsha plugin, we have to distinguish each user using this LTI plugin. So we use user info when user is authenticated otherwise we use a random uuid as "user_id".

## Proposal

- [x] Update LTI React component to send bind user info to get lti context
- [x] Update LTI React component to refetch lti context when user changes
- [x] Update LTI CMS Plugin to bind provided user info to lti content
- [x] Update LTI CMS Plugin to cache context per user/language/placeholder_id


tests/plugins/lti_consumer/test_api.py
```diff
@@ -48,15 +48,18 @@ def test_lti_consumer_api_get_context(self, mock_params):
         )
 
         response = self.client.get(
-            f"/api/v1.0/plugins/lti-consumer/{model_instance.pk}/context/"
+            f"/api/v1.0/plugins/lti-consumer/{model_instance.pk}/context/",
+            {"user_id": "richie"},
```
lunika, line 52: Can you use a uuid or a username generated with a lib like `factory_boy` ?

jbpenrath:
@lunika I fixed your comments.

Furthermore, I implement an idea of improvement in the last commit.
We could use react-query to cache lti context on the frontend side for 5 minutes. The idea is to limit to number of request to the server and also to speed up the lti iframe initialization time.

lunika (approved):

PR #1728, opened 2022-06-15
Add licence filtering and search/autocomplete

mbenadda:
## Purpose

Enable course filtering by licence. As we did this, we were able to add search/autocomplete for licences without much additional effort.

## Proposal

Description...

- [x] create a licences index so we can reuse `IndexableFilterDefinition` for licence-based course filtering
- [x] open endpoints for search & autocomplete on licences
- [x] add licence autocompletion to the general course search field


src/richie/apps/search/indexers/courses.py
```diff
@@ -603,6 +606,21 @@ def get_es_document_for_course(cls, course, index=None, action="index"):
...
+                    id__in=[
+                        plugin.id
+                        for plugin in course.extended_object.get_placeholders()
+                        .get(slot="course_license_content")
+                        .get_plugins()
+                    ]
+                )
+            ]
+        except (LicencePluginModel.DoesNotExist, Placeholder.DoesNotExist):
+            licences = None
```
sampaccoud: I proposed a fixup with an alternative after writing a test in `test_indexers_courses` which surfaced a duplicate result issue. There was actually a way to make the query directly on the Licence :sweat_smile:

PR #1756, opened 2022-08-05
📌(backend) pin all python dependencies

jbpenrath:
## Purpose

In one week, Richie master branch has been corrupted twice due to unpinned dependency updates (easy thumbnails 2.8.2 then django-cms 3.11). This is unlucky but we should take action to prevent this happen again.

Currently, python dependencies was not pinned so the main branch could be corrupted if a package update introduces a breaking change. This behaviour is weird. Renovate is in charge to update those packages each week, so we are able to upgrade those package securely and regularly.

## Proposal

- [x] Pin python dependencies

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -29,22 +29,22 @@ include_package_data = True
 # `matchPackageNames` (in the "ignored python dependencies" group) within
 # renovate.json file at the root of this repository
 install_requires =
-    arrow
-    Django
-    djangocms-file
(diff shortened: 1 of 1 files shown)


kernicPanel (approved):
I didn't check each dependency current version.

jbpenrath:
> I didn't check each dependency current version.

I use `pip list` to know the current version installed.

PR #1760, opened 2022-08-09
Feat/address form display error messages

jbpenrath:
## Purpose

In order to improve user experience, we want to display error messages into the `AddressForm` component. If we retrieve basically error message generated by yup, we get only english messages so in our cases we have to [customize errors returned by yup](https://github.com/jquense/yup/tree/pre-v1#using-a-custom-locale-dictionary) to be able to use `react-intl` to manage error message translations.

Furthermore, some error messages can be long, so we must be able to display those messages without breaking the layout that's why we revamp a little bit form fields to be able to display multilines messages.

## Proposal

- [x] Update form fields to accept multi lines messages
- [x] Display error messages into `AddressForm`

### DIFF
--- a/src/frontend/js/components/AddressesManagement/AddressForm.spec.tsx
+++ b/src/frontend/js/components/AddressesManagement/AddressForm.spec.tsx
@@ -1,0 +1,191 @@
+import { act } from '@testing-library/react-hooks';
+import { fireEvent, getByText, render, screen } from '@testing-library/react';
+import * as mockFactories from 'utils/test/factories';
+import { AddressFactory } from 'utils/test/factories';
+import countries from 'i18n-iso-countries';
+import { IntlProvider } from 'react-intl';
+import { Address } from 'types/Joanie';
+import { ErrorKeys } from './validationSchema';
+import AddressForm from './AddressForm';
+
+jest.mock('hooks/useAddresses', () => ({
+  useAddresses: () => ({
+    states: {
+      creating: false,
+      updating: false,
+    },
+  }),
+}));
+
+describe('AddressForm', () => {
+  const handleReset = jest.fn();
+  const onSubmit = jest.fn();
+
+  beforeEach(() => {
+    jest.resetAllMocks();
+    jest.resetModules();
+  });
+
+  it('renders a button with label "Use this address" when no address is provided', () => {
+    render(
+      <IntlProvider locale="en">
+        <AddressForm handleReset={handleReset} onSubmit={onSubmit} />
+      </IntlProvider>,
+    );
+
+    screen.getByRole('button', { name: 'Use this address' });
+    expect(screen.queryByRole('button', { name: 'Cancel' })).toBeNull();
+  });
+
+  it('renders a button with label "Use this address" and a cancel button when no address is provided', async () => {
+    const address: Address = AddressFactory.generate();
+    render(
+      <IntlProvider locale="en">
+        <AddressForm handleReset={handleReset} onSubmit={onSubmit} address={address} />
+      </IntlProvider>,
+    );
+
+    screen.getByRole('button', { name: 'Update this address' });
+
+    const $button = screen.getByRole('button', { name: 'Cancel' });
+    await act(async () => {
+      fireEvent.click($button);
+    });
+    expect(handleReset).toHaveBeenCalledTimes(1);
+  });
+
+  it('renders an error message when a value in the form is invalid', async () => {
+    render(
+      <IntlProvider locale="en">
+        <AddressForm handleReset={handleReset} onSubmit={onSubmit} />
+      </IntlProvider>,
+    );
+
+    screen.getByRole('form');
+    const $titleInput = screen.getByRole('textbox', { name: 'Address title' });
+    const $firstnameInput = screen.getByRole('textbox', { name: "Recipient's first name" });
+    const $lastnameInput = screen.getByRole('textbox', { name: "Recipient's last name" });
+    const $addressInput = screen.getByRole('textbox', { name: 'Address' });
+    const $cityInput = screen.getByRole('textbox', { name: 'City' });
+    const $postcodeInput = screen.getByRole('textbox', { name: 'Postcode' });
+    const $countryInput = screen.getByRole('combobox', { name: 'Country' });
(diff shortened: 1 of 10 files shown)


src/frontend/js/components/AddressesManagement/AddressForm.tsx
```diff
@@ -41,6 +42,22 @@ const AddressForm = ({ handleReset, onSubmit, address }: Props) => {
...
+    error: Maybe<
+      | string
+      | {
+          key: ErrorKeys;
+          values: Record<PropertyKey, string | number | Array<string | number>>;
+        }
+    >,
+  ) => {
+    if (!error) return undefined;
+    if (typeof error === 'string') return error;
```
Kitero, line 55: If it's directly an error you don't want to translate it? it should still be a key or it's only for the default errors of yup?

src/frontend/js/components/AddressesManagement/validationSchema.ts
```diff
@@ -1,6 +1,57 @@
...
+  },
+  [ErrorKeys.STRING_MIN]: {
+    id: 'components.AddressesManagement.validationSchema.stringMin',
+    defaultMessage: 'The minimum length is {min} {min, plural, one {char} other {chars}}.',
+    description: 'Error message displayed when a field value is too short.',
+  },
+});
+
+Yup.setLocale({
+  mixed: {
```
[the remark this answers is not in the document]
jbpenrath: The input related to this field is a checkbox so the value will be always a boolean, that's why there is no check about it.

src/frontend/js/components/Form/Inputs.tsx
```diff
@@ -11,7 +11,7 @@ import { forwardRef, useMemo } from 'react';
 
 interface FieldProps {
   error?: Boolean;
-  message?: string;
+  message?: ReactNode | string;
```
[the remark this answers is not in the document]
jbpenrath: Because in fact, `intl.formatMessage` returns a `ReactNode`.

src/frontend/js/components/AddressesManagement/AddressForm.tsx
jbpenrath: It is just in case, the error has not been customized, I would like to be able to conserve the default behaviour.

src/richie/apps/core/templates/richie/styleguide/index.html
```diff
@@ -460,15 +490,17 @@ <h4 class="h2 mb-3">Error state</h4>
...
+                      <label class="form-field__label" for="error_field_7">
+                          <span class="form-field__checkbox-control">
+                              <svg role="img">
+                                  <use href="#icon-check"></use>
+                              </svg>
+                          </span>
+                          Field 7 (Checkbox)
+                      </label>
+                    </span>
                     <span class="form-field__message">
```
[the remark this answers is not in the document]
jbpenrath: Form inputs style can be used within frontend widgets but also if need we are able to use classname to create forms within django templates. Styleguide is just a showcase of the available components styled to see how they look with the richie theme.

src/frontend/js/components/Form/Inputs.tsx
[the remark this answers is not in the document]
Kitero: Ah I didn't think so, it must return the react node <FormattedMessage/> with the props, or something like this i guess

Kitero (approved):

src/frontend/js/components/AddressesManagement/AddressForm.tsx
carofun, line 55:
```suggestion
    if (typeof error === 'string') return error;
```
I don't think this case is tested, it could be great to add a test for this purpose

jbpenrath: Finally, I returned a default localized message in the case the error message provided does not have corresponding i18n message.

carofun: we're just missing the test when a translation is missing

jbpenrath: Good catch ! It gave me a hard time but it's now tested : )

src/frontend/js/components/AddressesManagement/AddressForm.spec.tsx
```diff
@@ -0,0 +1,191 @@
...
+    screen.getByRole('button', { name: 'Update this address' });
+
+    const $button = screen.getByRole('button', { name: 'Cancel' });
+    await act(async () => {
+      fireEvent.click($button);
+    });
+    expect(handleReset).toHaveBeenCalledTimes(1);
+  });
+
+  it('renders error message when a value in the form is invalid', async () => {
```
carofun, line 57:
```suggestion
  it('renders an error message when a value in the form is invalid', async () => {
```

src/frontend/js/components/AddressesManagement/AddressForm.tsx
```diff
@@ -41,6 +42,31 @@ const AddressForm = ({ handleReset, onSubmit, address }: Props) => {
...
+
+    if (typeof error === 'string' || errorMessages[error.key] === undefined) {
+      // If the error has not been translated we return a default error message.
+      message = errorMessages[ErrorKeys.MIXED_INVALID];
+    } else {
+      message = errorMessages[error.key];
+      values = error.values;
+    }
+
+    return intl.formatMessage(message, values);
```
carofun, line 67:
```suggestion
 
    if (typeof error === 'string' || errorMessages[error.key] === undefined) {
      // If the error has not been translated we return a default error message.
      return intl.formatMessage( errorMessages[ErrorKeys.MIXED_INVALID]);
    }
    return intl.formatMessage(errorMessages[error.key], error.values);
```
I'm still not mad of declaring a variable without setting it and using it after, at the end I find that confusing, wdyt of this rework to clarify

src/frontend/js/components/AddressesManagement/AddressForm.tsx
carofun: really good one!

carofun (approved):
Big up for JB !!!

PR #1763, opened 2022-08-17
♿️(templates) add label on external video players

Leimi:
## Purpose

The main goal of the PR is to be a tiny more explicit about what the `iframe` elements on the website are about. This is mostly for screen reader users.

The added bonus of this PR is that it makes the website pass the RGAA test n° 2.1.

## Proposal

Add a `title` attribute on a video player `iframe` block.


jbpenrath (approved):
LGTM ✅

Leimi:
I don't understand why the cookiecutter-bootstrap test doesn't pass. Any idea @jbpenrath please?

jbpenrath:
That's because this job does not support branch from fork ...

PR #1765, opened 2022-08-26
🔧(search) settings to hide/reorder search filters

igobranco:
Add settings that allow to remove or change the order/position of each default search filter.

It makes more easy change the UI of the search filters, for example change the order of each search filter, or remove/disable completely a search filter.

For example, to hide the `pace` and move `levels` to be the last position, we only need to change the settings:
```
RICHIE_FILTERS_CONFIGURATION_PACE_ENABLED=False
RICHIE_FILTERS_CONFIGURATION_LEVELS_POSITION=8
```
So the bottom of the search filter UI would change to:
image

PR #1767, opened 2022-08-31
♻️(search) refactor the way filters are being configured

sampaccoud:
## Purpose

It was not easy to define the list of filters to display and their order.

PR https://github.com/openfun/richie/pull/1765 addressed this issue by introducing new settings, one per filter. 

This PR is an attempt to further simplify filters configuration and avoid adding too many settings, the existence of which is tied to what is defined in `FILTERS_CONFIGURATION`.

## Proposal

Introduce a new setting `FILTERS_PRESENTATION` and simplify the format of the existing `FILTERS_CONFIGURATION` setting.

With this new setting, it is easier to define which filters are displayed and their order.
Filter definitions became a dictionary which also makes it easier than previous list of tuples.

This requires upgrade attention only for sites that have customized their filters.

Now that filters configuration is easier, I added a documentation to explain what is possible.

### DIFF
--- a/src/frontend/js/components/SearchFiltersPane/index.spec.tsx
+++ b/src/frontend/js/components/SearchFiltersPane/index.spec.tsx
@@ -7,7 +7,7 @@ import { SearchFiltersPane } from '.';
 
 jest.mock('components/SearchFilterGroup', () => ({
   SearchFilterGroup: ({ filter }: any) => (
-    <span>{`Received filter title: ${filter.human_name}`}</span>
+    <span data-testid="filter">{`Received filter title: ${filter.human_name}`}</span>
   ),
 }));
 
@@ -25,7 +25,11 @@ describe('components/SearchFiltersPane', () => {
   ];
 
   it('renders all our search filter groups', () => {
-    const { getByText } = render(
+    const shuffledNames = ['Categories', 'Organizations', 'Persons'].sort(
+      () => Math.random() - 0.5,
+    );
+
+    const { getByText, getAllByTestId } = render(
       <IntlProvider locale="en">
         <HistoryContext.Provider value={makeHistoryOf({ limit: '999', offset: '0' })}>
           <SearchFiltersPane
@@ -37,7 +41,7 @@ describe('components/SearchFiltersPane', () => {
                 is_autocompletable: true,
                 is_searchable: true,
                 name: 'categories',
-                position: 0,
+                position: shuffledNames.indexOf('Categories'),
                 values: [],
               },
               organizations: {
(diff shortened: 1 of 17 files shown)


igobranco (approved):
I feel that, probably in future, the `FILTERS_LIST` need to be a `dict`. But in other end my thoughts could be over engineering it. For what we need now, I think it's good refactor.

Simpler for: 
1. add a new filter (doesn't need to copy paste all filter definitions that doesn't change)
2. remove some filter that shouldn't be included on the specific installation
3. reorder of the filters

I think we should at least document, in the docs, on how to remove and reorder the filters, the basic stuff.

sampaccoud:
> I feel that, probably in future, the `FILTERS_LIST` need to be a `dict`. But in other end my thoughts could be over engineering it. For what we need now, I think it's good refactor.

True! i put LIST because I forgot our discussion and didn't see why we didn't call it a list... Let's call it FILTERS_PRESENTATION as initially discussed.

> I think we should at least document, in the docs, on how to remove and reorder the filters, the basic stuff.

Good idea!

sampaccoud:
@igobranco The PR should be ready for further review. I have:
- renamed `FILTERS_LIST` to `FILTERS_PRESENTATION`
- refactored filter definitions to make them easier to override and clearer to understand (although it will never be trivial I'm afraid)
- added documentation about how to customize search filters

igobranco:
LGTM
I personally liked very much the docs. It has allowed me to understand better the configurations.

jbpenrath (commented):
It seems that `position` property is now [used only once](https://github.com/openfun/richie/blob/7ad9516c39249131ee0621bad6db92b8f76b384c/src/richie/apps/search/viewsets/courses.py#L85-L87). What do you think to fully remove this property instead then simply update the lambda to sort filters ?

```python
sorted(filters.items(), key=lambda f: FILTERS_PRESENTATION.index(f.name))
```

sampaccoud:
> It seems that `position` property is now used only once

i assumed that the fronted was using it but after your comment I checked and couldn't find it. So you confirm that the frontend only looks at the order in which the filters are returned?

jbpenrath:
> > It seems that `position` property is now used only once
> 
> i assumed that the fronted was using it but after your comment I checked and couldn't find it. So you confirm that the frontend only looks at the order in which the filters are returned?

Yes, I do.
I also checked from my side before writing this comment.

sampaccoud:
@jbpenrath following our discussion IRL where we decided to keep the "position" field to:
- make ordering explicit and robust
- not change the current API

I added a sorting close on the frontend and added a test to secure it.

src/frontend/js/components/SearchFiltersPane/index.spec.tsx
```diff
@@ -25,7 +25,12 @@ describe('components/SearchFiltersPane', () => {
   ];
 
   it('renders all our search filter groups', () => {
-    const { getByText } = render(
+    const shuffledNames = ['Categories', 'Organizations', 'Persons']
+      .map((value) => ({ value, sort: Math.random() }))
+      .sort((a, b) => a.sort - b.sort)
+      .map(({ value }) => value);
```
jbpenrath, line 31: Should do the job 
```suggestion
    const shuffledNames = ['Categories', 'Organizations', 'Persons']
      .sort(() => Math.random() - 0.5);
```

sampaccoud: :pray:

PR #1772, opened 2022-09-06
🐛(frontend) fix flaky tests using random currencies

jbpenrath:
## Purpose

Some components display prices, and we need to test price currencies are well displayed to the user. To empower our tests, price currency is randomly picked so each tests can display a different content. Thus ensure that content exists in document we try to find an element containing the formatted price using Intl.NumberFormat, but spaces seems not rendered in the same way between this method and the render method of testing-library/react. As a workaround, we replace non-breakable space by a "normal" space through `replace` method but we didn't think about currency containing several spaces. In order to fix that, we have to use `replaceAll` method instead of `replace`.

## Proposal

- [x] replace all `NBSP` by simple space


sampaccoud (approved):
legit

NathanVss (approved):
Seems ok to me :)

PR #1773, opened 2022-09-06
📌(backend) unpin python dependencies

jbpenrath:
## Purpose

We previously pinned all python dependencies to prevent some uncontrolled issues due to transparent upgrades. Unfortunately this workflow does not comply with library projects like Richie as user of this library should be able to install another version of a richie dependency and pinning all richie dependencies does not allow that that's why we have to roll back this change.

## Proposal

- [x] Rollback 798780746a996ec3bf9137acddddb590fdb79241

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -29,22 +29,22 @@ include_package_data = True
 # `matchPackageNames` (in the "ignored python dependencies" group) within
 # renovate.json file at the root of this repository
 install_requires =
-    arrow==1.2.2
+    arrow
     Django<4
-    djangocms-file==3.0.0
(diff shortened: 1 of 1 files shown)


sampaccoud:
@jbpenrath should we add a constraint on Django in the same PR?

jbpenrath:
> @jbpenrath should we add a constraint on Django in the same PR?

I let a constraint for Django (`Django<4`) to prevent the installation of Django 4 induced by Django CMS 3.11

jmaupetit (approved):

sampaccoud (approved):

PR #1774, opened 2022-09-07
🧑‍💻(frontend) add components to storybook

NathanVss:
## Purpose

The goal is to increase the coverage of the recently installed Storybook.

Components to create stories for: 
- `components/Banner`
- `components/Icon`
- `components/Form/Inputs/*`
- `components/Modal`

### DIFF
--- a/src/frontend/.storybook/preview.js
+++ b/src/frontend/.storybook/preview.js
@@ -1,9 +1,0 @@
-export const parameters = {
-  actions: { argTypesRegex: "^on[A-Z].*" },
-  controls: {
-    matchers: {
-      color: /(background|color)$/i,
-      date: /Date$/,
-    },
-  },
-}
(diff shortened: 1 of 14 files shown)


src/frontend/.storybook/preview.js
```diff
@@ -1,3 +1,6 @@
+import {FC, useState} from "react";
```
jbpenrath, line 1: FC seems not used.
```suggestion
import  { useState } from "react";
```

src/frontend/js/components/Modal/index.stories.tsx
```diff
@@ -0,0 +1,34 @@
+import { ComponentMeta, ComponentStory } from '@storybook/react';
```
jbpenrath, line 1: Just a nitpick. 
I think it could be interesting to add control for `shouldCloseOnOverlayClick` and `shouldCloseOnEsc` props as this is something interesting to know when we are using modal.

src/frontend/.storybook/preview.js
jbpenrath, line 1: Does storybook support ts modules ? If yes, could you convert this js file into a ts file. In this way, eslint should lint this file.

PR #1775, opened 2022-09-08
Cookiecutter/fix template

jbpenrath:
## Purpose

Currently, the workflow to create a project from cookiecutter is corrupted.
There are several aspects to fix :
- An incompatibility issue with docker compose > 2.17
- A dependency issue

Furthermore, as cookiecutter is crucial for newcomer to start a new project from richie, we should be aware of any problem about our template. -> **Will be address in a separate PR #1776.**

## Proposal

- [x] Fix current errors of our templates


sampaccoud (approved):
LGTM!

PR #1776, opened 2022-09-08
Cookiecutter/add job to check template consistency

jbpenrath:
## Purpose

As cookiecutter is crucial for newcomer to start a new project from richie, we should be aware of any problem about our template. That's why we decided to add a ci job to ensure that bootstrapping a new project through our cookiecutter template is working.

Furthermore, to not chase after issues, we should be able to use the branch version of richie and richie-education instead of the released versions to detect any bug introduce with a commit as soon as possible.

## Proposal

- [x] Write a job `cookiecutter-bootstrap` to ensure that our cookiecutter template is working fine with the branch codebase.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -760,6 +760,59 @@ jobs:
             echo "== Javascript =="
             ls ./build/js/*.js
 
+  cookiecutter-bootstrap:
+    docker:
+      - image: cimg/python:3.10-node
+        auth:
+          username: $DOCKER_USER
+          password: $DOCKER_PASS
+        environment:
+          RICHIE_SITE: ci
+    working_directory: ~/fun
+    steps:
+      # Activate docker-in-docker
+      - setup_remote_docker:
+          version: 19.03.13
+      - run:
+          name: Install cookiecutter
+          command: pip install cookiecutter
+      - run:
+          name: Create a project from richie cookiecutter template
+          command: cookiecutter gh:openfun/richie --checkout $CIRCLE_BRANCH --directory cookiecutter --no-input organization=fun
+      - run:
+          name: Enter the newly created project and add a new site "ci"
+          command: |
+            cd fun-richie-site-factory
+            cookiecutter template -o sites/ --no-input site=${RICHIE_SITE} domain=${RICHIE_SITE}.test
+      - run:
+          name: Create env files
+          command: |
+            cp fun-richie-site-factory/env.d/aws.dist fun-richie-site-factory/env.d/aws
+            cp fun-richie-site-factory/env.d/development.dist fun-richie-site-factory/env.d/development
+      - run:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -760,6 +760,38 @@ jobs:
...
+          username: $DOCKER_USER
+          password: $DOCKER_PASS
+    working_directory: ~/project
+    steps:
+      - run:
+          name: Install cookiecutter
+          command: pip install cookiecutter
+      - run:
+          name: Create a project from richie cookiecutter template
+          command: cookiecutter gh:openfun/richie --directory cookiecutter --no-input organization=fun
```
sampaccoud, line 776: you have to run cookiecutter on the template in the source code and not from github otherwise you get the one on master and not the one you are trying to merge :man_shrugging:

.circleci/config.yml
```diff
@@ -760,6 +760,56 @@ jobs:
             echo "== Javascript =="
             ls ./build/js/*.js
 
+  cookiecutter-bootstrap:
+    docker:
+      - image: cimg/python:3.10-node
+        auth:
+          username: $DOCKER_USER
+          password: $DOCKER_PASS
+    working_directory: ~/project
```
sampaccoud, line 769: shouldn't this be:
```suggestion
    working_directory: ~/fun
```

.circleci/config.yml
```diff
@@ -760,6 +760,56 @@ jobs:
...
+      - run:
+          name: Build front
+          command: |
+            cd fun-richie-site-factory/sites/ci/src/frontend
+            yarn install
+            yarn build-sass-production
+            yarn build-ts-production
+      - run:
+          name: Activate "ci" site
+          command: echo "export RICHIE_SITE=ci" >> $BASH_ENV
```
sampaccoud, line 805: you could set it for the whole job like it's done here for example:
https://github.com/openfun/richie/blob/master/.circleci/config.yml#L380

jbpenrath: Thanks !

jbpenrath:
@sampaccoud To address this part of the PR : 

> Furthermore, to not chase after issues, we should be able to use the branch version of richie and richie-education instead of the released one to detect any bug introduce with a commit as soon as possible.

I bring your attention on [this step of the job](https://github.com/openfun/richie/blob/bdcec7b62dc5c0ede540cce60b1bf197cb733e26/.circleci/config.yml#L797-L801).

sampaccoud:
> I bring your attention on [this step of the job](https://github.com/openfun/richie/blob/bdcec7b62dc5c0ede540cce60b1bf197cb733e26/.circleci/config.yml#L797-L801).

Smart :thinking: Maybe too smart... let's do that and see :+1:

sampaccoud (approved):

PR #1778, opened 2022-09-09
♻️(frontend) refactor modal component ( UX / DX ) 

NathanVss:
## Purpose

The modal close button wasn't included in the modal component, therefore it was required to define it each time when using the modal. So let's define a generic approach to modal close button for homogeneous UI.

Moreover, we decided to make homogeneous the behavior of the modal component, in terms in scroll and with a sticky header.

### DIFF
--- a/src/frontend/js/components/Modal/index.spec.tsx
+++ b/src/frontend/js/components/Modal/index.spec.tsx
@@ -1,4 +1,5 @@
 import { render } from '@testing-library/react';
+import { IntlProvider } from 'react-intl';
 import { Modal } from '.';
 
 describe('<Modal />', () => {
@@ -10,12 +11,14 @@ describe('<Modal />', () => {
 
   it('merges custom classNames of type string with the default classes', () => {
     render(
-      <Modal
-        isOpen={true}
-        className="custom-class"
-        bodyOpenClassName="custom-body-open-class"
-        overlayClassName="custom-overlay-class"
-      />,
+      <IntlProvider locale="en">
+        <Modal
+          isOpen={true}
+          className="custom-class"
+          bodyOpenClassName="custom-body-open-class"
+          overlayClassName="custom-overlay-class"
+        />
+      </IntlProvider>,
     );
 
     const $body = document.querySelector('body.has-opened-modal.custom-body-open-class');
@@ -29,19 +32,21 @@ describe('<Modal />', () => {
 
   it('merges custom classNames of type ReactModal.Classes with the default classes', () => {
     render(
-      <Modal
-        isOpen={true}
-        className={{
-          base: 'custom-class-base',
-          afterOpen: 'custom-class-afterOpen',
-          beforeClose: 'custom-class-beforeClose',
-        }}
-        overlayClassName={{
-          base: 'custom-overlay-class-base',
-          afterOpen: 'custom-overlay-class-afterOpen',
-          beforeClose: 'custom-overlay-class-beforeClose',
-        }}
-      />,
+      <IntlProvider locale="en">
+        <Modal
+          isOpen={true}
+          className={{
+            base: 'custom-class-base',
+            afterOpen: 'custom-class-afterOpen',
+            beforeClose: 'custom-class-beforeClose',
+          }}
+          overlayClassName={{
+            base: 'custom-overlay-class-base',
+            afterOpen: 'custom-overlay-class-afterOpen',
+            beforeClose: 'custom-overlay-class-beforeClose',
+          }}
+        />
+      </IntlProvider>,
     );
 
     const $modal = document.querySelector('.modal.custom-class-base');
--- a/src/frontend/js/components/Modal/index.spec.tsx
+++ b/src/frontend/js/components/Modal/index.spec.tsx
@@ -1,4 +1,4 @@
-import { render } from '@testing-library/react';
+import { render, screen } from '@testing-library/react';
 import { IntlProvider } from 'react-intl';
 import { Modal } from '.';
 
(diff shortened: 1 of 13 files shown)


src/frontend/js/components/Modal/index.tsx
```diff
@@ -43,6 +66,23 @@ export const Modal = ({
       overlayClassName={mergeClasses({ base: 'modal__overlay', classes: overlayClassName })}
       {...props}
     >
+      <div className={headerClasses.join(' ')}>
```
jbpenrath, line 69: In order to improve semantic what do you think to use a `header` tag here ?
```suggestion
      <header className={headerClasses.join(' ')}>
```

src/frontend/js/components/Modal/index.tsx
```diff
@@ -43,6 +66,23 @@ export const Modal = ({
...
+        <div>{title}</div>
```
jbpenrath, line 70: Same this inline content, I think a `p` tag is more appropriate in this case.

src/frontend/js/components/Modal/index.tsx
```diff
@@ -1,13 +1,31 @@
-import { PropsWithChildren, useMemo } from 'react';
+import { ReactNode, useMemo } from 'react';
 import ReactModal from 'react-modal';
+import { defineMessages, FormattedMessage, useIntl } from 'react-intl';
+
+const messages = defineMessages({
+  closeDialog: {
+    defaultMessage: 'Close modal',
```
jbpenrath, line 7: For non technical user, I believe dialog is more appropriate than modal (or we should use `dialog modal` 😅)
```suggestion
    defaultMessage: 'Close dialog',
```

src/frontend/js/components/Modal/index.tsx
NathanVss: What do you think about using a `h2` tag ? IMO it has more to do with a title than a paragraph

jbpenrath: Hmmm... if we consider modal content relevant for the document yep ˋh2` is the right heading level furthermore it allows us use lower heading tags within modal content. 

So let's do it ! 👍

NathanVss: All righty :)

src/frontend/js/components/Modal/index.spec.tsx
```diff
@@ -50,4 +55,50 @@ describe('<Modal />', () => {
...
+
+  it('displays close button', () => {
+    render(
+      <IntlProvider locale="en">
+        <Modal isOpen={true} />
+      </IntlProvider>,
+    );
+
+    const $closeButton = document.querySelector('.modal .modal__closeButton');
+    expect($closeButton).toBeInstanceOf(HTMLElement);
```
jbpenrath, line 67:
```suggestion
We are using [methods provided](https://testing-library.com/docs/queries/about#types-of-queries) by `testing-library` to select elements with accessibility in mind.

```suggestion
    // A close button should be displayed
    screen.getByRole('button', { name: 'Close dialog' });
```
```

File: src/frontend/js/components/Modal/index.spec.tsx [hunk 5]
```diff
@@ -50,4 +55,50 @@ describe('<Modal />', () => {
...
+  });
+
+  it('has no close button', () => {
+    render(
+      <IntlProvider locale="en">
+        <Modal isOpen={true} hasCloseButton={false} />
+      </IntlProvider>,
+    );
+
+    const $closeButton = document.querySelector('.modal .modal__closeButton');
```
jbpenrath, line 77: ditto

src/frontend/js/components/Modal/index.spec.tsx
```diff
@@ -50,4 +55,50 @@ describe('<Modal />', () => {
...
+  });
+
+  it('displays title', () => {
+    render(
+      <IntlProvider locale="en">
+        <Modal isOpen={true} title={<div className="custom__header">Hi there</div>} />
+      </IntlProvider>,
+    );
+
+    const $customHeader = document.querySelector('.modal .modal__header .custom__header');
```
jbpenrath, line 88: ditto

tips : You should put an attribute `data-testId` to your custom title element then retrieve the element through `getByTestId`.

src/frontend/js/components/Modal/index.spec.tsx
```diff
@@ -50,4 +55,50 @@ describe('<Modal />', () => {
...
+    render(
+      <IntlProvider locale="en">
+        <Modal isOpen={true} title="Hello world" />
+      </IntlProvider>,
+    );
+
+    const $header = document.querySelector('.modal .modal__header');
+    expect($header).toBeInstanceOf(HTMLElement);
+    expect($header?.firstChild).toBeInstanceOf(HTMLHeadingElement);
+    expect($header?.firstChild?.textContent).toEqual('Hello world');
```
jbpenrath, line 102: ditto

```suggestion
screen.getByRole('heading', { level: 2, name: 'Hello world' });
```

src/frontend/js/utils/StringHelper.spec.tsx
```diff
@@ -0,0 +1,19 @@
+import { StringHelper } from './StringHelper';
+
+describe('StringHelper', () => {
+  it('is a string', () => {
+    expect(StringHelper.isString('Hello')).toBeTruthy();
```
jbpenrath, line 5: Use `toBe(value)` instead of `toBeTruthy` or `toBeFalsy`. It is safer to limit the risk of false positive.

```suggestion
    expect(StringHelper.isString('Hello')).toBe(true);
```

src/frontend/js/components/SearchFilterGroupModal/index.tsx
```diff
@@ -257,8 +241,13 @@ export const SearchFilterGroupModal = ({ filter }: SearchFilterGroupModalProps)
       <Modal
         appElement={modalExclude}
         bodyOpenClassName="has-search-filter-group-modal"
-        className="search-filter-group-modal modal--stretched"
+        className="search-filter-group-modal"
         isOpen={modalIsOpen}
+        title={
+          <legend className="search-filter-group-modal__form__title">
```
jbpenrath, line 247: `legend` tag cannot be used here. This tag can be only used within a fieldset that is no more the case now.

src/frontend/js/utils/StringHelper.ts
```diff
@@ -0,0 +1,5 @@
+export class StringHelper {
```
NathanVss, line 1: Note for myself: create dedicated folder.

src/frontend/js/components/SearchFilterGroupModal/index.tsx
```diff
@@ -257,8 +241,13 @@ export const SearchFilterGroupModal = ({ filter }: SearchFilterGroupModalProps)
       <Modal
         appElement={modalExclude}
         bodyOpenClassName="has-search-filter-group-modal"
-        className="search-filter-group-modal modal--stretched"
+        className="search-filter-group-modal"
         isOpen={modalIsOpen}
+        title={
+          <h2>
+            <FormattedMessage {...messages.modalTitle} values={{ filterName: filter.human_name }} />
+          </h2>
```
jbpenrath, line 249: you can use the `intl.formatMessage` to retrieve a string
```suggestion
title={intl.formatMessage({...messages.modalTitle, values:{ filterName: filter.human_name }})}
```

jbpenrath (approved):
LGTM ✅

PR #1781, opened 2022-09-12
🧑‍💻(frontend) add style to disabled inputs

NathanVss:
## Purpose

Add a differentiable style to disabled inputs.

### DIFF
--- a/src/frontend/scss/objects/_form.scss
+++ b/src/frontend/scss/objects/_form.scss
@@ -53,6 +53,11 @@
     box-shadow: 0 0 0 2px r-theme-val(form, input-outline);
     outline: none;
   }
+
+  &[disabled] {
+    background-color: r-theme-val(form, input-disabled-background);
+    cursor: not-allowed;
+  }
 }
 
 .form {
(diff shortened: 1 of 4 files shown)


NathanVss:
@jbpenrath hi, I don't know why lint-front job is failing, it seems that form.scss is not well formatted but my IDE doesn't tell me anything about a wrong formatting and when I run `prettier` locally everything is ok .. and there is not verbosity on the CI so I can't do much more at this point. If you have an idea, it is welcomed :)

NathanVss:
@jbpenrath ok nevermind, I had to install a new plugin and update prettier :)

src/frontend/scss/objects/_form.scss
```diff
@@ -154,11 +179,16 @@
...
+      &[disabled] ~ label span {
+        background-color: r-theme-val(form, input-disabled-background);
+        cursor: not-allowed;
+      }
     }
 
     &__checkbox-input + &__label,
     &__radio-input + &__label {
-      color: r-theme-val(form, input-color) !important;
+      color: r-theme-val(form, input-color);
```
jbpenrath, line 181: Sure it has no impact ?

src/frontend/scss/objects/_form.scss
```diff
@@ -83,23 +83,13 @@
     position: relative;
 
     &:hover {
```
jbpenrath, line 85: I check the hover effect on storybook. Indeed, for text inputs and select, just fade the border color is enough.

But about radio and checkbox inputs, on hover, I'll also fade the label color (but not the message color).

src/frontend/scss/objects/_form.scss
jbpenrath: It has no impact let's remove it

src/frontend/scss/objects/_form.scss
```diff
@@ -154,11 +175,19 @@
...
       position: absolute;
       z-index: -1;
+
+      &[disabled] ~ label {
+        cursor: not-allowed;
+
+        span {
+          background-color: r-theme-val(form, input-disabled-background);
+        }
+      }
```
jbpenrath, line 185: Try to target element through class name instead of element to prevent futur issue.

```suggestion
      &[disabled] ~ .form-field__label {
        cursor: not-allowed;

        .form-field__checkbox-control {
          background-color: r-theme-val(form, input-disabled-background);
        }
      }
```

jbpenrath (approved):
One last nitpick then LGTM 🚀

PR #1782, opened 2022-09-12
🐛(admin) order languages choices in course run admin form

sampaccoud:
## Purpose

Languages choices should be ordered alphabetically in admin forms.

## Proposal

Ordering depends on the active language and only works if it is done directly in the form instance.

Add a test to secure the behavior.


jbpenrath (approved):
LGTM ✅

PR #1788, opened 2022-09-22
Feature/dashboard router

jbpenrath:
## Purpose

Through Joanie, we are now able to retrieve enrollment, order, preferences of the authenticated user. In this way, we are now able to implement a full featured dashboard within Richie.

As a first step, we just create the basic routing logic with i18n routes.

image

## Proposal

- [x] Create a DashboardRouter
- [x] Manage localized route paths and labels
- [x] Draft logic to restrict access to some routes
- [x] Write tests


src/frontend/js/utils/routers/dashboard/useDashboardRouter/index.tsx
```diff
@@ -0,0 +1,107 @@
+import { useMemo } from 'react';
+import { createBrowserRouter, Navigate, Outlet, RouteObject } from 'react-router-dom';
+import { useIntl } from 'react-intl';
+import Redirect from 'components/Redirect';
```
NathanVss: Error when using `make watch-ts`: Cannot find module 'components/Redirect' or its corresponding type declarations.

NathanVss (changes requested):

src/frontend/js/hooks/useRouteInfo/index.spec.tsx
```diff
@@ -0,0 +1,68 @@
...
+      pathname: '/posts',
+      params: {},
+      data: { posts: [] },
+      handle: undefined,
+    });
+
+    // Rerender to match the '/posts/:postId' route
+    rerender({ initialEntries: ['/posts/a-first-post'] });
+
+    await waitForNextUpdate();
```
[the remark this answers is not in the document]
jbpenrath: I'm using this method because I already use it just above when I call the `renderHook` method. In fact I think, `act` could be used here.

NathanVss (approved):
🤘

carofun (commented):
minor: Some commits don't have a majuscule and a dot
✨(frontend) add a useRouteInfo hook powerfull => powerful

🐛(backend) catch all dashboard routes
comment on commit messages:
t => T 
every all => all
which => that
django => Django
**T**he /dashboard instantiates a React component **that** uses under the hood
a router so **Django** should render the dashboard template for **all**
dashboard routes (dashboard/.*)

src/frontend/js/utils/routers/dashboard/useDashboardRouter/index.tsx
```diff
@@ -0,0 +1,106 @@
...
+  getDashboardRouteLabel,
+  getDashboardRoutePath,
+} from 'utils/routers/dashboard';
+import useRouteInfo from 'hooks/useRouteInfo';
+import { getDashboardBasename } from './getDashboardBasename';
+
+/**
+ * *Temporary
+ *
+ * A dummy route component for example which display all data related to the
```
carofun, line 15:
```suggestion
 * A dummy route component for example which displays all data related to the
```

src/frontend/js/hooks/useRouteInfo/index.ts
```diff
@@ -0,0 +1,22 @@
...
+ * ⚠️ This hook must be used within a Router.
+ */
+const useRouteInfo = () => {
+  const location = useLocation();
+  const matches = useMatches();
+
+  const match = matches.find((m) => location.pathname === m.pathname);
+
+  if (!match) {
+    throw new Error(`No route match the location ${location.pathname}`);
```
carofun, line 16:
```suggestion
    throw new Error(`No route match the location ${location.pathname}`);
```
I didn't find the corresponding test

src/frontend/js/utils/routers/dashboard/index.tsx
```diff
@@ -0,0 +1,105 @@
...
+    defaultMessage: 'My preferences',
+  },
+  [DashboardPaths.PREFERENCES_ADDRESS_EDITION]: {
+    id: 'components.Dashboard.DashboardRoutes.preferences.addresses.edition.label',
+    description: 'Label of the addresses edition view.',
+    defaultMessage: 'Billing addresses',
+  },
+  [DashboardPaths.PREFERENCES_CREDIT_CARD_EDITION]: {
+    id: 'components.Dashboard.DashboardRoutes.preferences.creditCards.label',
+    description: 'Label of the addresses edition view.',
```
carofun, line 72:
```suggestion
    description: 'Label of the credit card edition view.',
```

src/frontend/js/utils/routers/dashboard/index.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+/**
+ * Use `intl.formatMessage` to retrieve a path or label of a given route for the active locale.
+ *
+ * Currying function which takes first the type of attribute to translate (path or label)
+ * then takes the intl object retrieved from the IntlProvider.
+ * Finally, it takes a path and options to pass to `intl.formatMessage` method
+ */
+const getDashboardRouteAttribute =
+  (attribute: 'label' | 'path') =>
```
carofun, line 85:
```suggestion
  (attribute: 'label' | 'path') =>
```
maybe we could use const here

src/frontend/js/utils/routers/dashboard/index.tsx
```diff
@@ -0,0 +1,105 @@
...
+      values?: Record<string, PrimitiveType | FormatXMLElementFn<string, string>>,
+      opts?: IntlMessageFormatOptions,
+    ]
+  ) => {
+    const messages = attribute === 'label' ? dashboardRoutePaths : dashboardRouteLabels;
+    return intl.formatMessage(messages[path], ...options);
+  };
+
+/** Get the provided dashboard route label in the active locale */
+export const getDashboardRouteLabel = getDashboardRouteAttribute('label');
```
carofun, line 99:
```suggestion
export const getDashboardRouteLabel = getDashboardRouteAttribute('label');
```
ditto const

src/frontend/js/utils/routers/dashboard/index.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+/** Get the provided dashboard route path in the active locale */
+export const getDashboardRoutePath = getDashboardRouteAttribute('path');
```
carofun, line 102:
```suggestion
export const getDashboardRoutePath = getDashboardRouteAttribute('path');
```
ditto const

src/frontend/js/components/ProtectedOutlet/ProtectedOutlet.tsx
```diff
@@ -0,0 +1,26 @@
...
+import { location } from 'utils/indirection/window';
+
+export interface ProtectedOutletProps {
+  isAllowed: Boolean;
+  redirectTo?: string;
+}
+
+/**
+ * A component which renders <Outlet /> only if isAllowed is true,
+ * otherwise redirect to the provided reditect path.
```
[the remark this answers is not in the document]
jbpenrath: I used VS Code to work on this PR 😄

src/frontend/js/components/ProtectedOutlet/ProtectedOutlet.tsx
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * ⚠️ This component must be used within Routes.
+ */
+const ProtectedOutlet = ({ isAllowed, redirectTo = '/' }: ProtectedOutletProps) => {
+  useEffect(() => {
+    if (!isAllowed) {
+      location.replace(redirectTo!);
```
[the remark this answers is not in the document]
jbpenrath: Good catch !

src/frontend/js/utils/routers/dashboard/index.tsx
jbpenrath: Indeed I could use an enum.

src/frontend/js/hooks/useRouteInfo/index.ts
jbpenrath: As discussed, in the context where this hook can be called, this error never occurred so instead I remove this throw statement.

carofun (approved):

PR #1797, opened 2022-10-03
Improve/consumer lti plugin

jbpenrath:
## Purpose

Currently, an editor is not able to override values of `is_automatic_resizing` and `inline_ratio` fields when the plugin is based on a lti provider defined through `RICHIE_LTI_PROVIDERS` but it appears this feature could be useful is some cases.

From now, those fields defaults as follow:
- Plugin linked to a lti provider
  - Use `inline_ratio` specified by user or try get value from lti provider settings and finally default to `0.5625`
  - Use `is_automatic_resizing` specified by user or try get value from lti provider settings and finally default to `True` 
- Plugin not linked to a lti provider :
  - Use `inline_ratio` specified by user or default to `0.5625`
  - Use `is_automatic_resizing` specified by user or default to `True`

### New plugin options
image

### Screenshot of a problematic use case
The video has not a 16:9 aspect ratio so the layout is broken and currently editor is not able to change the aspect ratio.
image

### Screenshot with the suggested fix
image

## Proposal

- [x] Fix a localization issue related to rendering of `inline_ratio` percentage value
- [x] Make `inline_ratio` and `is_automatic_resizing`fields overridable when the plugin relies on a lti provider.


tests/plugins/lti_consumer/test_models.py
```diff
@@ -32,6 +32,43 @@ def test_lti_consumer_models_is_automatic_resizing_default(self):
         self.assertTrue(instance.is_automatic_resizing)
         self.assertEqual(instance.inline_ratio, 0.5625)
 
+    def test_lti_consumer_models_default_with_lti_provider(self):
+        """
+        The "inline_ratio" and "is_automatic_resizing" fields should default to the
+        lti provider values if there are otherwise respectively to 0.5625 and True.
```
sampaccoud, line 38: I don't see how is this superior to getting it from the settings? We are going to replicate the info in database and won't be able to change it for all at one?

jbpenrath:
@sampaccoud, I take your feedback into account. So from now, `inline_ratio` and `automatic_resizing` fields can be blanked then we use properties to retrieve the right value according the context.

Furthermore, in order to improve ux, I put those fields into a collapsed fieldset "Advanced settings".
image

src/richie/plugins/lti_consumer/models.py
```diff
@@ -43,11 +43,30 @@ class LTIConsumer(CMSPlugin):
...
-    inline_ratio = models.FloatField(
-        default=0.5625,  # default is 16/9
+    _is_automatic_resizing = models.BooleanField(
+        verbose_name=_("Is automatic resizing"),
+        blank=True,
+        null=True,
+        choices=[(None, _("Inherit")), (True, _("Yes")), (False, _("No"))],
+        help_text=_(
+            (
+                "If active, the lti viewport will automatically fits its content. Leave blank to"
```
sampaccoud, line 53:
```suggestion
                "If active, the lti viewport will automatically fit its content. Leave blank to"
```

src/richie/plugins/lti_consumer/models.py
```diff
@@ -76,15 +95,38 @@ def lti_provider(self):
             self.lti_provider_id, {}
         )
 
+    @property
+    def inline_ratio(self):
```
sampaccoud, line 99: I find it weird to name model fields with an underscore and would say the usual pattern in Django for this is what you see in drf or class based views for example:
The field is called `inline_ratio` and you get it via a method called `get_inline_ratio`

jbpenrath: Ok ! I wonder what is the best practice about that. I choose the syntax with `_` prefix as those values dot not aims to be used directly. I will use getter methods.

kernicPanel (approved):

src/richie/plugins/lti_consumer/models.py
```diff
@@ -76,15 +93,36 @@ def lti_provider(self):
...
         )
 
+    def get_inline_ratio(self):
+        """
+        Returns plugin "inline_ratio" or defaults to the lti provider setting
+        otherwise returns 0.5625 (16:9 aspect ratio).
+        """
+        inline_ratio = self.inline_ratio
+
+        if inline_ratio is None and self.lti_provider_id:
```
sampaccoud, line 103: That' a use case for the walrus operator! :hand_over_mouth: 
```suggestion
        if inline_ratio:=self.inline_ratio is None and self.lti_provider_id:
```

sampaccoud (approved):
Nice :clap:

src/richie/plugins/lti_consumer/models.py
jbpenrath: 🧠

PR #1798, opened 2022-10-05
🚨(frontend) enable rule to sort imports

jbpenrath:
## Purpose

In order to improve code structure we would like to sort imports as follows "builtin", "external", "internal", "parent", "sibling", "index". The eslint import plugin allow to do that. We also have to install a plugin to resolve webpack alias as internal imports.

## Proposal

- [x] Enable eslint `import/order` rule

### DIFF
--- a/src/frontend/.eslintrc.json
+++ b/src/frontend/.eslintrc.json
@@ -42,6 +42,13 @@
     "import/extensions": "off",
     "import/no-cycle": ["off"],
     "import/no-dynamic-require": "off",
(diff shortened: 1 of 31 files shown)


NathanVss (approved):
So far so good 😊

NB: The lint-front job seems to fail 😕

jbpenrath:
> So far so good 😊
> 
> NB: The lint-front job seems to fail 😕

Yup, there was something to reformat within `.eslintrc.json`.

PR #1803, opened 2022-10-17
(frontend) dashboard layout + sidebar + breadcrumbs

NathanVss:
## Purpose

- [x] Layout
- [x] Sidebar
    - [x] Basic component
    - [x] Sidebar active nav item desktop
    - [x] Sidebar select menu nav responsive
    - [x] Fake Avatar
- [x] Handle empty content
- [x] White background when deleting addresses
- [x] flex buttons
- [x] Breadcrumbs
    - [x] Basic component
    - [x] Responsive
    - [x] Ellispsis
    - [x] Var names
- [x] Homogeneous font size (rem)
- [x] Real username
- [x] Testing ( var in name breadcrumbs also ) 

## Sketch

Dashboard - Détails produits enroll 1

## This PR

Capture d’écran 2022-10-24 à 14 35 07


src/frontend/js/components/Dashboard/DashboardBreadcrumbs.tsx
```diff
@@ -0,0 +1,123 @@
...
+      <li>
+        <Button onClick={back} size="tiny">
+          <Icon name="icon-chevron-left" className="button__icon" />
+          <FormattedMessage {...messages.back} />
+        </Button>
+      </li>
+
+      {breadcrumbs.map((breadcrumb) => (
+        <li key={breadcrumb.pathname}>
+          <Button onClick={() => navigate(breadcrumb.pathname)} size="tiny">
```
jbpenrath, line 116: For a11y purpose, you should unfocus element on click otherwise when user clicks on the element, the button keeps the focus then screen reader will read notify user that there is on this button.

jbpenrath, line 116: I think there is an a11y issue to use a button here. This element should appear as a link.
```
          <Link to={breadcrumb.pathname} size="tiny">
```

src/frontend/js/components/Dashboard/DashboardSidebar.tsx
```diff
@@ -0,0 +1,91 @@
...
+            onChange={onSelectChange}
+            className="form-field__select-input"
+          >
+            {links.map((link) => (
+              <option key={link.to} value={link.to}>
+                {link.label}
+              </option>
+            ))}
+          </select>
+        </div>
```
jbpenrath, line 79: Ditto, you should unfocus this element when a value is selected. I think you should use downshift to build a accessible select.

src/frontend/js/components/Dashboard/_styles.scss
```diff
@@ -13,4 +32,120 @@
...
+      &__header {
+        padding: 4rem 1rem 1rem 1rem;
+        position: relative;
+        display: flex;
+        flex-direction: column;
+        align-items: center;
+
+        &__avatar {
+          position: absolute;
+          top: -50px;
```
jbpenrath, line 59: I'll use a variable $avatar-size: 100px; It will ease maintainability.

src/frontend/js/components/Dashboard/_styles.scss
```diff
@@ -13,4 +32,120 @@
...
+        }
+
+        li {
+          padding: 0.5rem;
+          border-top: 1px r-theme-val(dashboard-sidebar, base-border) solid;
+
+          a {
+            color: r-theme-val(dashboard-sidebar, base-color);
+            font-weight: 600;
+            font-family: 'Montserrat', sans-serif;
```
jbpenrath, line 97:
```suggestion
            font-family: $r-font-family-montserrat;
```

src/frontend/js/utils/routers/dashboard/index.tsx
```diff
@@ -66,22 +66,22 @@ const dashboardRouteLabels = defineMessages<DashboardPaths>({
   [DashboardPaths.PREFERENCES]: {
     id: 'components.Dashboard.DashboardRoutes.preferences.label',
     description: 'Label of the preferences view used in navigation components.',
-    defaultMessage: 'My preferences',
+    defaultMessage: 'My Preferences',
```
jbpenrath, line 69: I'll do inverse : `'My courses'` L59

src/frontend/js/components/Dashboard/DashboardBreadcrumbs.tsx
```diff
@@ -0,0 +1,123 @@
+import { useMatches, useNavigate } from 'react-router-dom';
```
jbpenrath, line 1: This file module does too much things. Can you split it in separate modules ?

`BreadcrumbProvider`, `useBreadcrumbPlaceholders`,  `useBreadcrumbParts` then `DashboardBreadcrumbs` or at least `BreadcrumbProvider` and `DasbhoardBreadcrumbs`

src/frontend/js/components/Dashboard/DashboardBreadcrumbs.tsx
```diff
@@ -0,0 +1,123 @@
...
+      location.replace('/');
+      return;
+    }
+    navigate(breadcrumbs[breadcrumbs.length - 2].pathname);
+  };
+
+  return (
+    <ul className="dashboard-breadcrumbs" data-testid="dashboard-breadcrumbs">
+      <li>
+        <Button onClick={back} size="tiny">
```
jbpenrath, line 108: Ditto for a11y purpose it should appear as a link.

src/frontend/js/components/Dashboard/DashboardBreadcrumbs.tsx
```diff
@@ -0,0 +1,123 @@
...
+      const { crumbLabel } = match.handle as { crumbLabel: MessageDescriptor };
+      return IntlHelper.doValuesExists(crumbLabel, context.meta.placeholders);
+    })
+    .map((match) => {
+      const { crumbLabel } = match.handle as { crumbLabel: MessageDescriptor };
+      return {
+        pathname: match.pathname,
+        name: intl.formatMessage(crumbLabel, context.meta.placeholders),
+      };
+    });
```
jbpenrath, line 90: What do you think to use `reduce` instead of 2 `filters` and one `map` ?
```suggestion
return matches.reduce((crumbs, match) => {
    const label: MessageDescriptor = (match.handle as any)?.crumbLabel;

    if (label && IntlHelper.doValuesExists(label, context.meta.placeholders)) {
      crumbs.push({
        name: intl.formatMessage(label, context.meta.placeholders),
        pathname: match.pathname,
      });
    }

    return crumbs;
  }, [] as Record<'pathname' | 'name', string>[]);
```

src/frontend/js/components/Dashboard/DashboardLayout.tsx
```diff
@@ -0,0 +1,30 @@
...
+import { DashboardSidebar } from 'components/Dashboard/DashboardSidebar';
+import {
+  DashboardBreadcrumbs,
+  DashboardBreadcrumbsProvider,
+} from 'components/Dashboard/DashboardBreadcrumbs';
+
+const LocationDisplay = () => {
+  const location = useLocation();
+  return <div data-testid="location-display">{location.pathname}</div>;
+};
```
jbpenrath, line 11: I think it's weird to add content to the dom only for testing purpose.

src/frontend/js/components/Dashboard/DashboardLayout.tsx
```diff
@@ -0,0 +1,30 @@
...
+export const DashboardLayout = () => {
+  return (
+    <DashboardBreadcrumbsProvider>
+      <div className="dashboard">
+        <DashboardSidebar />
+        <div className="dashboard__content">
+          <header>
+            <DashboardBreadcrumbs />
+          </header>
+          <main>
```
jbpenrath, line 22: One quick idea :
```suggestion
          <main data-testid={`location-display-${location.pathname}`}>
```

src/frontend/js/components/Dashboard/index.spec.tsx
```diff
@@ -40,16 +44,24 @@ describe('<Dashboard />', () => {
...
   beforeEach(() => {
     jest.resetAllMocks();
+    fetchMock.get('https://joanie.endpoint/api/orders/', []);
+    fetchMock.get('https://joanie.endpoint/api/credit-cards/', []);
+    fetchMock.get('https://joanie.endpoint/api/addresses/', []);
+  });
+
+  afterEach(() => {
+    jest.clearAllMocks();
+    fetchMock.restore();
```
jbpenrath, line 64: Why do you need to do this ? From my point of view, `beforeEach` is already in charge to clean those things.

src/frontend/js/utils/IntlHelper/index.spec.tsx
```diff
@@ -0,0 +1,26 @@
+import { IntlHelper } from 'utils/IntlHelper/index';
+
+describe('IntlHelper', () => {
```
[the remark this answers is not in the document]
NathanVss: Right, good catch

src/frontend/js/components/Dashboard/DashboardLayout.tsx
NathanVss: This should work I think

src/frontend/js/utils/routers/dashboard/index.tsx
NathanVss: I don't understand 😅

src/frontend/js/components/Dashboard/DashboardBreadcrumbs.tsx
NathanVss: 100% agreed

src/richie/apps/core/templates/richie/icons.html
```diff
@@ -24,6 +24,10 @@
       <path d="M9.2925,6.29376218 C8.9025,6.68376218 8.9025,7.31376218 9.2925,7.70376218 L13.1725,11.5837622 L9.2925,15.4637622 C8.9025,15.8537622 8.9025,16.4837622 9.2925,16.8737622 C9.6825,17.2637622 10.3125,17.2637622 10.7025,16.8737622 L15.2925,12.2837622 C15.6825,11.8937622 15.6825,11.2637622 15.2925,10.8737622 L10.7025,6.28376218 C10.3225,5.90376218 9.6825,5.90376218 9.2925,6.29376218 Z" id="Fill-1" transform="translate(12.292500, 11.583131) scale(-1, 1) rotate(-270.000000) translate(-12.292500, -11.583131) "></path>
     </symbol>
 
+    <symbol id="icon-chevron-left" viewBox="0 0 24 24">
+      <path d="M17.67 3.77L15.9 2L6 11.9L15.9 21.8L17.67 20.03L9.54 11.9L17.67 3.77Z" fill="currentColor"/>
+    </symbol>
+
```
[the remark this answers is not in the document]
NathanVss: They are not the same icons, let's talk about this in a meeting !

src/frontend/js/components/Dashboard/DashboardBreadcrumbs.tsx
NathanVss: TODO: Try to use href for a11y purpose.

src/frontend/js/components/Dashboard/DashboardSidebar.tsx
NathanVss: Add `aria-label` + unBlur

src/frontend/js/components/Dashboard/index.spec.tsx
NathanVss: I can suppress `jest.clearAllMock();` but I need to keep `fetchMock.restore();` to be able de define mocks before each tests.

src/frontend/js/components/Dashboard/DashboardBreadcrumbs.tsx
NathanVss: I really like this approach

src/frontend/js/components/RouterButton/index.tsx
```diff
@@ -0,0 +1,30 @@
...
+      onClick={(event: MouseEvent) => {
+        event.preventDefault();
+        if (props.href?.match('^http')) {
+          // We must do this to be able to tests redirects.
+          location.replace(props.href);
+          return;
+        }
+        navigate(props.href!);
+        ref.current?.blur();
+      }}
```
jbpenrath, line 26: In fact, if `props.href?.match('^http')` we could simply execute nothing.

```suggestion
onClick={(event: MouseEvent) => {
        const isDashboardRoute = !props.href?.match('^http');
        if (!isDashboardRoute) return;

        event.preventDefault();
        navigate(props.href!);
        ref.current?.blur();
      }}
```

src/frontend/js/utils/test/expectUrlMatchLocationDisplayed.ts
```diff
@@ -0,0 +1,6 @@
+import { screen } from '@testing-library/react';
+
+export const expectUrlMatchLocationDisplayed = (url: string) => {
+  // expect(screen.getByTestId('location-display').textContent).toMatch(new RegExp(url + '$'));
```
jbpenrath, line 4:
```suggestion
```

src/frontend/js/components/RouterButton/index.tsx
jbpenrath: `navigate` has not been called ?

NathanVss: I did the change, but unfortunately this forced me to mock it also for the first test ( "navigates inside the router" ) because I didn't succeed to mock it ONLY for the second test. 

So the consequence is that we lost the real navigation test during the first test. 

I spent the morning on it and I couldn't find a way to do it, so if you're willing to do a pair programming to find a way to do it, I'm you're guy because I tried everything Jest provides .. 😅

src/frontend/js/components/Dashboard/_styles.scss
```diff
@@ -0,0 +1,153 @@
+.richie-react--dashboard {
+  background-color: r-theme-val(dashboard, background-color);
+}
+
+div[data-testid='location-display'] {
+  display: none;
+}
```
jbpenrath, line 7: No more relevant
```suggestion
```

src/frontend/js/components/DashboardAddressesManagement/index.tsx
```diff
@@ -60,32 +60,33 @@ export const DashboardAddressesManagement = ({
...
+        {isLoading && <Spinner />}
+        {!isLoading && (
+          <>
+            {error && <Banner message={error} type={BannerType.ERROR} rounded />}
+            {!error && addressesList.length === 0 && (
+              <p className="dashboard-addresses__empty">
+                <FormattedMessage {...messages.emptyList} />
+              </p>
+            )}
+            {addressesList.map((address) => (
```
jbpenrath, line 75: Address list should be only displayed if there is no error right ?
```suggestion
            {!error && addressesList.map((address) => (
```

src/frontend/js/components/DashboardBreadcrumbs/DashboardBreadcrumbsProvider.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+export const DashboardBreadcrumbsContext = createContext<DashboardBreadcrumbsContextInterface>(
+  {} as any,
+);
+
+/**
+ * Provider that needs to wrap any component using the DashboardBreadcrumbs component.
+ * It provides a context used to make formatjs value available bottom up.
+ */
+export const DashboardBreadcrumbsProvider: FC = ({ children }) => {
```
jbpenrath, line 20: For the future ^^
```suggestion
export const DashboardBreadcrumbsProvider = ({ children }: PropsWithChildren<{}>) => {
```

jbpenrath (approved):
LGTM ✅

PR #1811, opened 2022-10-24
✨(search) allows different search and courses page

### LINKED ISSUE #1810: Incorrect course search links with different courses parent page and search page
## Bug Report

Issues when Richie is configured with different pages, one for the course parent page and another for the search page.

On NAU we have the `/en/courses` page, that allows to search the courses, but the parent courses page is the `/en/course`. This second page has the id `courses`, so Richie knows where to put the newer course pages, and it redirect to the first `/en/courses` page.

We have configured this because, this was how our older marketing site works. And it seems to be working fine.
Recently we have found an issue, that the organization detail page, categories page, etc, are refering to the correct courses page search. They are refering the `/en/course` page with additional query parameters. But the redirect on the page lost the query parameters.
So for the user, when he goes to the organization page and clicks to view all the courses for that organization, it's lost the organization filter during the redirect. 

I think that the major issue is that Richie has 2 different concepts mixed toogether:
1. the search page
2. the parent course page
On a simpler solution both could be the same page, but I think Richie should allow different configurations. For example the search could evolve to search more stuff, like programs and not only courses.

**Possible Solution**

Add a new Django setting that allow to change the `courses` search reverse id.

igobranco:
Add `RICHIE_SEARCH_REVERSE_ID` setting that allows to have different search page and courses parent page.
Closes #1810

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Added
 
+- Add `RICHIE_SEARCH_REVERSE_ID` setting that allows to have different search
+  page and courses parent page.
 - Add Dashboard router
 - Add generic dashboard component
 - Add dashboard components for Order, Enrollment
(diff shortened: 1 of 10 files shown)


igobranco:
Closing in favor of this PR https://github.com/django-cms/django-cms/pull/7489 to Django CMS.

PR #1549, opened 2021-12-17
✨(course) add course run catalog visibility

igobranco:
Add catalog visibility field to course run so it's possible to hide a
course run on catalog and/or hide it completely. Closes #1502

PR #1813, opened 2022-10-25
Fix synchronization with openedx

sampaccoud:
### Purpose

We were having course run synchronization issues on our sites since version 2.15.0.

After investigation, it appeared that the synchronization hook was misbehaving since PR https://github.com/openfun/richie/pull/1549. Indeed the no-update fields were removed from the payload for create as well as update queries which lead to 400 errors because the fields removed are required to create a course run.

### Proposal

Split "field conversion" and "removing no-update fields" in 2 separated actions to play them only when they are respectively required.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,11 +14,12 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 - Add generic dashboard component
 - Add dashboard components for Order, Enrollment
 - Add dashboard components for Credit Card management
+- Dashboard addresses management components
 
 ### Changed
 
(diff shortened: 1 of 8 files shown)


jbpenrath (approved):
LGTM ✅

jbpenrath:
@sampaccoud can you just refresh my mind on a point. Why do you not include migration file into this PR since a model has changed ?

sampaccoud:
> @sampaccoud can you just refresh my mind on a point. Why do you not include migration file into this PR since a model has changed ?

Well in this case I just added a `verbose_name` parameter which has absolutely no impact on the database schema... So there's no point adding one more migration file for nothing :man_shrugging:

PR #1815, opened 2022-10-27
Feature/joanie handler

jbpenrath:
## Purpose

Joanie will be soon able to synchronize course runs and products (interpolated as course run) into Richie. In this way, all products and course runs managed by Joanie will be indexed by Richie with ease. So from Richie, we have to implement an interface able to identify course runs managed by Joanie.

## Proposal

- [x] Implement JoanieBackend able to identify course runs managed by Joanie
- [x] Add template tag to identify course runs which are products
- [x] Write tests
- [x] Add all methods to JoanieBackend required to synchronize course runs

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -16,6 +16,27 @@ $ make migrate
 
 ## Unreleased
 
+## 2.16.x to 2.17.x
+
+- Richie is now able to delegate course enrollment to Joanie which will then synchronize its
+  course runs with Richie and provide advanced enrollment, ecommerce and multi-LMS features.
+  To set up this feature, you can add to `JOANIE_BACKEND` setting the following properties:
+  ```python
+  JOANIE_BACKEND = {
+    # ...
+    "BACKEND": "richie.apps.courses.lms.joanie.JoanieBackend",
+    "JS_BACKEND": "joanie",
+    # The regex to match a resource that can be handle by the lms.joanie.JoanieBackend interface
+    "COURSE_REGEX": r"^.*/api/(?P<resource_type>(course-runs|products))/(?P<resource_id>.*)/?$",
+    # The regex to match a resource that can be handle by the lms.joanie.JoanieBackend interface
+    "JS_COURSE_REGEX": r"^.*/api/(course-runs|products)/(.*)/?$",
+    # A list of course run properties to not update
+    "COURSE_RUN_SYNC_NO_UPDATE_FIELDS": [],
+    # The synchronization mode ("manual", "sync_to_public" or "sync_to_draft") 
+    "DEFAULT_COURSE_RUN_SYNC_MODE": "sync_to_public",
+  }
+  ```
+
 ## 2.15.x to 2.16.x
 
 - The way filters are configured has changed. There is now a separate setting
--- a/sandbox/settings.py
+++ b/sandbox/settings.py
@@ -618,6 +618,10 @@ class Base(StyleguideMixin, DRFMixin, RichieCoursesConfigurationMixin, Configura
             with sentry_sdk.configure_scope() as scope:
                 scope.set_extra("application", "backend")
 
+        # If a Joanie Backend has been configured, we add it into LMS_BACKENDS dict
+        if cls.JOANIE_BACKEND.get("BASE_URL") is not None:
+            cls.RICHIE_LMS_BACKENDS.append(cls.JOANIE_BACKEND)
+
 
 class Development(Base):
     """
@@ -663,7 +667,7 @@ class Test(Base):
 
 class ContinuousIntegration(Test):
     """
-    Continous Integration environment settings
+    Continuous Integration environment settings
 
     nota bene: it should inherit from the Test environment.
     """
--- a/sandbox/settings.py
+++ b/sandbox/settings.py
@@ -299,11 +299,35 @@ class Base(StyleguideMixin, DRFMixin, RichieCoursesConfigurationMixin, Configura
 
     # Joanie
     """
-    NB: Richie picks all Joanie's settings from the JOANIE namespace on the
-    settings, hence the nesting of all Joanie's values inside that prop
+    NB: Richie picks all Joanie's settings from the JOANIE_BACKEND namespace in the
+    settings, hence the nesting of all Joanie's values inside that prop.
+
+    If BASE_URL is defined, this setting is bound into RICHIE_LMS_BACKENDS to use Joanie
+    as a LMS BACKEND.
     """
-    JOANIE = {
+    JOANIE_BACKEND = {
         "BASE_URL": values.Value(environ_name="JOANIE_BASE_URL", environ_prefix=None),
+        "BACKEND": values.Value(
+            "richie.apps.courses.lms.joanie.JoanieBackend",
+            environ_name="JOANIE_BACKEND",
+            environ_prefix=None,
+        ),
+        "JS_BACKEND": values.Value(
+            "joanie", environ_name="JOANIE_JS_BACKEND", environ_prefix=None
+        ),
+        "COURSE_REGEX": values.Value(
+            r"^.*/api/(?P<resource_type>(course-runs|products))/(?P<resource_id>.*)/?$",
+            environ_name="JOANIE_COURSE_REGEX",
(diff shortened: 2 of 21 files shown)


src/richie/apps/courses/templatetags/extra_tags.py
```diff
@@ -23,7 +23,9 @@
 from cms.utils import get_site_id
 from cms.utils.plugins import get_plugins
 
+from .. import enums
```
sampaccoud, line 26: If I'm not mistaken this file does what we usually place in a  `defaults.py` file

src/richie/apps/courses/lms/__init__.py
```diff
@@ -24,8 +28,14 @@ def select_lms(url):
...
 
+        # First check if it matches a configured LMS
         for lms_configuration in settings.RICHIE_LMS_BACKENDS:
             if re.match(lms_configuration.get("COURSE_REGEX", r".*"), url):
                 return import_string(lms_configuration["BACKEND"])(lms_configuration)
 
+        # Second check if it matches Joanie
+        if is_joanie_enabled():
+            if re.match(settings.JOANIE.get("RESOURCE_REGEX", r"^$"), url):
+                return JoanieBackend(settings.JOANIE)
```
sampaccoud, line 39: Actually, this would not be necessary if Joanie was declared as a normal backend in settings...
Is it a bad idea that Joanie could just be an LMS backend like the others?

jbpenrath: Let's think about that ?

So to configure Joanie, we could use a `DictValue` which default to None ?
```
RICHIE_LMS_BACKENDS = [
        {
            "BASE_URL": "http://localhost:8073",
            "BACKEND": "richie.apps.courses.lms.edx.EdXLMSBackend",
            "COURSE_REGEX": r"^.*/courses/(?P<course_id>.*)/course/?$",
            "JS_BACKEND": "dummy",
            "JS_COURSE_REGEX": r"^.*/courses/(.*)/course/?$",
        },
        values.DictValue(None, JOANIE_LMS_BACKEND, environ_prefix=None)
    ]
```

Then to know if joanie is enabled we could check if we found within `RICHIE_LMS_BACKENDS` a backend with a `BACKEND` equals to `richie.apps.course.lms.joanieJoanieBackend` ?

jbpenrath: As discussed this morning, let's keep `JOANIE` settings for now then later we will move those settings into a model.

src/richie/apps/courses/lms/joanie.py
```diff
@@ -0,0 +1,37 @@
...
+        from the provided resource_link.
+        """
+        return self._extract_info(resource_link, "resource_type")
+
+    def extract_resource_id(self, resource_link):
+        """
+        Try to extract the resource id (products or course runs)
+        from the provided resource_link.
+        """
+        return self._extract_info(resource_link, "resource_id")
```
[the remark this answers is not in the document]
jbpenrath: What is the change ?

UPGRADE.md
```diff
@@ -16,6 +16,26 @@ $ make migrate
 
 ## Unreleased
 
+## 2.16.x to 2.17.x
+
+- Joanie is now able to call the richie course run synchronization webhook. To set up this feature,
```
sampaccoud, line 21:
```suggestion
- Richie is now able to delegate course enrollment to Joanie which will then synchronize its course runs with Richie and provide advanced enrollment, ecommerce and multi-LMS features. 
To set up this feature,
```

sandbox/settings.py
```diff
@@ -299,11 +299,29 @@ class Base(StyleguideMixin, DRFMixin, RichieCoursesConfigurationMixin, Configura
 
     # Joanie
     """
-    NB: Richie picks all Joanie's settings from the JOANIE namespace on the
-    settings, hence the nesting of all Joanie's values inside that prop
+    NB: Richie picks all Joanie's settings from the JOANIE_BACKEND namespace on the
```
sampaccoud, line 302:
```suggestion
    NB: Richie picks all Joanie's settings from the JOANIE_BACKEND namespace in the
```

src/richie/apps/core/templatetags/joanie.py
```diff
@@ -8,12 +8,12 @@
 @register.simple_tag()
 def is_joanie_enabled():
     """
-    Determines if Joanie is enabled
-
-    Within settings, JOANIE can be enabled/disabled by setting the value of
-    `JOANIE.BASE_URL` with the url of joanie endpoint.
+    Determines if Joanie is enabled by checking if there is a backend
+    with NAME sets to joanie within RICHIE_LMS_BACKENDS settings.
```
sampaccoud, line 12:
```suggestion
    with NAME set to joanie within RICHIE_LMS_BACKENDS settings.
```

sandbox/settings.py
```diff
@@ -594,6 +612,11 @@ def post_setup(cls):
             with sentry_sdk.configure_scope() as scope:
                 scope.set_extra("application", "backend")
 
+        # If a Joanie Backend has been configured, we add it into LMS_BACKENDS dict
+        if cls.JOANIE_BACKEND.get('BASE_URL') is not None:
+            cls.RICHIE_LMS_BACKENDS.append({ 'NAME': 'joanie', **cls.JOANIE_BACKEND })
```
sampaccoud, line 617: in order to avoid adding this "joanie" name, you could have a property "is_joanie = True" on the JoanieBackend so when you import the path of the backend given in JOANIE_BACKEND["BACKEND"], you will know it is one?

src/richie/apps/courses/lms/__init__.py
```diff
@@ -4,6 +4,10 @@
 from django.conf import settings
 from django.utils.module_loading import import_string
 
+from richie.apps.core.templatetags.joanie import is_joanie_enabled
+
+from ..lms.joanie import JoanieBackend
```
sampaccoud, line 9: unused?

sandbox/settings.py
jbpenrath: Yep indeed. But in this way, I have to import and instantiate each LMS backends configured to check if this is Joanie one.

src/richie/apps/courses/lms/joanie.py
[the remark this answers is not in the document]
jbpenrath: From my point of view it adds complexity to those methods for a tiny perf improvement. This class is instantiated for one shot usage, so it almost all cases, the regex will be executed.

sandbox/settings.py
sampaccoud: it can be a class attribute so you only need to import but not instantiate it.

src/richie/apps/courses/templatetags/extra_tags.py
```diff
@@ -227,6 +229,22 @@ def has_connected_lms(course_run):
...
+@register.filter()
+def is_joanie_product(course_run):
+    """
+    Check that the course run is managed by Joanie
+    then the resource type is equal to `products`.
+    """
+    handler = LMSHandler.select_lms(course_run.resource_link)
+
+    if not isinstance(handler, JoanieBackend):
+        return False
```
sampaccoud, line 241: Should we use the new class attribute?
```suggestion
    if not getattr(handler, "is_joanie", False):
        return False
```

jbpenrath: Yep. It improves cohesion.

src/richie/apps/courses/lms/__init__.py
```diff
@@ -12,6 +12,20 @@ class LMSHandler:
...
+        Return all enabled LMS classes.
+        """
+        lms_classes = set()
+        for lms_configuration in settings.RICHIE_LMS_BACKENDS:
+            try:
+                lms_classes.add(import_string(lms_configuration["BACKEND"]))
+            except ImportError:
+                pass
+
+        return lms_classes
```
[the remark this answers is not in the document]
jbpenrath: Indeed, in this way we have the same behaviour as `select_lms` method

sampaccoud (approved):
Nice! :handshake:

PR #1817, opened 2022-10-28
✅(frontend) fix flaky tests

jbpenrath:
## Purpose

The DashboardCreditCardsManagement test suite was flaky. In some case the test in charge to check if the right class is displayed when a credit card soon expire could generate an expiration date in the wrong interval.

## Proposal

- [x] Remove flaky test

### DIFF
--- a/src/frontend/js/components/DashboardCreditCardsManagement/index.spec.tsx
+++ b/src/frontend/js/components/DashboardCreditCardsManagement/index.spec.tsx
@@ -113,9 +113,10 @@ describe('<DashboardCreditCardsManagement/>', () => {
   });
 
   it('renders the correct label for an expiration date that will soon expire', async () => {
-    const limit = new Date();
-    limit.setMonth(limit.getMonth() + 1);
-    const date = faker.date.future(0.2, limit);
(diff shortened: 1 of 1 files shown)


src/frontend/js/components/DashboardCreditCardsManagement/index.spec.tsx
```diff
@@ -113,9 +113,7 @@ describe('<DashboardCreditCardsManagement/>', () => {
   });
 
   it('renders the correct label for an expiration date that will soon expire', async () => {
-    const limit = new Date();
-    limit.setMonth(limit.getMonth() + 1);
-    const date = faker.date.future(0.2, limit);
+    const date = faker.date.future(3 / 12); // [now < date < +3 months]
```
NathanVss, line 116:
```suggestion
   const offset = new Date();
    offset.setMonth(offset.getMonth() + 1);
    offset.setDate(1);
    const date = faker.date.future(2 / 12, offset);
```

To prevent flaky tests more prone during starts of months.

NathanVss (approved):
🥳

PR #1820, opened 2022-11-03
✅(frontend) fix flaky tests

NathanVss:
There was a flaky test for CreditCardHelper, it relates exactly to the same behavior that we previously fixed for DashboardCreditCardsManagement but we missed this one.


jbpenrath (approved):
LGTM :white:

PR #1824, opened 2022-11-09
🐛(frontend) fix courses badges css

NathanVss:
This bug was revealed by the 0bb70196c7c42bb3bceab4e06d60baadff8dc207 commit, by transforming all buttons to flex containers some css line that were not getting took into account before went to life resulting in this bug. So that means the existing code worked but ... by chance.


jbpenrath (commented):
Add an fixed entry in the changelog

jbpenrath (approved):

PR #1833, opened 2022-11-22
✨(api) allow bulk course run synchronization

sampaccoud:
## Purpose

When a course run is updated in Joanie, it may trigger the re-synchronization of many course runs in Richie (corresponding to a Joanie product). We don't want to have to call the synchronization hook too many times in a synchronous request.

## Proposal

Add bulk course synchronization to the existing API hook.

Single course run synchronization still works exactly the same as the existing tests can confirm.

### DIFF
--- a/src/richie/apps/courses/api.py
+++ b/src/richie/apps/courses/api.py
@@ -7,10 +7,13 @@ from django.db.models import Q
 from cms import signals as cms_signals
 from cms.models import Page
 from rest_framework.decorators import api_view
+from rest_framework.exceptions import ValidationError
 from rest_framework.permissions import BasePermission
 from rest_framework.response import Response
+from rest_framework.serializers import as_serializer_error
 from rest_framework.viewsets import ModelViewSet
 
+from .exceptions import MissingResourceLinkError
 from .lms import LMSHandler
 from .models import Course, CourseRun, CourseRunSyncMode
 from .serializers import CourseRunSerializer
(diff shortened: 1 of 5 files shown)


tests/apps/courses/test_api_course_run_sync_edx.py
```diff
@@ -1233,3 +1233,203 @@ def test_api_course_run_sync_catalog_visibility_about(self, mock_signal):
...
         self.assertFalse(mock_signal.called)
+
+    # Bulk
+
+    @override_settings(
+        RICHIE_DEFAULT_COURSE_RUN_SYNC_MODE="sync_to_public", TIME_ZONE="utc"
+    )
+    def test_api_course_run_sync_create_bulk_success(self, mock_signal):
+        """
+        It should be possible to synchronize a list of course runs in bulk.
```
jbpenrath, line 1244: You will have to move this test within `test_api_course_run_sync.py` file when you will update our branch.

src/richie/apps/courses/api.py
```diff
@@ -7,10 +7,13 @@
 from cms import signals as cms_signals
 from cms.models import Page
 from rest_framework.decorators import api_view
+from rest_framework.exceptions import ValidationError
 from rest_framework.permissions import BasePermission
 from rest_framework.response import Response
+from rest_framework.serializers import as_serializer_error
```
jbpenrath, line 13: 👍

jbpenrath (approved):
LGTM ✅

tests/apps/courses/test_api_course_run_sync_edx.py
sampaccoud: Done! Good catch :astonished:

PR #1835, opened 2022-11-22
Doc/cookiecutter use checkout argument

jbpenrath:
## Purpose

Currently our cookiecutter template is broken. The PR #1826 seems to fix the issue but this error highlights a weakness in our workflow. Indeed, anyone reading our cookiecutter documentation will fetch template from master branch and currently we are not able to check if the template is working properly on this branch.

So to prevent this kind of error, I suggest two improvements : 
- Upgrade cookiecutter documentation by adding `--checkout` argument sets with the latest tag release in commands to fetch cookiecutter template from the latest release instead of master.
- Run `cookiecutter-bootstrap` on master branch to be aware if cookiecutter template is broken on this branch.

## Proposal

- [x] Upgrade cookiecutter documentation 
- [x] Run cookiecutter-bootstrap against master branch.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,7 +12,11 @@ generate-version-file: &generate-version-file
         "$CIRCLE_PROJECT_REPONAME" \
         "$CIRCLE_BUILD_URL" > sandbox/version.json
 
-version: 2
+version: 2.1
+parameters:
+  nightly-round:
+    type: boolean
+    default: false
 
 aliases:
   - &checkout_fun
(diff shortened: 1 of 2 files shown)


jbpenrath:
@sampaccoud I followed your idea. I bumped Circle CI configuration file to version 2.1 to be able to run workflows conditionally. Then I created a schedule trigger to run a workflow called `nightly-round` each night at 3:00am. This workflow only run `cookiecutter-bootstrap` job.

I'll try later to improve this workflow to open an issue on github or push a notification into a slack channel when the job fail.

.circleci/config.yml
```diff
@@ -775,12 +782,12 @@ jobs:
...
       - run:
-         name: Is a release branch
+         name: Is a release or master branch
          command: |
-           # We want to run this job only on release branch, to guess that
+           # We want to run this job only on release or master branch, to guess that
            # we check if a new release entry has been added into the CHANGELOG
            # If there is no new release entry, git diff returns a 0 exit code
-           if git diff --quiet -G"^## \[[0-9.]*\]$" origin/master..HEAD CHANGELOG.md; then
+           if [ "$CIRCLE_BRANCH" != "master" ] | git diff --quiet -G"^## \[[0-9.]*\]$" origin/master..HEAD CHANGELOG.md; then
```
sampaccoud, line 790: Not sure why you need to change this? This will have the effect to run cookie cutter on all master builds which is not what you modified in the above comment?

jbpenrath: Before your idea to add a nightly round, I first updated circle ci configuration to run this job on master branch. I kept this change then I added a new commit to setup the nightly round.

sampaccoud: yes but my point is that the comment above should be:
```
# We want to run this job only when a release is merged to the master branch, to guess that
```

jbpenrath: Oh ok!!

jbpenrath: I updated the condition. No I stop the job if the branch is not master and there is no release commit.

sampaccoud (approved):

PR #1836, opened 2022-11-22
🔨(bin) stop using env variable to activate sites

sampaccoud:
## Purpose

The developper experience was not good as we had to run `bin/activate` and then export a variable. Many users forgot to export the variable expecting the activate script to do what it claims to do...

## Proposal

We now use the `.env` file to record the active site (similar to what is done for workspaces in Terraform). This allows a much better DX as running `bin/activate` now effectively activates the site.

We further improved DX by automatically activating the site without any prompt when there is only one site.

### DIFF
--- a/cookiecutter/{{cookiecutter.organization}}-richie-site-factory/Makefile
+++ b/cookiecutter/{{cookiecutter.organization}}-richie-site-factory/Makefile
@@ -1,5 +1,3 @@
-RICHIE_SITE ?= funmooc
-
 # -- Terminal colors
 COLOR_INFO    = \033[0;36m
 COLOR_RESET   = \033[0m
@@ -42,6 +40,7 @@ MANAGE = $(COMPOSE_RUN_APP) python manage.py
 default: help
 
 bootstrap: \
+  .env \
   env.d/aws \
   env.d/development \
   data/media/$(RICHIE_SITE)/.keep \
@@ -56,7 +55,7 @@ bootstrap: \
 bootstrap:  ## install development dependencies
 .PHONY: bootstrap
 
-add-site: ## add a new site to the site factory
+generate-site: ## generate a new site using cookiecutter
 	@docker run --rm -it -e LC_ALL=C.UTF-8 \
 		-u $(DOCKER_UID):$(DOCKER_GID) \
 		-v $(PWD):/app \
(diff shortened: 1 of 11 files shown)


cookiecutter/{{cookiecutter.organization}}-richie-site-factory/bin/activate
```diff
@@ -38,11 +43,5 @@ if [[ ${choice} -le 0 ]]; then
...
-RICHIE_SITE="${sites[$((choice-1))]}"
-
-if [[ "${BASH_SOURCE[0]}" != "${0}" ]] ; then
-    export RICHIE_SITE
-else
-    echo -e "\\n# Copy/paste ${RICHIE_SITE} activation command:"
-    echo "export RICHIE_SITE=${RICHIE_SITE}"
-fi
+# Save activated site to file
+echo "${sites[$((choice-1))]}" > .site
```
jbpenrath, line 47: From my point of view, I seems legit to put this file into env.d folder. WDYT?
```suggestion
echo "${sites[$((choice-1))]}" > env.d/site
```

jbpenrath (commented):
Is there no way to export an environment variable through a shell script ? 
Because currently there is one remaining drawback. We have to export RICHIE_SITE variable to be able to run docker compose commands not managed into Makefile.

cookiecutter/{{cookiecutter.organization}}-richie-site-factory/bin/activate
sampaccoud: Good idea :pray:

sampaccoud:
> Is there no way to export an environment variable through a shell script ? Because currently there is one remaining drawback. We have to export RICHIE_SITE variable to be able to run docker compose commands not managed into Makefile.

Your other comment about moving the `.site` file to `env.d` gave me an idea. This [new proposal](https://github.com/openfun/richie/pull/1836/commits/377abcb895c1aa913be8869fcdaa812c2ab1ecc9) makes use of the `.env` file to which docker-compose defaults to load environment variables. This way both the `docker-compose` and the `Makefile` work when a site is activated.

I also improved the scripts a bit to automatically activate the site when there is only one.

WDYT?

cookiecutter/{{cookiecutter.organization}}-richie-site-factory/docker-compose.yml
```diff
@@ -4,13 +4,13 @@ services:
   db:
     image: postgres:9.6
     environment:
-      - POSTGRES_DB=richie_${RICHIE_SITE}
+      - POSTGRES_DB=richie_${RICHIE_SITE:?}
```
jbpenrath, line 7: Not quit sure to understand the purpose of this syntax.

cookiecutter/{{cookiecutter.organization}}-richie-site-factory/Makefile
```diff
@@ -67,86 +65,89 @@ add-site: ## add a new site to the site factory
...
 
-stop: .site ## stop the development server
+stop: .env ## stop the development server
 	@$(COMPOSE) stop
 .PHONY: stop
 
-info: .site ## get activated site info
+info: .env ## get activated site info
 	@echo "RICHIE_SITE: $(COLOR_INFO)$(RICHIE_SITE)$(COLOR_RESET)"
 .PHONY: info
```
jbpenrath, line 112: All other commands seems to work but this one nope.
```
$> make info
RICHIE_SITE:
```

cookiecutter/{{cookiecutter.organization}}-richie-site-factory/docker-compose.yml
sampaccoud: it's to make it required. Otherwise it evaluates to an empty string and we start creating directories where we don't want to! :sweat_smile:

cookiecutter/{{cookiecutter.organization}}-richie-site-factory/Makefile
sampaccoud: Good catch!! Easy one :wink:

cookiecutter/{{cookiecutter.organization}}-richie-site-factory/docker-compose.yml
jbpenrath: Oh ! From my side if RICHIE_SITE is not set, docker compose returns an error. So with or without `:?`, I have the same behaviour.

PR #1840, opened 2022-11-30
♻️(frontend) update course product item

jbpenrath:
## Purpose

Now that Joanie push course runs into Richie, the CourseProductList component is no more useful. Instead we render CourseProductItem and retrieve product information directly.

## Proposal

- [x] Update course_detail template to render a CourseProductItem when a CourseRun is recognized as a Joanie product
- [x] Refactor CourseProductItem
- [x] Remove CourseProductList stuff
- [x] Update tests
- [x] Add new tests


src/frontend/js/utils/ObjectHelper/index.spec.ts
```diff
@@ -0,0 +1,19 @@
+import { ObjectHelper } from 'utils/ObjectHelper/index';
+
+describe('ObjectHelper', () => {
```
NathanVss, line 3: I would add some edge cases tests like: 

- `{foo: undefined}`
- `{foo: null}`
- `{foo: 0}`

src/frontend/js/components/CourseProductCourseRuns/EnrollableCourseRunList.tsx
```diff
@@ -110,14 +110,16 @@ const EnrollableCourseRunList = ({ courseRuns, order }: Props) => {
       if (relatedEnrollment) {
         await enrollment.methods.update({
           is_active: true,
-          course_run: selectedCourseRun.resource_link,
+          course_run: selectedCourseRun.id,
           id: relatedEnrollment.id,
+          was_created_by_order: relatedEnrollment.was_created_by_order,
```
NathanVss, line 111: Why not forcing it to true ? If the `relatedEnrollment` already exists and is re-activated through an order, then it must be considered to be created from an order, nope ?

src/frontend/js/utils/api/joanie.ts
```diff
@@ -211,19 +218,15 @@ const API = (): Joanie.API => {
             method: 'POST',
             body: JSON.stringify(payload),
           }).then(checkStatus),
-        get: async (arg1) => {
+        get: async (filters) => {
```
NathanVss, line 218: Nice, I'm going to re-use the way you refactored this function in the order PR too :)

src/frontend/js/components/CourseProductItem/index.tsx
```diff
@@ -11,46 +16,91 @@ const messages = defineMessages({
...
-  product: Joanie.Product;
-  order?: Joanie.OrderLite;
+  productId: Joanie.Product['id'];
+  courseCode: Joanie.Course['code'];
 }
 
-const CourseProductItem = ({ product, order }: Props) => {
-  const isOwned = order !== undefined;
+const CourseProductItem = ({ productId, courseCode }: Props) => {
+  const productQuery = useProduct(productId, { course: courseCode });
```
NathanVss, line 33: Why is `course` filter needed when the `productId` is already given?

src/frontend/js/components/CourseProductItem/index.tsx
```diff
@@ -11,46 +16,91 @@ const messages = defineMessages({
...
+  const product = productQuery.item;
+  const isOwned = useMemo(() => product?.orders.length > 0, [productQuery.item]);
+  const ordersQuery = useOrders(
```
NathanVss, line 36: We could use `useOrder` to take advantage of the `useResource` hook?

src/frontend/js/components/CourseProductItem/index.tsx
```diff
@@ -11,46 +16,91 @@ const messages = defineMessages({
...
+    { id: product?.orders[0] },
+    { enabled: product?.orders.length > 0 },
+  );
+  const order = useMemo(() => {
```
NathanVss, line 40: My previous point would impact the filtering you operate here.

src/frontend/js/data/CourseProductProvider/index.tsx
```diff
@@ -0,0 +1,41 @@
...
+  children,
+}: PropsWithChildren<CourseProductProviderProps>) => {
+  const value = useMemo(() => ({ productId, courseCode }), [productId, courseCode]);
+  return <Context.Provider value={value}>{children}</Context.Provider>;
+};
+
+/**
+ * A hook to use within `CourseProductProvider`. It returns the course code and product id context.
+ */
+export const useCourseProduct = () => {
```
NathanVss, line 33: is it really useful as React already throws such an error nope?

src/frontend/js/components/CourseProductItem/index.spec.tsx
```diff
@@ -67,30 +62,40 @@ describe('CourseProductItem', () => {
...
-    screen.getByRole('heading', { level: 3, name: product.title });
-    // the price shouldn't be a heading to prevent misdirection for screen reader users
+    // - A loader should be displayed while product information are fetching
+    screen.getByRole('status', { name: 'Loading product information...' });
+
+    productDeferred.resolve(product);
+
+    await waitFor(() => {
+      screen.getByRole('heading', { level: 3, name: product.title });
+    });
```
NathanVss, line 97: I think you could instead use 

```suggestion
await screen.findByRole('heading', { level: 3, name: product.title });
```

src/frontend/js/components/CourseProductItem/index.spec.tsx
```diff
@@ -117,47 +122,69 @@ describe('CourseProductItem', () => {
...
     render(
-      <Wrapper code="00000">
-        <CourseProductItem product={product} />
+      <Wrapper>
+        <CourseProductItem productId={product.id} courseCode="00000" />
       </Wrapper>,
     );
 
+    // Wait for product information to be fetched
+    await waitFor(() => {
```
NathanVss, line 139: Ditto

src/frontend/js/components/CourseProductItem/index.spec.tsx
```diff
@@ -117,47 +122,69 @@ describe('CourseProductItem', () => {
...
-      <Wrapper code="00000">
-        <CourseProductItem product={product} order={order} />
+      <Wrapper>
+        <CourseProductItem productId={product.id} courseCode="00000" />
       </Wrapper>,
     );
 
-    screen.getByRole('heading', { level: 3, name: product.title });
+    // Wait for product information to be fetched
+    await waitFor(() => {
```
NathanVss, line 169: Ditto

src/frontend/js/components/CourseProductItem/index.spec.tsx
```diff
@@ -178,23 +209,35 @@ describe('CourseProductItem', () => {
...
```
NathanVss, line 230: ditto

src/frontend/js/components/CourseProductItem/index.spec.tsx
```diff
@@ -178,23 +209,35 @@ describe('CourseProductItem', () => {
...
+      screen.getByRole('heading', { level: 3, name: product.title });
+    });
+
     // - In place of product price, a label should be displayed
-    const $enrolledInfo = screen.getByText('Enrolled');
-    expect($enrolledInfo.tagName).toBe('STRONG');
-    expect($enrolledInfo.classList.contains('h6')).toBe(true);
+    let $enrolledInfo: HTMLElement;
+    await waitFor(() => {
```
NathanVss, line 236: ditto

src/frontend/js/data/CourseProductProvider/index.spec.tsx
```diff
@@ -0,0 +1,33 @@
+import { renderHook } from '@testing-library/react';
+import { CourseProductProvider, useCourseProduct } from 'data/CourseProductProvider/index';
+import { noop } from 'utils';
+
+describe('useCourseProduct', () => {
+  beforeAll(() => {
+    jest.spyOn(console, 'error').mockImplementation(noop);
```
NathanVss, line 7: Don't you think that this spy is only useful in the error testing test ? It may hide some errors in the first test which does not test throwings.

sandbox/settings.py
```diff
@@ -316,12 +316,12 @@ class Base(StyleguideMixin, DRFMixin, RichieCoursesConfigurationMixin, Configura
             "joanie", environ_name="JOANIE_JS_BACKEND", environ_prefix=None
         ),
         "COURSE_REGEX": values.Value(
-            r"^.*/api/v1.0/(?P<resource_type>(course-runs|products))/(?P<resource_id>.*)/?$",
+            r"^.*/api/v1.0/(?P<resource_type>(course-runs|products))/(?P<resource_id>[^/]*)/?$",
```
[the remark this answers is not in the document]
jbpenrath: Previously this was working
`http://joanie.endpoint/api/v1.0/product/67146513-e3b5-44a7-ab22-edac8d18d63b`

but this one not 
`http://joanie.endpoint/api/v1.0/product/67146513-e3b5-44a7-ab22-edac8d18d63b/` as the capturing group resource_id received `67146513-e3b5-44a7-ab22-edac8d18d63b/` as value.

In fact, a better regex should be to catch only resource_id if this is an uuid.

sandbox/settings.py
```diff
@@ -316,12 +316,12 @@ class Base(StyleguideMixin, DRFMixin, RichieCoursesConfigurationMixin, Configura
...
             environ_name="JOANIE_COURSE_REGEX",
             environ_prefix=None,
         ),
         "JS_COURSE_REGEX": values.Value(
-            r"^.*/api/v1.0/(course-runs|products)/(.*)/?$",
+            r"^.*/api/v1.0/(course-runs|products)/([^/]*)/?$",
```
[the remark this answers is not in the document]
jbpenrath: Ditto 😛

src/frontend/js/utils/ObjectHelper/index.spec.ts
jbpenrath: Good idea!

src/frontend/js/components/CourseProductCourseRuns/EnrollableCourseRunList.tsx
jbpenrath: Indeed !

src/frontend/js/components/CourseProductItem/index.tsx
jbpenrath: It is totally optionnal but it ensures that the product requested is linked to the syllabus.

src/frontend/js/data/CourseProductProvider/index.tsx
jbpenrath: Nope, by default React will return the default value provided during context creation.

src/frontend/js/data/CourseProductProvider/index.spec.tsx
jbpenrath: Indeed!

src/frontend/js/data/CourseProductProvider/index.spec.tsx
```diff
@@ -24,6 +24,7 @@ describe('useCourseProduct', () => {
   });
 
   it('throws an error if it is not used within a CourseProductProvider', () => {
+    jest.spyOn(console, 'error').mockImplementation(noop);
```
NathanVss, line 27: Are you sure that you do not need to unmock this at the bottom of this test ? ( are there any risk that it hides error of other following test suites ? )

NathanVss (approved):

src/frontend/js/data/CourseProductProvider/index.spec.tsx
jbpenrath: This the purpose of the `jest.restoreAllMocks` within the `afterEach`

NathanVss: Right!

PR #1842, opened 2022-12-01
🧑‍💻(github) improve bug report issue template

sampaccoud:
## Purpose

When want to open an issue for a problem we encountered it is weird to first ask about the expected behavior before even describing the problem. I never know what to write and end-up changing the questions.

## Proposal

First ask about the but, then about what was expected.


jbpenrath (approved):
LGTM ✅

PR #1832, opened 2022-11-22
🐛(courses) fix course page style in edit mode

igobranco:
Fix style in edit mode on courses with catalog visibility with `course_only` or `course_and_search`.

A course run without title and a course visibility of `course_only` or `course_and_search` didn't display the small image svg near the run. So the course editor don't know that the course run isn't going to be displayed on the course page or the course run won't be included on the search.

Before:
image

After:
image

PR #1843, opened 2022-12-02
💄(course-detail) fix course run visibility state icon position

jbpenrath:
## Purpose

Recently [we fixed an issue about icon visibility state](#1832) of course detail page. It was not display when the course run had no title. As a workaround we positioned icon through `position: absolute`. But we forgot to add a `position: relative` attribute to the course run row element so all visibility icon stick to the top of the course run list element instead to stick to the top of the course run row.

### Before
Capture d’écran 2022-12-02 à 16 18 32

### After
Capture d’écran 2022-12-02 à 16 18 49

## Proposal

- [x] Add `position: relative` to the `.course-detail__run-descriptions` elemeent

### DIFF
--- a/src/frontend/scss/components/templates/courses/cms/_course_detail.scss
+++ b/src/frontend/scss/components/templates/courses/cms/_course_detail.scss
@@ -125,6 +125,8 @@
   }
 
   &__run-descriptions {
+    position: relative;
+
     dl {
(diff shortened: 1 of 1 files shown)


sampaccoud (approved):
Nice catch... a magnifier with a mask :laughing:

PR #1846, opened 2022-12-06
♻️(frontend) use versionned Joanie API routes

jbpenrath:
## Purpose

Recently, we versionned Joanie api endpoint. So we have to update joanie api interface to dialog with new endpoints.

## Proposal

- [x] Add version with joanie api routes.


NathanVss (commented):
Erratum: A test is failing :/

PR #1857, opened 2022-12-16
🔨(test) add script to run specific frontend test

igobranco:
Add `frontend-test` script that allows the developer to run a specific frontend test more easy.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -175,8 +175,9 @@ stop-storybook: ## Stop front-end's storybook
 	@$(COMPOSE) stop storybook
 .PHONY: stop-storybook
 
-test-front: ## run front-end tests
-	@$(YARN) test --runInBand
+test-front: ## run front-end tests, or specific test like `make test-front js/components/CourseRunEnrollment`
+	@args="$(filter-out $@,$(MAKECMDGOALS))" && \
+	$(YARN) test $${args:-${1}}
 .PHONY: test-front
 
 watch-sass: ## watch changes in Sass files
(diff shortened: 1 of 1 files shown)


jbpenrath:
Hi Ivo :wave:

We could take the opportunity to also add a `test-front` command into the Makefile

What do you think ?

igobranco:
@jbpenrath we already have a `test-front` Makefile target https://github.com/openfun/richie/blob/master/Makefile#L178-L180 what we don't have is an easy way to run a specific test.

jbpenrath:
> @jbpenrath we already have a `test-front` Makefile target https://github.com/openfun/richie/blob/master/Makefile#L178-L180 what we don't have is an easy way to run a specific test.

My intent was to update test-front make command to accept extra args. I find it simpler for developer to have only one command. What do you think 

```
test-front: ## run front-end tests
	@args="$(filter-out $@,$(MAKECMDGOALS))" && \
	bin/frontent-test $${args:-${1}}
.PHONY: test-front
```

```
test-front: ## run front-end tests
	@args="$(filter-out $@,$(MAKECMDGOALS))" && \
	$(YARN) test $${args:-${1}}
.PHONY: test-front
```

igobranco:
@jbpenrath  done.
I alo agree that we should use Makefile as the starting point for every command that a developer needs, because it is much simpler to run `make help` and view all the necessary commands for the project.
I've also included you code for the `test-back` Makefile target.

jbpenrath (approved):
Great 🚀 
In this way we are also able to run back tests from their name.
e.g: `make test-back -- -k test_models`

PR #1875, opened 2023-01-12
Fix/sale tunnel

jbpenrath:
## Purpose

OrderQuery invalidates itself on mutation success. By default all hooks
linked to this query triggered a rerender on this invalidation. About
the CourseProductItem that's problematic as this mutation is called
within the SaleTunnel, so the orderQuery fetching state switch to true then
the Spinner is show and SaleTunnel is unmounted. In order to prevent this
weird behavior, we rework the logic to display product information and
spinner.

## Proposal

- [x] Prevent CourseProductItem rerender on all orderQuery updates
- [x] Update some resource types
- [x] Add a `onSuccess` prop to the `SaleTunnel` component

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,8 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Fixed
 
+- Prevent product course run enrollment when user owns the product
+  through an invalid order
 - Display a message in the sales tunnel when at least one course has no course
   runs, to say that this product is not currently available for sale. 
 - Show error message when user tries to enroll or unroll to a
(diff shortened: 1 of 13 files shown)


CHANGELOG.md
```diff
@@ -22,6 +22,8 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Fixed
 
+- Prevent product course run enrollment when user owns the product
+  through an invalid order
```
sampaccoud, line 26: does the backend allow that?

jbpenrath: Nope

NathanVss (approved):
🥇

Issue #1871, opened 2023-01-09
Have the possibility to display an error message that comes from the backend


sandroscosta:
## Feature Request

**Current behavior**
The system returns a generic error message if the system errors out.
```javascript
{step === Step.ENROLLMENT_FAILED ? (
    <div className="course-run-enrollment__errortext">
       <FormattedMessage {...messages.enrollmentFailed} />
     </div>
) : null}
```
The message output is as follows *Your enrollment request failed.*.
We need to be able to provide better feedback to the user. If the user doesn't have privileges to enroll, he should get a clear message about what went wrong.

**Possible workaround**
On our Open edX instance, we added the possibility to return a payload message (already translated to the language defined by the user). 

> Example: The user is now allowed to enroll because the course is restricted only to members of a specific organization.
> The system returns a 400 status with a `you are not allowed to enroll in this course`.

If we can get something like:
```javascript
{step === Step.ENROLLMENT_FAILED ? (
    <div className="course-run-enrollment__errortext">
       { error?.message ?  <p>{error.message}</p> : <FormattedMessage {...messages.enrollmentFailed} />}
     </div>
) : null}
```

**Issues and Questions**
We have done some exploration exercises. As the system is as complex as is, we need to:

- Find a way to mock a request with status 400 and a payload message
- Get that error message back to the user if it exists
- Guarantee everything is backwards compatible

jbpenrath:
Hi @sandroscosta, if you have customized your Open edX instance to return explicit error messages, I think you should first create a custom LMS Api interface to customize enrollment api methods to parse error payload to extract message if needed.

Then as you suggest in your comment we will have to refactor the `CourseRunEnrollment` component to display an error message if one is provided and otherwise display a default error message.

sandroscosta:
@jbpenrath I've taken a closer look to this and have a question. At the moment, when you set the enrollment, you return a `isEnrolled`that is either true of false. 

```javascript
set: async (url: string, user: User): Promise<boolean> => {
        const courseId = extractCourseIdFromUrl(url);
        /*
            edx_csrf_token cookie is set through a SET_COOKIE header send from a previous request
            e.g: To be able to enroll user, you have to get the enrollment before
          */
        const csrfToken = Cookies.get(EDX_CSRF_TOKEN_COOKIE_NAME) || '';

        const isEnrolled = await fetch(ROUTES.enrollment.set, {
          method: 'POST',
          credentials: 'include',
          headers: {
            'Content-Type': 'application/json',
[...]

PR #1884, opened 2023-01-19
✨(CourseRunEnrollment) customized failure message

sandroscosta:
Related to #1871 

## Proposal
Give Richie the possibility of displaying custom error messages that come from an error payload

### DIFF
--- a/src/frontend/js/components/CourseRunEnrollment/index.tsx
+++ b/src/frontend/js/components/CourseRunEnrollment/index.tsx
@@ -9,6 +9,7 @@ import { Maybe, Nullable } from 'types/utils';
 import { handle } from 'utils/errors/handle';
 import { CommonDataProps } from 'types/commonDataProps';
 import useCourseEnrollment from 'data/useCourseEnrollment';
+import { HttpError } from 'utils/errors/HttpError';
 
 const messages = defineMessages({
   enroll: {
@@ -75,6 +76,7 @@ enum Step {
 enum ActionType {
   UPDATE_CONTEXT = 'UPDATE_CONTEXT',
   ENROLL = 'ENROLL',
+  ENROLLMENT_FAILED = 'ENROLLMENT_FAILED',
   ERROR = 'ERROR',
 }
 interface ReducerState {
@@ -84,12 +86,13 @@ interface ReducerState {
     courseRun: CourseRunEnrollmentProps['courseRun'];
     currentUser: Maybe<Nullable<User>>;
   };
-  error?: Error;
+  error?: HttpError;
 }
 type ReducerAction =
   | { type: ActionType.UPDATE_CONTEXT; payload: Partial<ReducerState['context']> }
   | { type: ActionType.ENROLL }
-  | { type: ActionType.ERROR; payload: { error: Error } };
+  | { type: ActionType.ENROLLMENT_FAILED; payload: { error: HttpError } }
+  | { type: ActionType.ERROR; payload: { error: HttpError } };
 
 const getStepFromContext = (
   { currentUser, courseRun, isEnrolled }: ReducerState['context'],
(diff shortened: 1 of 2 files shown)


sandroscosta:
@jbpenrath There seems to be an issue with `error` getting to the correct place. Apparently error is always undefined. Don't know if we need to change the API adapter to accept an error with a payload. Any hints on where to go next or if something is wrong.

src/frontend/js/components/CourseRunEnrollment/index.tsx
```diff
@@ -220,7 +230,7 @@ const CourseRunEnrollment: React.FC<CourseRunEnrollmentProps & CommonDataProps>
           </button>
           {step === Step.ENROLLMENT_FAILED ? (
             <div className="course-run-enrollment__errortext">
-              <FormattedMessage {...messages.enrollmentFailed} />
+              {error != undefined && error.code == 400 ? <>{error.message}</> : <FormattedMessage {...messages.enrollmentFailed} />}
```
jbpenrath, line 233: Use strict operators. Furthermore, you could use an optional chaining operator (`?`) instead to test if `error` is defined.

```suggestion
              {error?.code === 400 ? error.message : <FormattedMessage {...messages.enrollmentFailed} />}

```

src/frontend/js/components/CourseRunEnrollment/index.tsx
```diff
@@ -83,12 +85,14 @@ interface ReducerState {
     isEnrolled: Maybe<Nullable<Boolean>>;
     courseRun: CourseRunEnrollmentProps['courseRun'];
     currentUser: Maybe<Nullable<User>>;
+    error?: HttpError;
```
jbpenrath, line 88: There is already an `error` property outside context.

src/frontend/js/components/CourseRunEnrollment/index.tsx
```diff
@@ -132,6 +136,8 @@ const reducer = ({ step, context }: ReducerState, action: ReducerAction): Reduce
       const nextContext = { ...context, ...action.payload };
       return { step: getStepFromContext(nextContext, step), context: nextContext };
     }
+    case ActionType.ENROLLMENT_FAILED:
+      return { step: Step.ENROLLING, context: {...context, ...action.payload }};
```
jbpenrath, line 140:
```suggestion
      return { step: Step.ENROLLMENT_FAILED, context, ...action.payload };
```

src/frontend/js/components/CourseRunEnrollment/index.tsx
```diff
@@ -161,7 +166,13 @@ const CourseRunEnrollment: React.FC<CourseRunEnrollmentProps & CommonDataProps>
   const enroll = useCallback(async () => {
     dispatch({ type: ActionType.ENROLL });
     if (courseRun && currentUser) {
-      const isEnrolled = await setEnrollment().catch(() => undefined);
+      const isEnrolled = await setEnrollment().catch((error) => {
+        dispatch({
+          type: ActionType.ENROLLMENT_FAILED,
+          payload: { error },
+        });
+        return undefined;
```
jbpenrath, line 174: Return statement is useless.

```suggestion
```

src/frontend/js/components/CourseRunEnrollment/index.tsx
```diff
@@ -161,7 +166,13 @@ const CourseRunEnrollment: React.FC<CourseRunEnrollmentProps & CommonDataProps>
   const enroll = useCallback(async () => {
     dispatch({ type: ActionType.ENROLL });
     if (courseRun && currentUser) {
-      const isEnrolled = await setEnrollment().catch(() => undefined);
+      const isEnrolled = await setEnrollment().catch((error) => {
+        dispatch({
```
jbpenrath, line 170: I think you should use a `try...catch` here.
Furthermore there is something wrong as the action `UPDATE_CONTEXT` is always called. You should dispatch this action only when enroll request succeeded.

sandroscosta:
Had to change all Errors to be `HttpError` to be able to access code.

sandroscosta:
Deprecated to switch to a new approach

PR #1890, opened 2023-01-31
🐛(frontend) disable payment button if a product has no courses

NathanVss:
If a product has no target courses we need to prevent him from buying it.


src/frontend/js/components/SaleTunnel/index.spec.tsx
```diff
@@ -155,6 +155,28 @@ describe('SaleTunnel', () => {
     );
   });
 
+  it('render a sale tunnel with no courses', async () => {
```
jbpenrath, line 158:
```suggestion
  it('renders a sale tunnel with no courses', async () => {
```

jbpenrath (approved):
LGTM ✅

PR #1891, opened 2023-02-01
Fix/course product item

jbpenrath:
## Purpose

About CourseProductItem, previously, we guessed if current user has purchased the related product from the product response itself. Indeed, if the user has purchased the product, the product's property `orders` contains all pending and validated orders related to the product and the current user. But there was a bug as useProduct hook is not aware about session state. So in the case where the page was displayed without session state initialized, the product query was executed immediately then when the session state was updated, as the product query was not stale, it was not executed again.

In order to fix that, to guess if the authenticated user has purchased the related product, we use the `useOrders` hook which is aware of session state update.

Furthermore, if a course run targeted a Joanie product has been misconfigured, the product request can return error. Currently, nothing is display in this case that is weird. So we update the CourseProductItem to display an error message in this case with an error look'n feel.

image

## Proposal

- [x] Fix CourseProductItem issue on session state update
- [x] Display error message when `CourseProductItem` component has error

### DIFF
--- a/src/frontend/js/components/CourseProductItem/_styles.scss
+++ b/src/frontend/js/components/CourseProductItem/_styles.scss
@@ -8,6 +8,20 @@
   min-height: 100px;
   overflow: hidden;
 
+  &--has-error {
+    align-items: center;
+    border-color: r-theme-val(product-item, feedback-color);
+    display: flex;
+
+    & > .product-widget__content {
+      color: r-theme-val(product-item, feedback-color);
+      display: flex;
+      flex-direction: row;
+      gap: 1rem;
+      margin-bottom: 0;
+    }
+  }
+
   &:not(:last-child) {
     margin-bottom: 1.5rem;
   }
--- a/src/frontend/js/components/CourseProductItem/index.tsx
+++ b/src/frontend/js/components/CourseProductItem/index.tsx
@@ -6,7 +6,7 @@ import SaleTunnel from 'components/SaleTunnel';
 import type * as Joanie from 'types/Joanie';
 import { useProduct } from 'hooks/useProduct';
 import { Spinner } from 'components/Spinner';
-import { useOrder } from 'hooks/useOrders';
+import { useOrders } from 'hooks/useOrders';
 import { OrderState } from 'types/Joanie';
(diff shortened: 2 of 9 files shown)


src/frontend/js/components/CourseProductItem/_styles.scss
```diff
@@ -8,6 +8,21 @@
   min-height: 100px;
   overflow: hidden;
 
+  &--has-error {
+    align-items: center;
+    border-color: r-theme-val(product-item, feedback-color);
+    display: flex;
+    //background-color: rgba(r-theme-val(product-item, feedback-color), 0.1);
```
NathanVss, line 15: Lost comment here

src/frontend/js/components/CourseProductItem/index.tsx
```diff
@@ -32,16 +33,17 @@ export interface Props {
...
-    {
-      state: [OrderState.VALIDATED],
-    },
-    {
-      enabled: isOwned,
-    },
+  const ordersQuery = useOrders({
+    product: productId,
+    course: courseCode,
+    state: [OrderState.VALIDATED, OrderState.PENDING],
```
NathanVss, line 39: Why are you including `OrderState.PENDING` is the filters while, below at the line `42` you filter again to find the first `OrderState.VALIDATED` order? Why not this?


```suggestion
  const ordersQuery = useOrders({
    product: productId,
    course: courseCode,
    state: [OrderState.VALIDATED],
  });
```

NathanVss (approved):
🐛

src/frontend/js/components/CourseProductItem/index.tsx
jbpenrath: Because `isOwned` flag must be true if the user has a validated or a pending order. In this way, the user is not allow to open sale tunnel if it has a pending order.

But user should be allowed to enroll to course run only if it has a validated order.

jbpenrath: Oh! The `isOwned` was wrong! I added the test to test the behavior I described above.

NathanVss (approved):
🛰️

PR #1893, opened 2023-02-02
💄(frontend) allow to customise course enrollment colors in subheader

jbpenrath:
## Purpose

CourseRunEnrollment widget can be display into course detail subheader and also into course detail body. Each contexts have a different background color so developer must be able to customize the course run enrollment button and feedback colors for both context.

image

## Proposal

- [x] Allow to customize CourseRunEnrollment colors when displayed into course detail subheader


sampaccoud (approved):
:muscle:

PR #1894, opened 2023-02-02
Fix/sale tunnel

jbpenrath:
## Purpose

Currently, SaleTunnel component manages at same time the button to open the
payment modal and the modal itself that is too much complicated. So we decide to
extract <PurchaseButton /> from <SaleTunnel />

Then in some case, user can have an order in a pending state for a product. In order
to prevent user to open sale tunnel in this case we do not show the purchase
button. But the label currently display show "Enrolled" that is unclear. In
order to improve user understanding, we decide to display "Pending" label in
place of the current "Enrolled" label in this case.

## Proposal

- [x] Extract PurchaseButton from SaleTunnel
- [x] Display "Pending" label when user has purchased a product but its order is pending

### DIFF
--- a/src/frontend/js/components/CourseProductItem/PurchaseButton.spec.tsx
+++ b/src/frontend/js/components/CourseProductItem/PurchaseButton.spec.tsx
@@ -1,0 +1,154 @@
+import { act, fireEvent, render, screen } from '@testing-library/react';
+import fetchMock from 'fetch-mock';
+import { IntlProvider } from 'react-intl';
+import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
+import { ContextFactory as mockContextFactory, ProductFactory } from 'utils/test/factories';
+import { SessionProvider } from 'data/SessionProvider';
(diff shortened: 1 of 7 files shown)


src/frontend/js/components/CourseProductItem/PurchaseButton.spec.tsx
```diff
@@ -0,0 +1,154 @@
...
+
+jest.mock('utils/context', () => ({
+  __esModule: true,
+  default: mockContextFactory({
+    authentication: { backend: 'fonzie', endpoint: 'https://auth.endpoint.test' },
+    joanie_backend: { endpoint: 'https://joanie.test' },
+  }).generate(),
+}));
+
+describe('PurchaseButton', () => {
```
NathanVss, line 18: Might be a good idea to add a test for `pending` product. WDYT ?

NathanVss (approved):
Ok for me but I'm a bit concerned why the tests fails on my local machine 😅

PR #1901, opened 2023-02-06
(chore) generate joanie api client

rlecellier:
## Proposal

POC that add a simple script to generate joanie's api client with `openapi-typescript-codegen`


rlecellier:
note: we may be able to add react-query logic directly into a custom [request.ts](https://github.com/openfun/richie/pull/1901/files#diff-1507d9363a7dc679b0adb986721345d90576a40a467364afbce9e3295bf1ad63R1) file

PR #1912, opened 2023-02-15
✨(backend) implement feature flags from .env

rlecellier:
## Purpose

We need to activate / disable some features during development.
This will let us push partial features (only front,only back for example).

## Proposal

* configure feature flag in .env files with the prefix: "FEATURE_".
* add this flag in settings.py
* either use `settings.get('FEATURE_{MY_FEATURE}')` in python or use `{ feature_flags.feature 'MY_FEATURE' }` in jinja templates

## Improvement

I tried like to use a dict in .env file but i didn't find how to do it.
The idea is to set `settings.FEATURES` only once instead of editing `settings.py` each time we wanna add/remove a flag.

### DIFF
--- a/env.d/development/dev-ssl.dist
+++ b/env.d/development/dev-ssl.dist
@@ -4,3 +4,6 @@ AUTHENTICATION_BACKEND=openedx-hawthorn
 
 # LMS Backend
 EDX_BASE_URL=https://edx.local.dev:8073
+
+# Features
+FEATURES = {'JOANIE_WISHLIST': True}
--- a/sandbox/settings.py
+++ b/sandbox/settings.py
@@ -578,6 +578,9 @@ class Base(StyleguideMixin, DRFMixin, RichieCoursesConfigurationMixin, Configura
         environ_prefix=None,
     )
 
+    # Feature flags
+    FEATURES = values.DictValue(environ_name="FEATURES", environ_prefix=None)
+
     @classmethod
     def _get_environment(cls):
         """Environment in which the application is launched."""
--- a/src/richie/apps/core/templatetags/feature_flags.py
+++ b/src/richie/apps/core/templatetags/feature_flags.py
@@ -1,0 +1,15 @@
+"""Custom template tags related to FEATURES settings."""
+from django import template
+from django.conf import settings
+
+register = template.Library()
+
+
+@register.simple_tag()
+def is_feature_enabled(flag):
+    """
+    template tag to know if a feature flag is enable / disable
+    """
+    if hasattr(settings, "FEATURES"):
+        return settings.FEATURES.get(flag, False)
+    return False
--- a/tests/apps/core/test_templatetags_feature_flags.py
+++ b/tests/apps/core/test_templatetags_feature_flags.py
@@ -1,0 +1,52 @@
+"""
+Unit tests for the template tags related to FeatureFlags.
+"""
+from django.test import TestCase
+from django.test.utils import override_settings
+
+from richie.apps.core.templatetags.feature_flags import is_feature_enabled
+
+
+class FeatureFlagsTemplateTagsTestCase(TestCase):
+    """
+    Unit test suite to validate the behavior of the `is_feature_enabled` template tags
+    """
+
+    @override_settings(
+        FEATURES={
+            "TEST_FEATURE": True,
+        }
+    )
+    def test_templatetags_is_feature_enabled_true(self):
+        """
+        is_feature_enabled should return True if
+        an feature flags is set to True in settings.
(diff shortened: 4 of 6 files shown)


MorganeAD (changes requested):
Nice job !

I think it will be nice if you add an example entry of `FEATURES` in the file https://github.com/openfun/richie/blob/master/env.d/development/dev.dist

```
# FeatureFlags
# FEATURES={"FEATURE_NAME": Bool}
FEATURES={}
```

jbpenrath:
Write some tests to ensure that feature template tag is working as expected.
e.g : https://github.com/openfun/richie/blob/master/tests/apps/courses/test_templatetags_extra_tags_is_joanie_product.py

sandbox/settings.py
```diff
@@ -578,6 +578,9 @@ class Base(StyleguideMixin, DRFMixin, RichieCoursesConfigurationMixin, Configura
         environ_prefix=None,
     )
 
+    # Feature flags
+    FEATURES = values.DictValue(environ_name="FEATURES", environ_prefix=None)
```
jbpenrath, line 582: You could add default values into `dev.dist` and `dev-ssl.dist` env files. In this way by default.

src/richie/apps/core/templatetags/feature_flags.py
```diff
@@ -0,0 +1,15 @@
+"""Custom template tags related to Joanie."""
```
jbpenrath, line 1:
```suggestion
"""Custom template tags related to FEATURES settings."""
```

src/richie/apps/core/templatetags/feature_flags.py
```diff
@@ -0,0 +1,15 @@
...
+from django import template
+from django.conf import settings
+
+register = template.Library()
+
+
+@register.simple_tag()
+def features(flag):
```
jbpenrath, line 9: I should catch it sooner, but by convention we try to write method whose return a boolean as a question (`is_enabled`, `has_course_runs`, ...)

What do you think to rename this template tag as follow ?
```suggestion
def is_feature_enabled(flag):
```

tests/apps/core/test_templatetags_feature_flags.py
```diff
@@ -0,0 +1,52 @@
...
+"""
+from django.test import TestCase
+from django.test.utils import override_settings
+
+from richie.apps.core.templatetags.feature_flags import is_feature_enabled
+
+
+class FeatureFlagsTemplateTagsTestCase(TestCase):
+    """
+    Unit test suite to validate the behavior of the joanie template tags
```
jbpenrath, line 12:
```suggestion
    Unit test suite to validate the behavior of the `is_feature_enabled` template tags
```

src/richie/apps/core/templatetags/feature_flags.py
```diff
@@ -0,0 +1,15 @@
...
+from django import template
+from django.conf import settings
+
+register = template.Library()
+
+
+@register.simple_tag()
+def is_feature_enabled(flag):
+    """
+    template tag to enable / disable features
```
jbpenrath, line 11:
```suggestion
    template tag to know if a feature flag is enable / disable
```

tests/apps/core/test_templatetags_feature_flags.py
```diff
@@ -0,0 +1,52 @@
...
+        FEATURES={
+            "TEST_FEATURE": True,
+        }
+    )
+    def test_templatetags_features_enabled(self):
+        """
+        features should return True if
+        an feature flags is set to True in settings.
+        """
+        self.assertTrue(is_feature_enabled("TEST_FEATURE"))
```
jbpenrath, line 25: Do not forgot to rename test suite and docstrings according to the new template tag name.

```suggestion
    def test_templatetags_is_feature_enabled_true(self):
        """
        `is_feature_enabled` should return True if
        an feature flags is set to True in settings.
        """
        self.assertTrue(is_feature_enabled("TEST_FEATURE"))
```

jbpenrath (approved):
Last few comments then merge it 👏

env.d/development/dev-ssl.dist
```diff
@@ -4,3 +4,6 @@ AUTHENTICATION_BACKEND=openedx-hawthorn
 
 # LMS Backend
 EDX_BASE_URL=https://edx.local.dev:8073
+
+# Features
+FEATURES = {'JOANIE_WISHLIST': True}
```
jbpenrath, line 9: Missing new line

MorganeAD (approved):
:1st_place_medal:

PR #1914, opened 2023-02-21
🐛(frontend) fix enrollment not started test for CourseProductCourseRuns

PanchoutNathan:
## Purpose
The test passes randomly because of the random generation of the course_run

## Proposal
- [x] set the start date day to the today day


MorganeAD (approved):

src/frontend/js/components/CourseProductCourseRuns/index.spec.tsx
```diff
@@ -547,10 +547,11 @@ describe('CourseProductCourseRuns', () => {
       const today = new Date();
```
jbpenrath, line 547: You are updating the test so you can use ✅ as gitmoji.

jbpenrath (approved):
Just a little nitpick about the commit message otherwise LGTM ✅

PR #1918, opened 2023-03-01
💄(form) add * on select, checkbox and radio inputs when required

jbpenrath:
## Purpose

Currently, checkbox and select labels are not suffixed by an asterisk when they are required.

## Proposal

- [x] Fix Select, Checkbox and Radio fields
- [x] Add a param `Required` into input stories

### DIFF
--- a/src/frontend/js/components/Form/Inputs.tsx
+++ b/src/frontend/js/components/Form/Inputs.tsx
@@ -14,6 +14,7 @@ interface FieldProps {
   error?: Boolean;
   message?: ReactNode | string;
   fieldClasses?: string[];
+  required?: boolean;
 }
 
 // - Field
@@ -22,6 +23,7 @@ const Field = ({
   inputId,
   error,
   message,
+  required = false,
   children,
 }: React.PropsWithChildren<FieldProps>) => {
   const classList = useMemo(() => {
@@ -29,8 +31,11 @@ const Field = ({
     if (error) {
       classes.push('form-field--error');
     }
+    if (required) {
+      classes.push('form-field--required');
+    }
     return classes.join(' ');
-  }, [fieldClasses, error]);
+  }, [fieldClasses, required, error]);
 
   const iconHref = error ? '#icon-warning' : '#icon-info-rounded';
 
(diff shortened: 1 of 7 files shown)


jbpenrath:
Good catch @rlecellier! It should be good now.

src/frontend/js/hooks/useDashboardAddressForm.tsx
```diff
@@ -108,7 +108,7 @@ export const useDashboardAddressForm = (address?: Address) => {
       />
       <CountrySelectField
```
NathanVss, line 109: I was wondering if we must turn all input of addresses forms to required because atm the `*` is never shown. We only display errors afterwards when trying to submit which can be misleading.

src/frontend/js/hooks/useDashboardAddressForm.tsx
```diff
@@ -108,7 +108,7 @@ export const useDashboardAddressForm = (address?: Address) => {
...
         aria-invalid={!!formState.errors.country}
-        aria-required={false}
+        aria-required={true}
```
NathanVss, line 111: Considering my last comment IMO `aria-required` would not be needed anymore because `required` will be enough and seems to be supported by screen readers nowadays.

NathanVss (approved):

src/frontend/js/components/Form/Inputs.tsx
```diff
@@ -22,15 +23,19 @@ const Field = ({
...
   message,
+  required = false,
   children,
 }: React.PropsWithChildren<FieldProps>) => {
   const classList = useMemo(() => {
     const classes = ['form-field', ...fieldClasses];
     if (error) {
       classes.push('form-field--error');
     }
+    if (required === true) {
```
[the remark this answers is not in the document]
jbpenrath: The mood of the moment 🤷‍♂️. I'll fix that!

src/frontend/js/hooks/useDashboardAddressForm.tsx
jbpenrath: Yup, indeed !

PR #1926, opened 2023-03-10
♻️(frontend) rename product.certificate to certificate_definition

NathanVss:
Due to an update on Joanie's API we had to rename this attribute for semantic reasons.


jbpenrath (approved):
There is gitmoji for this kind of change `👽️`.
Also add an entry into the changelog please.

PR #1947, opened 2023-03-28
🔧(cookiecutter) fix frontend build

jbpenrath:
## Purpose

Currently, the nightly round job is failing because the cookiecutter fails to install frontend deps. Indeed, previously, webpack & webpack-cli were installed implicitly by `richie-education` package all was working fine. But now, it appears we have to installed explicitly those dependencies.

https://app.circleci.com/pipelines/github/openfun/richie/6136/workflows/33dcf424-4b4b-477d-93ab-735adee1d093/jobs/171755

## Proposal

- [x] Add `webpack` & `webpack-cli` packages into the list of devDependencies into the cookiecutter frontend app.


NathanVss (approved):
Legit :)

PR #1940, opened 2023-03-21
♻️(frontend) CourseGlimpse as generic component

rlecellier:
CourseGlimpse and CourseGlimpseList will be used in the teacher dashboard.
They need to become generic components and needed a story.

## Purpose

Create stories for `CourseGlimpse` and `CourseGlimpseList` to be able to apply modifications requested on [teacher dashboard](https://www.sketch.com/s/c0aaee1b-8058-4471-b744-771ed7356622/a/YG45waO).

## Proposal

Add a type `FullCourse` that include currently missing fields.

## Previews
GimpseList_Story
Gimpse_Story

PR #1948, opened 2023-03-29
♻️(frontend) split factories file into two

rlecellier:
## Purpose

Joanie and Richie apis share some model names.
We are in need of a proper separation to avoid
naming nightmare.

[Linked PR](https://github.com/openfun/richie/pull/1940) (shared CourseGlimpse between Joanie and Richie will introduce two Course types and factories).


UPGRADE.md
```diff
@@ -18,6 +18,8 @@ $ make migrate
 
 ## 2.20.1 to 2.21.0
 
+- utils/test/factories.ts have been split into multiple files: joanie.ts, richie.ts and 
+  reactQuery.ts. Update your import accordingly.
```
jbpenrath, line 22: Should be moved up into the `Unreleased` section

src/frontend/js/utils/test/factories/richie.ts
```diff
@@ -0,0 +1,62 @@
...
+});
+
+export const FonzieUserFactory = compose(
+  UserFactory,
+  createSpec({
+    access_token: btoa(faker.datatype.uuid()),
+  }),
+);
+
+export const RichieContextFactory = (context: Partial<CommonDataProps['context']> = {}) =>
```
jbpenrath, line 43: Why not conserve `ContextFactory` and prevent too much changes ? Furthermore, there is no duplication issue about this factory.
```suggestion
export const ContextFactory = (context: Partial<CommonDataProps['context']> = {}) =>
```

src/frontend/js/utils/test/factories/reactQuery.ts
jbpenrath, line 1: 🏅

src/frontend/js/utils/test/factories/richie.ts
rlecellier: `Context` is kind of a reserved word in React.
here we talk about the richie's application context and not a ReactContextFactory.

jbpenrath: Mokay

jbpenrath (approved):
LGTM ✅

PR #1949, opened 2023-03-29
✨(core) add js block into head

jbpenrath:
## Purpose

We recently added the support of multiple analytic providers. After a first use, it appears it miss a block within the template to add common code to all providers. Indeed, the only block available to override web_analytics template is one rendered within the loop which iterates over all analytic providers available.

## Proposal

- [x] Add new block `web_analytics_setup` within `web_analytics.html` template

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unrealeased]
 
+### Added
+
+- Add head_js block into base html template
+
 ### Fixed
 
 - Fix error `MultipleObjectsReturned` during synchronization due to missing
--- a/src/richie/apps/core/templates/richie/base.html
+++ b/src/richie/apps/core/templates/richie/base.html
@@ -60,6 +60,7 @@
         {% endif %}
         {% endspaceless %}
 
+        {% block head_js %}{% endblock head_js %}
         {% include "richie/web_analytics.html" with location='head' %}
 
         {% render_block "css" %}
@@ -268,7 +269,7 @@
             });
         </script>
         {% endif %}
-        {% include "richie/web_analytics.html" with location='footer' %}
         {% block body_js %}{% endblock body_js %}
+        {% include "richie/web_analytics.html" with location='footer' %}
     </body>
 </html>


igobranco:
Why don't you extend the `web_analytics.html` file and then override the `web_analytics_additional_providers` block?
You can use `my_custom_web_analytics` as the provider key and then the `location` as subkey to decide where to put the code in the html head or at the bottom of the body.
I don't understand why you need the `WEB_ANALYTICS_SETUP_LOCATION` setting, because each web analytics software can be put on a different location in the page.

```yaml
WEB_ANALYTICS: "{'google_universal_analytics': {'tracking_id': 'UA-12345678-9', 'location': 'head'}, 'my_custom_web_analytics': {'tracking_id': 'MY_TRACKING_CODE', 'location': 'head' } }"
```

File `web_analytics.html` file
```html
{% extends "richie/web_analytics.html" %}
{% block web_analytics_additional_providers %}
    <script>
       .... 
    </script>
{% endblock web_analytics_additional_providers %}
```

jbpenrath:
> Why don't you extend the `web_analytics.html` file and then override the `web_analytics_additional_providers` block? You can use `my_custom_web_analytics` as the provider key and then the `location` as subkey to decide where to put the code in the html head or at the bottom of the body. I don't understand why you need the `WEB_ANALYTICS_SETUP_LOCATION` setting, because each web analytics software can be put on a different location in the page.
> 
> ```yaml
> WEB_ANALYTICS: "{'google_universal_analytics': {'tracking_id': 'UA-12345678-9', 'location': 'head'}, 'my_custom_web_analytics': {'tracking_id': 'MY_TRACKING_CODE', 'location': 'head' } }"
> ```
> 
> File `web_analytics.html` file
> 
> ```
> {% extends "richie/web_analytics.html" %}
> {% block web_analytics_additional_providers %}
>     <script>
>        .... 
>     </script>
> {% endblock web_analytics_additional_providers %}
> ```

My use case is to be able to setup a cookie consent. I need to setup it once before web analytics initialization.

jbpenrath:
> Why don't you add 1 web analytics provider with both the cookie consent and your custom web analytics solution. Or alternatively use 2 different web analytics providers, but the 1st is the cookie consent and the 2nd is your really your custom web analytics.
> 
> ```
> {% extends "richie/web_analytics.html" %}
> {% block web_analytics_additional_providers %}
>     {% if provider == "custom_cookie_consent" %}
>     {% endif %}
>     {% if provider == "my_custom_web_analytics" %}
>     {% endif %}
> {% endblock web_analytics_additional_providers %}
> ```
> 
> Setting:
> 
> ```yaml
> WEB_ANALYTICS: "{'custom_cookie_consent': {'tracking_id': 'FAKE', 'location': 'footer'}, 'my_custom_web_analytics': {'tracking_id': 'MY_TRACKING_CODE', 'location': 'footer' } }"
> ```
> 
> Reviewing the Richie code of the `web_analytics.htm` file, I think that this line could probably be put inside of each web analytics block. https://github.com/openfun/richie/blob/40a40c0/src/richie/apps/core/templates/richie/web_analytics.html#L8 That is why I added the tracking_id `FAKE` value previously on the settings.

We start to tweak `WEB_ANALYTICS` setting I think. Furthermore, I need guarantee that cookie consent has been initialized before my analytic provider.

jbpenrath:
@igobranco according to our morning meeting, I'll update my PR with the implementation we talked about.

igobranco (approved):

PR #1954, opened 2023-04-04
💄(frontend) fix layout issue on CourseProductItem

jbpenrath:
## Purpose

We recently rework form styles and it introduces a layout issue on CourseProductItem.

## Proposal

- [x] Fix layout issue on CourseProductItem


NathanVss (approved):
LG 👌

PR #1966, opened 2023-04-21
✨(front) rework order layout and add new pages

NathanVss:
~~- [ ] Update to DashboardLayoutRoute to render only Outlet~~
- [x] Adapt renderLayout in routes
- [x] Use DashboardLayout inside Leaners dashboard
- [x] Create order routes
- [x] Make nested order routes for breadcrumbs
- [x] Exact matching to set active route on the sidebar
- [x] Breadcrumbs in new routes
- [x] Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 ### Added
 
 - Add cunningham and design system.
+- Specific sidebar for order related routes
 
 ### Changed
 
--- a/src/frontend/js/pages/DashboardCourses/index.spec.tsx
+++ b/src/frontend/js/pages/DashboardCourses/index.spec.tsx
@@ -10,7 +10,7 @@ import { EnrollmentFactory, OrderFactory, ProductFactory } from 'utils/test/fact
 import { createTestQueryClient } from 'utils/test/createTestQueryClient';
 import { SessionProvider } from 'contexts/SessionContext';
 import { LearnerDashboardPaths } from 'widgets/Dashboard/utils/learnerRouteMessages';
-import { Enrollment, Order, Product } from 'types/Joanie';
+import { CourseLight, Enrollment, Order, Product } from 'types/Joanie';
 import { expectNoSpinner, expectSpinner } from 'utils/test/expectSpinner';
 import { expectBannerError } from 'utils/test/expectBanner';
 import { Deferred } from 'utils/test/deferred';
@@ -52,6 +52,7 @@ describe('<DashboardCourses/>', () => {
   beforeEach(() => {
     fetchMock.get('https://joanie.endpoint/api/v1.0/addresses/', []);
     fetchMock.get('https://joanie.endpoint/api/v1.0/credit-cards/', []);
+    fetchMock.get('https://joanie.endpoint/api/v1.0/orders/', []);
   });
 
   afterEach(() => {
@@ -79,7 +80,10 @@ describe('<DashboardCourses/>', () => {
       const product: Product = ProductFactory().one();
       product.id = order.product;
       fetchMock.get(
-        'https://joanie.endpoint/api/v1.0/products/' + product.id + '/?course=' + order.course,
+        'https://joanie.endpoint/api/v1.0/products/' +
+          product.id +
+          '/?course=' +
+          (order.course as CourseLight).code,
         product,
       );
 
@@ -375,7 +379,7 @@ describe('<DashboardCourses/>', () => {
     await act(async () => userEvent.click(loadMoreButton));
     await waitFor(() => expectList(entities.slice(0, 250), products), { timeout: 30000 });
     expect(screen.queryByRole('button', { name: 'Load more' })).not.toBeInTheDocument();
-  }, 30000);
+  }, 300000);
 
   it('shows an error', async () => {
     jest.spyOn(console, 'error').mockImplementation(noop);
--- a/src/frontend/js/pages/DashboardOrderLayout/index.spec.tsx
+++ b/src/frontend/js/pages/DashboardOrderLayout/index.spec.tsx
@@ -1,0 +1,75 @@
+import { findByRole, render, screen, waitFor } from '@testing-library/react';
+import { QueryClientProvider } from '@tanstack/react-query';
+import { IntlProvider } from 'react-intl';
+import fetchMock from 'fetch-mock';
+import { RichieContextFactory as mockRichieContextFactory } from 'utils/test/factories/richie';
+import { Order } from 'types/Joanie';
+import { OrderFactory, TargetCourseFactory } from 'utils/test/factories/joanie';
+import { mockProductWithOrder } from 'utils/test/mockProductWithOrder';
+import { LearnerDashboardPaths } from 'widgets/Dashboard/utils/learnerRouteMessages';
+import { createTestQueryClient } from 'utils/test/createTestQueryClient';
+import { SessionProvider } from 'contexts/SessionContext';
+import { DashboardTest } from 'widgets/Dashboard/components/DashboardTest';
+import { expectUrlMatchLocationDisplayed } from 'utils/test/expectUrlMatchLocationDisplayed';
+
+jest.mock('utils/context', () => ({
+  __esModule: true,
+  default: mockRichieContextFactory({
+    authentication: { backend: 'fonzie', endpoint: 'https://demo.endpoint' },
+    joanie_backend: { endpoint: 'https://joanie.endpoint' },
+  }).one(),
+}));
+
+describe('<DashboardOrderLayout />', () => {
+  const WrapperWithDashboard = (route: string) => {
+    return (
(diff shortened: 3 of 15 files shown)


src/frontend/js/utils/test/mockProduct.ts
```diff
@@ -0,0 +1,13 @@
+import fetchMock from 'fetch-mock';
+import { Order, Product } from 'types/Joanie';
+import { ProductFactory } from 'utils/test/factories/joanie';
+
+export const mockProduct = (order: Order) => {
```
rlecellier, line 5: proposal `mockOrderProduct`

src/frontend/js/pages/DashboardOrderApplication/index.spec.tsx
```diff
@@ -0,0 +1,80 @@
...
+      { overwriteRoutes: true },
+    );
+
+    render(
+      WrapperWithDashboard(LearnerDashboardPaths.ORDER_APPLICATION.replace(':orderId', order.id)),
+    );
+    await expectBreadcrumbsToEqualParts(['Back', 'My courses', product.title, 'Application']);
+    expectUrlMatchLocationDisplayed(
+      LearnerDashboardPaths.ORDER_APPLICATION.replace(':orderId', order.id),
+    );
```
rlecellier, line 71: in other test i used : 
```js
const intl = createIntl({ locale: 'en' });
getDashboardRoutePath(intl)(TeacherDashboardPaths.ORDER_APPLICATION, {
  orderId: order.id,
})}
```
i don't know which one's better.

src/frontend/js/pages/DashboardOrderApplication/index.spec.tsx
```diff
@@ -0,0 +1,80 @@
...
+  }).generate(),
+}));
+
+describe('<DashboardOrderApplication />', () => {
+  const WrapperWithDashboard = (route: string) => {
+    return (
+      <QueryClientProvider client={createTestQueryClient({ user: true })}>
+        <IntlProvider locale="en">
+          <SessionProvider>
+            <DashboardTest initialRoute={route} />
```
rlecellier, line 32: I think we should directly use the page component  `DashboardOrderApplication` here instead of loading the whole router.

src/frontend/js/pages/DashboardOrderApplication/index.spec.tsx
```diff
@@ -0,0 +1,80 @@
...
+      WrapperWithDashboard(LearnerDashboardPaths.ORDER_APPLICATION.replace(':orderId', order.id)),
+    );
+    await expectBreadcrumbsToEqualParts(['Back', 'My courses', product.title, 'Application']);
+    expectUrlMatchLocationDisplayed(
+      LearnerDashboardPaths.ORDER_APPLICATION.replace(':orderId', order.id),
+    );
+
+    // click on back button.
+    const user = userEvent.setup();
+    await act(async () => user.click(screen.getByRole('link', { name: 'Back' })));
```
rlecellier, line 75: I see that a simple `await userEvent.click()` throw a act warning.
It may work with a await of a element in the new page. I try to found one to try but i didn't.

src/frontend/js/pages/DashboardOrderLayout/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+  const product = useProduct(order.item?.product, { course: order.item?.course });
+  const intl = useIntl();
+  const getRoutePath = getDashboardRoutePath(intl);
+  const getRouteLabel = getDashboardRouteLabel(intl);
+
+  const links = useMemo(
+    () => [
+      {
+        to: getRoutePath(LearnerDashboardPaths.ORDER, { orderId: params.orderId }),
+        label: getRouteLabel(LearnerDashboardPaths.ORDER_RUNS),
```
rlecellier, line 28: may be the label for `ORDER` is wrong in `LearnerDashboardPaths`

rlecellier (approved):
Every thing is :green_circle: , look's ok to me! 
:warning:  I wasn't able to live test it with data.

src/frontend/js/utils/test/mockProduct.ts
jbpenrath: `mockProductWithOrder` ?

src/frontend/js/utils/test/mockProduct.ts
```diff
@@ -0,0 +1,13 @@
...
+  const product: Product = ProductFactory.generate();
+  product.id = order.product;
```
jbpenrath, line 7: With the brand new factory 😎

```suggestion
  const product = ProductFactory({
    id: order.product,
  }).one();
```

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 - list teacher's organizations in the teacher dashboard sidebar.
 - Added Certificates in the dashboard
 - Dashboard infinite scroll of orders and enrollments
+- Specific sidebar for order related routes
```
jbpenrath, line 19: Think to move up into Unreleased section

src/frontend/js/pages/DashboardOrderApplication/index.spec.tsx
jbpenrath, line 75: `userEvent.click` does not work ? 🤔

jbpenrath:
Here are my feedbacks after testing : 
So I think we should not display "Applications", "Contract" and "Stats" menu entries.
Furthermore I'll rename "Sessions" entry into "General information".

Then according to mockups, the course title should not be put within the sidebar header but just above the menu entries.

Then I'll set a min-height to the sidebar in order to improve layout when there are one or two menu entries.

Here it is a layout suggestion to discuss about : 
image

src/frontend/js/utils/test/mockProduct.ts
```diff
@@ -0,0 +1,13 @@
...
+  fetchMock.get(
+    'https://joanie.endpoint/api/v1.0/products/' + product.id + '/?course=' + order.course,
```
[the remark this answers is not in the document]
NathanVss: 100% agree

src/frontend/js/pages/DashboardOrderApplication/index.spec.tsx
NathanVss: The goal is to use the router to see how the sidebar behaves here

src/frontend/js/pages/DashboardOrderLayout/index.tsx
NathanVss: Good catch !

NathanVss: It is voluntary 😅

CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ Versioning](https://semver.org/spec/v2.0.0.html).
```
jbpenrath, line 8: Within your commit message : 
```diff
-We need to display a custom sidebar when navigating insde orders
+We need to display a custom sidebar when navigating inside orders
```

src/frontend/js/pages/DashboardOrderApplication/index.spec.tsx
```diff
@@ -0,0 +1,81 @@
...
+  it('renders breadcrumb and back goes to order', async () => {
+    const order: Order = {
+      ...OrderFactory().one(),
+      target_courses: [TargetCourseFactory().one()],
+      enrollments: [],
+    };
+    const product = mockProductWithOrder(order);
+    fetchMock.get(
+      'https://joanie.endpoint/api/v1.0/orders/',
+      { results: [order], next: null, previous: null, count: null },
```
jbpenrath, line 61:
```suggestion
      { results: [order], next: null, previous: null, count: 1 },
```

src/frontend/js/pages/DashboardOrderLayout/index.spec.tsx
```diff
@@ -0,0 +1,84 @@
...
+    });
+    screen.getByRole('link', {
+      name: 'Application',
+    });
+    screen.getByRole('link', {
+      name: 'Stats',
+    });
+    screen.getByRole('link', {
+      name: 'Contracts',
+    });
```
jbpenrath, line 82: Those routes should be hidden for now.

src/frontend/js/pages/DashboardCourses/index.spec.tsx
```diff
@@ -375,7 +379,7 @@ describe('<DashboardCourses/>', () => {
     await act(async () => userEvent.click(loadMoreButton));
     await waitFor(() => expectList(entities.slice(0, 250), products), { timeout: 30000 });
     expect(screen.queryByRole('button', { name: 'Load more' })).not.toBeInTheDocument();
-  }, 30000);
+  }, 300000);
```
jbpenrath, line 382: 😢

jbpenrath (approved):
LGTM ✅

PR #1977, opened 2023-05-05
💚(circleci) fix build back job

jbpenrath:
## Purpose

The job build back currently fail because cairo package seems missing.

## Proposal

- [x] Install missing cairo package


docker-compose.yml
```diff
@@ -24,8 +24,8 @@ services:
     build:
       context: .
       target: development
-      args:
-        DOCKER_USER: ${DOCKER_USER:-1000}
+#      args:
+#        DOCKER_USER: ${DOCKER_USER:-1000}
```
lunika, line 28: Why adding comment ?

lunika:
I would suggest you to change your commit title with something like `fix build back job by installing cairo`

kernicPanel (approved):

PR #1980, opened 2023-05-09
🎨(frontend) icon in storybook

rlecellier:
We need to have a visualisation of existing icons on richie.

## Purpose

Simple story with all existing icon for richie.

## Proposal

image


src/frontend/js/components/Icon/index.tsx
```diff
@@ -3,7 +3,7 @@ import { PropsWithTestId } from 'types/utils';
 export type IconSize = 'small' | 'medium' | 'large';
 
 type Props = PropsWithTestId<{
-  name: IconType;
+  name: IconTypeEnum | IconType;
```
jbpenrath, line 6: I think we should only allow `IconTypeEnum`

PR #1982, opened 2023-05-16
🐛(frontend) fix factories initialization

rlecellier:
The given type as factory type parameter does not
work as expected.
In order to force the return type of factories's builder function, it need's to be directly define as the builder return type.


jbpenrath (approved):
LG(T):M ✅

PR #1984, opened 2023-05-16
✨(frontend) build page teacher dashboard classrooms

rlecellier:
## Purpose

Part of the Teacher dashboard : Add page for course run


janrenz:
So we do store fixed height values? Why is this needed?

PR #1996, opened 2023-06-05
♻️(frontend)generice use union resourse

rlecellier:
## Purpose

Use a generic code to merge entities from two different endpoints

### DIFF
--- a/src/frontend/js/hooks/useUnionResource/index.spec.tsx
+++ b/src/frontend/js/hooks/useUnionResource/index.spec.tsx
@@ -1,0 +1,319 @@
+import { renderHook, act, waitFor } from '@testing-library/react';
+import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
+import { IntlProvider } from 'react-intl';
+import fetchMock from 'fetch-mock';
+import { PropsWithChildren } from 'react';
+import { PaginatedResponse, PaginatedResourceQuery } from 'types/Joanie';
+import { History, HistoryContext } from 'hooks/useHistory';
+import { createTestQueryClient } from 'utils/test/createTestQueryClient';
+import { SessionProvider } from 'contexts/SessionContext';
+import { Deferred } from 'utils/test/deferred';
+
+import { noop } from 'utils';
+import { FetchEntityData } from './utils/fetchEntities';
+import { QueryConfig, FetchDataFunction } from './utils/fetchEntity';
+import useUnionResource from '.';
+
+interface TestDataA {
+  name: 'TestDataA';
+  id: string;
+  created_on: string;
+}
+interface TestDataAFilters extends PaginatedResourceQuery {
+  isFiltered?: boolean;
+}
+interface TestDataB {
+  name: 'TestDataB';
+  id: string;
+  created_on: string;
+}
+
+const mockPaginatedResponse: <Data>(
+  results?: Data[],
+  totalCount?: number,
+  haveNextPage?: boolean,
+) => PaginatedResponse<Data> = (results = [], totalCount = 0, haveNextPage = true) => {
+  return {
+    count: totalCount,
+    next: haveNextPage ? 'next' : null,
+    prev: null,
+    results,
+  };
+};
+
+const renderUseUnionResource = <
+  DataA extends FetchEntityData,
+  DataB extends FetchEntityData,
+  FiltersA extends PaginatedResourceQuery,
+  FiltersB extends PaginatedResourceQuery,
+>(
+  queryAConfig: QueryConfig<DataA, FiltersA>,
+  queryBConfig: QueryConfig<DataB, FiltersB>,
+) => {
+  const Wrapper = ({ client, children }: PropsWithChildren<{ client?: QueryClient }>) => {
+    const historyPushState = jest.fn();
+    const historyReplaceState = jest.fn();
+    const makeHistoryOf: (params: any) => History = () => [
+      {
+        state: { name: '', data: {} },
+        title: '',
+        url: `/`,
+      },
+      historyPushState,
+      historyReplaceState,
+    ];
+
+    return (
+      <QueryClientProvider client={client ?? createTestQueryClient({ user: true })}>
+        <IntlProvider locale="en">
+          <HistoryContext.Provider value={makeHistoryOf({})}>
+            <SessionProvider>{children}</SessionProvider>
+          </HistoryContext.Provider>
+        </IntlProvider>
+      </QueryClientProvider>
+    );
+  };
+
+  return renderHook(
+    () =>
+      useUnionResource<DataA, DataB, FiltersA, FiltersB>({
+        queryAConfig,
+        queryBConfig,
+        perPage: 3,
+      }),
+    { wrapper: Wrapper },
+  );
+};
+
+describe('useUnionResource', () => {
+  let perPage: number;
+  let dataAList: TestDataA[];
+  let dataBList: TestDataB[];
+  let queryAConfig: QueryConfig<TestDataA, TestDataAFilters>;
+  let queryBConfig: QueryConfig<TestDataB, PaginatedResourceQuery>;
+
+  beforeEach(() => {
+    perPage = 3;
+    dataAList = [
+      { name: 'TestDataA', id: '1', created_on: '2022-01-01' },
+      { name: 'TestDataA', id: '2', created_on: '2022-02-01' },
+      { name: 'TestDataA', id: '3', created_on: '2022-07-01' },
+
+      { name: 'TestDataA', id: '4', created_on: '2022-03-01' },
+      { name: 'TestDataA', id: '5', created_on: '2022-06-01' },
+      { name: 'TestDataA', id: '6', created_on: '2022-04-01' },
+
+      { name: 'TestDataA', id: '7', created_on: '2022-05-01' },
+    ];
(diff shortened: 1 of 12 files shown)


src/frontend/js/hooks/useUnionResource/utils/fetchEntity.ts
```diff
@@ -0,0 +1,73 @@
...
+  if (typeof eofRef.current[queryKeyString] === 'number' && page > eofRef.current[queryKeyString]) {
+    return;
+  }
+
+  const filters: Filters = {
+    ...(queryConfigFilters || {}),
+    page,
+    page_size: perPage,
+  };
+  const QUERY_KEY = ['user', ...queryKey, JSON.stringify(filters)];
```
jbpenrath, line 48: As this method is generic, it should not add `user` in for all queries.
```suggestion
  const QUERY_KEY = [...queryKey, JSON.stringify(filters)];
```

src/frontend/js/hooks/useUnionResource/utils/fetchEntity.ts
```diff
@@ -0,0 +1,73 @@
...
+  if (data) {
+    return data;
+  }
+  const res = await fn(filters);
+  queryClient.setQueryData(QUERY_KEY, res);
+
+  // If we reached the end of the list, we set the eof flag to prevent future requests.
+  if (!res.next) {
+    eofRef.current = { ...eofRef.current, [queryKeyString]: filters.page! };
+    // Eof is cached based, the same way, we cache the fetching data. Otherwise it there would
```
jbpenrath, line 68:
```suggestion
    // Eof is cached based, the same way, we cache the fetching data. Otherwise there would
```

src/frontend/js/hooks/useUnionResource/utils/hasIntegrity.spec.ts
```diff
@@ -0,0 +1,32 @@
...
+    const stack: IntegrityData[] = [{ id: '1' }];
+    const newStack: IntegrityData[] = [{ id: '1' }, { id: '2' }];
+    expect(hasIntegrity<IntegrityData>(stack, newStack, 2)).toBe(false);
+  });
+  it('should return false if `newStack` is smaller than requested integrity count', () => {
+    // type StackData = { name: string };
+    const stack: IntegrityData[] = [{ id: '1' }, { id: '2' }];
+    const newStack: IntegrityData[] = [{ id: '1' }];
+    expect(hasIntegrity<IntegrityData>(stack, newStack, 2)).toBe(false);
+  });
```
jbpenrath, line 31: Add blank lines between each tests.

src/frontend/js/hooks/useUnionResource/utils/hasIntegrity.ts
```diff
@@ -0,0 +1,18 @@
+export interface IntegrityData {
+  id: string;
+}
+
+export const hasIntegrity = <Data extends IntegrityData = IntegrityData>(
```
jbpenrath, line 5: Can you add a JSDoc comment above the method to explain its purpose ?

src/frontend/js/hooks/useUnionResource/utils/log.ts
```diff
@@ -0,0 +1,7 @@
+export const DEBUG = true;
```
jbpenrath, line 1: Should be defined within settings.
```
DEBUG = process.env.NODE_ENV != 'production' && true
```

src/frontend/js/hooks/useUnionResource/utils/syncIntegrityCount.ts
```diff
@@ -0,0 +1,122 @@
...
+}
+interface SyncIntegrityCountError<DataA, DataB> {
+  errorMessage: MessageDescriptor;
+  totalCount: number;
+  page: number;
+  stack: (DataA | DataB)[];
+  integrityCount: number;
+}
+
+export const syncIntegrityCount = async <
```
jbpenrath, line 30: Ditto do not forget to add a comment

src/frontend/js/hooks/useUnionResource/utils/syncIntegrityCount.ts
```diff
@@ -0,0 +1,122 @@
...
+      // eslint-disable-next-line no-console
+      console.error(e);
+      return {
+        errorMessage: {
+          id: 'hooks.useOrdersEnrollments.errorGet',
+          description:
+            'Error message shown to the user when orders or enrollments fetch request fails.',
+          defaultMessage:
+            'An error occurred while fetching orders and enrollments. Please retry later.',
+        },
```
jbpenrath, line 83: Something I missed previously I think, but this message should be declared at the module scope through `defineMessages` in order to be extracted by formatjs.

jbpenrath (commented):
Good start 👍

src/frontend/js/hooks/useUnionResource/index.ts
```diff
@@ -0,0 +1,135 @@
...
+
+  // Eof keeps track of the last page of orders and enrollments we fetched.
+  // Eof needs to be a ref as subsequent calls to fetchEntity needs to keep track of the latest
+  // value of eof. Which would be impossible with setState.
+  const queryClient = useQueryClient();
+  const eofRef = useRef<Record<string, number>>(queryClient.getQueryData(EOF_QUERY_KEY) ?? {});
+  log('eof', eofRef.current);
+
+  useEffect(() => {
+    async function fetchNewPage() {
```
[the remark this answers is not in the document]
rlecellier: This function is calling `await syncIntegrityCount` so the `async` is needed

src/frontend/js/settings.ts
```diff
@@ -36,3 +36,5 @@ export const PAYMENT_SETTINGS = {
 };
 
 export const MOCK_SERVICE_WORKER_ENABLED = false;
+
+export const DEBUG_UNION_RESOURCES_HOOK = process.env.NODE_ENV !== 'production' && true;
```
jbpenrath, line 40: My previous suggestion could be simplified I think 😅
```suggestion
export const DEBUG_UNION_RESOURCES_HOOK = process.env.NODE_ENV !== 'production';
```

src/frontend/js/hooks/useUnionResource/index.spec.tsx
```diff
@@ -0,0 +1,319 @@
+import { renderHook, act, waitFor } from '@testing-library/react';
```
jbpenrath, line 1: In our commit message : Move `Co-Authored-By` in the commit footer

src/frontend/js/hooks/useUnionResource/index.ts
```diff
@@ -0,0 +1,151 @@
...
+  perPage?: number;
+  errorGetMessage?: MessageDescriptor;
+}
+
+/**
+ * This hook is used to fetch two different paginated resources from the API.
+ *
+ * Good to know:
+ * - The first time useOrdersEnrollments is called, it will fetch the first page of orders and
+ *    enrollments automatically.
```
jbpenrath, line 44:
```suggestion
 * - The first time useUnionResource is called, it will fetch the first page of both resources automatically.
```

src/frontend/js/hooks/useUnionResource/index.ts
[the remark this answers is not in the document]
jbpenrath: 👍

jbpenrath (approved):
LGTM ✅
@NathanVss WDYT ?

src/frontend/js/pages/DashboardCourses/useOrdersEnrollments.tsx
```diff
@@ -1,28 +1,13 @@
...
 
-export enum DataType {
-  ENROLLMENT = 'enrollment',
-  ORDER = 'order',
-}
-
-export type Data = {
-  type: DataType;
-  item: Enrollment | Order;
+export const isOrder = (obj: Order | Enrollment): obj is Order => {
```
NathanVss, line 6: I get what you wanted to do by removing the `type` but I find that using attributes to determine the type is misleading and is more bug-prone for the future maintenance. As of now if the backend changes an attribute it must keep in mind that the frontend is detecting the type based on this, IMO it breaks the I/O principle ( and SRP ). WDYT all ?

NathanVss (approved):

src/frontend/js/settings.ts
rlecellier: I'll change it by `export const DEBUG_UNION_RESOURCES_HOOK = process.env.NODE_ENV !== 'production' && false;`
otherwise we'll have too many message in the browser consol

src/frontend/js/hooks/useUnionResource/utils/syncIntegrityCount.ts
```diff
@@ -0,0 +1,140 @@
...
+        stack: currentStack,
+        queryAConfig,
+        queryBConfig,
+        page: pageToLoad,
+        perPage,
+        eofQueryKey,
+        eofRef,
+        queryClient,
+      });
+      if (totalCount === undefined || res.totalCount > totalCount) {
```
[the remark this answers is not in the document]
rlecellier: I had cases where a subsequent `totalCount` arrive at 0. It may had come from broken test data during development tho, it shouldn't happen in real life.

PR #2018, opened 2023-07-07
🐛(frontend) make remaining_order_count optional

jbpenrath:
## Purpose

Currently, Joanie product endpoint does not return the `remaining_order_count`. In order to prevent to break the CourseProductItemWidget, we make this property optional.

## Proposal

- [x] Make `Product.remaining_order_count` optional


NathanVss (approved):
LGTM :)

PR #2029, opened 2023-07-20
Feat/bind features into context

jbpenrath:
## Purpose

Richie backend has a settings `FEATURES` to en.dis.able some feature in
    development. We are currently porting the course run list to React and this
    content contains a development feature `WISHLIST`. In order to be able to en
    .dis.able this feature, we need to get `FEATURES` settings through frontend, so
    we decided to bind this setting to `FRONTEND_CONTEXT`

## Proposal

- [x] Bind `settings.FEATURES` into `__richie_frontend_context__.context`

### DIFF
--- a/env.d/development/dev.dist
+++ b/env.d/development/dev.dist
@@ -5,6 +5,8 @@ AUTHENTICATION_BACKEND=dummy
 # LMS Backend
 EDX_BASE_URL=http://localhost:8073
 EDX_BACKEND=richie.apps.courses.lms.base.BaseLMSBackend
+EDX_COURSE_REGEX='^.*/courses/(?P<course_id>.*)/info/?$'
+EDX_JS_COURSE_REGEX='^.*/courses/(.*)/info/?$'
 EDX_JS_BACKEND=dummy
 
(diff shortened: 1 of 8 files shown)


env.d/development/dev.dist
```diff
@@ -5,7 +5,9 @@ AUTHENTICATION_BACKEND=dummy
 # LMS Backend
 EDX_BASE_URL=http://localhost:8073
 EDX_BACKEND=richie.apps.courses.lms.base.BaseLMSBackend
+EDX_COURSE_REGEX='^.*/courses/(?P<course_id>.*)/info/?$'
+EDX_JS_COURSE_REGEX='^.*/courses/(.*)/info/?$'
```
rlecellier, line 9: It look's like a other topic but i'm not sure that it deserve a commit.

rlecellier (approved):
Look's great!

env.d/development/dev.dist
jbpenrath: Yup, I could move it into a dedicated commit you're right

PR #2040, opened 2023-07-28
🐛(frontend) fix organization breadcrumb

rlecellier:
We have an error when organization is not in ReactQuery cache.


jbpenrath (approved):
LGTM ✅

PR #2054, opened 2023-08-09
🚚(frontend) purchase button as generic component

rlecellier:
## Purpose

We want to use `PurchaseButton` in the learner dashboard.
It need to be moved into components with all components/hooks/context that's used on it like `SaleTunnel`, `PaymentButton` and `PaymentInterface`.


jbpenrath (approved):
LGTM ✅

Issue #2063, opened 2023-08-28
Empty end date on course runs are considered minimum and shown


2018D:
## Bug Report

**Problematic Behavior**
Concerning the new way of showing course runs.
Empty end dates are considered as MINIMUM dates and show as Dec 31, 1969.
Course runs with empty course end dates and empty enrollment end date are archived instead of being open.

**Expected behavior/code**
Empty dates should be considered MAXIMUM dates and not show up like in v2.22.0

**Steps to Reproduce**
1. Create or modify an existing course run
2. Remove course end date
3. Dec 31, 1969 shows up in the course run detail. See screenshot 1

**Steps to Reproduce**
1. Create or modify an existing course run
2. Remove course end date and enrollment end date
3. Course run is archived. See screenshot 2

**Environment**
- Richie version: v2.23.0+
- Platform: Unix

**Possible Solution**

**Additional context/Screenshots**

I haven't checked what shows up on course glimpse yet.

![Screenshot1](https://github.com/openfun/richie/assets/17911549/e4d9dfe1-8e83-4773-b71c-cf12c3271e23)
-----------
![Screenshot2](https://github.com/openfun/richie/assets/17911549/9df82bd8-c976-4439-8eb9-82eae7e9874f)

2018D:
New commit onto the same branch to re-enable forever open course run on course page

NathanVss:
I'm working a the fix concerning the archived closed, it should be submitted today. Thank you for your report 🙏

NathanVss:
Here is it @2018D, let me know if it fixes your issue https://github.com/openfun/richie/pull/2083

2018D:
Thanks @NathanVss, It fixes the issue ! I included it before my own commit on my pull request because it allows me to test if the course is open when there's no end date and no enrollment_end date.

2018D:
Was fixed with commits from pull request #2083 and #2080

PR #2083, opened 2023-09-05
🐛(frontend) fix course run that should not be archived

NathanVss:
A miss in the code caused courses runs without enrollment end date or without end date to be considered as archived, where they are in reality ongoing.

Based on #2063


rlecellier (approved):
Maybe add a changelog: 
- Fix course run computed state. A run without end date must be ONGOING_OPEN.

PR #2084, opened 2023-09-06
⬆️(cookiecutter) fix cookiecutter circleci gitlint

igobranco:
Missing `requests` pip package when running `gitlint`, like the same have been done on: https://github.com/openfun/richie/blob/2e3d0f4078071243a026c51ffa57b27c975ac707/.circleci/config.yml#L65


sampaccoud (approved):
Nice thank you for the fix @igobranco :star_struck:

PR #2085, opened 2023-09-07
Rlecellier/order context

rlecellier:
## Purpose

`PaymentButton` is not working in the learner dashboard since it need `OrderProvider`.

## Proposal

To avoid duplicating the code of `CourseProductItem` in `ProductCertificateFooter` i decided to switch from a context to a hook `useProductOrder`.
This hook is in charge of returning the last order pending, submitted or validated for a product id and a course code.


src/frontend/js/components/PaymentButton/index.spec.tsx
```diff
@@ -467,60 +530,84 @@ describe('PaymentButton', () => {
...
-    );
-
-    expect(JSON.parse(fetchMock.lastOptions()!.body!.toString())).toEqual({
-      payment_id: paymentInfo.payment_id,
-    });
-
-    // - An error message should be displayed and focused (for screen reader users)
-    const $error = screen.getByText('An error occurred during payment. Please retry later.');
-    expect(document.activeElement).toBe($error);
+    // // - Route to create order should have been called
```
rlecellier, line 551: There is an error that i don't understand: 
```sh
Unmatched POST to https://joanie.test/api/v1.0/orders/1cc7077a-b494-47b6-9f43-b5dab3ba6f98/abort/
```
where this url is mock on line 508.

src/frontend/js/widgets/SyllabusCourseRunsList/components/CourseProductItem/index.tsx
```diff
@@ -64,7 +64,7 @@ const Header = ({ product, order, hasPurchased, isPendingState, compact }: Heade
   const formatDate = useDateFormat();
 
   const canShowMetadata = useMemo(() => {
-    return compact && !order;
+    return compact && (!order || [OrderState.SUBMITTED, OrderState.PENDING].includes(order!.state));
```
rlecellier, line 67: By reading tests this seems wrong.
Tests expect to show metadata until order is `VALIDATED`

src/frontend/js/components/PaymentButton/index.spec.tsx
```diff
@@ -36,42 +35,57 @@ jest.mock('utils/context', () => ({
...
     cleanup();
-    fetchMock.restore();
-    jest.clearAllTimers();
-    jest.resetAllMocks();
   });
 
   it('should render a payment button', async () => {
     const product: Joanie.Product = ProductFactory().one();
-
+    fetchMock.get(
```
NathanVss, line 85: Maybe I lack a bit of context: why do we need to rewrite this test? is it because of the workflow changes?

src/frontend/js/components/PaymentButton/index.spec.tsx
```diff
@@ -102,6 +116,10 @@ describe('PaymentButton', () => {
     const product: Joanie.Product = ProductFactory().one();
     const creditCard: Joanie.CreditCard = CreditCardFactory().one();
 
+    fetchMock.get(
```
NathanVss, line 119: Why is this mock added multiple times ?

src/frontend/js/components/PaymentButton/index.spec.tsx
```diff
@@ -168,14 +193,16 @@ describe('PaymentButton', () => {
...
-      <Wrapper order={undefined} client={createTestQueryClient({ user: { username: 'John Doe' } })}>
+      <Wrapper client={createTestQueryClient({ user: true })}>
         <PaymentButton
           billingAddress={billingAddress}
           product={product}
           onSuccess={handleSuccess}
         />
       </Wrapper>,
     );
+    nbApiCalls += 1; // fetch order for useProductOrder
```
NathanVss, line 204: I think we could ensure that the expected url has been called here, instead of testing counts it would be more accurate, WDYT?

src/frontend/js/components/PaymentButton/index.spec.tsx
```diff
@@ -205,7 +237,8 @@ describe('PaymentButton', () => {
     });
 
     // - Once payment succeeded, order should be refetch
-    expect(fetchMock.calls()).toHaveLength(3);
+    nbApiCalls += 1; // fetch validated order
```
NathanVss, line 240: ( the same goes for the other expects ofc )

src/frontend/js/hooks/useProductOrder/index.tsx
```diff
@@ -0,0 +1,31 @@
+import { useMemo } from 'react';
+import { useOrders } from 'hooks/useOrders';
+import { OrderState } from 'types/Joanie';
+
+interface UseProductOrderProps {
+  courseCode: string;
+  productId: string;
+}
+const useProductOrder = ({ courseCode, productId }: UseProductOrderProps) => {
```
NathanVss, line 9: WDYT about an isolated test of this hook?

src/frontend/js/widgets/SyllabusCourseRunsList/components/CourseProductItem/index.tsx
NathanVss: Yes + a bit of comment to make easier the reading of this condition would be great too

NathanVss (commented):
🫡

src/frontend/js/widgets/SyllabusCourseRunsList/components/CourseProductItem/index.tsx
rlecellier: Before: we simply create an `Order` with `OrderFactory`
Now: we mock the api call made by `useCurrentOrder`, it add the need of `JoanieSession` and all these small changes.

src/frontend/js/components/PaymentButton/index.spec.tsx
rlecellier: Before: we simply create an `Order` with `OrderFactory`
Now: we mock the api call made by `useCurrentOrder`, it add the need of `JoanieSession` and all these small changes.

src/frontend/js/components/PaymentButton/index.spec.tsx
rlecellier: it's added once by test, it's the call made by `useCurrentOrder`

src/frontend/js/components/PaymentButton/index.spec.tsx
rlecellier: I don't think that we need to check that this precise call is made in each test.
But we want to: 
* test at least once that each call have been made at the right time.
* test that no unwanted call have been made, by checking how many queries we've at each step.

NathanVss: It's up to you, I would have made sure that the URL is the expected one, as you want

NathanVss (approved):
Nice work!

PR #2091, opened 2023-09-12
🚚(frontend) rename teacher dashboard components

rlecellier:
TeacherDashboard must be a common prefix of teacher dashboard pages and components.

We had lot's of wrong names like "TeacherMyPageDashboard" instead of "TeacherDashboardMyPage"


NathanVss (approved):
All right !!

PR #2092, opened 2023-09-19
♻️(frontend) api.joanie: add buildApiUrl

rlecellier:
We create a function to handle url build that was repeat over and over.


PanchoutNathan (approved):
Just add a test and LGTM :)

PR #2098, opened 2023-09-26
♻️(frontend) upgrade Order type

rlecellier:
order have a new field `enrollment` that contain the enrollment from where the product certificate has been bought.
`order.enrollments` have been renamed `order.target_enrollments`.


src/frontend/js/types/Joanie.ts
```diff
@@ -190,7 +190,7 @@ export interface Order {
   id: string;
```
jbpenrath, line 190: Why you do not take opportunity to add the new field `enrollment` ?

rlecellier: It's not used yet and i didn't want to add it in case the backend change before we use it in richie.

PanchoutNathan: I agree, and what's more, it simplifies the understanding of renaming

jbpenrath (approved):
LGTM ✅

PR #2112, opened 2023-10-09
🧪(frontend) test that user token is refresh

rlecellier:
We've a production issue where all api calls to joanie return 401. This test reproduce this bug.


rlecellier:
Fix in https://github.com/openfun/richie/pull/2115

PR #2113, opened 2023-10-10
🧑‍💻(frontend) use enum for http status code

rlecellier:
## Purpose

Improve readability by using text instead of code for http status error 

## Proposal

add an enum `HttpStatusCode` and use it instead of raw code.


src/frontend/js/api/joanie.spec.ts
```diff
@@ -1,10 +1,11 @@
 import fetchMock from 'fetch-mock';
```
jbpenrath, line 1: I'll use a `🏷️` gitmoji for your commit message as this change impacts types 😉

jbpenrath (approved):
LGTM ✅

PR #2118, opened 2023-10-17
🐛(frontend) Fix sale tunnel

rlecellier:
:skull: The payment step of the sale tunnel is not accessible

Quickfix: 
Sale tunnel modal is render in a portal.
All our route are render inside a `<CunninghamProvider/>` but the modal's portal wasn't.

### DIFF
--- a/src/frontend/js/components/Modal/index.tsx
+++ b/src/frontend/js/components/Modal/index.tsx
@@ -1,7 +1,7 @@
 import { ReactNode, useMemo } from 'react';
 import ReactModal from 'react-modal';
 import { defineMessages, FormattedMessage, useIntl } from 'react-intl';
-import { Button } from '@openfun/cunningham-react';
+import { Button, CunninghamProvider } from '@openfun/cunningham-react';
 import { StringHelper } from 'utils/StringHelper';
 import { Icon, IconTypeEnum } from 'components/Icon';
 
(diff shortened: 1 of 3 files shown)


src/frontend/js/components/Modal/index.tsx
```diff
@@ -69,26 +69,28 @@ export const Modal = ({
...
-          >
-            <Icon name={IconTypeEnum.ROUND_CLOSE} />
-            <span className="offscreen">
-              <FormattedMessage {...messages.closeDialog} />
-            </span>
-          </Button>
-        )}
-      </header>
-      {children}
+      <CunninghamProvider>
```
jbpenrath, line 72: I think the issue is more global. You should declare the CunninghamProvider at the top level of the React app https://github.com/openfun/richie/blob/56fa25c9119168f3329cf43d54eb96df3c279857/src/frontend/js/index.tsx#L115-L123

jbpenrath (approved):
LGTM ✅

src/frontend/js/widgets/Dashboard/index.spec.tsx
jbpenrath, line 1: You must write a commit message.
As this commit is about testing you should use ✅ gitmoji

src/frontend/js/widgets/Dashboard/index.spec.tsx
```diff
@@ -41,10 +42,11 @@ jest.mock('hooks/useIntersectionObserver', () => ({
 }));
 
 describe('<Dashboard />', () => {
-  const DashboardWithUser = ({ user }: { user: Nullable<User> }) => {
+  let user: UserEvent;
```
jbpenrath, line 45: I would prefer to declare `const user = userEvent.setup()` in each test suite where it is needed.

NathanVss (approved):
🛩️

PR #2124, opened 2023-10-24
🐛(frontend) OrderStateMessage handle error on "validated" state

jbpenrath:
## Purpose

The OrderStateMessage does not handle validated state properly. Indeed this component does not have message to display for this state and it handles an error like this state was unknown. Of course, we only want to handle error in case an unknown state is provided but not when the provided state does not have message to display.

## Proposal

- [x] Call `handle` method only when unknown state is provided


NathanVss (approved):
🛫

PR #2127, opened 2023-10-26
🛂(backend) update default courses and organizations permissions

jbpenrath:
## Purpose

[django-filer 2.2.5](https://github.com/django-cms/django-filer/blob/master/CHANGELOG.rst#225-2023-06-11) has fixed a bug that allowed any user to access to folder without permissions. On our side, we took this behavior like a feature not a bug so we did not add corresponding filer permissions in default permissions applied to course and organization creation. Since this bug has been fixed, our staff users are not able anymore to access to their organization or course folder. In order to fix that we need add the missing filer permissions to the default permissions.

Furthermore, it appears we forgot to add also several permissions related to new plugins like LTIConsumerPlugin for example so we took opportunity of this fix to add those missing permissions.

## Proposal

- [x] Update courses and organizations default permissions


lunika (approved):

NamFra (approved):
Perfect! 😎🙏

PR #2136, opened 2023-11-08
🐛(front) fix payment in one click for dummy backend

NathanVss:
Previously the dummy payment with one click payment was causing maximum polling attempts due to non calling of notification URL. I reworked the architecture in order to delegate this decision down to the PaymentInterface.


src/frontend/js/components/PaymentButton/index.spec.tsx
```diff
@@ -434,6 +333,13 @@ describe('PaymentButton', () => {
       fireEvent.click($button);
     });
 
+    // - In real world condition the success callback is immediately called for one click payments.
+    // - but here we need to click manually.
+    const $success = screen.getByTestId('payment-success');
+    await act(async () => {
+      fireEvent.click($success);
+    });
+
```
jbpenrath, line 342: Should we update the mock of PaymentInterfaces instead ?

jbpenrath (approved):
LGTM ✅

PR #2144, opened 2023-11-21
👽️(frontend) update filter types

rlecellier:
All joanie's filters now use descriptive suffix like: _id, _code, _title, etc


NathanVss (approved):
LGTM :)

Issue #2157, opened 2023-12-04
React query cache issues on Contract signature


jbpenrath:
## Bug Report

**Problematic Behavior**
If the user navigates to its dashboard. In "training" list, if an order has a pending contract to sign then the user clicks on sign button, sign the contract, if it comes back to the course list, the button "sign" is still here.

**Expected behavior/code**
Once a contract has been signed, the related order should be updated and the button "Sign" should be hidden everywhere.

**Steps to Reproduce**
1. Purchase a product with a contract definition
2. Close the SaleTunnel and do not sign directly the contract
3. From the User Menu, go to the learner dashboard
4. From the course list, click on the "Sign" button of the freshly purchased order
5. Sign the contract
6. Comes back to course list
7. See that the order has not been updated.

**Environment**
- Richie version: next
- Platform: N/A

**Possible Solution**
Once a contract has been signed, orders query should be invalidated.

**Additional context/Screenshots**

https://github.com/openfun/richie/assets/9265241/65f65f36-7d46-4ae8-8b9b-83a5e14e8122

PR #2165, opened 2023-12-06
🐛(frontend) fix various cache order issues

NathanVss:
We had a bug where the cache invalidation triggered when a contract is signed was not impacting useOrdersEnrollments. This was causing the "Sign" button to be still displayed. A related bug where after buying a product from the syllabus page, the newly created order was not shown was also fixed by theses changes.


jbpenrath (commented):
You could add `Fix #2157` in the commit footer. In this way, the related issue will be closed on merge 😉 

LGTM ✅

jbpenrath (approved):

Issue #2158, opened 2023-12-04
Hide sign button on cancelled order ( or even simply hide cancelled orders ? )


NathanVss:
## Bug Report

**Problematic Behavior**
At the moment it is possible to click on the "sign" button of a cancelled order, but it triggers an API error.

**Expected behavior/code**
Two choices:
- Hide sign button on cancelled orders.
- Never show cancelled orders. 

**Additional context/Screenshots**
<img width="1409" alt="Capture d’écran 2023-12-01 à 15 43 31" src="https://github.com/openfun/richie/assets/9628870/186835ac-8ae0-4d99-b1af-6592266c8d4d">

jbpenrath:
I believed that canceled orders was not returned. If this is not the case, I think they should be hidden by default. We will see later if we should add filters allow user to get them.

PR #2166, opened 2023-12-06
🐛(frontend) exclude cancelled orders from dashboard

NathanVss:
Showing cancelled orders was causing various issues during API calls. We decided to simply stop showing them.


jbpenrath (approved):
Ditto you could mention the issue in the commit footer to link

`Resolve #2158`

LGTM ✅

PR #2167, opened 2023-12-06
💄(frontend) hide order sidebar, hide filters space

NathanVss:
We decided to simply hide the sidebar inside the order details because it was empty, and in the future we will just have to remove the display none clause to show it again. Also remove the empty space caused by the filters container, now we only show it if there are filters.

### DIFF
--- a/src/frontend/js/widgets/Dashboard/components/DashboardLayout/index.tsx
+++ b/src/frontend/js/widgets/Dashboard/components/DashboardLayout/index.tsx
@@ -1,22 +1,29 @@
 import { PropsWithChildren, ReactNode } from 'react';
 import { useLocation } from 'react-router-dom';
+import c from 'classnames';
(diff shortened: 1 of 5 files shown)


src/frontend/js/widgets/Dashboard/components/DashboardLayout/index.tsx
```diff
@@ -6,17 +6,23 @@ import { DashboardBreadcrumbs } from 'widgets/Dashboard/components/DashboardBrea
...
+export const DashboardLayout = ({
+  children,
+  sidebar,
+  filters,
+  className,
+}: DashboardLayoutProps) => {
   const location = useLocation();
   return (
-    <div className="dashboard">
+    <div className={['dashboard', className].filter(Boolean).join(' ')}>
```
jbpenrath, line 20: You could use `classnames` deps instead.

jbpenrath (approved):
LGTM ✅

PR #2170, opened 2023-12-07
🐛(frontend) multiple calls of pushBreadcrumbsPlaceholders

rlecellier:
## Purpose

We had multiple trigger of a useEffect that was enoying to test pages that use breadcrumb's placeholders.

### DIFF
--- a/src/frontend/js/hooks/useBreadcrumbsPlaceholders.tsx
+++ b/src/frontend/js/hooks/useBreadcrumbsPlaceholders.tsx
@@ -9,5 +9,5 @@ export const useBreadcrumbsPlaceholders = (data: DashboardBreadcrumbsPlaceholder
   const context = useContext(DashboardBreadcrumbsContext);
   useEffect(() => {
     context.pushBreadcrumbsPlaceholders(data);
-  }, [data]);
+  }, [JSON.stringify(data)]);
 };


src/frontend/js/hooks/useBreadcrumbsPlaceholders.tsx
```diff
@@ -5,9 +5,12 @@ import { DashboardBreadcrumbsContext } from 'widgets/Dashboard/contexts/Dashboar
...
-export const useBreadcrumbsPlaceholders = (data: DashboardBreadcrumbsPlaceholders) => {
+export const useBreadcrumbsPlaceholders = (
+  data: DashboardBreadcrumbsPlaceholders,
+  dependencies: any[],
+) => {
   const context = useContext(DashboardBreadcrumbsContext);
   useEffect(() => {
     context.pushBreadcrumbsPlaceholders(data);
-  }, [data]);
+  }, dependencies);
```
NathanVss, line 15: What about replacing old `[data]` by `[JSON.stringify(data)]` it would maybe work avoid us to define `dependencies` for each time we use `useBreadcrumbsPlaceholders`, WDYT?

NathanVss (commented):
Nice finding! Check out my proposal, if it works it would be more concise ✌️

NathanVss (approved):
LGTM 🤌

PR #2174, opened 2023-12-11
♻️(react) use Cunningham button everywhere

NathanVss:
The legacy Button component was still used in some places, including
RouterButton. This commit deletes this legacy and use Cunningham
Button everywhere, with some size changes for visual coherence.

Capture d’écran 2023-12-11 à 17 10 14
Capture d’écran 2023-12-11 à 17 10 08
Capture d’écran 2023-12-11 à 17 10 02


src/frontend/js/widgets/Dashboard/components/DashboardBreadcrumbs/_styles.scss
```diff
@@ -19,6 +19,7 @@
   li {
     margin-right: 1rem;
     padding-right: 0.5rem;
+    height: 1.5rem;
```
jbpenrath, line 22: Sure? Why not use padding instead ?

src/frontend/scss/tools/_buttons.scss
```diff
@@ -273,3 +273,8 @@
   margin: inherit;
   padding: inherit;
 }
+
+// Without this, bootstrap overrides Cunningham buttons style.
+a.c__button:hover {
+  text-decoration: none;
+}
```
jbpenrath, line 280: I think it should moved into `richie/src/frontend/scss/trumps/_bootstrap.scss`

jbpenrath (approved):
Great !

jbpenrath:
Add an entry into the CHANGELOG please.

Furthermore, you should add an entry in the UPDATE.md to mention that the react component Button has been removed in favor of Cunningham Button component.

src/frontend/js/widgets/Dashboard/components/DashboardBreadcrumbs/_styles.scss
NathanVss: Yes the icon break the height and resulting in a broken alignement :/

PR #2182, opened 2023-12-12
Frontend - Use full_name if it exists

jbpenrath:
## Purpose

Since fonzie 0.5.0, authentication server can now return user full name. So in this case, this property is defined, we should display it instead of username.

With PR openfun/joanie#489, Joanie create dev demo script now generate several users which stick to realistic use cases. Learner, teacher, organization owner are created so on Richie, in order to improve DX, we should be able to switch easily between those profiles. That's why we update the dummy api interface. We hardcode all existing users created by dev demo joanie script then we implement a method `getUserInfo` to return the right payload just by providing a username.

## Proposal

- [x] Display fullname instead of username if it exists
- [x] Improve DX of dummy authentication backend

### DIFF
--- a/src/frontend/js/api/lms/dummy.ts
+++ b/src/frontend/js/api/lms/dummy.ts
@@ -5,6 +5,49 @@ import { APILms } from 'types/api';
 import { Enrollment, OpenEdXEnrollment } from 'types';
 import { location } from 'utils/indirection/window';
 import { RICHIE_USER_TOKEN } from 'settings';
+import { base64Decode } from 'utils/base64Parser';
+
+type JWTPayload = {
+  email: string;
+  exp: number;
+  full_name: string;
+  iat: number;
+  jti: string;
+  language: string;
+  token_type: 'access';
+  username: string;
+};
+
+/* All JWT tokens will expire the 12 Dec 2024 ! */
+const JOANIE_DEV_DEMO_USER_JWT_TOKENS = {
+  admin:
+    'eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJ0b2tlbl90eXBlIjoiYWNjZXNzIiwiZXhwIjoxNzMzOTI4MjE0LCJpYXQiOjE3MDIzNjU1MjMsImp0aSI6IjRhMzQxZWVmMmVhOTRkNGFiMzQ5OThkOWE4ZDM5MTI0IiwiZW1haWwiOiJhZG1pbkBleGFtcGxlLmNvbSIsImxhbmd1YWdlIjoiZW4tdXMiLCJ1c2VybmFtZSI6ImFkbWluIiwiZnVsbF9uYW1lIjoiIn0.rT8nymp8f4T7tIIXO-M5-ahXBwxoDNVqtaZIrb_GHuk',
+  user0:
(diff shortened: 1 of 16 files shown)


src/frontend/js/widgets/Dashboard/components/LearnerDashboardSidebar/index.tsx
```diff
@@ -46,7 +47,7 @@ export const LearnerDashboardSidebar = (props: Partial<DashboardSidebarProps>) =
   return (
     <DashboardSidebar
       menuLinks={links}
-      header={intl.formatMessage(messages.header, { name: user?.username })}
+      header={intl.formatMessage(messages.header, { name: UserHelper.getName(user!) })}
```
NathanVss, line 50: In `TeacherDashboardProfileSidebar/index.tsx` you used `header={user ? UserHelper.getName(user) : ''}`, why not here?

src/frontend/js/widgets/UserLogin/index.not.isJoanieEnabled.spec.tsx
```diff
@@ -111,7 +111,7 @@ describe('<UserLogin />', () => {
 
     await userEvent.click(button);
 
-    screen.getByText(user.username);
+    screen.getByText(user.full_name!);
```
NathanVss, line 114: Here it is equivalent to: `screen.getByText(undefined);`. Is it what you want? 🤔

src/frontend/js/api/lms/dummy.ts
```diff
@@ -6,6 +6,79 @@ import { Enrollment, OpenEdXEnrollment } from 'types';
...
+
+    {
+      access_token:
+        'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ0b2tlbl90eXBlIjoiYWNjZXNzIiwiZXhwIjoxNzMzOTI4MjE0LCJpYXQiOjE3MDIzOTIyMTQsImp0aSI6ImNkZjAyMGM4ODdjOTQxYzU5ZmExN2FkZGExNjNjMDIzIiwiZW1haWwiOiJqZWFuLWJhcHRpc3RlLnBlbnJhdGgrc3R1ZGVudF91c2VyQGZ1bi1tb29jLmZyIiwibGFuZ3VhZ2UiOiJmci1mciIsInVzZXJuYW1lIjoic3R1ZGVudF91c2VyIiwiZnVsbF9uYW1lIjoiXHUwMGM5dHVkaWFudCJ9.JMdnC2VXwq2VbNPrIYxj8PEq0oJJ4LZZT_ywWyE1lBM',
+      username: 'student_user',
+      full_name: 'Étudiant',
+    },
+  ];
+
+  return JOANIE_DEV_DEMO_USERS.find((u) => u.username === username);
```
NathanVss, line 79: You could have used an object indexed by the tokens username in order to O(1) find the according token, WDYT?

NathanVss (approved):
💯

src/frontend/js/widgets/UserLogin/index.not.isJoanieEnabled.spec.tsx
jbpenrath: Nope I just say I'm sure that the user has a full_name property defined and I want to retrieve this text in the rendered content.

src/frontend/js/api/lms/dummy.ts
jbpenrath: Indeed

src/frontend/js/widgets/Dashboard/components/LearnerDashboardSidebar/index.tsx
jbpenrath: Yup right let's use that here tto

src/frontend/js/widgets/UserLogin/index.not.isJoanieEnabled.spec.tsx
NathanVss: Yep, but on line 56 you wrote `const user: User = UserFactory({ full_name: undefined }).one();` 🤔

NathanVss: Oh crap !! I missed the Github code wrapping, nevermind 😂

PR #2183, opened 2023-12-13
🏷️(frontend) update Contract

rlecellier:
Contract are signed twice, once by the studdent then once by a
    organization owner.
    Here we update our Contract type


jbpenrath (approved):
LGTM ✅

PR #2190, opened 2023-12-14
♻️(frontend) refactor joanie api endpoints

jbpenrath:
## Purpose

Currently an error can occured when a user can access to a course from its organization access but it does not course/training accesses on this course. Indeed, the user will be able to access to the course/training detail page but the api will return a 404... 

To work properly, in the context of an organization, course/training information should be fetched from the nested organization endpoints.

| Before | After |
|--------|-------|
|Capture d’écran 2023-12-14 à 16 23 19|Capture d’écran 2023-12-14 à 16 22 51|
|Capture d’écran 2023-12-14 à 16 23 19|Capture d’écran 2023-12-14 à 16 22 33|
## Proposal

- [x] Improve buildAPIUrl method to support several url string params
- [x] Fetch course and course product relations from nested organizations or root endpoint according to teacher dashboard context.

### DIFF
--- a/src/frontend/js/api/joanie.spec.ts
+++ b/src/frontend/js/api/joanie.spec.ts
@@ -12,21 +12,47 @@ describe('api/joanie', () => {
     expect(responseBody).toBe('');
     fetchMock.restore();
   });
+
   it('buildApiUrl should build api url', () => {
     interface TestApiFilters extends ResourcesQuery {
+      addresses: string[];
+      age?: number;
       firstname: string;
+      gender: string;
+      is_active: boolean;
       lastname: string;
-      gender: string[];
+      links: string[];
+      profile: string | null;
     }
     const apiFilters: TestApiFilters = {
       id: 'DUMMY_TEST_API_ID',
+      addresses: ['1', '2'],
+      age: undefined,
       firstname: 'John',
+      gender: '',
+      is_active: false,
       lastname: 'Do',
-      gender: ['male', 'robot'],
+      links: [],
+      profile: null,
     };
-    const url = buildApiUrl('test/:id/', apiFilters);
+    const url = buildApiUrl('http://example.com/test/:id/', apiFilters);
     expect(url).toEqual(
-      'test/DUMMY_TEST_API_ID/?firstname=John&gender=male&gender=robot&lastname=Do',
+      'http://example.com/test/DUMMY_TEST_API_ID/?addresses=1&addresses=2&firstname=John&is_active=false&lastname=Do',
     );
+  });
+
+  it('buildApiUrl should build api url with nested resource', () => {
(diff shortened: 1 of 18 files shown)


src/frontend/js/api/joanie.spec.ts
```diff
@@ -12,21 +12,24 @@ describe('api/joanie', () => {
...
     }
     const apiFilters: TestApiFilters = {
       id: 'DUMMY_TEST_API_ID',
       firstname: 'John',
       lastname: 'Do',
       gender: ['male', 'robot'],
+      is_active: true,
     };
-    const url = buildApiUrl('test/:id/', apiFilters);
+    const url = buildApiUrl('http://example.com/test/:id/', apiFilters);
```
rlecellier, line 30: It would be nice to have a test with different type like: `null`, `undefined` and `''`

src/frontend/js/widgets/SyllabusCourseRunsList/components/CourseProductItem/index.tsx
```diff
@@ -160,12 +160,10 @@ const Content = ({ product, order }: { product: Product; order?: CredentialOrder
...
-  const { item: courseProductRelation, states: productQueryStates } = useCourseProduct(
-    course.code,
-    {
-      productId,
-    },
-  );
+  const { item: courseProductRelation, states: productQueryStates } = useCourseProduct({
+    product_id: productId,
+    course_id: course.code,
+  });
```
rlecellier, line 166: part of the previous commit

src/frontend/js/api/joanie.ts
```diff
@@ -191,17 +191,31 @@ export const getRoutes = () => {
...
-  if (id) url = url.replace(':id', id);
-  else url = url.replace(':id/', '');
-
-  if (!ObjectHelper.isEmpty(queryParameters)) {
-    url += '?' + queryString.stringify(queryParameters);
+  // eslint-disable-next-line compat/compat
+  const url = new URL(raw_url);
+  if (filters) {
+    Object.entries(filters).forEach(([key, value]) => {
+      if (url.pathname.search(`:${key}`) > 0) {
```
NathanVss, line 201: You could first verify that `value` is not `undefined` at the very start of the callback instead of verifying it two times, WDYT?

NathanVss, line 201: You could first verify that value is not undefined at the very start of the callback instead of verifying it two times, WDYT?

NathanVss (approved):
LGTM ! 😊

src/frontend/js/api/joanie.ts
jbpenrath: Indeed !

NathanVss (approved):
✅

PR #2193, opened 2023-12-18
🩹(frontend) clean billing address before order submit

jbpenrath:
## Purpose

It appears there was a typing issue that lead to pass a wrong payload. This could break payment logic when working the dummy payment backend. Indeed, the billing address should not contain id and is_main fields but currently the address was not cleaned up before putting it in the payload. We now use the new `ObjectHelper.omit` utils to remove id and is_main fields from the selected address.

## Proposal

- [x] Remove `id` and `is_main` properties from address before passing it to `order.submit`

### DIFF
--- a/src/frontend/js/components/PaymentButton/index.tsx
+++ b/src/frontend/js/components/PaymentButton/index.tsx
@@ -14,6 +14,7 @@ import { CourseProductEvent } from 'types/web-analytics';
 import useProductOrder from 'hooks/useProductOrder';
 import { useTerms } from 'components/PaymentButton/hooks/useTerms';
 import { useSaleTunnelContext } from 'components/SaleTunnel/context';
+import { ObjectHelper } from 'utils/ObjectHelper';
 import PaymentInterface from './components/PaymentInterfaces';
 
 const messages = defineMessages({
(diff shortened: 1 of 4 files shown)


src/frontend/js/utils/ObjectHelper/index.spec.ts
```diff
@@ -1,19 +1,27 @@
...
-  it('should return true if array is empty', () => {
-    expect(ObjectHelper.isEmpty([])).toBe(true);
+    it('should return true if array is not empty', () => {
+      expect(ObjectHelper.isEmpty(['foo'])).toBe(false);
+    });
   });
 
-  it('should return true if array is not empty', () => {
-    expect(ObjectHelper.isEmpty(['foo'])).toBe(false);
+  describe('omit', () => {
```
NathanVss, line 22: We could maybe add a case which makes sure that there is no errors when trying to omit attributes not present? 

e.g. `ObjectHelper.omit({ foo: 1, bar: 2, baz: 3 }, 'foo', 'bazinga')`

src/frontend/js/components/PaymentButton/index.tsx
```diff
@@ -153,10 +154,12 @@ const PaymentButton = ({ billingAddress, creditCard, onSuccess }: PaymentButtonP
       let paymentInfos = payment;
 
       if (!paymentInfos) {
+        const cleanedBillingAddress = ObjectHelper.omit(billingAddress!, 'id', 'is_main');
```
NathanVss, line 157: Suggestion about naming: `billingAddressPayload` or `billingAddressData`, the term `cleaned` does not make me think of the content of an API request instantly. It's up to you 😊

NathanVss (approved):
📨

src/frontend/js/utils/ObjectHelper/index.spec.ts
jbpenrath: With the signature function, a typescript error is raised 😉

NathanVss: Nice 😏

PR #2195, opened 2023-12-18
✨(frontend) add ContractHelper to compute state

jbpenrath:
## Purpose

A contract expects to be signed by both learner and organization, in some of our
 views we need to get the signature state. That's the purpose of this
 ContractHelper util. It is able to return the raw state or a human readable
 label. Furthermore, according to point of view of the user (learner or
 organization), label should be adapted.

## Proposal

- [x] Create a ContractHelper to easily get signature state.
- [x] Update ContractStatus to get state from the new ContractHelper

### DIFF
--- a/src/frontend/js/types/Joanie.ts
+++ b/src/frontend/js/types/Joanie.ts
@@ -416,6 +416,12 @@ export interface CourseProductRelationQueryFilters extends PaginatedResourceQuer
   id?: CourseProductRelation['id'];
   organization_id?: Organization['id'];
 }
+
+export enum ContractState {
+  UNSIGNED = 'unsigned',
+  LEARNER_SIGNED = 'half_signed',
+  SIGNED = 'signed',
+}
 export interface ContractFilters extends PaginatedResourceQuery {
   organization_id?: Organization['id'];
   course_id?: CourseListItem['id'];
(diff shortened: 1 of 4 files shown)


src/frontend/js/types/Joanie.ts
```diff
@@ -416,6 +416,12 @@ export interface CourseProductRelationQueryFilters extends PaginatedResourceQuer
   id?: CourseProductRelation['id'];
   organization_id?: Organization['id'];
 }
+
+export enum ContractState {
+  UNSIGNED = 'unsigned',
+  HALF_SIGNED = 'half_signed',
+  SIGNED = 'signed',
+}
```
rlecellier, line 424:
```suggestion
export enum ContractState {
  UNSIGNED = 'unsigned',
  STUDDENT_SIGNED = 'half_signed',
  SIGNED = 'signed',
}
```

src/frontend/js/utils/ContractHelper/index.spec.ts
```diff
@@ -0,0 +1,70 @@
+import { createIntl } from 'react-intl';
+import { ContractFactory } from 'utils/test/factories/joanie';
+import { ContractState } from 'types/Joanie';
+import { ContractHelper, ContractStatePoV } from '.';
+
+const unsignedContractFactory = ContractFactory({
```
rlecellier, line 6: When i first read i would have feel more confortable wiith `UnsignedContractFactory`, but it's a instanciated factories so may be not.

rlecellier (approved):
It work !
But it's not easy to see how useful it is without usage example.

src/frontend/js/types/Joanie.ts
jbpenrath: Finally I named it `LEARNER_SIGNED`

Issue #2187, opened 2023-12-13
Frontend - Learner Dashboard - Use banner to mention that user has no enrollment / order to show


jbpenrath:
On dashboard learner, currently when the user has no enrollment / order, a simple text is displayed. But on certificate or training contract views a banner is used and the look'n feel looks better !

We should use also a banner in this view.

<img width="900" alt="image" src="https://github.com/openfun/richie/assets/9265241/939cd315-8ba1-4d2d-a9f2-8f5d3c33c5c5">
<img width="900" alt="image" src="https://github.com/openfun/richie/assets/9265241/5749562d-60a2-4673-a9da-8cfe59378a28">

PR #2198, opened 2023-12-19
💄(frontend) Banner on dashboard/courses

rlecellier:
issue: [#2187](https://github.com/openfun/richie/issues/2187)

Use a banner component to display the "empty list" message in the orders and enrollments listing page.

Before: 
Capture d’écran du 2023-12-19 14-09-37
After:
Capture d’écran du 2023-12-19 14-10-31

### DIFF
--- a/src/frontend/js/pages/DashboardCourses/index.spec.tsx
+++ b/src/frontend/js/pages/DashboardCourses/index.spec.tsx
@@ -19,7 +19,7 @@ import { SessionProvider } from 'contexts/SessionContext';
 import { LearnerDashboardPaths } from 'widgets/Dashboard/utils/learnerRouteMessages';
 import { CourseLight, CourseProductRelation, Enrollment, CredentialOrder } from 'types/Joanie';
 import { expectNoSpinner, expectSpinner } from 'utils/test/expectSpinner';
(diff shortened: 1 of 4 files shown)


src/frontend/js/pages/DashboardCourses/index.tsx
```diff
@@ -51,7 +52,7 @@ export const DashboardCourses = () => {
         <>
           {count === 0 && (
             <p className="dashboard__courses__empty">
-              <FormattedMessage {...messages.emptyList} />
+              <Banner message={intl.formatMessage(messages.emptyList)} />
```
jbpenrath, line 55: Can you just update the related test to use the `expectBanner` utils ?

src/frontend/js/pages/DashboardCourses/index.spec.tsx
```diff
@@ -138,7 +138,7 @@ describe('<DashboardCourses/>', () => {
     });
   };
 
-  it('renders an empty placeholder', async () => {
+  it.only('renders an empty placeholder', async () => {
```
NathanVss, line 141: Forgot to remove the `only` clause?

NathanVss (commented):
<- Is the merge commit expected? ( Instead of rebase? 👀 )

Issue #2168, opened 2023-12-07
Frontend - Student Dashboard - Load more button displayed on resource loading


jbpenrath:
## Bug Report

**Problematic Behavior**
On student dashboard, when enrollment and orders are loading, the "LOAD MORE" button is displayed

**Expected behavior/code**
During fetching, only the loader should be displayed

**Steps to Reproduce**
1. Clear the react-query cache
2. Go to http://localhost:8070/en/dashboard

**Environment**
- Richie version: master

**Possible Solution**
Hide the "LOAD MORE" button when order/enrollment are loading

**Additional context/Screenshots**
<img width="1216" alt="Capture d’écran 2023-12-07 à 11 43 52" src="https://github.com/openfun/richie/assets/9265241/a79f8c33-959f-4812-907d-d844d8819dad">

PR #2199, opened 2023-12-19
🐛(frontend) Fix useUnionResource hasLoadMore

rlecellier:
issue: [#2168](https://github.com/openfun/richie/issues/2168)

In the orders and enrollments listing page where we use this useUnionResources, we had a "Load more" button on the first render, when data where loading.
hasLoadMore must be false until we retrieve data for the first time.


jbpenrath (approved):
LGTM ✅

PR #2204, opened 2023-12-21
🔥(backend) remove unused course_enrollment_widget_props extra tag

jbpenrath:
## Purpose

Since we port course run list rendering into React, the extra tag `course_enrollment_widget_props` is not used anymore. So we remove it.


NathanVss (approved):
LGTM 🚀

PR #2205, opened 2023-12-21
🧑‍💻(frontend) fix storybook

rlecellier:
Since cunnigham is used, most of our stories were broken because CunninghamProvider was missing.

### DIFF
--- a/src/frontend/.storybook/preview.tsx
+++ b/src/frontend/.storybook/preview.tsx
@@ -1,7 +1,8 @@
 import { useState } from 'react';
 import { IntlProvider } from 'react-intl';
-import './__mocks__/utils/context';
+import { CunninghamProvider } from '@openfun/cunningham-react';
 import { useAsyncEffect } from 'hooks/useAsyncEffect';
+import './__mocks__/utils/context';
 
 export const parameters = {
   actions: { argTypesRegex: '^on[A-Z].*' },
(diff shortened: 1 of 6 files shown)


src/frontend/.storybook/preview-body.html
```diff
@@ -1 +1 @@
-<div id="modal-exclude" style="min-height: 0;"></div>
\ No newline at end of file
+<div id="modal-exclude" style="height: auto;"></div>
```
jbpenrath, line 1: A blank line is missing

src/frontend/.storybook/preview.tsx
```diff
@@ -34,7 +35,9 @@ export const decorators = [
   (Story) => (
     <IntlProvider locale="en">
       <IconsWrapper>
-        <Story />
+        <CunninghamProvider>
+          <Story />
+        </CunninghamProvider>
```
jbpenrath, line 40: Did you check that some stories does not import CunninghamProvider internally ?

jbpenrath (approved):
LGTM ✅

src/frontend/.storybook/preview.tsx
rlecellier: i hadn't, it's done now some stories needed to have the provider removed.

PR #2211, opened 2024-01-04
🩹(frontend) fix dashboard sidebar zindex issue

jbpenrath:
## Purpose

On mobile, the dashboard sidebar is displayed at the top of the page. But currently its z-index is less than the content so when user open the select menu
 to navigate this one appears below the content.

## Proposal

- [x] Increase z-index of `dashboard-sidebar`


NathanVss (approved):
🪩

PR #2212, opened 2024-01-04
🩹(frontend) minor fixes

jbpenrath:
## Purpose

Currently, the content of the dashboard sidebar header was not align to center so with long content, this one was left aligned that was weird.

| Before | After |
|--------|-------|
|Capture d’écran 2024-01-04 à 13 36 05|image|

Furthermore, some messages have been declared without using the react-intl defineMessages
method so those messages have not been uploaded to crowdin.

## Proposal

- [x] Align dashboard sidebar header text content to center
- [x] Use `defineMessages` where it is missing


PanchoutNathan (approved):
Phew, I thought I wasn't going to get away with this HUGE review!

PR #2215, opened 2024-01-05
🐛(frontend) fix getCourseGlimpseProps intl then release v2.25.0-beta.1

jbpenrath:
## Purpose

In some case, the `getCourseGlimpseProps` method needs to use an intl interface
to compute the course_url prop. Currently, we were creating the intl object
inside this method but we did not provide messages so this was not work. In
order to fix that, we are now passing an intl object to this method instead of
locale to afterward create an intl object.

#### Fixed
- Fix `getCourseGlimpseProps` method to handle localized course urls

## Proposal

- [x] Fix `getCourseGlimpseProps` method with localized `course_url` props
- [x] Release richie v2.25.0-beta.1


CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ Versioning](https://semver.org/spec/v2.0.0.html).
```
NathanVss, line 8: body: 

"... so this was not workING*"

NathanVss (approved):
LGTM :) 🇫🇷

rlecellier (approved):

PR #2224, opened 2024-01-12
🩹(frontend) await logout request

jbpenrath:
## Purpose

The authentication logout is async but we do not wait the promise resolve when the user triggers logout method. But in some case it could lead to bug. Indeed, if the user query invalidation trigger a redirect, not await the promise resolve, lead to refetch user data before the authentication server has really logout the user so the session is populated with outdated user information...

## Proposal

- [x] Await logout authentication request


NathanVss (approved):
🐣 !

PR #2225, opened 2024-01-15
🚧(frontend) Split useContract into useContract and useOrganizationContract 

rlecellier:
## Purpose

We have three routes to interact with contracts:
* /contracts/id : return current user orders's contrats
* /organizations/id/contracts/id: return contract for a organization, check `OrganizationAccess`.
* /courses/id/contracts/id: return contract for a organization, check `CourseAccess`.

The current implementation of theses route use a unique hook that check filters to decide which route to use between contract and organizations/contracts. 
In order to implement courses's contract page and action, we need to interact with `courses/contracts` route and it's filters like : 
`courses/course_id/contracts/contract_id/?organization_id=organization_id`.
This isn't possible with the current implementation of contracts.get routes.

I decide here to split `useContract` hook in three to get one hook per routes

### DIFF
--- a/src/frontend/js/api/joanie.ts
+++ b/src/frontend/js/api/joanie.ts
@@ -177,6 +177,9 @@ export const getRoutes = () => {
       products: {
         get: `${baseUrl}/courses/:course_id/products/:id/`,
       },
+      contracts: {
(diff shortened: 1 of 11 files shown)


rlecellier:
duplicated by https://github.com/openfun/richie/pull/2228

PR #2228, opened 2024-01-16
♻️(frontend) Split useContract into useContract and useOrganizationContract 

rlecellier:
## Purpose

We have three routes to interact with contracts:
* /contracts/id : return current user orders's contrats
* /organizations/id/contracts/id: return contract for a organization, check `OrganizationAccess`.
* /courses/id/contracts/id: return contract for a organization, check `CourseAccess`.

The current implementation of theses route use a unique hook that check filters to decide which route to use between contract and organizations/contracts. 
In order to implement courses's contract page and action, we need to interact with `courses/contracts` route and it's filters like : 
`courses/course_id/contracts/contract_id/?organization_id=organization_id`.
This isn't possible with the current implementation of contracts.get routes.

I decide here to split `useContract` hook in three to get one hook per routes

### DIFF
--- a/src/frontend/js/hooks/useContracts/index.tsx
+++ b/src/frontend/js/hooks/useContracts/index.tsx
@@ -1,6 +1,6 @@
 import { defineMessages } from 'react-intl';
 import { useJoanieApi } from 'contexts/JoanieApiContext';
-import { useResource, useResources, UseResourcesProps } from 'hooks/useResources';
+import { QueryOptions, useResource, useResources, UseResourcesProps } from 'hooks/useResources';
 import { API, Contract, ContractFilters } from 'types/Joanie';
 
 const messages = defineMessages({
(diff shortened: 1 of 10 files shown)


src/frontend/js/hooks/usePendingContractCount/index.ts
jbpenrath, line 1: The filename must match the name of the hook.

src/frontend/js/hooks/usePendingContractCount/index.ts
```diff
@@ -0,0 +1,57 @@
...
+      contracts,
+      pendingContractCount: meta?.pagination?.count ?? 0,
+      canSignContracts: contractAbilities.can(ContractActions.SIGN),
+    };
+  }
+  return {
+    contracts: [],
+    pendingContractCount: 0,
+    canSignContracts: false,
+  };
```
jbpenrath, line 54: From my point of view, this is not the purpose of this hook to check if the authenticated user has the right to sign contract.

src/frontend/js/hooks/usePendingContractCount/index.ts
```diff
@@ -0,0 +1,57 @@
...
+import { PER_PAGE } from 'settings';
+import { ContractState, CourseProductRelation, Organization } from 'types/Joanie';
+import { ContractActions } from 'utils/AbilitiesHelper/types';
+
+interface UsePendingContractCountProps {
+  organizationId?: Organization['id'];
+  courseProductRelationId?: CourseProductRelation['id'];
+}
+
+const useTeacherPendingContracts = ({
```
jbpenrath, line 14: I'll rename this hook `useTeacherPendingContractsCount` to explicit its purpose.

src/frontend/js/hooks/useContracts/index.tsx
```diff
@@ -22,8 +22,45 @@ const props: UseResourcesProps<Contract, ContractFilters, API['user']['contracts
...
+
+export const useOrganizationContract = (
+  id: string,
+  filters: ContractFilters,
+  queryOptions?: QueryOptions<Contract>,
+) => {
+  return useResource(organizationProps)(id, filters, {
+    ...queryOptions,
+    enabled:
+      !!filters?.organization_id && (queryOptions?.enabled === undefined || queryOptions.enabled),
```
jbpenrath, line 53: If no id is provided, the query should be disabled.

```suggestion
     !!id && !!filters?.organization_id && (queryOptions?.enabled === undefined || queryOptions.enabled),
```

jbpenrath (approved):

PR #2235, opened 2024-01-18
🐛(circle) fix tag check in npm job

lunika:
## Purpose

In the NPM job, the check on the CIRCLE_TAG was wrong and we don't pass in it if we are in a master branch.
We change it with the solution found in
https://stackoverflow.com/a/13864829

## Proposal

- [x] fix tag check in npm job

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -928,7 +928,7 @@ jobs:
         auth:
           username: $DOCKER_USER
           password: $DOCKER_PASS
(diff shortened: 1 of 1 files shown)


lunika:
@jbpenrath I also updated the `yarn.lock` file in the cookicutter. The update was made using the package `"richie-education": "2.25.0-beta.1"`.

jbpenrath (approved):
LGTM ✅

PR #2239, opened 2024-01-19
✨(plugin) add notifications to courses

### LINKED ISSUE #2238: Course warning and notifications
## Feature Request

We want to have the ability to notify users when some issue occurs on the course. 
An example of this can be:
* Some contents are missing;
* The certificates are only issued when the course ends;
* There's an issue with the course;

This messages messages can also be use to provide information, like:
* The course start date has been pushed a few days.

It should be a simple element.

**Solution**
Implement a simple Notification plugin to display information to students when needed.
Information can be of type:
* Warning
* Information

This will have an effect on how the notification box is rendered.

**Mockup**
This is the implementation idea I have in mind.
image

sandroscosta:
Add a new plugin to create warning or information notifications. 
This plugin is only available on course detail pages.

image

Closes #2238


jbpenrath:
Hey Sandro, look's interesting. I'll have few suggestion to improve your contribution.

First, there is already an existing banner component and I think you could be use it here.
https://github.com/openfun/richie/blob/fcff6bc248d21f3acb7a0c8dcb5b1ef09c18cdbb/src/richie/apps/courses/templates/courses/cms/fragment_error_detail_template_banner.html#L1-L14

https://github.com/openfun/richie/blob/37e79952570fa5fac41d0bfa9e69850a7c644916/src/frontend/scss/objects/_banner.scss#L1

Then, I wonder if this the right place to put the notification.
What do you think to create a static placeholder (like footer) in the global page template ? 
image

sandroscosta:
Do you think it makes sense for us to meet in the middle? This banner is more "informative" than anything else. The one you show in the images seems to be related more to errors. When I created this, I focused on the course. The idea is to pass messages to the students inside the course page, hence why I chose to put it where it is.

I can refactor with my idea in mind. What do you think? Would a mix of both make sense?

sandroscosta:
> What do you think to create a static placeholder (like footer) in the global page template ?

I don't dislike the idea at all. I just want to restrict this kind of notifications to the courses and programs, for now.

jbpenrath:
> > What do you think to create a static placeholder (like footer) in the global page template ?
> 
> I don't dislike the idea at all. I just want to restrict this kind of notifications to the courses and programs, for now.

Yup sorry, I realized that the example of the footer placeholder was not a good one.
My suggestion was just to add the notification banner above the course content but of course this placeholder should be unique for each page.

sandroscosta:
> My suggestion was just to add the notification banner above the course content but of course this placeholder should be unique for each page.

Make sense. I'll check and work on it. I might create some additional styles to have different levels of alert states.

PR #2245, opened 2024-01-22
✨(frontend) add signature polling description

### LINKED ISSUE #2243: Frontend - Signature polling information
## Feature Request

After user has signed contract, your signature provider can take time to notify our backend. In order to wait this notification, we have a polling logic but currently there is no label to explain to the user why it is waiting for that is weird.

During the polling signature step, we should display a label to explain to the user why it is waiting and that this step can take up few minutes.

NathanVss:
We want to add a bit of explanations to the user becasue this step can take some time.

Capture d’écran 2024-01-22 à 17 10 38

Closes #2243

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Fixed
 
+- Add signature polling description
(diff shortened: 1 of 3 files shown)


jbpenrath (approved):
Do not forget to add an entry in the CHANGELOG

PR #2246, opened 2024-01-23
🐛(frontend) fix Enroll now translation

igobranco:
When using an external LMS, the CTA `Enroll now` on course runs isn't being translated.

Screenshot on french, is still showing the `Enroll now` on english.
image

After applying this fix and translate it on `src/frontend/i18n/locales/fr-FR.json`:
```json
"components.SyllabusCourseRun.enroll": {
    "description": "CTA for users who can enroll in the course run or could enroll if they logged in.",
    "message": "S’inscrire maintenant"
  },
```
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 - Switch from setup.cfg to pyproject.toml 
 
+### Fixed
+
+- Fix frontend translation of Enroll now for external LMS backend
+
 ## [2.25.0-beta.1]
 
 ### Added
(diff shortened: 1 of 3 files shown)


jbpenrath (commented):
Those labels should have been translated by the backend 🤔 
https://github.com/openfun/richie/blob/f47f5b0c383bdca38a274e70cccbd98837315c98/src/richie/apps/courses/models/course.py#L55-L56

igobranco:
@jbpenrath I think it doesn't come from backend, but from the frontend.
The CourseState is calculated on the frontend and not on the backend.
So I think it just need to translate the `call_to_action` Strings from the `computeState` function:
https://github.com/openfun/richie/blob/eae94fe7a4efb852fa55d6938c2167e51b17c25d/src/frontend/js/utils/CourseRuns/index.ts#L12-L79

igobranco:
Replaced by https://github.com/openfun/richie/pull/2553

PR #2248, opened 2024-01-23
🧑‍💻(frontend) ignore dev user selection

rlecellier:
We've to change dev user during development and it's common to forget about it on commit.
To avoid redundant task of switching dev user on each new branch, we moved this choice in an ignored file `settings.dev.ts`

* copy `settings.dev.dist.ts` as `settings.dev.ts`
* choose your user in `settings.dev.ts`
* enjoy

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -101,3 +101,6 @@ docker/files/etc/nginx/ssl/*
 
 # Storybook
 storybook-static
+
+# Frontend dev settings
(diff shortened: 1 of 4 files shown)


NathanVss (approved):
I love this idea!

jbpenrath (approved):
Lot's of idea for the next step of this work

PR #2252, opened 2024-01-24
fix cache enrollment product

NathanVss:
The payment button was still shown after buying the associcated product certificate. So this refactor aims to do some optimistic UI updates in order to hide it after SaleTunnel payment.

### DIFF
--- a/src/frontend/js/components/SaleTunnel/index.tsx
+++ b/src/frontend/js/components/SaleTunnel/index.tsx
@@ -50,13 +50,14 @@ const focusCurrentStep = (container: HTMLElement) => {
   }
 };
 
-interface SaleTunnelProps {
+export interface SaleTunnelProps {
   isOpen: boolean;
   onClose: () => void;
   course?: CourseLight;
   enrollment?: Enrollment;
   product: CredentialProduct | CertificateProduct;
   orderGroup?: OrderGroup;
+  onFinish?: (order: Order) => void;
 }
 
 const SaleTunnel = ({
@@ -66,6 +67,7 @@ const SaleTunnel = ({
   onClose,
   enrollment,
   orderGroup,
+  onFinish,
 }: SaleTunnelProps) => {
   const intl = useIntl();
   const {
@@ -111,6 +113,7 @@ const SaleTunnel = ({
           invalidateOrders();
           refetchOmniscientOrders();
           queryClient.invalidateQueries({ queryKey: ['user', 'enrollments'] });
+          onFinish?.(order!);
         },
         onExit: () => {
           handleModalClose();
--- a/src/frontend/js/components/SaleTunnel/index.tsx
+++ b/src/frontend/js/components/SaleTunnel/index.tsx
@@ -1,5 +1,6 @@
 import { useEffect, useMemo, useRef, useState } from 'react';
 import { defineMessages, useIntl } from 'react-intl';
+import { useQueryClient } from '@tanstack/react-query';
 import { Modal } from 'components/Modal';
 import {
   CourseLight,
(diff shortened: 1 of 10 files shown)


jbpenrath (approved):

src/frontend/js/hooks/useUnionResource/index.ts
```diff
@@ -87,6 +90,20 @@ const useUnionResource = <
...
+    setTotalCount(undefined);
+    setError(undefined);
+    setCursor(perPage);
+    setIntegrityCount(0);
+  };
+
+  if (refetchOnInvalidation) {
+    useQueryKeyInvalidateListener(queryAConfig.queryKey, reset);
+    useQueryKeyInvalidateListener(queryBConfig.queryKey, reset);
+  }
```
rlecellier, line 105: It's weird to have a conditional hook.
May be use a `enable` flag in `useQueryKeyInvalidateListener`

src/frontend/js/components/TeacherDashboardCourseList/index.spec.tsx
```diff
@@ -102,7 +102,7 @@ describe('components/TeacherDashboardCourseList', () => {
     );
```
rlecellier, line 102: in commit msg: something like: 

`useUnionResources` do not use react-query cache but it's own custom cache mechanic.
This custom mechanic didn't handle the refetch part of react-query.
There a update that handle queries refetch, as it should be, in our custom useUnionResource hook.

src/frontend/js/hooks/useUnionResource/index.ts
```diff
@@ -132,7 +149,7 @@ const useUnionResource = <
       setIsSyncing(true);
       fetchNewPage();
     }
-  }, [cursor]);
+  }, [cursor, stack.length]); // stack.length is added in the dependency array to force a new fetch on reset.
```
rlecellier, line 152: will this double the amount of execution of this useEffect  ? 
* one when we set cursor to "stack + per_page"
* one when integrity is ok and that we set stack with new fetched entites (l.135)

I see two other way to do it:
* in `reset`, to use `setCursor(0)` and add a new `useEffect` that will set back cursor to `per_page` when `cursor === 0`.
* have `useQueryKeyInvalidateListener` return the `refershFlag` and use this `refreshFlag` there to trigger a new sync.

src/frontend/js/components/SaleTunnel/index.tsx
```diff
@@ -76,6 +77,7 @@ const SaleTunnel = ({
   const key = `${product.type === ProductType.CREDENTIAL ? course!.code : enrollment!.id}+${
     product.id
   }`;
+  const queryClient = useQueryClient();
```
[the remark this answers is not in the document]
NathanVss: I find it much more verbose

src/frontend/js/hooks/useUnionResource/index.ts
NathanVss: Yes I get what you mean, the key point to keep in mind is that we cannot change this flag afterwards, or it will automatically trigger an error. I prefer having this pattern to avoid having to create listeners where we don't need them, for performance purpose.

src/frontend/js/components/TeacherDashboardCourseList/index.spec.tsx
NathanVss: I will provide more context

src/frontend/js/hooks/useUnionResource/index.ts
NathanVss: Yes but the second will not be triggering any additional render.

Using `useQueryKeyInvalidateListener` will not be possible as we want this hook to be optional, and setting cursor to 0 looks a bit odd semantically in my opinion. What really reflects the reality is that we simply trash the stack, so as this hook must guarantee a correct output, we must take into account the mutation of `stack` IMO

src/frontend/js/hooks/useUnionResource/index.ts
rlecellier: ok, let's say that no one will dynamically change the value of `refetchOnInvalidation`

PR #2256, opened 2024-01-26
🐛(frontend) fix logout issue

NathanVss:
We realized that, on logout, setting react query ["user"] to null was taking too much time to be persisted in SessionStorage resulting in a redirection to the homepage before the data being flushed into the SessionStorage. (It is caused by throttleTime of the persister ) Sometimes the user was not able to logout due to this issue.


src/frontend/js/api/lms/dummy.ts
jbpenrath, line 1: What do you think to store in local storage the username of the current user ?

So in this way, the `me` method, just have to retrieve the username and retrieve the related data.
Then the logout will remove the key.
And finally login and register methods will set it.

From my point of view, It's more logic 🤔

jbpenrath (approved):
LGTM ✅

Issue #2227, opened 2024-01-16
🐛(frontend) useUnionResources reset page scroll


rlecellier:
## Bug Report

**Problematic Behavior**
On the teacher dashboard course listing page.
When all the pages have been fetch, the list switch to the beginning of the page.

**Expected behavior/code**
We should stay at the end of the liste when finishing scrolling bottom.

**Steps to Reproduce**
1. go on the course listing page
2. scroll bottom
3. observe that you get teleporte to the beginig of the page.

**Environment**
- Richie version: 2.25.0-beta.1
- Platform: linux / firefox

**Additional context/Screenshots**
[Capture vidéo du 16-01-2024 16:06:48.webm](https://github.com/openfun/richie/assets/139848/9f12e436-fe6d-4212-9fd8-996027ebcd15)

NathanVss:
I can't find a way to reproduce this bug on MacOS, maybe it is 100% related to your OS 🤔

PR #2261, opened 2024-01-30
🐛(frontend) fix teacher courses list scroll top

rlecellier:
We had a bug where the courses list scroll was reset at the top of the page.
`useUnionResource` `isLoading` is switching between `true` and `false` each time it load a new chunk of results.
As `isLoading` was used in `TeacherDashboardCourseList` to switch display between the spinner and the courses list, the page scroll was sometime reset when the list disappear.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,8 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Fixed
 
+- Fix on teacher dashboard courses list. During infinit scroll loading the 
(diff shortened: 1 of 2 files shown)


NathanVss (approved):
Nice one, LGTM :)

jbpenrath:
You could amend our commit to add into the footer:
`Resolve #2227`

jbpenrath (changes requested):
Remove the merge commit

jbpenrath (approved):

PR #2262, opened 2024-01-30
✅(frontend) fix flaky test

rlecellier:
`DashboardCreditCardsManagement` test about credit card expiration date use a date in less than 3 month in the future.
This date some time happen to be exactly 3 month, just a few hours before the treshold of 3 month.
As `CreditCardHelper.isExpiredSoon` only check year and month and not hours, the test failed.


rlecellier:
duplicated fix with #2258

PR #2265, opened 2024-01-30
🐛(frontend) fix useContractFilters

rlecellier:
initialize filter had wrong value when using the first user's
organization. It happened because `useOrganization` fetch didn't update
`initialFilters`.

This fix also introduce `useDefaultOrganizationId` hook that centralize
the choice logic between route organization, query organization and
first user's organization.
This hook will be used in a second place on 'course bulk sign'
development, in order to use default organization on `CourseSidebar` when
route organization isn't define.


src/frontend/js/pages/TeacherDashboardContractsLayout/TeacherDashboardContracts/index.spec.tsx
```diff
@@ -12,7 +12,7 @@ import { createTestQueryClient } from 'utils/test/createTestQueryClient';
 import { ContractFactory, OrganizationFactory } from 'utils/test/factories/joanie';
```
jbpenrath, line 12: In the commit message :
```diff
-TeacherDashboardContracts/index.spec.tsx had some warnign that could
+TeacherDashboardContracts/index.spec.tsx had some warning that could
-lead to fake possitive tests.
+lead to fake positive tests.
```

src/frontend/js/pages/TeacherDashboardContractsLayout/components/ContractActionsBar/index.spec.tsx
```diff
@@ -87,10 +87,7 @@ describe('TeacherDashboardContractsLayout/ContractActionsBar', () => {
```
jbpenrath, line 87: In the commit message
```diff
-Therefor, we hide this button on a training context
+Therefore, we hide this button on a training context
```

jbpenrath (approved):

PR #2274, opened 2024-02-01
♻️(frontend) normalize download file feature

rlecellier:
## Purpose

We had the same code at three different places. It's a easy pick to
normalize this browser download feature.

### DIFF
--- a/src/frontend/js/api/joanie.ts
+++ b/src/frontend/js/api/joanie.ts
@@ -23,13 +23,21 @@ interface CheckStatusOptions {
   ignoredErrorStatus: (number | HttpStatusCode)[];
 }
 
+export async function getFileFromResponse(response: Response): Promise<File> {
+  const filenameRegex = /filename="(.*)";/;
+  const dispositionHeader = response.headers.get('Content-Disposition');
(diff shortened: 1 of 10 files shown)


src/frontend/js/api/joanie.ts
```diff
@@ -23,13 +23,21 @@ interface CheckStatusOptions {
   ignoredErrorStatus: (number | HttpStatusCode)[];
 }
 
+export async function getFileFromResponse(response: Response): Promise<File> {
+  const filenemRegex = /filename="(.*)";/;
```
jbpenrath, line 27:
```suggestion
  const filenameRegex = /filename="(.*)";/;
```

src/frontend/js/utils/download.ts
```diff
@@ -0,0 +1,38 @@
+import { handle } from './errors/handle';
+
+type DownloadFunction = () => Promise<File>;
+
+export const browserDownloadFromBlob = async (
```
jbpenrath, line 5: Can you just add a JS Doc to explain what is the purpose of this method.

jbpenrath (approved):
LGTM ✅

PR #2276, opened 2024-02-02
✨(frontend) link to terms and conditions page

jbpenrath:
## Purpose

Currently, into the sale tunnel when there is a contract to sign, the user must accept our terms and conditions. We allow the user to download a preview of the contract which include those terms and conditions.

This is not a proper solution for ux. Instead we decide to create a dedicated page on our platform then we just have to link this page in the sale tunnel. Simpler, better.

## Proposal

- [x] Bind terms-and-conditions url to frontend context
- [x] Replace the download contract preview logic by a link to our terms and conditions page.

### DIFF
--- a/src/frontend/js/components/PaymentButton/hooks/useTerms.tsx
+++ b/src/frontend/js/components/PaymentButton/hooks/useTerms.tsx
@@ -3,7 +3,7 @@ import { useState } from 'react';
 import { defineMessages, useIntl } from 'react-intl';
 import { PaymentErrorMessageId } from 'components/PaymentButton/index';
 import { Product } from 'types/Joanie';
-import { useJoanieApi } from 'contexts/JoanieApiContext';
+import context from 'utils/context';
 
 const messages = defineMessages({
   termsMessage: {
@@ -33,7 +33,6 @@ export const useTerms = ({
   error?: PaymentErrorMessageId;
 }) => {
   const intl = useIntl();
-  const api = useJoanieApi();
   const [termsAccepted, setTermsAccepted] = useState(false);
   const validateTerms = () => {
     if (!product.contract_definition) {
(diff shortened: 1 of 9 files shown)


src/frontend/js/components/PaymentButton/hooks/useTerms.tsx
```diff
@@ -66,12 +53,14 @@ export const useTerms = ({
             label={
               <>
                 {intl.formatMessage(messages.termsMessage)}{' '}
-                <button
-                  onClick={openContract}
+                <a
+                  href={context.site_urls.terms_and_conditions ?? '#'}
+                  target="_blank"
+                  rel="noopener noreferrer"
```
NathanVss, line 59: In theory the contract will be hosted on our own infrastructure, why do you set those two `noopener noreffered`? Is this just a habit or for another specific reason?

src/frontend/js/components/PaymentButton/hooks/useTerms.tsx
```diff
@@ -66,12 +53,14 @@ export const useTerms = ({
             label={
               <>
                 {intl.formatMessage(messages.termsMessage)}{' '}
-                <button
-                  onClick={openContract}
+                <a
+                  href={context.site_urls.terms_and_conditions ?? '#'}
```
NathanVss, line 57: In which cases this attribute could not be defined?

NathanVss (approved):
Noïce :)

src/frontend/js/components/PaymentButton/hooks/useTerms.tsx
lunika: IMO with `target="_blank"` you set this `rel`. No matter the targeted url.

lunika (approved):

src/frontend/js/components/PaymentButton/hooks/useTerms.tsx
jbpenrath: If the page does not exist.

PR #2283, opened 2024-02-07
🧑‍💻(frontend) Split settings file

rlecellier:
All settings may be changeable in development without interfering with
production settings.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -103,4 +103,4 @@ docker/files/etc/nginx/ssl/*
 storybook-static
 
 # Frontend dev settings
-src/frontend/js/settings.dev.ts
+src/frontend/js/settings/settings.dev.ts
--- a/src/frontend/js/settings/index.ts
+++ b/src/frontend/js/settings/index.ts
@@ -1,0 +1,38 @@
+import { mergeWith } from 'lodash-es';
+import * as prodSettings from './settings.prod';
+import * as testSettings from './settings.test';
+
+let settingsOverride = {};
+if (process.env.NODE_ENV === 'development') {
+  try {
+    settingsOverride = require('./settings.local.ts');
+  } catch {
+    // no local settings found, do nothing
+  }
+} else if (process.env.NODE_ENV === 'test') {
+  settingsOverride = testSettings;
+}
+
+try {
+  settingsOverride = require('./settings.local.ts');
+} catch {
+  // no local settings found, do nothing
+}
+
+const settings = mergeWith({}, prodSettings, settingsOverride);
+
+export const {
+  API_LIST_DEFAULT_PARAMS,
+  EDX_CSRF_TOKEN_COOKIE_NAME,
+  RICHIE_USER_TOKEN,
+  RICHIE_LTI_ANONYMOUS_USER_ID_CACHE_KEY,
+  JOANIE_API_VERSION,
+  REACT_QUERY_SETTINGS,
+  PAYMENT_SETTINGS,
+  CONTRACT_SETTINGS,
+  CONTRACT_DOWNLOAD_SETTINGS,
+  PER_PAGE,
+  MOCK_SERVICE_WORKER_ENABLED,
+  DEBUG_UNION_RESOURCES_HOOK,
+  CURRENT_JOANIE_DEV_DEMO_USER,
+} = settings;
--- a/src/frontend/js/settings/settings.dev.dist.ts
+++ b/src/frontend/js/settings/settings.dev.dist.ts
@@ -1,0 +1,25 @@
+// This configuration file can be overridden with settings.dev.ts file.
+// `$ cp settings.local.dist.ts settings.local.ts`, then change what's needed
+// in local environment in settings.local.ts.
+import { DevDemoUser } from 'api/lms/dummy';
+
+// disable react query cache
+// export const REACT_QUERY_SETTINGS = {
+//   staleTimes: {
+//     session: 0,
+//     sessionItems: 0,
+//   },
+// };
+
+/**
+ * Available users:
+ *  * admin
+ *  * user0
+ *  * user1
+ *  * user2
+ *  * user3
+ *  * user4
+ *  * organization_owner
+ *  * student_user
+ */
+export const CURRENT_JOANIE_DEV_DEMO_USER: DevDemoUser = 'admin';
--- a/src/frontend/js/settings/settings.prod.ts
+++ b/src/frontend/js/settings/settings.prod.ts
@@ -1,5 +1,3 @@
-import * as defaultDevSettings from './settings.dev.dist';
-
 export const API_LIST_DEFAULT_PARAMS = {
   limit: '21',
   offset: '0',
@@ -61,15 +59,5 @@ export const PER_PAGE = {
   useOrdersEnrollments: DEFAULT_PER_PAGE,
 };
 
-let devSettings;
-try {
-  devSettings = require('settings.dev.ts');
-} catch {
-  devSettings = defaultDevSettings;
-}
-
 export const MOCK_SERVICE_WORKER_ENABLED = false;
-
-export const DEBUG_UNION_RESOURCES_HOOK = process.env.NODE_ENV !== 'production' && false;
-
-export const { CURRENT_JOANIE_DEV_DEMO_USER } = devSettings;
+export const DEBUG_UNION_RESOURCES_HOOK = false;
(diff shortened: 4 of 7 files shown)


src/frontend/js/settings.ts
```diff
@@ -1,75 +1,10 @@
...
-  devSettings = defaultDevSettings;
+  localSettings = false;
 }
 
-export const MOCK_SERVICE_WORKER_ENABLED = false;
-
-export const DEBUG_UNION_RESOURCES_HOOK = process.env.NODE_ENV !== 'production' && false;
-
-export const { CURRENT_JOANIE_DEV_DEMO_USER } = devSettings;
+export default (process.env.NODE_ENV !== 'production' && localSettings) || prodSettings;
```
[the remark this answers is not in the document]
jbpenrath: Yup and in the dev we should be able to declare only keys that we want override.

src/frontend/js/settings.ts
```diff
@@ -1,75 +0,0 @@
...
-  devSettings = require('settings.dev.ts');
-} catch {
-  devSettings = defaultDevSettings;
-}
-
-export const MOCK_SERVICE_WORKER_ENABLED = false;
-
-export const DEBUG_UNION_RESOURCES_HOOK = process.env.NODE_ENV !== 'production' && false;
-
-export const { CURRENT_JOANIE_DEV_DEMO_USER } = devSettings;
```
rlecellier, line 75: Should i put back this demo user in our new prod settings or should i remove the test about dummy's demo user ?

src/frontend/js/settings/index.ts
```diff
@@ -0,0 +1,33 @@
+import { isArray, mergeWith } from 'lodash-es';
+import prodSettings from './settings.prod';
+
+let localSettings;
+try {
+  localSettings = require('./settings.local.ts');
```
jbpenrath, line 6: I liked the term `dev` over `local`

src/frontend/js/settings.ts
jbpenrath: Does that mean, we should create a dedicated settings file for test ?

src/frontend/js/settings/index.ts
```diff
@@ -0,0 +1,33 @@
...
+  localSettings = {};
+}
+
+const settings = !['production', 'test'].includes(process.env.NODE_ENV || 'prod')
+  ? mergeWith({}, prodSettings, localSettings.default, (_objValue, sourceObj) => {
+      if (isArray(sourceObj)) {
+        return sourceObj;
+      }
+    })
+  : prodSettings;
```
jbpenrath, line 17: Why do you need to use a `customizer` to check if the sourceObj is an array ?
For our case, I think the `merge` method does the job nope ?

Furthermore, why not choose `prodSettings` has `source` ?

src/frontend/js/settings/settings.prod.ts
```diff
@@ -0,0 +1,61 @@
...
+    useUnionResources: 50,
+    useCourseProductUnion: 50,
+    useOrdersEnrollments: 50,
+  },
+  MOCK_SERVICE_WORKER_ENABLED: false,
+  DEBUG_UNION_RESOURCES_HOOK: false,
+  CURRENT_JOANIE_DEV_DEMO_USER: false,
+};
+
+export default settings;
```
jbpenrath, line 61: I'll export all settings through const like before.
In this way, no matter the file a dev will open it always see what `settings.ts` is exporting.

src/frontend/js/settings/index.ts
rlecellier: I choose local because i thought that it could be used in some production instances, if we get some spécific configuration per instance (like scss theme  ?)

src/frontend/js/settings.ts
rlecellier: We may, but i think that's overkill.
We want the test to be :green_circle: with production settings.

src/frontend/js/settings/index.ts
rlecellier: `customizer` allow use to replace arrays instead of merge them.
With:
prod: `MY_ARRAY_SETTING: ['a', 'b']`
and local: `MY_ARRAY_SETTING: ['c']`
We want the final result to be `MY_ARRAY_SETTING: ['c']` and not `MY_ARRAY_SETTING: ['a', 'b', 'c']`

The first obj of lodash merge function will be modify like if we used `assign`. I feel more comfortable not to change `prodSettings` and `localSettings` objects

src/frontend/js/settings.ts
rlecellier: I've added a test settings after all.
We may have more testing tooling to test with tweaked settings.

src/frontend/js/settings/index.ts
jbpenrath: Not quit sure to understand. Furthermore, we are merging this settings only when we are not in a production environment.

src/frontend/js/settings/index.ts
```diff
@@ -0,0 +1,40 @@
...
+
+let settingsOverride = {};
+if (process.env.NODE_ENV === 'development') {
+  let localSettings = {};
+  try {
+    localSettings = require('./settings.local.ts');
+  } catch {
+    // no local settings found, do nothing
+  }
+  settingsOverride = localSettings;
```
jbpenrath, line 13: This looks work:

```suggestion
  try {
    settingsOverride = require('./settings.local.ts');
  } catch {
    // no local settings found, do nothing
  }
```

src/frontend/js/settings/index.ts
rlecellier: i'll rename.
But it feel like these settings are local, it can be for dev usage or for deployment usage.

src/frontend/js/settings/settings.test.ts
```diff
@@ -0,0 +1,16 @@
...
+ *  * admin
+ *  * user0
+ *  * user1
+ *  * user2
+ *  * user3
+ *  * user4
+ *  * organization_owner
+ *  * student_user
+ */
+export const CURRENT_JOANIE_DEV_DEMO_USER = 'admin' as DevDemoUser;
```
jbpenrath, line 16:
```suggestion
export const CURRENT_JOANIE_DEV_DEMO_USER: DevDemoUser = 'admin';
```

src/frontend/js/settings/settings.prod.ts
```diff
@@ -61,15 +62,7 @@ export const PER_PAGE = {
...
-  devSettings = defaultDevSettings;
-}
-
 export const MOCK_SERVICE_WORKER_ENABLED = false;
+export const DEBUG_UNION_RESOURCES_HOOK = false;
 
-export const DEBUG_UNION_RESOURCES_HOOK = process.env.NODE_ENV !== 'production' && false;
-
-export const { CURRENT_JOANIE_DEV_DEMO_USER } = devSettings;
+export const CURRENT_JOANIE_DEV_DEMO_USER: Nullable<DevDemoUser> = null;
```
jbpenrath, line 68: I don't think that variables only for debug or dev purposes should be declared here.

.gitignore
```diff
@@ -103,4 +103,4 @@ docker/files/etc/nginx/ssl/*
 storybook-static
 
 # Frontend dev settings
-src/frontend/js/settings.dev.ts
+src/frontend/js/settings/settings.local.ts
```
jbpenrath, line 106: Can be reverted

src/frontend/js/settings/settings.local.dist.ts
```diff
@@ -0,0 +1,25 @@
+// This confiuration file can be override with settings.local.ts file.
```
jbpenrath, line 1:
```suggestion
// This configuration file can be overridden with settings.dev.ts file.
```

src/frontend/js/settings/settings.local.dist.ts
jbpenrath, line 1: According to our discussion, this file should be named `settings.dev.dist.ts`

src/frontend/js/settings/settings.dev.dist.ts
```diff
@@ -0,0 +1,25 @@
...
+// in local environment in settings.local.ts.
+import { DevDemoUser } from 'api/lms/dummy';
+
+export const REACT_QUERY_SETTINGS = {
+  // disable react query cache
+  staleTimes: {
+    session: 0,
+    sessionItems: 0,
+  },
+};
```
jbpenrath, line 12: Not a good idea I think to disable react query cache by default. At least session.

jbpenrath (approved):

PR #2285, opened 2024-02-09
SaleTunnel various cache issues

NathanVss:
🐛(fix) SaleTunnel several cache issues

We encountered several issues in specific use cases with the SaleTunnel:

- Open and close the payment interface, when trying to open again the
frontend tries to create a new order, but there is already an existing
order: we should simply re-submit it. 
Fixed by: invalidate cache after aborting.

- Open and close payment interface and SaleTunnel, the PurchaseButton
is not displayed: it should because the order is pending.
Fixed by: invalidate cache after aborting.

- If we open a new tab and there is already an existing pending order,
the last step of the SaleTunnel (resume) crashes because the order
variable it uses is not defined.
Fixed by: Bring up useProductOrder to SaleTunnel in order to make the
existing order available accross the context.

### DIFF
--- a/src/frontend/js/api/joanie.ts
+++ b/src/frontend/js/api/joanie.ts
@@ -306,7 +306,7 @@ const API = (): Joanie.API => {
       },
       orders: {
         abort: async ({ id, payment_id }) => {
-          fetchWithJWT(ROUTES.user.orders.abort.replace(':id', id), {
+          return fetchWithJWT(ROUTES.user.orders.abort.replace(':id', id), {
             method: 'POST',
             body: payment_id ? JSON.stringify({ payment_id }) : undefined,
           }).then(checkStatus);
--- a/src/frontend/js/components/PaymentButton/index.spec.tsx
+++ b/src/frontend/js/components/PaymentButton/index.spec.tsx
@@ -1,6 +1,6 @@
 import { act, cleanup, fireEvent, render, screen, waitFor } from '@testing-library/react';
 import fetchMock from 'fetch-mock';
-import { PropsWithChildren, useMemo, useState } from 'react';
+import { PropsWithChildren, useMemo } from 'react';
 import { IntlProvider } from 'react-intl';
 import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
 import { faker } from '@faker-js/faker';
@@ -25,7 +25,6 @@ import {
   OrderCredentialCreationPayload,
   OrderState,
   ProductType,
-  Order,
   OrderGroup,
   CertificateProduct,
   CredentialProduct,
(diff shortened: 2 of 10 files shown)


src/frontend/js/api/joanie.ts
```diff
@@ -297,12 +297,11 @@ const API = (): Joanie.API => {
...
       },
       orders: {
-        abort: async ({ id, payment_id }) => {
+        abort: async ({ id, payment_id }) =>
           fetchWithJWT(ROUTES.user.orders.abort.replace(':id', id), {
             method: 'POST',
             body: payment_id ? JSON.stringify({ payment_id }) : undefined,
-          }).then(checkStatus);
-        },
+          }).then(checkStatus),
```
jbpenrath, line 304: What do you think to use an explicit return statement ?

```suggestion
        abort: async ({ id, payment_id }) => {
          return fetchWithJWT(ROUTES.user.orders.abort.replace(':id', id), {
            method: 'POST',
            body: payment_id ? JSON.stringify({ payment_id }) : undefined,
          }).then(checkStatus);
        },
```

src/frontend/js/components/PaymentButton/index.spec.tsx
```diff
@@ -62,6 +62,14 @@ jest.mock('utils/context', () => ({
...
 jest.mock('./components/PaymentInterfaces');
 
+type WrapperProps = PropsWithChildren<{
+  // hallucination:
+  // eslint-disable-next-line react/no-unused-prop-types
+  client?: QueryClient;
+  product: CredentialProduct | CertificateProduct;
+  orderGroup?: OrderGroup;
+}>;
+
```
jbpenrath, line 72: This is not an hallucination.
The `SaleTunnelWrapper` never uses the `client` prop. In order to fix that you should create two types.

```ts
type SaleTunnelWrapperProps = PropsWithChildren<{
  product: CredentialProduct | CertificateProduct;
  orderGroup?: OrderGroup;
}>;
type WrapperProps = PropsWithChildren<
  SaleTunnelWrapperProps & {
    client?: QueryClient;
  }
>;
```

src/frontend/js/widgets/SyllabusCourseRunsList/components/CourseProductItem/index.tsx
```diff
@@ -171,6 +171,7 @@ const CourseProductItem = ({ productId, course, compact = false }: CourseProduct
   });
 
   const order = productOrder as CredentialOrder;
+  // TODO: Rename this, how having no order is a pending state? It's hard to understand imo.
   const isPendingState = !order || order.state === OrderState.PENDING;
```
jbpenrath, line 175: From developer point of view, we could rename it : `isPurchasingState`. WDYT ?

jbpenrath (approved):
Nice work ! Some nitpicks otherwise LGTM ✅

src/frontend/js/components/PaymentButton/index.spec.tsx
NathanVss: I used `Exclude`

PR #2286, opened 2024-02-09
✨(frontend) require terms consent on all purchase

jbpenrath:
## Purpose

Previously, only product having a contract definition link required to accept terms before pay. Since terms and conditions are available from a dedicated page
 on our catalog, it seems relevant to require this consent on any kind of
 product.

Related work : https://github.com/openfun/joanie/pull/617

## Proposal

- [x] Require terms consent on any kind of product
- [x] Upload order creation payload to send the consent state

### DIFF
--- a/src/frontend/js/components/PaymentButton/index.tsx
+++ b/src/frontend/js/components/PaymentButton/index.tsx
@@ -114,12 +114,7 @@ const PaymentButton = ({ billingAddress, creditCard, onSuccess }: PaymentButtonP
   });
 
   const isReadyToPay = useMemo(() => {
-    return (
-      (course || enrollment) &&
-      product &&
-      billingAddress &&
-      (termsAccepted || !product.contract_definition)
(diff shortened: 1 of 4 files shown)


src/frontend/js/components/PaymentButton/index.tsx
```diff
@@ -120,12 +120,7 @@ const PaymentButton = ({ billingAddress, creditCard, onSuccess }: PaymentButtonP
   });
 
   const isReadyToPay = useMemo(() => {
-    return (
-      (course || enrollment) &&
-      product &&
-      billingAddress &&
-      (termsAccepted || !product.contract_definition)
-    );
+    return (course || enrollment) && product && billingAddress && termsAccepted === true;
```
NathanVss, line 123: is `=== true` really required?

NathanVss (approved):
📝 🍀

PR #2287, opened 2024-02-09
🩹(frontend) prevent to clear address select field

jbpenrath:
## Purpose

In the sale tunnel, currently the select is clearable but it does not unselect the address so the UX is weird. In order to improve that, we disable the clear feature of the select.

https://github.com/openfun/richie/assets/9265241/5306b81c-ccce-4929-abe9-8cbfa9e6c59d

## Proposal

- [x] Disable select clear feature


NathanVss (approved):
👌

PR #2296, opened 2024-02-21
🧑‍💻(frontend) jest/setup: create modal div

rlecellier:
Our modal needs a div with a specific id to be render. This test configuration is better now in a global configuration than repeted everywhere.


jbpenrath (approved):
LGTM ✅

PR #2297, opened 2024-02-21
✨(frontend) auto color for ListAvatar background

rlecellier:
## Purpose

This idea of feature is introduce to use default avatar as real avatar, in order to have a easy visual way to differentiate user line with they avatar.
There for we've here a way to generate colors from a string.

Learner list:
image

Stories:
image)


jbpenrath:
I think this could be a good feature but from my point of view, we should clamp the hue to keep consistent color palette and prevent a rainbow effect that IMO is weird.

rlecellier:
I should make a storybook story to let anyone give it's point of view on it.
I like lot's of hue, you don't and we must found the right choice who's mostly subjective but may be some can have wining arguments.

jbpenrath:
As discussed, let's use the full range of hue but we have to find the right saturation and luminosity.

PR #2305, opened 2024-02-27
🧪(frontend) fix credit card flacky test

NathanVss:
The past date generated was sometimes not within the previous month, which was making the test to fail.

### DIFF
--- a/src/frontend/js/pages/DashboardCreditCardsManagement/index.spec.tsx
+++ b/src/frontend/js/pages/DashboardCreditCardsManagement/index.spec.tsx
@@ -68,7 +68,10 @@ describe('<DashboardCreditCardsManagement/>', () => {
   });
 
   it('renders the correct label for expired date', async () => {
-    const date = faker.date.past();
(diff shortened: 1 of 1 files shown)


src/frontend/js/pages/DashboardCreditCardsManagement/index.spec.tsx
```diff
@@ -68,7 +68,10 @@ describe('<DashboardCreditCardsManagement/>', () => {
   });
 
   it('renders the correct label for expired date', async () => {
-    const date = faker.date.past();
+    const now = new Date();
+    const date = faker.date.past({
+      refDate: new Date(now.getFullYear(), now.getMonth(), 0, 23, 59, 59),
```
jbpenrath, line 73:
```suggestion
      refDate: new Date(now.getFullYear(), now.getMonth(), 0, 0, 0, 0),
```

Otherwise, you can still get a date in the current month.

jbpenrath (approved):

PR #2306, opened 2024-02-28
💄(frontend) teacher dashboard menu item

rlecellier:
We decide to put teacher dashboard user menu link as first link. We also change it design a tiny bit to make it more visible than other menu elements.

image


jbpenrath (approved):
Could take opportunity of this work to also rename this entry.
```diff
-Teacher Dashboard
+Course administration
```

jbpenrath (approved):
LGTM ✅

PR #2307, opened 2024-02-29
🚑 (frontend) invalidate order queries on submit call

jbpenrath:
## Purpose

Currently, when we submit an order, related queries are not invalidated so
the query client is not update and could not know the order. But in the last
step of the SaleTunnel we need to get the order id but sometimes order
is not defined so an error is raised...


NathanVss (approved):
LGTM 🕵️‍♂️

PR #2308, opened 2024-03-01
🐛(frontend) useUnionResource refetch on filter change

rlecellier:
`useUnionResource` should fetch new data when queries's filters change.

### DIFF
--- a/src/frontend/js/hooks/useUnionResource/index.ts
+++ b/src/frontend/js/hooks/useUnionResource/index.ts
@@ -90,6 +90,8 @@ const useUnionResource = <
   const eofRef = useRef<Record<string, number>>(queryClient.getQueryData(eofQueryKey) ?? {});
   log('eof', eofRef.current);
 
+  const [unionQueryKey, setUnionQueryKey] = useState<string>();
+
   const reset = () => {
     setStack([]);
     setPage(0);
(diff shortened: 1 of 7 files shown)


src/frontend/js/utils/test/wrappers/BaseJoanieAppWrapper.tsx
```diff
@@ -0,0 +1,20 @@
+import { PropsWithChildren } from 'react';
+
+import JoanieSessionProvider from 'contexts/SessionContext/JoanieSessionProvider';
+import { IntlWrapper } from './IntlWrapper';
+import { ReactQueryWrapper } from './ReactQueryWrapper';
+import { AppWrapperProps } from './types';
+
+export const BaseJoanieAppWrapper = ({
```
jbpenrath, line 8: Let's call it `JoanieAppWrapper` WDYT ?

In this way we will be homogeneous over the code base. 
About session provider we have two versions : 
- `BaseSessionProvider`
- `JoanieSessionProvider`

src/frontend/js/hooks/useUnionResource/index.ts
```diff
@@ -90,6 +90,8 @@ const useUnionResource = <
   const eofRef = useRef<Record<string, number>>(queryClient.getQueryData(eofQueryKey) ?? {});
   log('eof', eofRef.current);
 
+  const [unionQueryKey, setUnionQueryKey] = useState<string>();
```
jbpenrath, line 93: I'm not quit sure you need to use a state here.

Instead you could use a useMemo.

jbpenrath, line 93: You should populate the `useState` with the current filters to prevent to trigger a reset on mount.

jbpenrath (approved):

PR #2304, opened 2024-02-26
✨(frontend) teacher courses search with query

rlecellier:
note: this branch is rebased on https://github.com/openfun/richie/pull/2308

## Purpose

Add search field on teacher courses listing pages

image

[Capture vidéo du 28-02-2024 15:47:50.webm](https://github.com/openfun/richie/assets/139848/99129b5e-5ccc-4429-8207-630c56a3c552)

PR #2309, opened 2024-03-01
✨(frontend) learner courses search with query

rlecellier:
Note: rebased on ✨[(frontend) teacher courses search with query](https://github.com/openfun/richie/pull/2304)

## Purpose

on learner courses listing page, we can now perform textual research.

image

update: 
Capture d’écran du 2024-03-12 14-50-49

### DIFF
--- a/src/frontend/js/hooks/useLearnerCoursesSearch/index.tsx
+++ b/src/frontend/js/hooks/useLearnerCoursesSearch/index.tsx
@@ -1,0 +1,77 @@
+import { useEffect, useState } from 'react';
+import { useSearchParams } from 'react-router-dom';
+import { Enrollment, CredentialOrder, OrderState, ProductType } from 'types/Joanie';
+import { Maybe, Nullable } from 'types/utils';
+import { useOrdersEnrollments } from 'pages/DashboardCourses/useOrdersEnrollments';
+
+const useLearnerCoursesSearch = () => {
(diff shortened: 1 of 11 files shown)


src/frontend/js/hooks/useLearnerCoursesSearch/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+    setSearchParams(searchParams);
+    if (!newQuery) {
+      setCount(undefined);
+    }
+  };
+
+  return {
+    submitSearch,
+    data: orderAndEnrollmentList,
```
NathanVss, line 68: why not using `data` directly?

rlecellier: `data` is the return or `useOrdersEnrollments` where `orderAndEnrollmentList` is an insternal state.
`orderAndEnrollmentList` isn't up to date each time data is, it's not a bug but a feature:
When `query` chagne,  `useOrdersEnrollments` queries are invalidated and `data` is temporary set to `0`.
There we don't want the list to disepear but to fade until data get a new value. That the usage of the internal state.

NathanVss (approved):
Nice work 👏

Just small comment, I find the gap between the search and the alert too big, it feels odd to me:

Capture d’écran 2024-03-19 à 17 15 56

PR #2315, opened 2024-03-04
🐛(frontend) useUnionResource refetch on invalidate

rlecellier:
rebased on https://github.com/openfun/richie/pull/2308

## Purpose

fix a bug where when stack.length === 0, invalidate the query will not refetch.

### DIFF
--- a/src/frontend/js/hooks/useUnionResource/index.ts
+++ b/src/frontend/js/hooks/useUnionResource/index.ts
@@ -90,8 +90,10 @@ const useUnionResource = <
   const eofRef = useRef<Record<string, number>>(queryClient.getQueryData(eofQueryKey) ?? {});
   log('eof', eofRef.current);
 
-  const [unionQueryKey, setUnionQueryKey] = useState<string>();
+  const [update, setForceUpdate] = useState<number>();
 
+  // to force execution of useEffect::fetchNewPage(),
+  // reset need to generate a uniq key that is part of it's dependencies.
   const reset = () => {
     setStack([]);
     setPage(0);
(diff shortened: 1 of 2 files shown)


src/frontend/js/hooks/useUnionResource/index.ts
```diff
@@ -90,9 +91,15 @@ const useUnionResource = <
   const eofRef = useRef<Record<string, number>>(queryClient.getQueryData(eofQueryKey) ?? {});
   log('eof', eofRef.current);
 
-  const [unionQueryKey, setUnionQueryKey] = useState<string>();
+  const [unionQueryKey, setUnionQueryKey] = useState<Nullable<string>>();
+  const handleSetUnionQueryKey = (filterA: FiltersA, filterB: FiltersB) => {
+    setUnionQueryKey(
+      [JSON.stringify(filterA), JSON.stringify(filterB), new Date().getTime()].join('-'),
+    );
```
jbpenrath, line 98: Are you adding date into the query key to invalidate the previous query each time you are calling this method ?

If yes, you should invalidate the query key instead I think.

rlecellier: It's not a react-query key.
It's juste an inner home made key in this component that trigger new fetch when it change.

jbpenrath: Yup but I do not still understand why you need bind datetime in this query key ?

What is the case where this method could be called but filters did not change ?

rlecellier: We need the `useEffect` that containe `fetchNewPage` to be executed in order to update `useUnionResource` results.
To achive that we need to handle two corder cases that weren't done at start.
* When filters change, we update an internal key that is part of this `useEffect` dependencies.
* When a query is invalidate, we call the `reset` function that will now also update the internal key with a uniqid (timestamp).

note: previously reset rely on setting `stack` at `[]`, but this isn't enought as an invalidated query can be empty and there for stack chagne from `[]` to `[]` and the `useEffet` is not executed.

src/frontend/js/hooks/useUnionResource/index.ts
```diff
@@ -90,32 +90,32 @@ const useUnionResource = <
...
+  // to force execution of useEffect::fetchNewPage(),
+  // reset need to generate a uniq key that is part of it's dependencies.
   const reset = () => {
     setStack([]);
     setPage(0);
     setTotalCount(undefined);
     setError(undefined);
     setCursor(perPage);
     setIntegrityCount(0);
+    setForceUpdate(new Date().getTime());
```
jbpenrath, line 104: Just shorter
```suggestion
    setForceUpdate(Date.now());
```

jbpenrath (approved):
LGTM ✅

PR #2321, opened 2024-03-05
🐛(frontend) fix useUnionResource loadMore

### LINKED ISSUE #2317: "Load more" remains displayed at the bottom of the list on teacher dashboard
## Bug Report

**Problematic Behavior**

"Load more" remains displayed at the bottom of the list at `http://localhost:8070/en/dashboard/teacher/courses?query=Course`

Capture d’écran 2024-03-05 à 10 48 05

**Expected behavior/code**

It should disappear once the list is fully loaded

**Steps to Reproduce**
1. Go to `http://localhost:8070/en/dashboard/teacher/courses?query=Course`

rlecellier:
Because our Eof wasn't reset correctly, useUnionResource would never fetch the seconds pages and always return loadMore flag as true.

resolve #2317

### DIFF
--- a/src/frontend/js/hooks/useUnionResource/index.ts
+++ b/src/frontend/js/hooks/useUnionResource/index.ts
@@ -94,7 +94,11 @@ const useUnionResource = <
 
   // to force execution of useEffect::fetchNewPage(),
   // reset need to generate a uniq key that is part of it's dependencies.
-  const reset = () => {
+  const reset = (eofKey?: string) => {
+    if (eofKey) {
+      delete eofRef.current[eofKey];
+    }
+
     setStack([]);
     setPage(0);
     setTotalCount(undefined);
@@ -107,8 +111,12 @@ const useUnionResource = <
   // we manualy observe key invalidation to trigger new search
   // by generating a new update key
   if (refetchOnInvalidation) {
-    useQueryKeyInvalidateListener(queryAConfig.queryKey, reset);
-    useQueryKeyInvalidateListener(queryBConfig.queryKey, reset);
+    useQueryKeyInvalidateListener(queryAConfig.queryKey, () => {
+      reset(queryAConfig.queryKey.join('-'));
+    });
+    useQueryKeyInvalidateListener(queryBConfig.queryKey, () => {
+      reset(queryBConfig.queryKey.join('-'));
+    });
   }
 
   // filters have changes, new results will be fetch.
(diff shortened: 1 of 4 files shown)


NathanVss (approved):
Neat + adding a test would be nice

src/frontend/js/hooks/useUnionResource/utils/fetchEntity.ts
```diff
@@ -65,14 +65,11 @@ export const fetchEntity = async <
...
-    // If we reached the end of the list, we set the eof flag to prevent future requests.
-    if (!res.next) {
-      eofRef.current = { ...eofRef.current, [queryKeyString]: filters.page! };
-      // Eof is cached based, the same way, we cache the fetching data. Otherwise there would
-      // be request to non existing pages after reload.
-      queryClient.setQueryData(eofQueryKey, eofRef.current);
-    }
+    // Eof is cached based, the same way, we cache the fetching data. Otherwise there would
+    // be request to non existing pages after reload.
+    const totalPages = Math.ceil(res.count / perPage);
```
NathanVss, line 70: Can you confirm that this change around removing `if(!res.next)` is purely to make the code simpler or it does solve an issue?

src/frontend/js/hooks/useUnionResource/index.ts
```diff
@@ -94,7 +94,11 @@ const useUnionResource = <
 
   // to force execution of useEffect::fetchNewPage(),
   // reset need to generate a uniq key that is part of it's dependencies.
-  const reset = () => {
+  const reset = (eofKey?: string) => {
```
NathanVss, line 97: What I suggest is to not have a param passed to the `reset` function, to make sure it stay simple. And to reset both list each time an invalidation occur, WDYT? After all we can consider that if there is one invalidation we should reset the entire list as they nearly the same entity

NathanVss (approved):

src/frontend/js/hooks/useUnionResource/utils/fetchEntity.ts
rlecellier: I don't remember it fixing a bug. But i felt better having `eofRef.current` up to date all along on debuging and i saw no reason not to keep it like this at the end.

NathanVss: Allright

src/frontend/js/hooks/useUnionResource/index.ts
NathanVss: Ok deal if it works well

PR #2322, opened 2024-03-06
🐛(frontend) manage certificate linked to enrollment

jbpenrath:
## Purpose

Some certificate can be linked to an enrollment. Currently we do not manage
those cases so an error is displayed when we meet this use case.

## Proposal

- [x] Manage certificate linked to enrollment

### DIFF
--- a/src/frontend/js/types/Joanie.ts
+++ b/src/frontend/js/types/Joanie.ts
@@ -98,12 +98,20 @@ export interface CertificateDefinition {
   description: string;
 }
 
-export interface Certificate {
+export type Certificate = {
   id: string;
   issued_on: string;
   certificate_definition: CertificateDefinition;
-  order: NestedCertificateOrder | NestedCredentialOrder;
-}
+} & (
+  | {
+      order: NestedCertificateOrder | NestedCredentialOrder;
+      enrollment: null;
+    }
+  | {
+      enrollment: EnrollmentLight;
+      order: null;
+    }
+);
 
 // - Organization
 export interface OrganizationLight {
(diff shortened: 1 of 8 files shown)


src/frontend/js/types/Joanie.ts
```diff
@@ -102,7 +102,8 @@ export interface Certificate {
   id: string;
   issued_on: string;
   certificate_definition: CertificateDefinition;
-  order: NestedCertificateOrder | NestedCredentialOrder;
+  order: Nullable<NestedCertificateOrder | NestedCredentialOrder>;
```
NathanVss, line 105: WDYT about trying to make an exclusion type like "if `order` is defined then `enrollment` is not" and vice-versa? I'm not saying that's easy though 😅

src/frontend/js/widgets/Dashboard/components/DashboardItem/Certificate/index.tsx
```diff
@@ -20,40 +21,43 @@ export const DashboardItemCertificate = ({
   productType,
   mode,
 }: DashboardItemCertificateProps) => {
-  if (certificate) {
-    if (certificateDefinition) {
-      throw new Error('certificate and certificateDefinition are mutually exclusive');
+  const getCourseFromCertificate = () => {
```
NathanVss, line 24: In the future we will maybe have to make this function available to other components, WDYT about make a standalone function? And then writing a unit test forit?

NathanVss (approved):
📜

src/frontend/js/widgets/Dashboard/components/DashboardItem/Certificate/index.tsx
jbpenrath: Ok, I created a `CertificateHelper` implementing a `getCourse` method. Let's take a look to the dedicated commit.

src/frontend/js/types/Joanie.ts
jbpenrath: Please review the fixup commit to check if the proposed solution is fine for you.

PR #2323, opened 2024-03-06
🚚(frontend) api/joanie as joanieLegacyClient

rlecellier:
We'll implement a api client generated from joanie's openApi schema. The current client needs to change place.


rlecellier:
impossible to rebase.

PR #2339, opened 2024-03-22
🐛(frontend) filter learner list by organization

rlecellier:
Users mustn't see learner for organization they don't have access to. This will be handle by joanie's backend but for now, we need to add a filter on all available organization id for the logged user

FIXME:
Joanie don't have needed filter on organization_ids ([see filters on joanie](https://github.com/openfun/joanie/blob/main/src/backend/joanie/core/filters/client.py#L197-L229))

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,8 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Changed
 
+- Learner list in teacher dashboard are always fitered by
+  organization id.
 - Upgrade docker compose to v2.24.5
 - remove dashboard i18n routing
 - Complete `course_detail` RDFa markups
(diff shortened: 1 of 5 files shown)


jbpenrath (commented):
You do not manage the case where the user is not a member of an organization.

CHANGELOG.md
```diff
@@ -32,7 +32,9 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Changed
 
+- Learner list in teacher dashboard are always fitered by
 - Upgrade docker compose to v2.24.5
+  organization id.
```
jbpenrath, line 37:
```suggestion
- Learner list in teacher dashboard are always filtered by
  organization id.
- Upgrade docker compose to v2.24.5
```

src/frontend/js/pages/TeacherDashboardCourseLearnersLayout/hooks/useCourseLearnersFilters/index.ts
```diff
@@ -24,13 +25,16 @@ const useCourseLearnersFilters = () => {
     };
   }, Array.from(searchParams.entries()));
 
+  // default orgnizationId between (ordered by priority): route, query, first user's organization.
```
jbpenrath, line 28:
```suggestion
  // default organizationId between (ordered by priority): route, query, first user's organization.
```

jbpenrath (approved):

PR #2340, opened 2024-03-22
♻️(frontend) remove i18n routing

NathanVss:
At first we though it was a good idea but in pratice we realized it was causing too much pain to maintain. The ultime issue was that if the URL were not 100% translated for a given language then the dashboard was crashing due to the fact that we cannot set child route that has a different starting portion than its parent. And this was happening in the following example:

Parent: /teacher/cours/:courseId/produits/:courseProductRelationId
Child:  /teacher/courses/:courseId/products/:courseProductRelationId/learners

Resulting error: An absolute child route path must start with the combined path of all its parent routes.


src/frontend/js/widgets/Dashboard/utils/teacherRouteMessages.tsx
```diff
@@ -6,109 +6,22 @@ import { defineMessages } from 'react-intl';
...
+  ORGANIZATION_COURSE_CONTRACTS = '/teacher/organizations/:organizationId/courses/:courseId/contracts',
+  ORGANIZATION_PRODUCT_CONTRACTS = '/teacher/organizations/:organizationId/courses/:courseId/products/:courseProductRelationId/contracts',
+  ORGANIZATION_COURSE_PRODUCT_LEARNER_LIST = '/teacher/organizations/:organizationId/courses/:courseId/products/:courseProductRelationId/learners',
+  ORGANIZATION_COURSE_GENERAL_INFORMATION = '/teacher/organizations/:organizationId/courses/:courseId/information',
+  COURSE = '/teacher/courses/:courseId',
+  COURSE_GENERAL_INFORMATION = '/teacher/courses/:courseId/information',
+  COURSE_PRODUCT = '/teacher/courses/:courseId/products/:courseProductRelationId',
+  COURSE_PRODUCT_LEARNER_LIST = '/teacher/courses/:courseId/products/:courseProductRelationId/learners',
+  COURSE_CONTRACTS = '/teacher/courses/:courseId/contracts',
+  COURSE_PRODUCT_CONTRACTS = '/teacher/courses/:courseId/products/:courseProductRelationId/contracts',
```
jbpenrath, line 22: Now, should we use template string to prevent to repeat paths and reduce the risk of typo ?

```js
  ORGANIZATION = '/teacher/organizations/:organizationId',
  ORGANIZATION_CONTRACTS = `${ORGANIZATION}/contracts`,
```

src/frontend/js/widgets/Dashboard/utils/teacherRoutes.tsx
```diff
@@ -32,35 +32,26 @@ export function getTeacherDashboardRoutes() {
...
+          path: getRoutePath(TeacherDashboardPaths.COURSE),
           children: [
             {
               index: true,
               element: (
                 <NavigateWithParams
-                  to={getRoutePath(TeacherDashboardPaths.COURSE_GENERAL_INFORMATION, {
-                    courseId: ':courseId',
-                  })}
+                  to={getRoutePath(TeacherDashboardPaths.COURSE_GENERAL_INFORMATION)}
```
jbpenrath, line 41: Not quit sure getRoutePath is useful anymore. Indeed, we could declare routes into an enum right now.

NathanVss: It is useful only if we need to replace variables in the URL

src/frontend/js/widgets/Dashboard/utils/teacherRouteMessages.tsx
NathanVss: Tbh I think having plain string gives a faster way to read the route full url, improving DX. But .. that's only my %, let's talk about it during the next daily?

lunika: I agree with @jbpenrath suggestion

src/frontend/js/widgets/Dashboard/utils/dashboardRoutes.tsx
```diff
@@ -86,8 +78,21 @@ const getDashboardRouteAttribute =
   ) => {
     const isLabel = attribute === RouteAttributes.LABEL;
 
-    const message = isLabel ? DASHBOARD_ROUTE_LABELS[path] : DASHBOARD_ROUTE_PATHS[path];
-    return intl.formatMessage(message, ...options);
+    if (isLabel) {
+      return intl.formatMessage(DASHBOARD_ROUTE_LABELS[path], ...options);
+    }
+
+    let outputPath = path as string;
```
rlecellier, line 85: I would just get ride of `getDashboardRouteAttribute` and have theses two functions: 
```ts
/** Get the provided dashboard route label in the active locale */
export const getDashboardRouteLabel =
  (intl: IntlShape) =>
  (
    path: DashboardPath,
    ...options: [
      values?: Record<string, PrimitiveType | FormatXMLElementFn<string, string>>,
      opts?: IntlMessageFormatOptions,
    ]
  ) => {
    return intl.formatMessage(DASHBOARD_ROUTE_LABELS[path], ...options);
  };

/** Get the provided dashboard route path in the active locale */
export const getDashboardRoutePath = (originalPath: string, params?: {} | undefined) => {
  return generatePath(originalPath, params);
};
```

We can do a bit more and remove `getDashboardRoutePath` by replacing usage of `getRoutePath` by `generatePath` everywhere.
```sh
$ git grep 'getRoutePath(' | wc -l
42
```
And also get ride of `getDashboardRouteLabel` and replace it with a classic `intl.formatMessage` everywhere.

Also, `TEACHER_DASHBOARD_ROUTE_LABELS` and `LEARNER_DASHBOARD_ROUTE_LABELS` may be renamed as `teacherRouteMessages` and `learnerRouteLabel`.
I would also move them in a dedicated`utils/routeMessages.ts`.
They're used to display route link's text and can now be separeted to the routing logic.

src/frontend/js/widgets/Dashboard/utils/teacherRoutes.tsx
rlecellier: we can use [react-router generatePath](https://reactrouter.com/en/main/utils/generate-path) to replace route's params in path.

src/frontend/js/widgets/Dashboard/utils/teacherRouteMessages.tsx
rlecellier: The downside of this proposal is that broken string are harder to grep. Also a grep of `COURSE_PRODUCT` to found all usage of this route will give false postive usage like 
```ts
const COURSE_PRODUCT_LEARNER_LIST = `${COURSE_PRODUCT}/learners`
```
The only advantage is that we avoid typo.

At first i like the idea but it seems not needed at the end.

CHANGELOG.md
```diff
@@ -33,6 +33,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 - Contract list in the teacher dashbaord are now filtered by 
   courseProductRelationId instead of courseId and productId.
 - Switch from setup.cfg to pyproject.toml 
+- remove dashboard i18n routing
```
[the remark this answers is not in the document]
NathanVss: you are right!

src/frontend/js/widgets/Dashboard/utils/dashboardRoutes.tsx
NathanVss: Interesting, I will implement this approach regarding the functions 👌

NathanVss:
@jbpenrath @rlecellier please do not hesitate to take few minutes to monkey test the navigation 🙏

src/frontend/js/components/CourseGlimpse/utils.ts
```diff
@@ -14,17 +14,16 @@ const getCourseGlimpsePropsFromCourseProductRelation = (
...
   organizationId?: string,
 ): CourseGlimpseCourse => {
-  const getRoutePath = getDashboardRoutePath(intl);
   const courseRouteParams = {
     courseId: courseProductRelation.course.id,
     courseProductRelationId: courseProductRelation.id,
   };
   const courseRoute = organizationId
-    ? getRoutePath(TeacherDashboardPaths.ORGANIZATION_PRODUCT, {
+    ? getDashboardRoutePath(TeacherDashboardPaths.ORGANIZATION_PRODUCT, {
```
rlecellier, line 22: Why not using `generatePath` directly ?

src/frontend/js/widgets/Dashboard/components/TeacherDashboardCourseSidebar/index.tsx
```diff
@@ -86,12 +85,11 @@ export const TeacherDashboardCourseSidebar = () => {
...
 
   const getMenuLinkFromPath = (basePath: TeacherDashboardPaths) => {
-    const path = getRoutePath(basePath, {
-      organizationId: routeOrganizationId,
-      courseId: routeCourseId,
-      courseProductRelationId: routeCourseProductRelationId,
+    const path = getDashboardRoutePath(basePath, {
+      organizationId: routeOrganizationId ?? '',
+      courseId: routeCourseId ?? '',
+      courseProductRelationId: routeCourseProductRelationId ?? '',
```
rlecellier, line 91: I tried to replace with :

```suggestion
      organizationId: routeOrganizationId,
      courseId: routeCourseId,
      courseProductRelationId: routeCourseProductRelationId,
```

and everything work's fine.

src/frontend/js/widgets/Dashboard/utils/dashboardRoutes.tsx
```diff
@@ -51,31 +44,15 @@ export function getDashboardRoutes() {
...
-  ...LEARNER_DASHBOARD_ROUTE_PATHS,
-  ...TEACHER_DASHBOARD_ROUTE_PATHS,
-};
+
 type DashboardPath = LearnerDashboardPaths | TeacherDashboardPaths;
 
-const getDashboardRouteAttribute =
-  (attribute: RouteAttributes) =>
+/** Get the provided dashboard route label in the active locale */
+export const getDashboardRouteLabel =
```
rlecellier, line 55: This function is just alias for `intl.formatMessage` and therefor not needed

src/frontend/js/widgets/Dashboard/utils/dashboardRoutes.tsx
```diff
@@ -84,14 +61,10 @@ const getDashboardRouteAttribute =
...
-    return intl.formatMessage(message, ...options);
+    return intl.formatMessage(DASHBOARD_ROUTE_LABELS[path], ...options);
   };
 
-/** Get the provided dashboard route label in the active locale */
-export const getDashboardRouteLabel = getDashboardRouteAttribute(RouteAttributes.LABEL);
-
 /** Get the provided dashboard route path in the active locale */
-export const getDashboardRoutePath = getDashboardRouteAttribute(RouteAttributes.PATH);
+export const getDashboardRoutePath = (originalPath: string, params: any = {}) => {
```
rlecellier, line 68: This function is just alias for `generatePath` and therefor not needed

src/frontend/js/widgets/Dashboard/utils/learnerRouteMessages.tsx
```diff
@@ -5,71 +5,17 @@ import { defineMessages } from 'react-intl';
...
-  },
-  [LearnerDashboardPaths.PREFERENCES_CREDIT_CARD_EDITION]: {
-    id: 'components.Dashboard.DashboardRoutes.preferences.creditCards.edition.path',
-    description: 'The path to display the credit cards edition view.',
-    defaultMessage: '/preferences/credit-cards/{creditCardId}',
-  },
-});
-
 // Translations of dashboard route labels
 export const LEARNER_DASHBOARD_ROUTE_LABELS = defineMessages<LearnerDashboardPaths>({
```
rlecellier, line 20: Proposal:  
Move `LearnerDashboardPaths` into `utils.learnerRoutes.tsx`
This way `learnerRoutes` will include routing logic and `learnerRouteMessages` will includes translations logic.

src/frontend/js/widgets/Dashboard/components/TeacherDashboardCourseSidebar/index.tsx
NathanVss: Did you run the tests?

src/frontend/js/components/CourseGlimpse/utils.ts
NathanVss: I preferred to keep a bit of encapsulation, what do you think @jbpenrath? I'll be democratic on this

src/frontend/js/widgets/Dashboard/utils/dashboardRoutes.tsx
NathanVss: Same, encapsulation purpose, you're vote @jbpenrath ?

src/frontend/js/widgets/Dashboard/utils/dashboardRoutes.tsx
NathanVss: old commit here

src/frontend/js/widgets/Dashboard/utils/learnerRouteMessages.tsx
NathanVss: Personaly I prefer having both data ( path + labels ) within the same file it avoid having to go back and forth between two files, democratos @jbpenrath?

I think it is good to have the opinions of everyone on these choices as they are related to architecture

src/frontend/js/components/CourseGlimpse/utils.ts
jbpenrath: As now, this method acts as a proxy, I'll use `generatePath` directly.

src/frontend/js/widgets/Dashboard/utils/dashboardRoutes.tsx
jbpenrath: In this case, I think I prefer the encapsulation as I think it's better for dx. This is simpler to understand to have to call `getDashboardRouteLabel` than to have to import `DASHBOARD_ROUTE_LABELS` objects then use into a `intl.formatMessage`

src/frontend/js/widgets/Dashboard/utils/learnerRouteMessages.tsx
jbpenrath: Personally I do not like to have to navigate between too many modules so IMO here this single module is fine, it gathers all learners routing logic.

src/frontend/js/widgets/Dashboard/components/TeacherDashboardCourseSidebar/index.tsx
rlecellier: i will, i had build-ts cmd to check if everything was fine.

NathanVss: This does not run tests

src/frontend/js/widgets/Dashboard/hooks/useDashboardRouter/index.tsx
```diff
@@ -38,7 +38,7 @@ export const useDashboardNavigate = () => {
         if (typeof to === 'number') {
           return navigate(to);
         }
-        return navigate(getDashboardRoutePath(to, values), routerOptions);
+        return navigate(generatePath(to, values as any), routerOptions);
```
jbpenrath, line 41: Maybe we could use the type expected by `generatePath` to type `values` parameter ?

NathanVss: Nope we can't here as the type is inferred statically from the string, but here the values differs as it is generic ^^'

jbpenrath: This seems to work 😉 
```
return useMemo(
    () =>
      <Path extends number | LearnerDashboardPaths | TeacherDashboardPaths>(
        to: Path,
        params?: Parameters<typeof generatePath<Exclude<Path, number>>>[1],
        routerOptions?: NavigateOptions,
      ) => {
        if (typeof to === 'number') {
          return navigate(to);
        }
        return navigate(generatePath(to, params), routerOptions);
      },
    [],
  );
```

jbpenrath (approved):

src/frontend/js/widgets/Dashboard/utils/learnerRoutesPaths.tsx
```diff
@@ -0,0 +1,70 @@
+import { defineMessages } from 'react-intl';
+
+/**
+ * All existing dashboard paths.
+ */
+export enum LearnerDashboardPaths {
```
rlecellier, line 6: You renamed `TeacherDashboardPaths` into `TeacherRoutesPaths` but keep `LearnerDashboardPaths` as is.
Choose one and use it for both ?

rlecellier (approved):
Nice, look's great to me !

src/frontend/js/widgets/Dashboard/utils/learnerRoutesPaths.tsx
NathanVss: Good catch! I choose to keep `TeacherDashboardPaths` because "path" and "route" are quite synonyms.

rlecellier: dashboard or routes doesn't matter, but it's nice to have the same naming both ways.

PR #2343, opened 2024-03-25
🧪(frontend) add useUnionResource loadMore test

rlecellier:
Our listing page always display the "loadMore" button where it should continue to request for results.


rlecellier:
includes in #2321

PR #2357, opened 2024-04-08
✨(back) management command to migrate course runs course link to joanie

lunika:
## Purpose

To use joanie as backend we have to migrate the course_link from edx to joanie. A management command is made for that making a bulk migration.

## Proposal

- [x] management command to migrate course runs course link to joanie

### DIFF
--- a/src/richie/apps/courses/management/commands/migrate_course_run_resource_link.py
+++ b/src/richie/apps/courses/management/commands/migrate_course_run_resource_link.py
@@ -1,0 +1,67 @@
+"""Command migrating richie course_runs course_link to joanie uri."""
+
+import logging
+from http import HTTPStatus
+
+from django.conf import settings
+from django.core.management.base import BaseCommand
+
+import requests
+from requests.exceptions import RequestException
+
+from richie.apps.courses.models import CourseRun
+
+logger = logging.getLogger("richie.commands.core.migrate_course_run_course_link")
+
+
+class Command(BaseCommand):
+    """
+    Migrate course run course link from edx to joanie uri.
+    """
+
+    help = __doc__
+
+    def handle(self, *args, **options):
+        """
+        Handle the command migrating existing course run with the course link return by joanie.
+        """
+        joanie_backend = settings.JOANIE_BACKEND
+        course_runs_to_update = []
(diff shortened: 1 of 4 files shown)


src/richie/apps/courses/management/commands/migrate_course_run_resource_link.py
```diff
@@ -0,0 +1,62 @@
+"""Command migratind richie course_runs course_link to joanie uri."""
```
kernicPanel, line 1:
```suggestion
"""Command migrating richie course_runs course_link to joanie uri."""
```

src/richie/apps/courses/management/commands/migrate_course_run_resource_link.py
```diff
@@ -0,0 +1,62 @@
...
+                    course_run.id,
+                    response.status_code,
+                    response.text,
+                )
+                continue
+
+            joanie_course_run = response.json()
+
+            course_run.resource_link = joanie_course_run["uri"]
+            course_run.save(update_fields=["resource_link"])
```
kernicPanel, line 62: Did you try it with a large number of course runs ?
Would it be more efficient with a bulk_update ?

src/richie/apps/courses/management/commands/migrate_course_run_resource_link.py
```diff
@@ -0,0 +1,62 @@
...
+    """
+
+    help = __doc__
+
+    def handle(self, *args, **options):
+        """
+        Handle the command migrating existing course run with the course link return by joanie.
+        """
+        joanie_backend = settings.JOANIE_BACKEND
+        for course_run in CourseRun.objects.all():
```
jbpenrath, line 29: Just for memory optimization, I'll use `iterator` instead of `all`.

```suggestion
        for course_run in CourseRun.objects.iterator():
```

jbpenrath (approved):
LGTM ✅

src/richie/apps/courses/management/commands/migrate_course_run_resource_link.py
lunika: You are right, I update the command using `bulk_update` but with a `batch_size` parameter.

PR #2359, opened 2024-04-08
✅ (frontend) fix false positive test

jbpenrath:
## Purpose

An await before an asynchrone assertion was missing. After adding this missing `await` it appears the assertion was failing.

## Proposal

- [x] Fix a false positive frontend test

### DIFF
--- a/src/frontend/js/components/PurchaseButton/index.spec.tsx
+++ b/src/frontend/js/components/PurchaseButton/index.spec.tsx
@@ -181,7 +181,7 @@ describe('PurchaseButton', () => {
     expect(button).toBeDisabled();
 
     // Further, a message is displayed to explain why the CTA is disabled
-    screen.findByText('This product is full, is currently not available for sale');
+    await screen.findByText('There are no more places available for this product.');
   });
 
   it.each([
@@ -318,7 +318,7 @@ describe('PurchaseButton', () => {
       // No alert message
       expect(
         screen.queryByText(
-          'The course run is not active, this product is not currently available for sale',
+          'The course run is not active. This product is not currently available for sale.',
         ),
       ).not.toBeInTheDocument();
     },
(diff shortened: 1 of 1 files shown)


src/frontend/js/components/PurchaseButton/index.spec.tsx
```diff
@@ -181,7 +181,7 @@ describe('PurchaseButton', () => {
     expect(button).toBeDisabled();
 
     // Further, a message is displayed to explain why the CTA is disabled
-    screen.findByText('This product is full, is currently not available for sale');
+    await screen.findByText('There are no more places available for this product.');
```
rlecellier, line 184: We've an other one in the same file: 
```js
$ git grep -A 3 '  screen.find'
src/frontend/js/components/PurchaseButton/index.spec.tsx:    screen.findByText('This product is full, is currently not available for sale');
src/frontend/js/components/PurchaseButton/index.spec.tsx-  });
src/frontend/js/components/PurchaseButton/index.spec.tsx-
src/frontend/js/components/PurchaseButton/index.spec.tsx-  it.each([
--
src/frontend/js/components/PurchaseButton/index.spec.tsx:    screen.findByText(
src/frontend/js/components/PurchaseButton/index.spec.tsx-      'At least one course has no course runs, this product is not currently available for sale',
src/frontend/js/components/PurchaseButton/index.spec.tsx-    );
src/frontend/js/components/PurchaseButton/index.spec.tsx-  });

```

rlecellier (approved):

NathanVss (approved):
LGTM :)

src/frontend/js/components/PurchaseButton/index.spec.tsx
jbpenrath: Good catch, this is also a false positive ..

PR #2360, opened 2024-04-08
💄(frontend) improve order group message color and layout

jbpenrath:
## Purpose

Currently, the order group message in CourseProductItem has no color set. That mean we are not able to customize it from our factory that is weird. We decide to use the base color of the CourseProductItem. Furthermore, the spacing around the message is pretty weird so we propose an improvement of margin.

| Before | After |
|--------|-------|
|image|image|

## Proposal

- [x] Set a color for order group message element
- [x] Propose an improvement of spacing in CourseProductItem footer.


rlecellier (approved):
look's good !

PR #2363, opened 2024-04-09
✨(frontend) add fullname form in saletunnel 

rlecellier:
add a story for the sale tunnel. It make it easy to update it.

BEFORE
image
image
=====
AFTER
image
image

### DIFF
--- a/src/frontend/js/components/OpenEdxFullNameForm/index.tsx
+++ b/src/frontend/js/components/OpenEdxFullNameForm/index.tsx
@@ -1,0 +1,174 @@
+import { Button, Input } from '@openfun/cunningham-react';
+import { FormattedMessage, defineMessages, useIntl } from 'react-intl';
+import { FormProvider, useForm } from 'react-hook-form';
+import * as Yup from 'yup';
+import { yupResolver } from '@hookform/resolvers/yup';
+import { useEffect, useMemo, useRef, useState } from 'react';
+import { useSession } from 'contexts/SessionContext';
+import useOpenEdxProfile from 'hooks/useOpenEdxProfile';
+import Form, { getLocalizedCunninghamErrorProp } from 'components/Form';
+import { Spinner } from 'components/Spinner';
+import Banner, { BannerType } from 'components/Banner';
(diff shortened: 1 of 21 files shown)


src/frontend/js/components/OpenEdxFullNameForm/index.tsx
```diff
@@ -0,0 +1,165 @@
...
+  } = useOpenEdxProfile({
+    username: user!.username,
+    onUpdateSuccess: () => {
+      setIsSuccess(true);
+    },
+  });
+
+  const defaultValues = {
+    name: (openEdxProfileData ? UserHelper.getName(openEdxProfileData) : '')?.trim(),
+  } as OpenEdxFullNameFormValues;
```
jbpenrath, line 75: Currently this value does not seems to be refreshed when user updates its fullname.

jbpenrath (commented):
Few ideas to improve the layout : 

https://github.com/openfun/richie/assets/9265241/138047e5-420c-4e47-8537-41205d6cd7c8

1. What do you think to put the save button within the input through the rightIcon props ?
2. Then I'll display a spinner during the mutation is pending

Furthermore, Cunningham guidelines suggests to use only one primary button per view. In this view, this should be the "Pay" button. IMO other buttons should be secondary.

src/frontend/js/components/OpenEdxFullNameForm/index.tsx
```diff
@@ -79,10 +76,12 @@ const OpenEdxFullNameForm = () => {
...
     resolver: yupResolver(validationSchema),
   });
+
   const { register, handleSubmit, reset, formState } = form;
 
-  const onSubmit = (values: OpenEdxFullNameFormValues) => {
-    update(values);
+  const onSubmit = async (values: OpenEdxFullNameFormValues) => {
+    await update(values);
+    buttonRef.current?.blur();
```
jbpenrath, line 84: IMO, when the username is updated, the default Values should be updated so the submit should be disabled. Currently this is weird as there is no feedback that the user full name has been updated.

jbpenrath (approved):

PR #2367, opened 2024-04-11
🐛(frontend) fix "Go to course" link in syllabus

rlecellier:
"Go to course" link displayed under enrollment when a user have purchased it were targeting the resource link that is an api endpoint. Here we replace the target by the learner dashboard order details page.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,9 +12,14 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 - Add two new sections in DjangoCMS courses: Accessibility and Required
   Equipment.
+- Ongoing product displayed on the syllabus without active enrollment 
+  now link to the order details page in the learner dashboard.
 
 ## Fixed
 
+- Ongoing product displayed on the syllabus with an active enrollment 
+  now link to the order details page in the learner dashboard instead of 
+  OpenEdx course.
 - Downloaded contract archive on product page was broken, it now works
   properly by sending the course_product_relation_id in the archive 
   creation request.
@@ -22,6 +27,9 @@ Versioning](https://semver.org/spec/v2.0.0.html).
   message was displayed on successful download.
 
 ## [2.25.1]
+
+## Added
+
 
 ### Fixed
 
--- a/src/frontend/js/hooks/useCourseRunOrder/index.tsx
+++ b/src/frontend/js/hooks/useCourseRunOrder/index.tsx
@@ -1,0 +1,27 @@
+import { isJoanieEnabled } from 'api/joanie';
+import { extractResourceMetadata } from 'api/lms/joanie';
+import { useSession } from 'contexts/SessionContext';
+import { useOrders } from 'hooks/useOrders';
+import { CourseRun } from 'types';
+
+const useCourseRunOrder = (courseRun: CourseRun) => {
+  const { user } = useSession();
+  const resourceLinkResources = extractResourceMetadata(courseRun.resource_link);
+  const isProduct = !!(resourceLinkResources?.course && resourceLinkResources?.product);
+
+  if (!isJoanieEnabled || !isProduct || !user) {
+    return { item: undefined, undefined, states: { fetching: false, isFetched: true } };
+  }
+
+  const {
+    items: orders,
+    states: { fetching, isFetched },
+  } = useOrders({
+    course_code: resourceLinkResources?.course,
+    product_id: resourceLinkResources?.product,
+  });
+
+  return { item: orders.length > 0 ? orders[0] : undefined, states: { fetching, isFetched } };
+};
+
+export default useCourseRunOrder;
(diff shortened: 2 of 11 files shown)


src/frontend/js/widgets/SyllabusCourseRunsList/components/CourseProductItem/components/CourseProductCourseRuns/EnrolledCourseRun.tsx
```diff
@@ -91,7 +94,7 @@ const EnrolledCourseRun = ({ enrollment }: Props) => {
           {isStarded && (
             <Button
               size="small"
-              href={enrollment.course_run.resource_link}
+              href={`/dashboard${generatePath(LearnerDashboardPaths.ORDER, { orderId })}`}
```
jbpenrath, line 97: I think you should use the same logic than here.
https://github.com/openfun/richie/blob/c32b64e742fffd56acc2817f3fd271ed205025ad/src/frontend/js/components/SaleTunnel/components/SaleTunnelStepResume/index.tsx#L74-L77

jbpenrath, line 97: You should only do that if the related course run is a product. Otherwise, we should continue to use the `resource_link` as link.

src/frontend/js/utils/test/factories/joanie.ts
```diff
@@ -437,12 +439,14 @@ export const CredentialOrderWithOneClickPaymentFactory = factory(
...
 
 export const CertificateOrderFactory = factory((): CertificateOrder => {
-  return {
+  const order = {
     ...AbstractOrderFactory().one(),
     course: undefined,
     target_courses: [],
     enrollment: EnrollmentLightFactory().one(),
   };
+  delete order.course;
```
jbpenrath, line 448: `course` will exist but will have a `null` value.

src/frontend/js/utils/test/factories/joanie.ts
```diff
@@ -410,11 +410,13 @@ const AbstractOrderFactory = factory((): Order => {
 });
 
 export const CredentialOrderFactory = factory((): CredentialOrder => {
-  return {
+  const order = {
     ...AbstractOrderFactory().one(),
     course: CourseLightFactory().one(),
     enrollment: undefined,
   };
+  delete order.enrollment;
```
jbpenrath, line 418: `enrollment` exists but it will have `null` value

CHANGELOG.md
```diff
@@ -10,6 +10,11 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [2.25.1]
 
+## Added
+
+- Ongoing product displayed on the syllabus are now link to the order
+  details page in the learner dashboard.
+
```
jbpenrath, line 17: IMO this is a fix as a link was displayed but a wrong one.

src/frontend/js/hooks/useCourseRunOrder/index.tsx
```diff
@@ -0,0 +1,25 @@
+import { isJoanieEnabled } from 'api/joanie';
+import { extractResourceMetadata } from 'api/lms/joanie';
+import { useSession } from 'contexts/SessionContext';
+import { useOrders } from 'hooks/useOrders';
+import { CourseRun } from 'types';
+import { CourseLight } from 'types/Joanie';
+
+const useCourseRunOrder = (courseRun: CourseRun, course?: CourseLight) => {
+  const { user } = useSession();
+  const resourceLinkResources = extractResourceMetadata(courseRun.resource_link);
```
jbpenrath, line 10: If I'm right you should retrieve the course code here ?

jbpenrath (approved):

CHANGELOG.md
rlecellier: It's both a fix and a add:
* when we had an active enrollment, the wrong link was displayed.
* when we had an order but no enrollment, we didn't had link.

src/frontend/js/hooks/useCourseRunOrder/index.tsx
rlecellier: It's look like we can have both (see joanie's [CourseProductRelationViewSet](https://github.com/openfun/joanie/blob/main/src/backend/joanie/core/api/client/__init__.py#L104-L116)).
But you're right, the one used when we sync the product as a richie's course run is the code (see joanie's [CourseProductRelation](https://github.com/openfun/joanie/blob/main/src/backend/joanie/core/models/courses.py#L743-L756) model)

I would not say that i'm 100% `extractResourceMetadata` will return a code but now that i know it's the case here, i could remove this code property chain.

rlecellier:
```shell
$ git grep 'courses/.*code.*/products' | wc -l
40
$ git grep 'courses/.*id.*/products' | wc -l
27
```

jbpenrath: You're right, it looks risky to `extractResourceMetadata` to get course code.

PR #2380, opened 2024-04-19
✨(frontend) refetch user profile after edition

rlecellier:
On click on modify profile, we open a new window in OpenEdx profile page.
When modification are done, we would like to refresh informations display on this page.

This onWindowFocus event will be trigger the first time the user came back from OpenEdx profile page to refresh the informations. /!\ It's not perfect as it'll not work it he go and back from one tab to the other.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Added
 
+- User profile in the learner dashboard is now always synchronized with 
+  openEdx profile informations.
 - Add a tab in learner dashboard certificate page to render both order
   certificate and enrollment certificate lists.
 - Add Lyra payment interface
--- a/src/frontend/js/pages/DashboardCreditCardsManagement/DashboardEditCreditCard.spec.tsx
+++ b/src/frontend/js/pages/DashboardCreditCardsManagement/DashboardEditCreditCard.spec.tsx
@@ -97,6 +97,7 @@ describe('<DahsboardEditCreditCard/>', () => {
   });
 
   it('updates a credit card name', async () => {
+    fetchMock.get('https://endpoint.test/api/v1.0/user/me', richieUser);
     const creditCard = CreditCardFactory().one();
     const creditCards = [...CreditCardFactory().many(5), creditCard];
     const creditCardUpdated = CreditCardFactory().one();
@@ -195,6 +196,7 @@ describe('<DahsboardEditCreditCard/>', () => {
   });
 
   it('sets credit card as main', async () => {
+    fetchMock.get('https://endpoint.test/api/v1.0/user/me', richieUser);
     const creditCard = CreditCardFactory().one();
     const creditCards = [...CreditCardFactory().many(5), creditCard];
     const creditCardUpdated = { ...creditCard };
@@ -282,6 +284,7 @@ describe('<DahsboardEditCreditCard/>', () => {
   });
 
   it('deletes a credit card', async () => {
+    fetchMock.get('https://endpoint.test/api/v1.0/user/me', richieUser);
     const creditCard = CreditCardFactory().one();
     let creditCards = [...CreditCardFactory().many(5), creditCard];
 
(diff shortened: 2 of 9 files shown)


CHANGELOG.md
```diff
@@ -20,6 +20,8 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Fixed
 
+- User profile in the learner dashboard is now always synchornized with
```
jbpenrath, line 23:
```suggestion
- User profile in the learner dashboard is now always synchronized with 
```

src/frontend/js/pages/DashboardOpenEdxProfile/index.tsx
```diff
@@ -111,7 +111,24 @@ export const DEFAULT_DISPLAYED_FORM_VALUE = ' - ';
...
+  };
+
+  // Because the user can leaver the page to update his profile in openEdx.
+  // if we invalidate before leaving the page, it's directly refetched.
+  // We need to invalidate openEdx's profile each time we reload this page.
+  // it's the only way to have user data synchronised between richie's frontend
+  // and openEdx's database.
+  useEffect(() => {
+    invalidate();
+  }, []);
```
jbpenrath, line 131: Did you try to trigger invalidation on `beforeunload` event emits by window ?
```suggestion
  const onClickModifyProfile = async () => {
    window.addEventListener('beforeunload', invalidate, { once : true });
    window.location.replace(`${context.authentication.endpoint}/account/settings`);
  };
```

rlecellier: I tried multiple new solution including this beforeunload.
This event do the same as other ways to trigger invalidate, it invalidate then refetch.
I tried to use `removeQueries` and/or `resetQueries` that execute when the query "enabled" switch to `false`.
It work without the `location.replace`: i trigger the reset, no refresh happened, i change route then comme back, the data is refetch.
But it didn't work with the `location.replace`, the reset is trigger, i change page, edit the profile, come back to richie home then profile -> it found the data from somewhere and don't trigger a new fetch.

src/frontend/js/index.tsx
```diff
@@ -102,7 +102,7 @@ async function render() {
     document.body.append(rootContainer);
 
     // React-query
-    const queryClient = createQueryClient({ logger: true, persister: true });
+    const queryClient = createQueryClient({ logger: true, persister: false });
```
jbpenrath, line 105: Debug residue ?

src/frontend/js/pages/DashboardCreditCardsManagement/DashboardEditCreditCard.spec.tsx
```diff
@@ -97,6 +97,7 @@ describe('<DahsboardEditCreditCard/>', () => {
   });
 
   it('updates a credit card name', async () => {
+    fetchMock.get('https://endpoint.test/api/v1.0/user/me', richieUser);
```
jbpenrath, line 100: You could add it in the `beforeEach` nope ?

src/frontend/js/pages/DashboardOpenEdxProfile/index.spec.tsx
```diff
@@ -32,9 +32,11 @@ describe('pages.DashboardOpenEdxProfile', () => {
   setupJoanieSession();
 
   it('should render profile informations', async () => {
+    const richieUser = UserFactory().one();
+    fetchMock.get('https://endpoint.test/api/v1.0/user/me', richieUser);
```
jbpenrath, line 36: Ditto ?

src/frontend/js/pages/DashboardOpenEdxProfile/index.tsx
```diff
@@ -225,13 +231,9 @@ const DashboardOpenEdxProfile = () => {
...
       <Form.Footer>
-        <RouterButton
-          fullWidth
-          target="_blank"
-          href={`${context.authentication.endpoint}/account/settings`}
-        >
+        <Button fullWidth onClick={onClickModifyProfile}>
           <FormattedMessage {...messages.editButtonLabel} />
-        </RouterButton>
+        </Button>
```
jbpenrath, line 236: For a11y purpose, I think we could use href attribute to generate a link.
Does following suggestion work ?

```suggestion
        <Button
          fullWidth
          href={`${context.authentication.endpoint}/account/settings`}
          onClick={clearReactQueryCache}
        >
          <FormattedMessage {...messages.editButtonLabel} />
        </Button>
```

src/frontend/js/pages/DashboardCreditCardsManagement/DashboardEditCreditCard.spec.tsx
rlecellier: it's not needed everywhere (not on the last test).

jbpenrath (approved):

PR #2390, opened 2024-05-03
⬆️(backend) upgrade to Django 4.x (4.0, 4.1, <5)

jonathanreveille:
## Purpose

**From Django version 3.2.25 to 4.0 :**
The project was using **Django 3.2.25**, it now uses **Django 4.0**, this involved : 
- upgrading postgres image to version 12 (drop support for postgres < 12)
- shift from `pytz` library to `datetime.timezone` instead.
- updated the template tags `{% ifequal %}` in favor of `{% if %}`.
- `USE_L1ON` is deprecated in settings if value is `True`

**From Django 4.0 to 4.1 :**
- Update project.toml file to pin version to 4.1 of Django.
- Adjust admin view tests 

**From Django 4.1 to 4.2:**
- Update project.toml file to pin version Django <5
- Change staticfile settings constant in configuration file.
- Dropped support for mysql < 8
- Dropped circleci test-back-msql-5

## Proposal

- resources : 
1. https://docs.djangoproject.com/en/5.0/releases/4.0/
2. https://docs.djangoproject.com/en/5.0/releases/4.1/
3. https://docs.djangoproject.com/en/5.0/releases/4.2/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Changed
 
+- Upgrade to Django version 4.2 (pin version < 5)
 - Migrate to Sentry SDK 2.0
 
 ## Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
   the user isn't authorized.
 - Add two new sections in DjangoCMS courses: Accessibility and Required
   Equipment.
-- Ongoing product displayed on the syllabus without active enrollment 
+- Ongoing product displayed on the syllabus without active enrollment
   now link to the order details page in the learner dashboard.
 
 ## Changed
@@ -24,11 +24,11 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Fixed
 
-- Ongoing product displayed on the syllabus with an active enrollment 
-  now link to the order details page in the learner dashboard instead of 
+- Ongoing product displayed on the syllabus with an active enrollment
+  now link to the order details page in the learner dashboard instead of
   OpenEdx course.
 - Downloaded contract archive on product page was broken, it now works
-  properly by sending the course_product_relation_id in the archive 
+  properly by sending the course_product_relation_id in the archive
   creation request.
 - Fix a bug in the learner dashboard contract page where an alert
   message was displayed on successful download.
@@ -46,7 +46,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Added
 
-- Add a form to edit user fullname in the payment step of the sale 
+- Add a form to edit user fullname in the payment step of the sale
   tunnel.
 - Add ready-only user profile in the learner dashboard preferences page.
   This profile display data from LMS profile.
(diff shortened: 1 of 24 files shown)


CHANGELOG.md
jbpenrath, line 1: You have a custom linter in the backend I think ^^

CHANGELOG.md
```diff
@@ -15,20 +15,26 @@ Versioning](https://semver.org/spec/v2.0.0.html).
...
   now link to the order details page in the learner dashboard.
 
 ## Changed
 
+- Update configuration setting `STATICFILES_STORAGE` and
+  `DEFAULT_FILE_STORAGE` with `STORAGES` dictionary configuration
+- Pin Django version < 5 and dropped support for MySQL version < 8
+- Upgrade to Django 4.1 version
+- Upgrade to Django 4.0 version and postgres to version 12 in
+  docker compose file
```
jbpenrath, line 28: IMO, it does not make sense to add an entry to each upgrade of Django. The final result of your work is that you upgrade to Django 4.2, the CHANGELOG should not reflect the fact you updated progressively.

jbpenrath (approved):

CHANGELOG.md
jonathanreveille: Okay, I will adjust this file right away. 
I thought it would be good to insert a line for each commit within this Pull Request.

kernicPanel (approved):

tests/apps/courses/test_admin_course_run.py
```diff
@@ -209,12 +208,12 @@ def test_admin_course_run_change_view_get_superuser_draft(self):
...
-        self.assertContains(response, "id_enrollment_start_", count=3)
-        self.assertContains(response, "id_enrollment_end_", count=3)
-        self.assertContains(response, "id_languages", count=2)
-        self.assertContains(response, "id_enrollment_count", count=2)
+        self.assertContains(response, "id_start_", count=2)
+        self.assertContains(response, "id_end_", count=2)
+        self.assertContains(response, "id_enrollment_start_", count=2)
+        self.assertContains(response, "id_enrollment_end_", count=2)
+        self.assertContains(response, "id_languages", count=3)
+        self.assertContains(response, "id_enrollment_count", count=3)
```
lunika, line 216: Why all the count are changing ?

jonathanreveille: When I set the version 4.1 of DJango, this test failed by itself. It seems like django might have change their template frendering for admin views. By counting in the response.content, we can find that change. I will count them again to confirm you the count found between version 3.2, 4.0, and 4.1.

wilbrdt (approved):
LGTM! :tada:

tests/apps/courses/test_admin_course_run.py
jonathanreveille: After my investigation, once we pass to the versionn 4.1 of Django, this test suddenly fails and needs some adjustments. 
The new count is accurate.  Plus, I've checked the creation on the django admin side of Richie and it works to create course runs in each versions (3.2, 4.0, 4.1, and 4.2.13). All the fields in the forms work and my resource is created at the end of the form. There are no errors and the page is accessible.

jonathanreveille: Using the DOM with the inspector:

When the count decreases by 1: In version 4.1, the label ("Début de cours" - Course start) no longer has the string of the type id_start_, totaling 2 occurrences, for example, for the 2 fields to fill containing the searched string. When we are below version 4.1, we have the label and the 2 fields that have the searched string, totaling 3 occurrences.

When the count increases by 1: In version 4.1, a new tag is added to the template for certain fields of the type chainerecherché_helptext that we do not have in version < 4.1 (for example: id_languages_helptext or id_enrollment_count_helptext). In versions <4.1, the string was simply "help".

lunika (approved):

wilbrdt (approved):

lunika (approved):

jbpenrath (approved):

PR #2412, opened 2024-05-23
✨(frontend) retrieve estimated payment schedule in sale tunnel

jbpenrath:
## Purpose

In further development, we will stop to sell training with one shot payment. Instead we will schedule payments. From frontend, that means, before the user purchase a training, it should be able to know what is the planned schedule. A first step is to retrieve the estimated payment schedule and display it in the sale tunnel.

CleanShot 2024-06-12 at 16 51 08@2x

## Proposal

- [x] Add payment schedule api route to Joanie API module
- [x] Create a hook to retrieve payment schedule through react-query
- [x] Display payment schedule in the sale tunnel

### DIFF
--- a/src/frontend/js/utils/OrderHelper/index.ts
+++ b/src/frontend/js/utils/OrderHelper/index.ts
@@ -5,6 +5,7 @@ import {
   OrderState,
   ContractDefinition,
   NestedCourseOrder,
+  PaymentScheduleState,
 } from 'types/Joanie';
 
 export enum OrderStatus {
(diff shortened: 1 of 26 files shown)


kernicPanel (approved):

src/frontend/js/utils/OrderHelper/index.ts
```diff
@@ -95,6 +96,8 @@ export class OrderHelper {
   }
 
   static getFailedInstallment(order: Order) {
-    return order.payment_schedule?.find((installment) => installment.state === 'refused');
+    return order.payment_schedule?.find(
+      (installment) => installment.state === PaymentScheduleState.FAILED,
```
lunika, line 100: The state defined in the backend are this one : 

```
PAYMENT_STATE_PENDING = "pending"
PAYMENT_STATE_PAID = "paid"
PAYMENT_STATE_REFUSED = "refused"
```

Failed is not a valid one.

jbpenrath: Oh yes good catch !

lunika (approved):

PR #2415, opened 2024-05-28
🐛(frontend) put CunninghamProvider at the root level

jbpenrath:
## Purpose

Currently some views are broken as there are using Cunningham context but there
are not wrapped into the CunninghamProvider. To prevent this issue, we decide to
wrap the whole app into the CunninghamProvider.

## Proposal

- [x] Put `CunninghamProvider` at root level
- [x] Refactor test wrappers


NathanVss (approved):
Nothing to say, shiiip ⛴️

PR #2416, opened 2024-05-30
🐛(frontend) reset payment state on payment failure

jbpenrath:
## Purpose

Unlike Payplug, with Lyra, until the user has submit its payment, nothing will
happend on the backend side to notify that a payment has expired or has failed.
So we have to manage all those aspects on the frontend side.

## Proposal

- [x] Improve Lyra error management to prevent to stuck user during payment
- [x] Fix some bugs found on the road (mainly resubmit order after payment failure)

### DIFF
--- a/src/frontend/js/components/SaleTunnel/GenericPaymentButton/index.tsx
+++ b/src/frontend/js/components/SaleTunnel/GenericPaymentButton/index.tsx
@@ -104,7 +104,7 @@ export const GenericPaymentButton = ({ buildOrderPayload }: Props) => {
   const { methods: orderMethods } = useOrders(undefined, { enabled: false });
   const [payment, setPayment] = useState<PaymentInfo>();
   const [state, setState] = useState<ComponentStates>(ComponentStates.IDLE);
-  const [error, setError] = useState<PaymentErrorMessageId>(PaymentErrorMessageId.ERROR_DEFAULT);
+  const [error, setError] = useState<PaymentErrorMessageId>();
   const hasPaymentId = (p: Maybe<Payment>): p is Extract<Payment, PaymentWithId> => {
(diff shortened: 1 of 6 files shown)


src/frontend/js/components/SaleTunnel/GenericPaymentButton/index.tsx
```diff
@@ -33,6 +33,13 @@ const messages = defineMessages({
     description: 'Error message shown when payment creation request failed.',
     id: 'components.PaymentButton.errorDefault',
   },
+  errorPollingLimit: {
```
NathanVss, line 36: Great, I used the same approach on my PR

src/frontend/js/components/SaleTunnel/GenericPaymentButton/index.tsx
```diff
@@ -235,8 +242,9 @@ export const GenericPaymentButton = ({ buildOrderPayload }: Props) => {
     const checkOrderValidity = async () => {
       if (round >= PAYMENT_SETTINGS.pollLimit) {
         timeoutRef.current = undefined;
-        orderMethods.abort({ id: payment!.order_id, payment_id: paymentId });
-        setState(ComponentStates.ERROR);
+        // @TODO Instead of displaying a raw error message, we could call onSuccess with a "special" argument
```
NathanVss, line 249: Why not doing it in this PR?

NathanVss (approved):
LGTM :)

src/frontend/js/components/SaleTunnel/GenericPaymentButton/index.tsx
jbpenrath: Yup indeed, I'll could do that right now !

PR #2425, opened 2024-06-05
⚡️(cdn) load react JS bundle from CDN

### LINKED ISSUE #2423: React JS bundles not loading from CDN domain
## Bug Report

It isn't a bug, it is more of an improvement of performance.

**Problematic Behavior**
On a Richie site with a CDN for static assets, the React JS bundles are being loaded directly from Richie domain, but they could be downloaded using the CDN domain.

**Expected behavior/code**

The page https://www.fun-mooc.fr/en/ loads:
- https://www.fun-mooc.fr/static/richie/js/build/36108.643fcedba17297f192e3.index.js

But the course search page loads:
- https://dz03nossv7tsm.cloudfront.net/static/richie/js/build/36108.643fcedba17297f192e3.index.js

**Environment**
- Any

**Possible Solution**

The block:
https://github.com/openfun/richie/blob/2951503e3f62480e64be8410ed3a7977cb7b2db2/src/richie/apps/search/templates/search/search.html#L5-L10

I think it should be included to the base.html of the Richie site.

```html
{% block meta_public_path %}
    {% if CDN_DOMAIN %}
    <meta name="public-path" value="{{ CDN_DOMAIN }}" />
    {% endif %}
{% endblock meta_public_path%}
```

Add any other context about the problem here. If applicable, add screenshots to help explain.

igobranco:
If Richie site is configured to use a CDN_DOMAIN then the React JS Bundle should be downloaded from CDN instead of Richie site domain.

resolves #2423

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,10 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 - PurchaseButton instantiated a hook conditionnaly
 - Make UserHelper manages OpenEdxProfile object to extract name
 
+### Added
+
+- Load React JS bundles from CDN on all pages
+
 ## [2.27.0]
 
 ### Added
--- a/src/richie/apps/core/templates/richie/base.html
+++ b/src/richie/apps/core/templates/richie/base.html
@@ -51,6 +51,12 @@
             {% endblock meta_rdfa_context %}
             {% endblock meta_rdfa %}
 
+            {% block meta_public_path %}
+                {% if CDN_DOMAIN %}
+                    <meta name="public-path" value="{{ CDN_DOMAIN }}" />
+                {% endif %}
+            {% endblock meta_public_path%}
+
         {% endblock meta %}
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 - Add `SaleTunnelNotValidated` step into `SaleTunnel`
 - Add WarningIcon component
+- Load React JS bundles from CDN on all pages
```
jbpenrath, line 23: You put this entry in the `2.27.0` release. Do not forgot to move up into `Unrealeased` section

jbpenrath (approved):
Nice catch! Good to see you @igobranco

CHANGELOG.md
igobranco: Yes, sorry

igobranco:
> Nice catch! Good to see you @igobranco

Thanks! We are going to migrate to Kubernetes using Open edX Tutor approach of deployment and we came up with this inconsistency!

PR #2427, opened 2024-06-06
🐛(backend) fix Category.get_children_categories in public mode

jbpenrath:
## Purpose

Currently, calling `get_children_categories` for a published category return all category children even draft ones. In order to prevent that, if the category is not in draft, we explicitly exclude unpublished categories.


igobranco:
@jbpenrath  I can reproduce the problem on the demo site, so I'm having difficult to give you feedback on this PR.

kernicPanel (approved):

lunika (approved):

PR #2428, opened 2024-06-07
✨(frontend) compact syllabus for self-paced

### LINKED ISSUE #1969: Richie course page: change order of info and simplify date presentation
## Feature Request

We believe that the course page can be improved, to prevent cases where the student don’t understand how much effort and the duration of the course. We have found cases, on support and comments on the social networks, where they simply don’t understand because they view longer start and end dates and they though that the course is very long, but in the reality the course is only a couple of hours in self pace mode. The majority of NAU courses is self pace and I think this situation is more dramatic in our case than in FUN installation, where there are a more balanced mix of self pace and instructor pace. That is why we propose the next improvement on Richie, if we don’t agree we could add it only on our installation, but if we agree on a compromise we could improve the platform for everyone.

### Restrictions
For:
- self pace course
- single open to enroll course run
- current date is in the enrollment period - course in open for enroll

Then on the course run block:
- do not show any enrollment dates.
- show only the end date - on a compact style.

### Language

I think another problem is the language per course run. It is a good feature to have different runs using different languages, but in reality almost all course runs for the course has the same language(s).
What we propose is if all course runs for the course has the same language(s) then we should present that information only once. For an example where a run is almost finish and another is starting up.
One location for this single language presentation could be after the duration, effort and pace; but before the enrollment count.
So, the course languages should be an union of all course run languages.

### Change order of Catch phrase, duration, effort and pace

If we compare Richie with other Course catalog systems, like [edx.org](http://edx.org/) or Coursera, we see that the catch phrase is right below the course title. Only after the catch phrase, it has more information, for example it shows the course duration, effort and pace. We believe that we should change the order of the information to be similar to those platforms.

We propose that we change the order of those elements in Richie. To be like: course title, catch phrase, duration, effort, pace, languages and enrollment count. Maintaining the duration, effort and pace on the same line.

<Course Title>
<catch phrase>
<duration> <effort> <pace>
<languages>
<enrollment count>

### Screen captures and Mockups

Current course presentation from the demo site:
image

(Mockup) The same demo course presented with the suggestion changes:
course_page

On NAU this change would have this effect:
nau_course_card

Implement part on NAU Richie site factory

Yes.

JoaoGarcao:
1. If all runs of an opened course have the same languages, we hide the languages on the course run card, because those languages are shown after the course description.
2. On self-paced courses with single opened run, display a simplified version of the syllabus.

fix #1969

Change the message of the enrollment period for a course with single opened run when it is self-paced, and
hide the languages of a course when every run of that course has the same languages as all the others.

## Opened self-paced course:
This PR shows a compacted version of the information.
Before:

After:

## Opened course with the same languages as all the other runs of that course
As all runs for the course as the same languages, showing the same exact information multiple times is unnecessary, so this PR removes that languages on each run syllabus.
This behavior is the same for both self-paced and instructor-paced courses.
Before:

After:

## Instructor course with different languages, stays the same
If a course isn't self-paced (instructor pace) and the languages of all runs of that course aren't equal, then the course run card stays the same as it was:

## Compacted version for self-paced & same languages for all runs
If an opened course run is self-paced and the languages of all runs of that course are equal, then the course run card will be as displayed bellow.
This is the normal/majority of the courses in [NAU](https://www.nau.edu.pt/).

## Compacted version for self-paced, same languages for all runs & no course end date
if we have the case described above, but also there is no course end date,  then the course run card will be:

FYI @igobranco @sandroscosta

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,11 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 - csrftoken in richie context
 
+### Added
+
+- On self-paced courses with single run display a simplified
+  version of the syllabus.
+
 ## [2.28.1]
 
 ### Fixed
--- a/src/frontend/js/types/index.ts
+++ b/src/frontend/js/types/index.ts
@@ -16,6 +16,12 @@ export enum CourseRunDisplayMode {
   DETAILED = 'detailed',
 }
 
+export interface PacedCourse {
+  id: string;
+  code: string;
+  is_self_paced: boolean;
+}
+
 export interface CourseRun {
   id: number;
   resource_link: string;
--- a/src/frontend/js/utils/CourseRunHelper/index.spec.ts
+++ b/src/frontend/js/utils/CourseRunHelper/index.spec.ts
@@ -1,0 +1,35 @@
+import { CourseRun, Priority } from 'types';
+import { CourseRunHelper } from 'utils/CourseRunHelper/index';
+import { CourseRunFactoryFromPriority } from 'utils/test/factories/richie';
+
+describe('CourseRunHelper', () => {
+  it('should return true if all course runs have the same languages', () => {
+    const courseRuns: CourseRun[] = [
+      CourseRunFactoryFromPriority(Priority.ONGOING_OPEN)({
+        languages: ['en'],
+      }).one(),
+      CourseRunFactoryFromPriority(Priority.ARCHIVED_OPEN)({
+        languages: ['en'],
+      }).one(),
+      CourseRunFactoryFromPriority(Priority.FUTURE_OPEN)({
+        languages: ['en'],
+      }).one(),
+    ];
+    expect(CourseRunHelper.IsAllCourseRunsWithSameLanguages(courseRuns)).toEqual(true);
+  });
+
+  it('should return false if all course runs have different languages', () => {
+    const courseRuns: CourseRun[] = [
+      CourseRunFactoryFromPriority(Priority.ONGOING_OPEN)({
+        languages: ['en'],
+      }).one(),
+      CourseRunFactoryFromPriority(Priority.ARCHIVED_OPEN)({
+        languages: ['fr'],
+      }).one(),
+      CourseRunFactoryFromPriority(Priority.FUTURE_OPEN)({
+        languages: ['it'],
+      }).one(),
+    ];
+    expect(CourseRunHelper.IsAllCourseRunsWithSameLanguages(courseRuns)).toEqual(false);
+  });
+});
--- a/src/frontend/js/utils/CourseRunHelper/index.ts
+++ b/src/frontend/js/utils/CourseRunHelper/index.ts
@@ -1,0 +1,13 @@
+import { CourseRun } from 'types';
+
+export class CourseRunHelper {
+  /**
+   * Checks if all given runs of a course have the same languages.
+   */
+  static IsAllCourseRunsWithSameLanguages = (courseRuns: CourseRun[]) => {
+    const languages = courseRuns[0].languages.sort().join();
+    return courseRuns
+      .map((courseRun) => courseRun.languages.sort().join())
+      .every((runLanguages) => runLanguages === languages);
+  };
+}
--- a/src/frontend/js/widgets/SyllabusCourseRunsList/components/SyllabusAsideList/index.tsx
+++ b/src/frontend/js/widgets/SyllabusCourseRunsList/components/SyllabusAsideList/index.tsx
@@ -1,9 +1,9 @@
 import { defineMessages, FormattedMessage, useIntl } from 'react-intl';
 import React from 'react';
-import { CourseRun, Priority } from 'types';
+import { PacedCourse, CourseRun, Priority } from 'types';
+import { CourseRunHelper } from 'utils/CourseRunHelper';
 import { SyllabusSimpleCourseRunsList } from 'widgets/SyllabusCourseRunsList/components/SyllabusSimpleCourseRunsList';
 import { SyllabusCourseRun } from 'widgets/SyllabusCourseRunsList/components/SyllabusCourseRun';
-import { CourseLight } from 'types/Joanie';
 
 const messages = defineMessages({
   otherCourseRuns: {
@@ -57,7 +57,7 @@ export const SyllabusAsideList = ({
   maxArchivedCourseRuns,
 }: {
   courseRuns: CourseRun[];
-  course: CourseLight;
+  course: PacedCourse;
   maxArchivedCourseRuns: number;
 }) => {
   const intl = useIntl();
@@ -106,6 +106,8 @@ export const SyllabusAsideList = ({
     [Priority.ARCHIVED_CLOSED].includes(run.state.priority),
   );
 
+  const showLanguages = CourseRunHelper.IsAllCourseRunsWithSameLanguages(courseRuns);
+
   return (
     <>
       <h2 className="course-detail__title">
@@ -134,7 +136,12 @@ export const SyllabusAsideList = ({
           className="course-detail__row course-detail__runs course-detail__runs--open"
         >
           {openedRuns.map((run) => (
-            <SyllabusCourseRun key={run.id} courseRun={run} course={course} />
+            <SyllabusCourseRun
+              key={run.id}
+              courseRun={run}
+              course={course}
+              showLanguages={showLanguages}
+            />
           ))}
         </div>
       )}
--- a/src/frontend/js/widgets/SyllabusCourseRunsList/components/SyllabusCourseRunCompacted/index.tsx
+++ b/src/frontend/js/widgets/SyllabusCourseRunsList/components/SyllabusCourseRunCompacted/index.tsx
@@ -1,5 +1,4 @@
 import { defineMessages, FormattedMessage, useIntl } from 'react-intl';
-import { Button } from '@openfun/cunningham-react';
 import { CourseRun, CourseRunDisplayMode, PacedCourse } from 'types';
 import useDateFormat from 'hooks/useDateFormat';
 import { extractResourceId, isJoanieResourceLinkProduct } from 'api/lms/joanie';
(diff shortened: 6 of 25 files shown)


igobranco:
@JoaoGarcao please fix the lint's.

```bash
make lint-back
make lint-front
```

On the last image scenario, _Compacted version for self-paced & same languages for all runs_, please remove the _Enrollment_.
This means that the header with the _ Enrollment_ as text shouldn't be displayed on the compacted version.

igobranco:
@jbpenrath - I haven't present to you @JoaoGarcao , but he is a student that have been doing a curricular internship on NAU.
I've already reviewed this PR with him, but I think you or someone from FUN should review this PR, before we merge it.
This PR has impact on almost 100% of NAU course pages, but on you side as lower impact ( < 50% opened are self paced).

src/frontend/js/types/Joanie.ts
```diff
@@ -204,9 +205,10 @@ export interface AbstractCourse {
   organizations: OrganizationLight[];
   title: string;
   course_runs: CourseRun[];
+  is_self_paced: boolean;
 }
 
-export type CourseLight = Pick<AbstractCourse, 'id' | 'code' | 'title'> & {
+export type CourseLight = Pick<AbstractCourse, 'id' | 'code' | 'title' | 'is_self_paced'> & {
   cover: Nullable<JoanieFile>;
 };
```
jbpenrath, line 213: Until this PR, the `CourseLight` type can be shared. But since you add the field `is_self_paced` you have to create a dedicated type.

Indeed, We are using `CourseLight` type to type Joanie API Responses but from Joanie the response will never return `is_self_paced` field.

IMO you should create a Course type within `types/index.ts`

CHANGELOG.md
```diff
@@ -73,7 +78,6 @@ Versioning](https://semver.org/spec/v2.0.0.html).
   Equipment.
 - Ongoing product displayed on the syllabus without active enrollment
   now link to the order details page in the learner dashboard.
-- Add new sale tunnel
```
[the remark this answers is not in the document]
JoaoGarcao: Change reverted.

src/frontend/js/types/Joanie.ts
JoaoGarcao: As you said, i created a Course type (and named it as `PacedCourse`) within `types/indes.ts`. It has a `courseLight` and the boolean variable `is_self_paced`.

src/frontend/js/utils/CourseRunLanguages/index.ts
```diff
@@ -0,0 +1,15 @@
...
+    const runLanguages = courseRuns[i].languages.sort();
+    if (
+      !(runLanguages.length === languages.length) ||
+      !runLanguages.every((value, index) => value === languages[index])
+    ) {
+      return false;
+    }
+  }
+  return true;
+};
```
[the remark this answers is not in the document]
JoaoGarcao: I looked at `CertificateHelper` and tried to do a similar thing, hopefully it looks good to you.

src/frontend/js/widgets/SyllabusCourseRunsList/components/SyllabusAsideList/index.tsx
```diff
@@ -106,6 +106,8 @@ export const SyllabusAsideList = ({
     [Priority.ARCHIVED_CLOSED].includes(run.state.priority),
   );
 
+  const runsWithSameLanguages = CourseRunHelper.IsAllCourseRunsWithSameLanguages(courseRuns);
```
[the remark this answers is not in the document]
JoaoGarcao: Change done.

src/frontend/js/utils/CourseRunHelper/index.ts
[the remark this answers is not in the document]
JoaoGarcao: Tests made.

src/frontend/js/widgets/SyllabusCourseRunsList/components/SyllabusCourseRunCompacted/index.tsx
```diff
@@ -0,0 +1,112 @@
...
+  selfPaceRunPeriod: {
+    id: 'components.SyllabusCourseRun.selfPaceCoursePeriod',
+    description: 'Course date of an opened and self paced course run block',
+    defaultMessage: 'Available until {endDate}',
+  },
+  selfPaceNoEndDate: {
+    id: 'components.SyllabusCourseRun.selfPaceNoEndDate',
+    description: 'Self paced course run block with no end date',
+    defaultMessage: 'Available',
+  },
```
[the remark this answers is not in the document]
JoaoGarcao: Replacement done.

src/frontend/js/types/index.ts
```diff
@@ -16,6 +17,11 @@ export enum CourseRunDisplayMode {
   DETAILED = 'detailed',
 }
 
+export interface PacedCourse {
+  courseLight: CourseLight;
+  is_self_paced: boolean;
+}
```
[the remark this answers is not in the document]
JoaoGarcao: I did that change as you said, but i also changed the type of some variables, because there were fields that ```PacedCourse``` couldn't have.

src/frontend/js/utils/CourseRunHelper/index.spec.ts
```diff
@@ -0,0 +1,35 @@
+import { CourseRun, Priority } from 'types';
+import { CourseRunHelper } from 'utils/CourseRunHelper/index';
+import { CourseRunFactoryFromPriority } from 'utils/test/factories/richie';
+
+describe('CourseRunHelper', () => {
+  it('All courses have the same languages', () => {
```
jbpenrath, line 6:
```suggestion
  it('should return true if all course runs have the same languages', () => {
```

src/frontend/js/utils/CourseRunHelper/index.spec.ts
```diff
@@ -0,0 +1,35 @@
...
+        languages: ['en'],
+      }).one(),
+      CourseRunFactoryFromPriority(Priority.FUTURE_OPEN)({
+        languages: ['en'],
+      }).one(),
+    ];
+    expect(CourseRunHelper.IsAllCourseRunsWithSameLanguages(courseRuns)).toEqual(true);
+  });
+
+  it('All courses haev diferent languages', () => {
```
jbpenrath, line 21:
```suggestion
  it('should return false if all course runs have different languages', () => {
```

src/frontend/js/widgets/SyllabusCourseRunsList/index.tsx
```diff
@@ -41,7 +43,7 @@ const SyllabusCourseRunsList = ({
   maxArchivedCourseRuns,
 }: {
   courseRuns: CourseRun[];
-  course: CourseLight;
+  course: CourseLight | PacedCourse;
```
jbpenrath, line 46:
```suggestion
  course: PacedCourse;
```

With your changes, this component should now accept only `PacedCourse`

src/frontend/js/widgets/SyllabusCourseRunsList/index.tsx
```diff
@@ -81,11 +85,26 @@ const SyllabusCourseRunsList = ({
           </div>
         </div>
       )}
-      {openedRuns.length === 1 && (
-        <div className="course-detail__row course-detail__runs course-detail__runs--open">
-          <SyllabusCourseRun courseRun={openedRuns[0]} course={course} />
-        </div>
-      )}
+      {openedRuns.length === 1 &&
+        ('is_self_paced' in course &&
```
jbpenrath, line 89: As now course is a CoursePaced, you are able to remove this first check

jbpenrath (approved):
Last comments then looks good to me ✅👏

Just an overall comment for your next PRs try to split your work in smaller commits. E.g: IMO this PR could have been split into 3 commits.
1. Commit which udpate the extra tags to self_paced property in the backend application
2. Commit focused on the showLanguages feature
3. Commit focused on the new component for self paced course runs.

src/frontend/js/utils/CourseRunHelper/index.spec.ts
JoaoGarcao: Change done.

src/frontend/js/utils/CourseRunHelper/index.spec.ts
JoaoGarcao: Change done.

src/frontend/js/widgets/SyllabusCourseRunsList/index.tsx
JoaoGarcao: I've made some changes according to what I've understood.

src/frontend/js/types/Joanie.ts
```diff
@@ -182,6 +182,16 @@ export interface CourseProductRelation {
...
 }
+
+export interface PacedCourseProductRelation {
+  id: string;
+  course: PacedCourse;
+  organizations: Organization[];
+  product: Product;
+  created_on: string;
+  order_groups: OrderGroup[];
+}
```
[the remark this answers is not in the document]
JoaoGarcao: Change done.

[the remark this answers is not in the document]
jbpenrath: Just for information, click on the ♻️ icon to notify me that you are waiting for a new review.
image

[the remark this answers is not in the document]
JoaoGarcao: Thank you, i'll start using it for now on.

src/frontend/js/utils/test/factories/joanie.ts
```diff
@@ -277,6 +277,14 @@ export const CourseFactory = factory((): CourseListItem => {
...
 });
 
+export const PacedCourseFactory = factory((): PacedCourse => {
+  return {
+    id: faker.string.uuid(),
+    code: faker.string.alphanumeric(5),
+    is_self_paced: false,
+  };
+});
+
```
jbpenrath, line 287: Can you just move this factory into `js/utils/test/factories/richie.ts` please.

src/frontend/js/types/Joanie.ts
```diff
@@ -182,6 +182,7 @@ export interface CourseProductRelation {
   created_on: string;
   order_groups: OrderGroup[];
 }
+
```
jbpenrath, line 185:
```suggestion
```

jbpenrath (approved):
Last minor comment then looks good to merge 👏

igobranco (approved):

PR #2440, opened 2024-06-20
🐛(frontend) fix ProductCertificateFooter state issue

jbpenrath:
## Purpose

Currently, if the learner has a submitted or pending order, the label displayed is the one as it has a validated order. Instead for submitted order, we want to prevent that the user tries to purchase again the product and with pending order
 the learner should be able to retry to purchase the certificate.

### DIFF
--- a/src/frontend/js/widgets/Dashboard/components/DashboardItem/Enrollment/ProductCertificateFooter/index.spec.tsx
+++ b/src/frontend/js/widgets/Dashboard/components/DashboardItem/Enrollment/ProductCertificateFooter/index.spec.tsx
@@ -115,7 +115,7 @@ describe('<ProductCertificateFooter/>', () => {
     },
   ])(
     "shouldn't display purchase button for a closed course run without order (state $courseRunStateData.priority).",
-    async ({ courseRunStateData }) => {
+    ({ courseRunStateData }) => {
       render(
         <ProductCertificateFooter
           product={product}
@@ -132,7 +132,7 @@ describe('<ProductCertificateFooter/>', () => {
     },
   );
 
-  it('should display download button for a course run with certificate.', async () => {
+  it('should display download button for a course run with certificate.', () => {
     const order = OrderEnrollmentFactory({
       certificate_id: 'FAKE_CERTIFICATE_ID',
       state: OrderState.VALIDATED,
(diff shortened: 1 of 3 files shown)


src/frontend/js/widgets/Dashboard/components/DashboardItem/Enrollment/ProductCertificateFooter/index.spec.tsx
```diff
@@ -147,10 +147,40 @@ describe('<ProductCertificateFooter/>', () => {
...
+      product_id: product.id,
+      state: OrderState.PENDING,
+    }).one();
+    const enrollment = EnrollmentFactory({
+      orders: [order],
+      course_run: CourseRunFactory({ course }).one(),
+    }).one();
+    render(<ProductCertificateFooter product={product} enrollment={enrollment} />);
+    expect(await screen.queryByRole('button', { name: 'Download' })).not.toBeInTheDocument();
+    expect(screen.queryByTestId('PurchaseButton__cta')).toBeInTheDocument();
```
lunika, line 181:
```suggestion
    expect(screen.getByTestId('PurchaseButton__cta')).toBeInTheDocument();
```

src/frontend/js/widgets/Dashboard/components/DashboardItem/Enrollment/ProductCertificateFooter/index.spec.tsx
```diff
@@ -147,10 +147,40 @@ describe('<ProductCertificateFooter/>', () => {
...
+      certificate_id: undefined,
+      product_id: product.id,
+      state: OrderState.PENDING,
+    }).one();
+    const enrollment = EnrollmentFactory({
+      orders: [order],
+      course_run: CourseRunFactory({ course }).one(),
+    }).one();
+    render(<ProductCertificateFooter product={product} enrollment={enrollment} />);
+    expect(await screen.queryByRole('button', { name: 'Download' })).not.toBeInTheDocument();
```
lunika, line 180: Hmmm `queryBy` doesn't return a promise. You should use `findByRole` if you want a promise. But in your case it should not work.

kernicPanel (approved):
Seems ok to me.
Just an idea: could we abstract the order state usage away by using something like
`order?.hasCertificate` instead of `order?.state === OrderState.VALIDATED` ?
I'm thinking of the incoming order state refactoring here.

src/frontend/js/widgets/Dashboard/components/DashboardItem/Enrollment/ProductCertificateFooter/index.spec.tsx
jbpenrath: `getByTestId` throw an error when the element is not found to the test fails.

jbpenrath: Good catch

jbpenrath:
> Seems ok to me. Just an idea: could we abstract the order state usage away by using something like `order?.hasCertificate` instead of `order?.state === OrderState.VALIDATED` ? I'm thinking of the incoming order state refactoring here.

Good idea, I get it for the coming refactoring

lunika (approved):

PR #2457, opened 2024-07-03
✨(frontend) sale tunnel v2

jbpenrath:
## Purpose

On Joanie, the order workflow has been totally revamped. So we have to rework the Sale Tunnel to work with this new workflow.

The learner is now invited to sign its training agreement first if there is one, then if the order is not free, it is invited to define a payment method which will be used to pay upcoming installments.

https://github.com/user-attachments/assets/66d06a05-63d6-44f9-8b91-960541e42dbe


src/frontend/js/components/PurchaseButton/index.spec.tsx
```diff
@@ -102,11 +103,14 @@ describe('PurchaseButton', () => {
   it('shows cta to open sale tunnel when user is authenticated', async () => {
     const courseCode = '00000';
     const product = ProductFactory().one();
+    const orderQueryParameters = {
+      course_code: courseCode,
+      product_id: product.id,
+      state: [...ACTIVE_ORDER_STATES, ...PURCHASABLE_ORDER_STATES],
```
kernicPanel, line 109: As `[...ACTIVE_ORDER_STATES, ...PURCHASABLE_ORDER_STATES]` is used many times, what about adding `NON_CANCELED_ORDER_STATES` in [types](https://github.com/openfun/richie/blob/65bb019557cae6ff4d7548e9186b905e888e6076/src/frontend/js/types/Joanie.ts#L296) ?

kernicPanel (commented):
Just small comments.

lunika (approved):

PR #2459, opened 2024-07-09
🥅(frontend) improve lyra error catching

jbpenrath:
## Purpose

Currently, when an error occurred during a payment using lyra pop-in, a generic error message is displayed. This is weird as the message is not explicit and user does not understand if the problem comes from the application from its bank.

image

## Proposal

- [x] Improve Lyra error management


kernicPanel (approved):
I suppose we have no tests for this.

PR #2464, opened 2024-07-24
🔥(core) remove csrftoken in frontend context processor

lunika:
## Purpose

The csrftoken present in frontend context processor is not used anymore by the front application. Still using it can lead to increase the django cms cache because lot of csrftoken can be issued and lead to increase significantly the cache.

## Proposal

- [x] remove csrftoken in frontend context processor

### DIFF
--- a/src/frontend/js/types/commonDataProps.ts
+++ b/src/frontend/js/types/commonDataProps.ts
@@ -23,7 +23,6 @@ enum FEATURES {
 
 export interface RichieContext {
   authentication: AuthenticationBackend;
-  csrftoken: string;
   environment: string;
   features: Partial<Record<FEATURES, boolean>>;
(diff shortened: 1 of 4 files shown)


src/frontend/js/types/commonDataProps.ts
```diff
@@ -23,7 +23,6 @@ enum FEATURES {
 
 export interface RichieContext {
   authentication: AuthenticationBackend;
-  csrftoken: string;
```
jbpenrath, line 26: Do not forgot to remove this property from the context factory.

https://github.com/openfun/richie/blob/a55cc68d79acbf75a97af1c6f020bcc107d62151/src/frontend/js/utils/test/factories/richie.ts#L155

PR #2468, opened 2024-08-01
🐛(ci) fix npm development version

lunika:
## Purpose

setuptools_scm generates a development version which can not be compatible with the semver syntax used by npm. For a development version, the version and prerelease tag should be seperate with a dash, setuptools_scm use a period as separation. We have to transform the version with one compatible with NPM.

## Proposal

- [x] fix npm development version


.circleci/config.yml
```diff
@@ -849,6 +849,11 @@ jobs:
             if [[ -z "${CIRCLE_TAG+x}" ]]; then
               python -m pip install --upgrade setuptools_scm
               NEXT_VERSION=$(python -m setuptools_scm)
+              # Match version like 1.2.3.dev123
+              if [[ $NEXT_VERSION =~ ([[:digit:]]+?.[[:digit:]]+?.[[:digit:]]+)?.([[:alnum:]]+) ]]; then
```
jbpenrath, line 853: You should escape the `.` char as currently you are matching any chars.

```suggestion
              if [[ $NEXT_VERSION =~ ([[:digit:]]+?\.[[:digit:]]+?\.[[:digit:]]+)?\.([[:alnum:]]+) ]]; then
```

.circleci/config.yml
```diff
@@ -849,6 +849,11 @@ jobs:
...
+                # transform it in 1.2.3-dev123
+                NEXT_VERSION="${BASH_REMATCH[1]}-${BASH_REMATCH[2]}"
```
jbpenrath, line 855: Nice, one thing learned today !

kernicPanel (approved):

.circleci/config.yml
lunika: Oh you are right. I though that the question mark can be used for this in bash. It is not the case

kernicPanel:
Just a question, do we always get `dev` word in the version ? If so, maybe it would be easier to use sed.
```
NEXT_VERSION=$(python -m setuptools_scm | sed 's/.dev/-dev/')
```

lunika:
> Just a question, do we always get `dev` word in the version ? If so, maybe it would be easier to use sed.
> 
> ```
> NEXT_VERSION=$(python -m setuptools_scm | sed 's/.dev/-dev/')
> ```

No it can be other world like `alpha`, `beta`, etc

kernicPanel:
> > Just a question, do we always get `dev` word in the version ? If so, maybe it would be easier to use sed.
> > ```
> > NEXT_VERSION=$(python -m setuptools_scm | sed 's/.dev/-dev/')
> > ```
> 
> No it can be other world like `alpha`, `beta`, etc

Too bad

PR #2477, opened 2024-08-14
🔧(cookiecutter) fix ignore static process regex

jbpenrath:
## Purpose

The custom storage class currently ignores js files from its process as those files are already containing a hash during its generation. But the regex was not
 improperly configured so those files were matched and their process recently
 leads to issue.

## Proposal

- [x] Fix `STATIC_POSTPROCESS_IGNORE_REGEX` in storage.py


kernicPanel (commented):
In the commit message:
`improperly` -> `properly` ?

kernicPanel (approved):

PR #2479, opened 2024-08-14
👔(frontend) refresh credit cards query on payment method definition

jbpenrath:
## Purpose

Currently, when the user add a new credit card, the related query is not refetched so the new credit card does not appear on the list and credit card information displayed are staled.

## Proposal

- [x] Refresh credit card query on payment method definition


src/frontend/js/components/SaleTunnel/SaleTunnelSavePaymentMethod/index.tsx
```diff
@@ -90,23 +88,18 @@ const SaleTunnelSavePaymentMethod = () => {
 
   useEffect(() => {
     if (!payment) {
```
lunika, line 90: `payment` should not be a dependency ?

jbpenrath: Nope, we just want to trigger this effect when creditCard items has changed.

PR #2493, opened 2024-09-16
✨(backend) Added new page extension 'IndexPage'

sveetch:
## Purpose

This new extension cover the need of  some options for pages in main menu. These options are:

- Enable to show a dropdown of children page;
- Add a custom color class for the page on hover event in menu;

## Proposal

The only proper way to carry these option data on a CMS page is through a page extension. 

However we can not add these options to the existing page extensions since they are not used on the pages listed on the main menu because these pages are just some index to list children pages (like categories, news, courses, etc..) and so these index pages don't have any extension yet.

So we would add this new extension 'IndexPage' that will carry new options and we would create this extension on page that are listed in the main menu, commonly they are page at level 0 of the CMS tree.

This PR include almost everything to implement their usage:

- A new model `IndexPage` with its migration;
- A toolbar class `IndexPageExtensionToolbar` to display (or not) the toolbar menu entry to create or edit the extension on a page; 
- A wizard class `IndexPageWizard` to create a new page with the new extension;
- An admin class `IndexPageAdmin` use by the related wizard;
- A new factory class `IndexPageFactory` to create a page with the new extension;
- A menu modifier class `MenuWithIndexPage` to patch menu to include the new extension data (since on default the menu node are not page objects and lack of many page attributes);
- New settings to manage IndexPage behaviors in toolbar and menu;
- Modified the main menu template to implement usage of IndexPage options;
- Modified the main menu Sass component to start implementation of menu entry hover color;
- Some tests around toolbar to cover its specific behavior (compared to other toolbar classes);


jbpenrath (commented):
As a first review, it looks to work great!

About the look'n feel of the menu, as there is a need of interaction I think we should manage those menus through a React component.

src/richie/apps/courses/cms_wizards.py
jbpenrath, line 1: I'm not quit sure it is useful to add an entry in the Wizard.
IMO, it's fine to be able to disable this feature globally through settings or enable it for every pages. I feel this entry in the wizard will add confusion.

src/richie/apps/courses/settings/__init__.py
```diff
@@ -577,3 +577,14 @@ def richie_placeholder_conf(name):
...
+# If true the toolbar item will already be showed. If false only a page which already
+# have the extension will have the toolbar item and users won't be able to add
+# pageindex extension on existing page, only create new page with index extension
+# through the wizard.
+RICHIE_PAGEINDEX_ALLOW_CREATION = False
+
+# Define which node level can be processed to search for pageindex extension. You can
+# set it to 'None' for never processing any node.
+# This is a limit against performance issues to avoid making querysets for nothing.
+RICHIE_PAGEINDEX_MENU_ALLOWED_LEVEL = 0
```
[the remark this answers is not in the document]
sveetch: Done

src/richie/apps/courses/models/index.py
```diff
@@ -0,0 +1,51 @@
...
+from django.db import models
+from django.utils.translation import gettext_lazy as _
+
+from cms.extensions.extension_pool import extension_pool
+
+from ...core.models import BasePageExtension
+from .. import defaults
+
+
+class IndexPage(BasePageExtension):
```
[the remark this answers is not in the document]
sveetch: Ok, finally i've renamed it to MainMenuEntry, this seems more accurate and follow the scheme from other extension model which does not have "Page" in their name. Every other mentions and variables have been adjusted to the new name.

The related migration have been altered so your local database is probably broken if you already have executed this migration.

src/richie/apps/courses/defaults.py
```diff
@@ -382,3 +386,16 @@
...
+# Define possible hover color that can be choosen for an MainMenuEntry and to apply on
+# its menu item
+MENU_ENTRY_COLOR_CLASSES = getattr(
+    settings,
+    "RICHIE_MENU_ENTRY_COLOR_CLASSES",
+    # (("", "None"),)
+    (
+        ("", "None"),
+        ("primary", "Primary"),
+        ("warning", "Warning"),
```
jbpenrath, line 399: Those labals should be internationalized.

cookiecutter/{{cookiecutter.organization}}-richie-site-factory/template/{{cookiecutter.site}}/src/backend/{{cookiecutter.site}}/settings.py
```diff
@@ -659,6 +659,13 @@ class Base(StyleguideMixin, DRFMixin, RichieCoursesConfigurationMixin, Configura
         "body > svg, #main-menu, .body-footer, .body-mentions"
     )
 
+    # Wheither you can create PageIndex extension on page through toolbar if true or
+    # just editing existing extension if false
+    RICHIE_PAGEINDEX_ALLOW_CREATION = False
+
+    # Define which node level can be processed to search for pageindex extension
+    RICHIE_PAGEINDEX_MENU_ALLOWED_LEVEL = 0
```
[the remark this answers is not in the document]
sveetch: Done

src/richie/apps/courses/defaults.py
sveetch: I've only make it for the 'None' item since the other ones were just for local development, i've removed them so there won't be confusion again.

src/frontend/scss/components/_header.scss
```diff
@@ -208,6 +208,18 @@
...
+    @if r-theme-val(topbar, item-hover-color) {
+      --r--menu--item--hover--color: #{r-theme-val(topbar, item-hover-color)};
+    }
+    // WARNING: Those are currently some variant samples during development
+    &--primary {
+      --r--menu--item--hover--color: #{$primary};
+    }
+    &--warning {
+      --r--menu--item--hover--color: #{$warning};
+    }
```
jbpenrath, line 221: Can be remove

jbpenrath (commented):
Can you just add an entry in the UPGRADE.md please

src/frontend/scss/components/_header.scss
sveetch: Done

PR #2498, opened 2024-09-19
🔍(backend) fix duplicated h1 in blogpost detail page

jbpenrath:
## Purpose

On blogpost detail page, there is currently a SEO major issue. Indeed, there are
 two H1 on the same page. The first is the news label on the subheader and the
 second is the blogpost title itself. The post title is the most relevant one
 and the subheader "News" is not really relevant that's why we decide to fix
 this issue to simply remove the subheader on the blogpost detail template.

| Before | After |
|--------|-------|
|image|image|

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,11 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unrealeased]
 
+### Changed
+
+- Remove News subheader in blogpost detail page
+
+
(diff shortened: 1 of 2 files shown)


jonathanreveille (approved):

jonathanreveille:
Maybe add an entry into the CHANGELOG, and it looks all good to me.

kernicPanel (approved):

kernicPanel (approved):

PR #2499, opened 2024-09-20
💄(plugins) added 'large-banner' plugin template variant 'compacted-banner'

sveetch:
## Purpose

It was required to have a compacted block available for page contents and that allowed to insert a title with the search bar and some buttons.

Mockup for expected display was this:

Screenshot_20240920_161346

## Proposal

Since expected block contents was nearly identical to the ones from plugin 'large-banner', we just implemented this compacted block with existing a new template variant (and related CSS) for this plugin.

This new feature includes:

- New template variant 'Compact' for plugin 'large-banner';
- New CSS part in Sass source to implement variant layout;

## Notices

As required, plugin image and logo are ignored from this template because the compacted form is expected to be simple.

The variant 'hero-intro' was styling a specific class `hero-intro__title` that was required to append in HTML content to work, so existing content may have this and since of BEM scope it would still work inside compacted variant but this is unwanted since it is not very compact.

So the CSS class `.compacted-banner` include a rule `.hero-intro__title` to disable it, but this can not work with the text font-size. I recommend we assume that writers will be made aware about it for migrating existing banner to the compacted variant and will just remove it and so we can just remove it from the CSS. For compatibility there is also a rule `.compacter-banner__title` that is identical to `.hero-intro__title`, we may probably remove it also.


src/frontend/scss/components/templates/richie/large_banner/_large_banner.scss
```diff
@@ -223,3 +223,153 @@
...
   }
 }
+
+// A compact banner without image contents
+// Opposed to 'hero-intro' this variant delegates the top padding for topbar in 'over'
+// mode in content component
+//
+.compacted-banner {
+  position: relative;
+  padding: 0;
```
jbpenrath, line 233: What do you think to move this add into a dedicated file ? `large_banner/_compacter_banner.scss` ?

src/richie/plugins/large_banner/defaults.py
```diff
@@ -11,5 +11,6 @@
     [
         ("richie/large_banner/large_banner.html", _("Default")),
         ("richie/large_banner/hero-intro.html", _("Hero introduction")),
+        ("richie/large_banner/compacted.html", _("Compacted")),
```
jbpenrath, line 14:
```suggestion
        ("richie/large_banner/compacted.html", _("Compacted introduction")),
```

src/frontend/scss/colors/_theme.scss
```diff
@@ -128,6 +128,18 @@ $r-theme: (
...
+    title-alt-color: r-color('firebrick6'),
+    content-color: r-color('black'),
+    search-input-background: rgba(r-color('white'), 0.7),
+    search-btn-background: r-color('firebrick6'),
+    search-icon-fill: r-color('white'),
+    cta-variant-from: rgba(r-color('white'), 0.8),
+    cta-variant-to: rgba(r-color('white'), 0.8),
+    cta-color: r-color('firebrick6'),
+    cta-border: r-color('white'),
+  ),
```
jbpenrath, line 142: Can you add an entry in the UPGRADE.md, to remind to override `compacted-banner` theme for site which overrides `hero-intro` ?

src/richie/plugins/large_banner/defaults.py
sveetch: Done

src/frontend/scss/components/templates/richie/large_banner/_large_banner.scss
sveetch: Done

src/frontend/scss/colors/_theme.scss
sveetch: Done

UPGRADE.md
```diff
@@ -16,6 +16,24 @@ $ make migrate
 
 ## Unreleased
 
+## 2.29.2 to 2.29.3
```
jbpenrath, line 19:
```suggestion
```

jbpenrath (approved):

UPGRADE.md
sveetch: Done

PR #2508, opened 2024-10-08
🐛(frontend) fix logic to know if user has purchased an ongoing product

jbpenrath:
## Purpose

The logic to know if a user is enrolled to an ongoing product was wrong and it could lead to display that the user is enrolled when not...


src/frontend/js/widgets/SyllabusCourseRunsList/components/CourseRunItemWithEnrollment/index.product.spec.tsx
```diff
@@ -73,6 +73,39 @@ describe("CourseRunItemWithEnrollment for joanie's product's course run", () =>
...
+      mockPaginatedResponse([], 0, false),
+    );
+
+    render(<CourseRunItemWithEnrollment item={courseRun} />, {
+      wrapper: BaseJoanieAppWrapper,
+      queryOptions: { client: createTestQueryClient({ user }) },
+    });
+    await expectNoSpinner();
+
+    // Only dates should have been displayed.
```
lunika, line 87:
```suggestion
    // Only dates should be displayed.
```

lunika (approved):

PR #2513, opened 2024-10-12
✨(plugins) added grid columns and gutter option to section plugin

sveetch:
## Purpose

It has been required a feature to allow displaying Section plugin contents as a grid with a choice for predefined sizes.

## Proposal

This pull request implements the grid with two new options on the Section plugin:

- Grid: This is a choice list where you can select a set of predefined column sizes;
- Grid gutter: A boolean to apply a gutter between grid columns and rows.

This grid layout part is implemented with CSS Grid for its ability and efficiency to manage real grid.

### Grid columns

The choices of predefined column sizes is defined from settings and implement the same ones as with some page templates for multiple columns:

- `None` where grid is disabled and apply the default current behavior with flexbox  where items can define their own size;
- `33% 33% 33%` where contents will be equally distributed on three columns per row;
- `50% 50%` where contents will be equally distributed on two columns per row;
- `25% 75%` where contents are distributed on two columns, the first column takes 25% of width size and the second one takes 75%;
- `75% 25%` where contents are distributed on two columns, the first column takes 25% of width size and the second one takes 75%;

### Grid gutter

The gutter size can not be choosen from plugin options. It can only be enabled (default) or disabled.

I'm not sure about this option, i did it because the gutter mode need to neutralize some content object rules but i can see a case where it would be better to let content object to define their own margin. What do you think ?

## Todo

- [x] Prototype grid options on section plugin templates;
- [x] Add grid options to the Section plugin model;
- [x] Add grid rules to ".section" CSS component and update its related template;
- [x] Add grid rules to ".section-tiles" CSS component and update its related template;
- [x] Remove every temporary prototyping stuff in sources;

## Screenshots

Screenshots of samples on the About page from demo, all section are the same in all available grid choice and with yellow background for section using the grid and pale white for the one in the current default behavior without grid. The background is temporary only to show the topography.

In each section there is many various plugins,  respectively Organization, Course, Organization (yet again), Link, Simple text, Plain text and Post. 

As you will notice, the Simple text is divided in two column since it is two paragraphs so the grid assume them as two items, on the contrary the plain text is assumed a single item (would be different with some whitespaces).

### Large screen

desktop_layout

### Mobile screen

mobile_layout


sandroscosta:
I was thinking about implementing something like this for a bit more personalization. This is top notch. Great work @sveetch

Makefile
```diff
@@ -347,6 +347,24 @@ data/static/.keep:
...
+	@$(COMPOSE_UP_WAIT) -d ${DB_HOST}
+	@$(MANAGE) check
+.PHONY: check
+
+check-migrations: ## WARNING: Temporary while prototyping
+	@echo "Start and wait for ${DB_HOST} to be up..."
+	@$(COMPOSE_UP_WAIT) -d ${DB_HOST}
+	@$(MANAGE) makemigrations --check --dry-run -v 3 section
+.PHONY: check
+
```
jbpenrath, line 367: IMO, we could keep `migrations` command at least, it can be useful.

sveetch: As seen with you, tasks `migrations` and `check-migrations` is part of this pull request, i've cleaned the `check` one

jbpenrath (approved):
@NamFra can you just check that available options about columns width is ok ?

CHANGELOG.md
```diff
@@ -8,6 +8,12 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unrealeased]
 
+### Added
+
+- Grid options for the Section plugin
+- Add 'check-migrations' task to check pending migrations to create
+- Add 'migrations' task to create all pending migrations
```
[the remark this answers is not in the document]
sveetch: Done

PR #2525, opened 2024-10-29
✨(frontend) display checkbox to waive withdrawal right if needed

jbpenrath:
## Purpose

We want to prevent a user to subscribe to a training without waiving withdrawal right if the training will start before the end of the withdrawal period. So according the property `has_withdrawal_period` of a product, we ask to user to check a box to explicitly waive its withdrawal right.

CleanShot 2024-10-29 at 17 46 03@2x

CleanShot 2024-10-29 at 17 46 28@2x

### DIFF
--- a/src/frontend/js/components/SaleTunnel/WithdrawRightCheckbox/index.tsx
+++ b/src/frontend/js/components/SaleTunnel/WithdrawRightCheckbox/index.tsx
@@ -1,0 +1,59 @@
+import { Alert, Checkbox, VariantType } from '@openfun/cunningham-react';
+import { useCallback, useEffect, useState } from 'react';
+import { defineMessages, FormattedMessage } from 'react-intl/lib';
+import { useSaleTunnelContext } from 'components/SaleTunnel/GenericSaleTunnel';
+
+const messages = defineMessages({
+  waiveCheckboxExplanation: {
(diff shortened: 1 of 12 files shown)


wilbrdt (approved):
LGTM!

src/frontend/js/components/SaleTunnel/WithdrawRightCheckbox/index.tsx
```diff
@@ -0,0 +1,60 @@
+import { Alert, Checkbox, VariantType } from '@openfun/cunningham-react';
+import { useCallback, useEffect, useState } from 'react';
+import { defineMessages, FormattedMessage } from 'react-intl/lib';
+import { useSaleTunnelContext } from 'components/SaleTunnel/GenericSaleTunnel';
+
+const messages = defineMessages({
+  waiveCheckboxExplanation: {
+    defaultMessage:
+      'This training will start before the end of your withdrawal period. You must waive your right of withdrawal to subscribe.',
```
kernicPanel, line 9: Maybe we could simplify a bit this explanation:
```suggestion
      'This training will start before the end of your withdrawal period. You must waive it to subscribe.',
```

kernicPanel (approved):

PR #2534, opened 2024-11-14
✨(frontend) change 'CKEditorPlugin' for new field `variant`

sveetch:
## Purpose

This pull request introduces a new feature for the included plugin `CKEditorPlugin` ("Simple text" from the plugin selector) to allow for variant styling.

## Proposal

Now `CKEditorPlugin` will have a new field `variant` that is a choice list of variant CSS classnames.

This new field allows to choose a variant style for the 'simple-text' component generated by `CKEditorPlugin`.

Variants are defined from Sass component source `_simpletext.scss` and enabled from setting `RICHIE_SIMPLETEXT_VARIANTS`.

On default, no variant is enabled and `CKEditorPlugin` is still in the same old behavior without any `div` element around content to maintain compatibility with already existing contents.

When a choice is selected, the content is now enclosed within a 'div' element with the related variant classes.

Currently the few shipped default variants make a rounded box with their own colorization because it was the main required goal.

Also, note that `CKEditorPlugin` has been added to the common allowed plugins from helper function `richie_placeholder_conf` because this feature was required for both columned templates (that are driven by this helper function) and section plugin.

### Desktop

desktop

### Mobile

mobile


src/richie/plugins/simple_text_ckeditor/defaults.py
```diff
@@ -1,5 +1,17 @@
...
+SIMPLETEXT_VARIANTS = getattr(
+    settings,
+    "RICHIE_SIMPLETEXT_VARIANTS",
+    [
+        ("", _("None")),
+        ("undefined", _("Undefined variant")),
+        ("blue-on-white", _("Blue on white")),
+        ("white-on-blue", _("White on blue")),
+    ],
+)
```
jbpenrath, line 17: This comment is related to my overall comment.

Make variant more generic
```suggestion
SIMPLETEXT_VARIANTS = getattr(
    settings,
    "RICHIE_SIMPLETEXT_VARIANTS",
    [
        ("", _("None")),
        ("undefined", _("Box")),
        ("fulfill", _("Fulfilled box")),
        ("stroked", _("Stroked box")),
    ],
)
```

src/frontend/scss/components/templates/richie/simpletext/_simpletext.scss
```diff
@@ -0,0 +1,44 @@
...
+    h6 {
+      color: $primary;
+    }
+  }
+
+  &__variant-white-on-blue {
+    color: $white;
+    border-color: $primary;
+    background: $primary;
+  }
```
jbpenrath, line 43: This comment is related to my overall comment.

```suggestion
  &__variant-stroked {
    border-color: r-theme-val(simpletext-variant-box, primary-color);
    background: r-theme-val(simpletext-variant-box, secondary-color);

    h1,
    h2,
    h3,
    h4,
    h5,
    h6 {
      color: r-theme-val(simpletext-variant-box, primary-color);
    }
  }

  &__variant-fulfilled {
    color: r-theme-val(simpletext-variant-box, secondary-color);
    border-color: r-theme-val(simpletext-variant-box, primary-color);
    background: r-theme-val(simpletext-variant-box, primary-color);
  }
```

jbpenrath (commented):
I feel the way you create variant is a little bit too specific so it will be "hard" to override it in our site factory.

IMO, you could keep your approach but just make variant more generic then create a theme palette for the box variant. In this way, we still have the possibility to create new variant but if we just want to change color of those existing variants it will be easy.

jbpenrath (approved):

PR #2543, opened 2024-11-25
✨(plugins) added Slider plugin

sveetch:
## Purpose

This pull request implement a new plugin to contain multiple content that can be slided.

Each content would include a required title, a required image, an optional rich content (with CKEditor) and an optional link (with an option to open it in a blank window).

## Proposal

The Slider is done through two plugins:

- The SliderPlugin that is just a container to include slide items;
- The SliderItemPlugin for a slide item.

This has been done using parent/children options from CMS plugin, the container only accepts SliderItemPlugin as children and item plugin can only be included in the SliderPlugin.

Item image is published in multiple sizes for responsiveness (using the same sizes than the 'Large banner' plugin).

This commit only include the backend, CSS and template implementation.

There is still a JavaScript component to develop so currently the slider is static, only the current slide is showed (but the other ones are still properly generated in HTML).

### Plugin in page structure

slider_structure

### Item plugin form

slider_item_admin

### Sample front

slider


src/richie/plugins/slider/factories.py
```diff
@@ -0,0 +1,41 @@
...
+    class Meta:
+        model = SlideItem
+
+    @factory.lazy_attribute
+    def link_url(self):
+        """
+        Set a random url
+        """
+        instance = faker.Faker()
+        return instance.url()
```
jbpenrath, line 41: This doesn't work ?

```suggestion
    link_url = factory.Faker("url")

    class Meta:
        model = SlideItem
```

src/richie/plugins/slider/models.py
```diff
@@ -0,0 +1,68 @@
...
+
+class Slider(CMSPlugin):
+    """
+    Slide container plugin for Slide item plugins.
+    """
+
+    title = models.CharField(_("title"), max_length=255)
+
+    def __str__(self):
+        return Truncator(self.title).words(6, truncate="...")
```
jbpenrath, line 21: Why do you need to truncate the str representation ?

src/richie/plugins/slider/models.py
```diff
@@ -0,0 +1,68 @@
...
+        help_text=_("Make the slide as a link with an URL."),
+    )
+    link_open_blank = models.BooleanField(
+        _("open new window"),
+        default=False,
+        help_text=_("If checked the link will be open in a new window"),
+    )
+
+    def __str__(self):
+        return Truncator(self.title).words(6, truncate="...")
```
jbpenrath, line 64: Ditto

jbpenrath (commented):
As we will display the slider from a react component, there is no need to create a html template with a slider structure. Instead you should render a div with a proper `id` and a `data-props` attribute that will be used by React then.

https://github.com/openfun/richie/blob/1095f78401e5200f21eddccbf5dd624fc2ee46fe/docs/django-react-interop.md?plain=1#L41

src/richie/plugins/slider/factories.py
sveetch: It was an workaround for an old factoryboy bug i forgotten it was fixed long ago. Done.

src/richie/plugins/slider/models.py
sveetch: Done

src/richie/plugins/slider/models.py
sveetch: Done

sveetch:
I've made an update to remove the template from HTML, now it only include the slider data through a payload that can be loaded from JavaScript frontend.

Since the slider payload can be huge it is holded in a `<script>` element with an unique ID and linked the slider container element to it with attribute `data-props-source` (alike in the cms menu template).

src/richie/plugins/slider/models.py
```diff
@@ -0,0 +1,111 @@
...
+            "slides": [
+                {
+                    "pk": plugin.id,
+                    "title": plugin.title,
+                    "image": plugin.image.url,
+                    "content": plugin.content,
+                    "link_url": plugin.link_url,
+                    "link_open_blank": plugin.link_open_blank,
+                }
+                for plugin in slides
```
jbpenrath, line 66: Is the slide order guaranteed ? Otherwise could we compute a position property?

sveetch: Here the slides order has already been from the cms and the order is defined from the plugin items in the page structure, so yes the order is guaranteed.

jbpenrath (approved):

PR #2547, opened 2024-11-27
💬 Update withdrawal texts then release 2.32.0

jbpenrath:
## Purpose

Update existing messages to clarify the conditions under which the right
of withdrawal is waived for certificate product then prepare a minor release.


NamFra (commented):
GTM! Thanks @jbpenrath :rocket:

PR #2548, opened 2024-11-28
Unpin/djangocms link

jbpenrath:
## Purpose

Django CMS Link 5 has been released and introduces some breaking change. We have to update our demo site script to work with this new version. (This version has been already used on our preproduction and production environment and related migrations have been applied so we have no other choice to upgrade...)

~Furthermore currently some packages are not pinned and this may lead to introduce bugs
during deployment that we are not able to detect during development phase
that is really weird.~ -> Will be addressed in a dedicated PR

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 - Add Slider plugin
 
+### Fixed
+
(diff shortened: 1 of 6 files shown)


lunika:
This repo is a lib, is it a good idea to pin all dependencies ? I think it should be better to set a range of compatible versions instead.

kernicPanel (approved):

PR #2551, opened 2024-12-02
📌(backend) pin all packages

jbpenrath:
## Purpose

Currently some packages are not pinned and this may lead to introduce bugs during deployment that we are not able to detect during development phase that is really weird. As richie is a library package, we cannot pin exactly all deps but to prevent to deploy silently packages containing breaking changes, we use a stragegy to install only minor and patch versions of all packages. Major versions will be upgraded by renovate.


jbpenrath:
@igobranco, @sandroscosta please review this PR to be sure you are aligned with this new strategy to manage dependency versions and mainly ensure that it will not breaks anything from our side.

igobranco:
@jbpenrath no problem for us. We have been just copying the upstream dependencies from cookiecutter or seeing what you have been put in yours [funmooc](https://github.com/openfun/fun-richie-site-factory/tree/main/sites/funmooc/requirements) site.

igobranco (approved):

kernicPanel (approved):

sandroscosta:
Same response from as as @igobranco. Everything that we're able to do, we'll do it upstream. It still makes sense. As for upgrades, if this help better manage upgrade on richie-site-factory, it's a great idea.

sandroscosta (approved):

claudusd (approved):

PR #2561, opened 2024-12-16
🐛(backend) skip hidden course runs from state calculation,

Tiago-Salles:
- On the state property skip the hidden ones


igobranco (approved):
LGTM ✔️ 

@jbpenrath this looks good for me, but I think someone from FUN should also review this PR before merging.

tests/apps/courses/test_templates_category_detail.py
```diff
@@ -715,3 +718,46 @@ def test_template_category_detail_additional_info_page_id(self):
 
         self.assertEqual(response.status_code, 200)
         self.assertNotContains(response, message)
+
+    def test_templates_category_detail_cms_published_hidden_courses(self):
+        """
+        Ensures that changing the course run `catalog_visibility` paratemer
+        will prevent it from showing on the category detail page
+        """
```
jbpenrath, line 726:
```suggestion
        """
        Ensures that changing the course run `catalog_visibility` to `hidden`                                 
        parameter will prevent it from being take in account to compute
        the course state on the category detail page
        """
```

tests/apps/courses/test_templates_category_detail.py
jbpenrath, line 1: Why you don't write a single test on the course state method to ensure that hidden course run are exluded to the state computation?

Tiago-Salles: Seems appropriate, I will consider it.

Tiago-Salles:
**WARNING**:exclamation:

This branch has a `fixup` commit, it must be marged with **Squash and merge**.

jbpenrath (approved):
⚠️ You should not squash and merge. You should squash our two commits locally then push force on your branch

Tiago-Salles:
Yes, I just realized even being squashed into one commit, which is our interest, **Squash and merge** is not the correct approach regarding the `fixup` resource.

PR #2567, opened 2025-01-23
♿️(search) Add label to the search input

sandroscosta:
Add a label to the `SearchInput` component to improve accessibility.

## Purpose

As of our latest accessibility review, we needed to add a `label` to all search fields on Richie.
This PR adds a label to the `SearchInput` input field.


sandroscosta:
@jbpenrath I changed the tests to fetch the input value by `testId`, because now we have 2 "Search" texts being rendered in this component.

src/frontend/js/components/SearchInput/index.spec.tsx
```diff
@@ -11,13 +11,13 @@ describe('<SearchInput />', () => {
...
   it('renders with the input field and button', () => {
-    const { container, getByText } = render(
+    const { container, getByTestId } = render(
       <IntlProvider locale="en">
         <SearchInput context={contextProps} inputProps={inputProps} />
       </IntlProvider>,
     );
 
-    getByText('Search');
+    getByTestId('search-input-btn');
```
jbpenrath, line 20: As the button is now accessible, it could be relevant to try to find it though its accessible role and name ? WDYT?

`getByRole('button', { name: 'Search' })`

sandroscosta: It's indeed much cleaner that my approach.

src/frontend/js/components/SearchInput/index.spec.tsx
```diff
@@ -10,14 +10,15 @@ describe('<SearchInput />', () => {
...
+  it('renders with the input field, label and button', () => {
+    const { container, getByRole, getByLabelText } = render(
       <IntlProvider locale="en">
         <SearchInput context={contextProps} inputProps={inputProps} />
       </IntlProvider>,
     );
 
-    getByText('Search');
+    getByRole('button', { name: 'Search' });
+    getByLabelText('Search');
```
jbpenrath, line 21: IMO the `getByRole` is enough

jbpenrath (approved):
LGTM ✅

PR #2580, opened 2025-02-19
🔥(base) removed storage and cache from site-factory

Tiago-Salles:
This PR removes storage and cache from site-factory and apply the responsability of these resources to Richie .

# Purpose

In order to avoid repeating unnecessary and critical implementations for each instance, the `storage.py` and `cache.py` was replaced from the site-factory to Richie. This change descreases the complexity of maintaing a new site-factory, since these resources most of time are not modified, but need to be maintained.

It does not mean a new instance can't implement by its way, if any customization is necessary, it gets up to the site-factory implementation, otherwise, just leave it to Richie to handle it.

# Validating the changes

To test install the `richie` package from this branch
```
git+https://github.com/Tiago-Salles/richie@Tiago-Salles/issues/296-upstream-storage-and-cache#egg=richie
```

make all the process of generating a site-factory and then execute it, you will be able to confirm it works normally as expected.


CHANGELOG.md
```diff
@@ -11,6 +11,10 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 ### Changed
 
 - BREAKING CHANGE - Upgrade to React 19
+- Removed storage and cache from site-factory and applied to Richie
+in order to avoid repeating unnecessary and critical implementations
+for each instance. If any customization is necessary, it is up to the
+site-factory implementation. Otherwise, as expected, it keeps up to Richie.
```
igobranco, line 17: Review format. I think you should add 2 white spaces before each additional line.

cookiecutter/{{cookiecutter.organization}}-richie-site-factory/template/{{cookiecutter.site}}/src/backend/base/tests/test_cache.py
```diff
@@ -49,7 +49,7 @@ def test_get_redis_cache(self, redis_cache_mock, fallback_cache_mock):
         redis_cache_mock.assert_called_once()
         fallback_cache_mock.assert_not_called()
 
-    @mock.patch("base.cache.logger")
+    @mock.patch("richie.base.cache.logger")
```
igobranco, line 52: Why the test_cache.py wasn't also moved to Richie?

igobranco (changes requested):
@Tiago-Salles 
Add a comment to UPGRADE.md file. So operators/developer of Richie site factories know that they should have special attention to the cache configuration for the next upgrade.

cookiecutter/{{cookiecutter.organization}}-richie-site-factory/template/{{cookiecutter.site}}/src/backend/base/tests/test_cache.py
Tiago-Salles: The first moment I thought about it, I concluded it was not necessary, and after making the changes, I did not repeat the necessary considerations to make here. Yes, the tests were also replaced to be together with their respectives files.

igobranco (approved):

igobranco:
@jbpenrath can you review this PR?

jbpenrath (commented):
What do you think to move `base` files into the richie core app?

jbpenrath:
> > What do you think to move `base` files into the richie core app?
> 
> It seems better replaced, but since core already has a cache structure, and this change aims to the site-factory resources, was created a folder called `site_factory_base` where it is explicit the reason those files exist.

Sorry but IMO this is not a good idea to put the concept of site factory here. We should see those cache and storage classes as optional configuration stuff. We decide to put them in the core application as they are really useful in our site factories. I don't see any reason to not put that in the core application even if they are not used at a first place

Tiago-Salles:
> > > What do you think to move `base` files into the richie core app?
> > 
> > 
> > It seems better replaced, but since core already has a cache structure, and this change aims to the site-factory resources, was created a folder called `site_factory_base` where it is explicit the reason those files exist.
> 
> Sorry but IMO this is not a good idea to put the concept of site factory here. We should see those cache and storage classes as optional configuration stuff. We decide to put them in the core application as they are really useful in all of our site factories so it makes sens to gather this code. I don't see any reason to not put that in the core application even if they are not used at a first place. WDYT?

Yes, I see the concern about not mixing concepts here. Regarding it, do you beliave the best approach is to merge the two `cache.py` files making the code to be served from the same context? It seems fine to me, since the context is about cache it respects the single resposibility.

jbpenrath:
> > > > What do you think to move `base` files into the richie core app?
> > > 
> > > 
> > > It seems better replaced, but since core already has a cache structure, and this change aims to the site-factory resources, was created a folder called `site_factory_base` where it is explicit the reason those files exist.
> > 
> > 
> > Sorry but IMO this is not a good idea to put the concept of site factory here. We should see those cache and storage classes as optional configuration stuff. We decide to put them in the core application as they are really useful in all of our site factories so it makes sens to gather this code. I don't see any reason to not put that in the core application even if they are not used at a first place. WDYT?
> 
> Yes, I see the concern about not mixing concepts here. Regarding it, do you beliave the best approach is to merge the two `cache.py` files making the code to be served from the same context? It seems fine to me, since the context is about cache it respects the single resposibility.

It looks fine for me. If you take a look into the core/cache module, you will see that there is a cache client and a cache middleware so this module already don't have a single reponsability. Then if the cache.py module starts to be too large, we could split it into chunks.

jbpenrath (approved):

PR #2590, opened 2025-03-13
🐛(backend) fix header_menu template

jbpenrath:
## Purpose

Since dropdown menu, there are several case where we want to apply `topbar__item--selected` class to a `topbar__item`. First the current child has property `selected` sets to `True` then when the current child has been configured to redirect to another page and the current page is the target of this redirect. But in some case, the page display has an empty path and in this case the current behavior applied the selected modified class. So in order to fix that we need to ensure that a redirect url is sets before comparing it to the current page url.


src/richie/apps/core/templates/menu/header_menu.html
```diff
@@ -2,43 +2,38 @@
...
+                {% if child.sibling %} topbar__item--sibling{% endif %}
+                {% if child.descendant %} topbar__item--descendant{% endif %}
+                {% if child.selected %} topbar__item--selected
+                {% elif redirect_url and page_url == redirect_url %} topbar__item--selected
+                {% endif %}"
+            >
+                {% comment %}Dropdown menu for children are only for page with index page
+                extension with a specific option enabled{% endcomment %}
+                {% if menu_options.allow_submenu and child.children %}
+                        <button aria-expanded="false" aria-haspopup="true" aria-controls="dropdown-{{ child.id }}" id="dropdown-button-{{ child.id }}">
```
kernicPanel, line 19: FMI, why removing line breaks here ?

kernicPanel (approved):

src/richie/apps/core/templates/menu/header_menu.html
jbpenrath: Just my IDE that reformat the file after I changed indentation and the readability was poor so I put this tag inline but indeed I could revert this.

Tiago-Salles (approved):
LGTM!

src/richie/apps/core/templates/menu/header_menu.html
kernicPanel: No need to revert :+1:

jbpenrath: Too late 😄

PR #2598, opened 2025-03-19
🐛(frontend) fix edge case with topbar dropdown menu

jbpenrath:
## Purpose

In some case, the dropdown menu does not need to be full width
but it has not enough remaining space on its right to be displayed
properly. So we manage this case and align the menu on left in those case.

https://github.com/user-attachments/assets/e6884e15-afab-4c26-8a2b-58a4564da8a8

### DIFF
--- a/src/frontend/scss/components/_header.scss
+++ b/src/frontend/scss/components/_header.scss
@@ -241,12 +241,8 @@
           display: none;
           position: absolute;
           transform: translateX(-8px);
+          width: max-content;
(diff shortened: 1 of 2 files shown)


jonathanreveille (approved):
Just a question, will it work on a smartphone as well ?

jbpenrath:
> Just a question, will it work on a smartphone as well ?

Yes of course. Actually on mobile, there is a burger menu so this fix will not be used on this platform

PR #2621, opened 2025-04-16
🐛(video_player) fix lazy load video player for programs

igobranco:
When showing a video player on programs with the
`RICHIE_VIDEO_PLUGIN_LAZY_LOADING` setting enabled it was raising a Javascript error.
This change makes the element swap more resilient for the both cases courses and programs page.


Tiago-Salles (approved):
LGTM

kernicPanel (approved):

PR #2631, opened 2025-05-16
✨(backend) exclude non listed courses from list

### LINKED ISSUE #2630: Enable bulk unlisting of archived courses across CMS and search pages
## Feature Request

We need the ability to bulk-hide a number of catalog pages because their latest course sessions have been archived for a long time.
While a feature to hide courses already exists, it must be activated page by page and only affects course glimpses in the search, not CMS pages.

The simplest approach would be to use the `is_listed` flag, introduced here https://github.com/openfun/richie/pull/1177 on the course model.
Given the number of courses involved, being able to update them all at once via a script would be ideal 😇.

However, while this flag filters courses out of Elasticsearch, they still appear on CMS pages (organizations, categories, persons, etc.). If possible, we'd like this feature to be extended to all Richie pages so that course glimpses are hidden globally when `is_listed` is set to `false`.

If using `is_listed` to hide course glimpses across all pages turns out to be problematic (possibly for you, @igobranco?), @KernicPanel suggested introducing a dedicated attribute such as `is_hidden` instead.

- JB: If it's just a matter of applying a bulk update with well-defined criteria, we could run a one-shot command in production along with a sync, or even set up a cron job to make the initiative sustainable (e.g., automatically hide any course that hasn’t run in over X time).
    - A list of affected syllabus pages will be provided for the update 🙏.
- Importantly, these pages must **not** be unpublished, to avoid 404 errors and negative SEO impact.
- For a medium-term improvement (assuming users still access these outdated pages), 2 possible paths have been discussed:
    - finalize the wishlist feature already implemented on the backend
    - or work on a landing page solution: https://github.com/openfun/richie/issues/1446
- This request could also be linked to the following CMS improvement issue: https://github.com/openfun/richie/issues/1834

kernicPanel:
## Purpose

When a course is hidden from the search results, it should also be hidden from lists.

Fixes #2630

## Proposal

Add a filter `is_listed=True` to the models get_courses queries.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unrealeased]
 
+### Changed
+
+- Hide courses that are not `is_listed` in public pages
+
 ## [3.1.2] - 2025-05-22
 
 ### Added
--- a/cookiecutter/{{cookiecutter.organization}}-richie-site-factory/template/{{cookiecutter.site}}/src/backend/{{cookiecutter.site}}/settings.py
+++ b/cookiecutter/{{cookiecutter.organization}}-richie-site-factory/template/{{cookiecutter.site}}/src/backend/{{cookiecutter.site}}/settings.py
@@ -545,7 +545,12 @@ class Base(StyleguideMixin, DRFMixin, RichieCoursesConfigurationMixin, Configura
                 "level": "ERROR",
                 "handlers": ["console"],
                 "propagate": False,
-            }
+            },
+            "richie": {
+                "level": values.Value("ERROR", environ_name="RICHIE_LOGGING_LEVEL"),
+                "handlers": ["console"],
+                "propagate": False,
+            },
         },
     }
 
--- a/src/richie/apps/courses/models/category.py
+++ b/src/richie/apps/courses/models/category.py
@@ -71,7 +71,10 @@ class Category(EsIdMixin, BasePageExtension):
         """
         return self.get_reverse_related_page_extensions(
             "course", language=language, include_descendants=include_descendants
-        ).filter(extended_object__node__parent__cms_pages__course__isnull=True)
+        ).filter(
+            extended_object__node__parent__cms_pages__course__isnull=True,
+            is_listed=True,
+        )
 
     def get_blogposts(self, language=None, include_descendants=True):
         """
--- a/src/richie/apps/courses/models/course.py
+++ b/src/richie/apps/courses/models/course.py
@@ -180,7 +180,7 @@ class Course(EsIdMixin, BasePageExtension):
     is_listed = models.BooleanField(
         default=True,
         verbose_name=_("is listed"),
-        help_text=_("Tick if you want the course to be visible on the search page."),
+        help_text=_("Tick if you want the course to be visible."),
     )
 
     is_self_paced = models.BooleanField(
--- a/src/richie/apps/courses/models/organization.py
+++ b/src/richie/apps/courses/models/organization.py
@@ -147,7 +147,10 @@ class Organization(EsIdMixin, BasePageExtension):
         """
         return self.get_reverse_related_page_extensions(
             "course", language=language
-        ).filter(extended_object__node__parent__cms_pages__course__isnull=True)
+        ).filter(
+            extended_object__node__parent__cms_pages__course__isnull=True,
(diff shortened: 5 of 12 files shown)


kernicPanel:
> IMO you should also hide from the Programs page.
> 
> Change this line:
> 
> https://github.com/openfun/richie/blob/56e1b5b25d3b677617500d81bc95be4dd9b32e49/src/richie/apps/courses/templates/courses/cms/program_detail.html#L194
> 
> To use the model based querying like organizations, to use more an Object Oriented code use.
> 
> https://github.com/openfun/richie/blob/56e1b5b25d3b677617500d81bc95be4dd9b32e49/src/richie/apps/courses/templates/courses/cms/organization_detail.html#L135
> 
> https://github.com/openfun/richie/blob/56e1b5b25d3b677617500d81bc95be4dd9b32e49/src/richie/apps/courses/models/organization.py#L143
> 
> So every model can have courses associated, having the `get_courses` function defined.
> 
> Don't forget to include a line on the CHANGELOG.md.

Thanks for your feedback!
Meanwhile, @NamFra and I discussed, and we decided to use a new attribute `is_hidden` , so your usage will still be working.

igobranco:
> Thanks for your feedback! Meanwhile, @NamFra and I discussed, and we decided to use a new attribute `is_hidden` , so your usage will still be working.

My problem is that we don't know if we want to show them on that page.... I only found out that they were being shown on that page today. Personally I would prefer a simpler implementation a flag that hides the page everywhere. But on our not technical team could be other visions...

kernicPanel:
> > Thanks for your feedback! Meanwhile, @NamFra and I discussed, and we decided to use a new attribute `is_hidden` , so your usage will still be working.
> 
> My problem is that we don't know if we want to show them on that page.... I only found out that they were being shown on that page today. Personally I would prefer a simpler implementation a flag that hides the page everywhere. But on our not technical team could be other visions...

OK, as I'm currently adding `is_hidden` on the search feature, I'll pause until we get all point of views.
Meanwhile, I'll go back to the unique attribute, and filter the program page, as you pointed out.

kernicPanel:
@igobranco @Tiago-Salles can you have a look at the program page please ?
I tried to follow your instructions (sorry I'm not familiar with django cms…) but I believe I've done something wrong, as I don't see how to manage the program's course anymore.

igobranco:
Hi @kernicPanel,
We, @Tiago-Salles and @sandroscosta , talked during our morning DEV meeting and we think that this code resolve your question:
```html
{% if current_page.publisher_is_draft or not current_page|is_empty_placeholder:"program_courses" %}
        {% if request.toolbar.edit_mode_active or not request.toolbar.edit_mode_active and current_page.program.get_courses|length > 0 %}
            <div class="program-detail__courses program-detail__block">
                <div class="program-detail__row">
                    <section class="course-glimpse-list">
                        <div class="course-glimpse-list__header">
                            <h2 class="program-detail__title">{% trans "Related courses" %}</h2>
                        </div>
                        <div class="course-glimpse-list__content">
                            {% if request.toolbar.edit_mode_active %}
                                {% placeholder "program_courses" %}
                            {% else %}
                                {% for course in current_page.program.get_courses %}
                                    {% include "courses/cms/fragment_course_glimpse.html" with course=course header_level=3 %}
                                {% endfor %}
                            {% endif %}
                        </div>
                    </section>
                </div>
            </div>
        {% endif %}
    {% endif %}
```

Example: 1 program with 4 courses, one of them is not listed. On view mode shows 3 courses, on edit mode shows 4 courses.

kernicPanel:
Hello @igobranco , @Tiago-Salles 
Could you to review this PR please ?

igobranco:
@kernicPanel IMO we need a test to secure the courses not listed aren't shown inside the Program detail page. I added this test on a new commit directly on your branch.
`test_templates_program_detail_course_is_listed`
If you don't agree, just remove it :smile: 

Fix the CHANGELOG conflict and :rocket:

igobranco (approved):

kernicPanel:
> @kernicPanel IMO we need a test to secure the courses not listed aren't shown inside the Program detail page. I added this test on a new commit directly on your branch. `test_templates_program_detail_course_is_listed` If you don't agree, just remove it 😄
> 
> Fix the CHANGELOG conflict and 🚀

Thanks for the test @igobranco :+1:

Tiago-Salles (approved):
LGTM

PR #2636, opened 2025-05-21
📌(backend) pin easy-thumbnails to 2.9.0 to avoid save JPEG error

wilbrdt:
## Purpose

The `bootstrap_elasticsearch` command is failing due to an error from `easy-thumbnails` when trying to generate thumbnails for the a gif (used in person portraits). The error is: "OSError: cannot write mode P as JPEG".

## Proposal

This is a known issue introduced in `easy-thumbnails` 2.10.0 (see https://github.com/SmileyChris/easy-thumbnails/issues/653). We pin the version to be < 2.10 as a temporary workaround.


kernicPanel (approved):
Nice one!

kernicPanel:
We should add an entry to the changelog, I think.

kernicPanel:
@wilbrdt I pushed a fixup for the changelog, just to avoid conflicts on an upcoming PR.
Feel free to reword it if you want.

wilbrdt:
@kernicPanel you did the right thing!

PR #2640, opened 2025-06-02
Dev/discounts

kernicPanel:
## Purpose

Description...

## Proposal

Description...

- [] item 1...
- [] item 2...


kernicPanel:
Hello @igobranco 
I added you as a reviewer, just to make sure that nothing will bother you. I believe that all I modified wont be used by Nau.

env.d/development/localtunnel.dist
```diff
@@ -0,0 +1,9 @@
+## Locatunnel configuration
+
+LOCALTUNNEL_PORT=8070
+LOCALTUNNEL_SUBDOMAIN=dev-${USER}-richie
```
wilbrdt, line 4: Is this `USER` defined? I could not find it

src/frontend/js/widgets/SyllabusCourseRunsList/components/CourseProductItem/CourseProductItemFooter/index.tsx
```diff
@@ -42,6 +42,7 @@ const CourseProductItemFooter = ({
       <PurchaseButton
         course={course}
         product={courseProductRelation.product as CredentialProduct}
+        courseProductRelation={courseProductRelation}
```
wilbrdt: Have you consider passing the price from here, instead of the relation?
Such as: `{courseProductRelation.discounted_price || courseProductRelation.product.price}`

Why did you go for passing the `courseProductRelation` instead?

wilbrdt (approved):
For what it's worth, looks good to me! :tada:

igobranco:
> Hello @igobranco I added you as a reviewer, just to make sure that nothing will bother you. I believe that all I modified wont be used by Nau.

@kernicPanel Yes, we don't use Joanie, so IMO NAU won't be affected.

env.d/development/localtunnel.dist
kernicPanel: `USER` should come from local environment.
But, I believe this is not really working, as I hardcoded mine in `env.d/development/localtunnel`

src/frontend/js/widgets/SyllabusCourseRunsList/components/CourseProductItem/CourseProductItemFooter/index.tsx
kernicPanel: Yes, the idea here is to match the existing pattern on the component props.
But, I presume I could remove course and product, which are contained in the relation…

wilbrdt: For this purchase button, it looks like we're already extracting all the necessary information from the relation, such as the product, organizations, whether it is withdrawable etc.
I’d suggest either avoiding passing the entire relation further down the line or simply leaving it as is, since the current approach seems OK to me.

PR #2644, opened 2025-06-16
✨(cdn) add specific setting for serving media

wilbrdt:
## Purpose

At FUN, we plan to move away from AWS S3 + Cloudfront.
Static files will be served directly from our servers, while media files will be served from Scaleway S3 + Edge services.
However, the current configuration uses a single domain for both static and media files.

## Proposal

Introducing a new setting `MEDIA_CUSTOM_DOMAIN` to enable serving media files from a separate CDN domain, independent of static files.


kernicPanel (approved):

wilbrdt:
Hi @igobranco, thanks for the review! You're absolutely right, I'll switch the storage backend to `S3Boto3Storage` and configure it directly using the `AWS_` settings!
So in this PR, I've made the `STORAGES` setting configurable, following the approach you used on your side, but I’ve dropped the previously proposed changes. Let me know if this looks good to you, or if there’s anything else you’d like me to adjust!

PR #2655, opened 2025-06-24
🐛(frontend) discounted price in sales tunnel for certificate product

### LINKED ISSUE #2645: Certificate discount not reflected in displayed total price
## Bug Report

**Problematic Behavior**
When a discount is applied to a certificate product, the full price is still displayed in the total during checkout. The user is charged the correct (discounted) price, but the total shown does not reflect it.

**Expected behavior/code**
The total price displayed during checkout should match the discounted price.

**Steps to Reproduce**
1. Create a discount (OfferRule) for a certificate product
2. Open the course page as a user and start the purchase
3. Check the total price shown during checkout: it displays the full price

**Environment**
- Richie version: dev
- Platform: Richie preprod

On the screenshot below, the actual price paid is 550 €, but the total shown is 600 €:

Image

This only affects **certificate products with a discount applied**.

jonathanreveille:
## Purpose

When we open the sales tunnel from the student's dashboard to purchase a certificate in a course that we have enrolled, when a discount is present we would not see the discounted price. Now the sales tunnel information has the data if there is a discounted price on the product certificate. Otherwise, it works has usual.

There is actually two entry points to see the salesTunnel, and we don't use the same endpoints in both.

So my certificate costs normally **600.00.**
Consider that we have applied an OfferRule that takes 10% off the price, we should have **540.00** as last price.

Shot before :
Capture d’écran du 2025-06-25 19-11-49

Shot after : 
Capture d’écran du 2025-06-25 19-11-27

Fixes #2645

## Proposal

- [x] fetch the discounted price for certificate products when opening the sales tunnel information

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,8 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 ### Fixed
 
 - Prevent indexation failure when any thumbnail generation fails
+- Display discounted price for the purchase of a certificate product
+  from student dashboard in sale tunnel informations
 
 ## [3.1.2] - 2025-05-22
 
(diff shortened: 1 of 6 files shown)


kernicPanel (approved):

liamls (approved):
Tests passed, code seems ok for me.

PR #2657, opened 2025-06-24
🐛(backend) corrected ACTIVATED_DB default value

Tiago-Salles:
In [this](https://github.com/openfun/richie/pull/2652#event-18276910430) PR was merged a wrong information, which was attributing `postgres` as default value for `ACTIVATED_DB`, while the file is named as `postgresql`.  

- Corrected `postgresql` as default value for `ACTIVATED_DB` instead of only `postgres`.

It must to match the file name, such as `docker-compose-postgresql.yml`

image

and then, in the `docker-compose.yml` it includes the corresponding service.

```
include:
  - docker-compose-${ACTIVATED_DB:-postgresql}.yml
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 - Allow configuration of the `STORAGES` setting via environment variables
 - Rename `CourseProductRelation` to `Offer`
 - Only display payment schedule for credential products
-- Provide postgres as default value for `ACTIVATED_DB` variable.
+- Provide `postgresql` as default value for `ACTIVATED_DB` variable.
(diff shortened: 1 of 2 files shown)


Tiago-Salles:
Hi @NamFra, sorry about the situation, things that happen in the life... Doing many things at the same time, I had a lack of attention and did not realize that we were merging the variable value as `postgres` and not as it expects `postgresql`. 

Another option over merging this PR is to rewrite master, and either edit [this](https://github.com/openfun/richie/pull/2652/commits/636a892d3f9421c9087a75eb415f96f062a2946e) commit or remove it from the history. Otherwise it keeps both, which in my vision is not a pretty thing , but also not that bad.

Tiago-Salles:
Hello @NamFra, it is correct, thank you!

PR #2659, opened 2025-07-01
Discounts

kernicPanel:
## Purpose

Description...

## Proposal

Description...

- [] item 1...
- [] item 2...


src/richie/apps/demo/management/commands/create_dev_data.py
```diff
@@ -0,0 +1,1194 @@
...
+    we visually don't want all our courses to have an icon.
+    We make it a convenience method so that it can be mocked in tests.
+    """
+    return random.choice(  # nosec
+        [0] * NB_OBJECTS["course_icons"] + list(range(NB_OBJECTS["course_icons"]))
+    )
+
+
+def get_internal_link(page):
+    return f"cms.page:{page.id}"
```
jonathanreveille, line 244: No docstring needed ?

src/richie/apps/demo/management/commands/create_dev_data.py
```diff
@@ -0,0 +1,1194 @@
...
+    courses,
+    licences,
+    log,
+    organizations,
+    pages_created,
+    persons,
+    programs,
+    subjects,
+):
+    placeholder = pages_created["home"].placeholders.get(slot="maincontent")
```
jonathanreveille, line 261: Ditto for the other methods in this file. Could be good if someone needs to maintain it

src/richie/apps/demo/management/commands/create_dev_data.py
```diff
@@ -0,0 +1,1194 @@
...
+            glimpse_data["image"].save()
+            add_plugin(
+                language=language,
+                placeholder=placeholder,
+                plugin_type="GlimpsePlugin",
+                target=sponsor_section,
+                **glimpse_data,
+            )
+
+    # course_run = factories.CourseRunFactory(
```
jonathanreveille: If it has a purpose, maybe add a comment line when we should activate those commented lines ? L822 to L840

src/richie/apps/demo/management/commands/create_dev_data.py
```diff
@@ -0,0 +1,1194 @@
...
+                            **item_info,
+                        )
+                else:
+                    # Create link at first level
+                    create_footer_link(
+                        pages_created,
+                        language=language,
+                        placeholder=footer_placeholder,
+                        **footer_info,
+                    )
```
jonathanreveille, line 1019:
```suggestion
                if "items" not in footer_info:
                    # Create link at first level
                    create_footer_link(
                        pages_created,
                        language=language,
                        placeholder=footer_placeholder,
                        **footer_info,
                    )
                if "items" in footer_info:
                    # Create the first level items for main columns
                    nest_column_plugin = add_plugin(
                        footer_placeholder,
                        plugin_type="NestedItemPlugin",
                        language=language,
                        content=footer_info.get("title", ""),
                    )

                    # Create the second level items for links
                    for item_info in footer_info.get("items", []):
                        create_footer_link(
                            pages_created,
                            language=language,
                            placeholder=footer_placeholder,
                            target=nest_column_plugin,
                            **item_info,
                        )
                    
```

tests/apps/courses/test_admin_course_run.py
```diff
@@ -463,6 +467,10 @@ def _prepare_change_view_post(self, course_run, course, status_code, check_metho
             "certificate_price": "29.98",
             "sync_mode": "manual",
             "display_mode": "detailed",
+            "discounted_price": "39.98",
+            "discount": "-1000 €",
```
jonathanreveille: The discount seems a little bit over the price ? no ?

src/richie/apps/demo/management/commands/create_dev_data.py
jonathanreveille: it's a matter of taste, I feel like if we can avoid the `else` statement it's sometimes better.

kernicPanel: But it would need two edits if the condition changes.

Tiago-Salles:
> Hello @Tiago-Salles are you planning to review this PR today (I've seen your 👀 under the description) ? Unfortunately, I need to merge this tomorrow…

Hello @kernicPanel! I came across this PR and I have only one doubt. I was not able to see the necessity of having the variables `discounted_price`, `discount`, `certificate_discounted_price` and `certificate_discount`. I imagine that for the `discount` the `discounted_price` would be the final price after the calculation, and also the value of `discount` will be shown on the UI (with percentage if it is required). Right?

```
    discounted_price="19.99",
    discount="-10 €",
    certificate_discounted_price="27.98",
    certificate_discount="-30%",
```

With this brief understanding I had reading the code, I would like to propose a different approach for the situation.  We could have the two fields `discount_percentage` and `discount_value` as a `Decimal` in the admin form. With the definition of a priority in the case where both are filled, we could use the `discounted_price` as a property of the model, which using a piece of code we could calculate its value, and also a property such as `discount_display` for the information of UI aspects, "-10 €" or "-30%".

Something like this
```
    def get_discount_amount(self):
        if self.discount_value:
            return self.discount_value
        elif self.discount_percentage:
            discount_percentage = self.discount_percentage / 100
            return self.price * discount_percentage
        return 0

    @property
    def discounted_price(self):
        price = self.price - self.get_discount_amount()
        return price

    @property
    def discount_display(self):
        if self.discount_value:
            return f"-{self.discount_value} €"
        elif self.discount_percentage:
            return f"-{self.discount_percentage}%"
        return None
```

With this approach we would have two fields in the admin that determine the amount to be discounted, and by filling one or both of them, the properties could trigger the calculation. It avoids the possibility of human errors on filling or calculating the values.

Is it a wrong view of your requirements?

Tiago-Salles (approved):

PR #2667, opened 2025-07-10
Display certificate and normal discount price for Syllabus Course Run

liamls:
## Purpose

We want to display the certificate and normal discount price for syllabus course run

## Proposal

Here's the proposed design for the component. 
image

### DIFF
--- a/src/frontend/js/widgets/SyllabusCourseRunsList/components/SyllabusCourseRun/index.tsx
+++ b/src/frontend/js/widgets/SyllabusCourseRunsList/components/SyllabusCourseRun/index.tsx
@@ -111,7 +111,9 @@ const OpenedCourseRun = ({
   let courseOfferMessage = null;
   let certificationOfferMessage = null;
   let enrollmentPrice = '';
+  let enrollmentDiscountedPrice = '';
   let certificatePrice = '';
+  let certificateDiscountedPrice = '';
 
   if (courseRun.offer) {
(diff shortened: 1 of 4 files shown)


src/frontend/js/widgets/SyllabusCourseRunsList/components/SyllabusCourseRun/index.tsx
```diff
@@ -111,7 +111,9 @@ const OpenedCourseRun = ({
   let courseOfferMessage = null;
   let certificationOfferMessage = null;
   let enrollmentPrice = '';
+  let originalEnrollmentPrice = '';
   let certificatePrice = '';
+  let originalCertificatePrice = '';
```
kernicPanel: I'd rather use those variables for clarity, and to match the CourseRun model:

```suggestion
  let enrollmentPrice = '';
  let enrollmentDiscountedPrice = '';
  let certificatePrice = '';
  let certificateDiscountedPrice = '';
```

liamls: Good idea, just modified my commit to have these attributes

kernicPanel (approved):
👍 Nice
Could you also add an entry to the changelog please ?

liamls:
> 👍 Nice Could you also add an entry to the changelog please ?

Done !

PR #2678, opened 2025-08-08
Voucher feature

liamls:
## Purpose

We want to handle voucher codes in sale tunnel that will be used for "normal" users and also for the users with a group purchase voucher given by their company.

Fixes #1152 
Fixes #1153 

## Proposal

Add a voucher code field in the sale tunnel and validate button. The total price and the payment schedule will be updated if there's a voucher.

image


src/frontend/js/components/PurchaseButton/index.spec.tsx
kernicPanel, line 1: Those edits should belong to the commit that changes the api endpoint.

src/frontend/js/components/SaleTunnel/index.credential.spec.tsx
kernicPanel, line 1: ditto

src/frontend/js/components/SaleTunnel/index.full-process.spec.tsx
kernicPanel, line 1: ditto

src/frontend/js/components/SaleTunnel/index.stories.tsx
kernicPanel, line 1: ditto

src/frontend/js/components/SaleTunnel/index.spec.tsx
kernicPanel, line 1: ditto

kernicPanel (commented):
I think that the majority of the test edits should belong to the first commit, and I'm afraid that the second and third ones are not tested enough.

liamls:
New tests were added:
- Submit voucher and display discounted price in full process test
- Submit invalid voucher and display error banner

The initial commit is now separated in 3 commits : use payment-plan, handle voucher code in api and add the voucher code field in sale tunnel + each commit have dedicated tests

Do you see anything else to modify in this PR ? @kernicPanel

src/frontend/js/components/SaleTunnel/SaleTunnelInformation/index.tsx
```diff
@@ -72,13 +121,19 @@ export const SaleTunnelInformation = () => {
         </div>
       </div>
       <div>
-        {product.type === ProductType.CREDENTIAL && <PaymentScheduleBlock />}
-        <Total />
+        {product.type === ProductType.CREDENTIAL && <PaymentScheduleBlock schedule={schedule} />}
```
kernicPanel:
```suggestion
        {product.type === ProductType.CREDENTIAL && schedule && (
          <PaymentScheduleBlock schedule={schedule} />
        )}
```

src/frontend/js/components/SaleTunnel/SaleTunnelInformation/index.tsx
```diff
@@ -100,45 +155,126 @@ const Email = () => {
...
+            <FormattedMessage {...messages.paymentSchedule} />
+          </h4>
+          <div className="mt-t">
+            <PaymentScheduleGrid schedule={schedule} />
+          </div>
+        </>
+      )}
     </div>
   );
 };
```
kernicPanel:
```suggestion
const PaymentScheduleBlock = ({ schedule }: { schedule: PaymentSchedule }) => {
  return (
    <div className="payment-schedule">
      <h4 className="block-title mb-t">
        <FormattedMessage {...messages.paymentSchedule} />
      </h4>
      <div className="mt-t">
        <PaymentScheduleGrid schedule={schedule} />
      </div>
    </div>
  );
};
```

src/frontend/js/components/SaleTunnel/index.full-process.spec.tsx
```diff
@@ -267,7 +267,36 @@ describe('SaleTunnel', () => {
     const $totalAmount = screen.getByTestId('sale-tunnel__total__amount');
     expect($totalAmount).toHaveTextContent(
       'Total' +
-        priceFormatter(product.price_currency, product.price).replace(/(\u202F|\u00a0)/g, ' '),
+        priceFormatter(product.price_currency, paymentPlan.price).replace(/(\u202F|\u00a0)/g, ' '),
```
kernicPanel: I know it's out of this PR scope, but WDYT about moving `.replace(/(\u202F|\u00a0)/g, ' ')`` to `priceFormatter` ?
This way we could use 
```suggestion
        priceFormatter(product.price_currency, paymentPlan.price),
```
everywhere in this file.

kernicPanel (commented):
I added some proposals.
WDYT ?

src/frontend/js/components/SaleTunnel/SaleTunnelInformation/index.tsx
kernicPanel:
```
const PaymentScheduleBlock = ({ schedule }: { schedule: PaymentSchedule }) => (
  <div className="payment-schedule">
    <h4 className="block-title mb-t">
      <FormattedMessage {...messages.paymentSchedule} />
    </h4>
    <div className="mt-t">
      <PaymentScheduleGrid schedule={schedule} />
    </div>
  </div>
);
```

src/frontend/js/components/SaleTunnel/SaleTunnelInformation/index.tsx
liamls: Is there a benefit of using `&& ( <Component />) `instead of using directly the component with ` && <Component />` ?

src/frontend/js/components/SaleTunnel/SaleTunnelInformation/index.tsx
liamls: Since `schedule` value is get from a fetch, it can be null or undefined. To use a mandatory schedule props in PaymentSheculeBlock, we have also to check schedule when calling the component 

```
{product.type === ProductType.CREDENTIAL && schedule && (
          <PaymentScheduleBlock schedule={schedule} />
        )}
```

liamls: I agree that it is better with checking before the component than in the component

src/frontend/js/components/SaleTunnel/SaleTunnelInformation/index.tsx
kernicPanel: My point is to simplify `PaymentScheduleBlock` which, in my opinion, should not be rendered if no schedule exists.

liamls: My bad I did not saw the `&&schedule` added...

src/frontend/js/components/SaleTunnel/index.full-process.spec.tsx
liamls: Sure, it will be better

src/frontend/js/components/SaleTunnel/SaleTunnelInformation/index.tsx
```diff
@@ -100,45 +155,126 @@ const Email = () => {
...
+}) => {
+  const intl = useIntl();
+  const { voucherCode, setVoucherCode } = useSaleTunnelContext();
+  const [voucher, setVoucher] = useState('');
+  const handleVoucher = (e: ChangeEvent<HTMLInputElement>) => setVoucher(e.target.value);
+  const submitVoucher = () => {
+    setVoucherError(false);
+    setVoucherCode(voucher);
+    setVoucher('');
+  };
```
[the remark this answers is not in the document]
liamls: Null or undefined voucher code won't trigger API call so it's not needed IMO

src/frontend/js/components/SaleTunnel/SaleTunnelInformation/index.tsx
```diff
@@ -52,10 +53,58 @@ const messages = defineMessages({
...
+  const price = query.data?.price || props.paymentPlan?.price;
+  const discountedPrice = query.data?.discounted_price || props.paymentPlan?.discounted_price;
+  const discount = query.data?.discount || props.paymentPlan?.discount;
+
+  useEffect(() => {
+    if (query.error && voucherCode) {
+      setVoucherCode('');
+      setVoucherError(true);
+    }
+  }, [query.error, voucherCode, setVoucherCode]);
```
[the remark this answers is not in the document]
liamls: Query is only used for this payment-plan so I think it's not needed to add more code

liamls:
> When a voucher is applied, I think we should remove the input field: image image
> 
> Also, during voucher validation, if it fails, the form is jumping
> Kooha-2025-09-08-15-41-33.webm

I think it can be confusing to hide the input for users, but I think it's good to disable the input and button when a voucher code was submited.

PR #2682, opened 2025-08-28
Fix CourseProductItem when no rule in offering

liamls:
The offering may be sometimes null, which created an error. (Detected during the latest MTP).

## Purpose

We want to handle the possibly null offering and rules in the offering.

## Proposal

Description...

- Add `offering?.rules?.` to handle null values

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Fixed
+
+- Handle null offering rules on CourseProduct item
+
 ## [3.2.0] - 2025-08-26
 
 ### Added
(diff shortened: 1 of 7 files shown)


kernicPanel (commented):
Please add at least a test where `rules` is missing.

CHANGELOG.md
```diff
@@ -39,6 +39,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 - Display discounted price for the purchase of a certificate product
   from student dashboard in sale tunnel informations
 - Clear the page and search caches when syncing a course run
+- Handle null offering rules on CourseProduct item
```
kernicPanel: This line should go under `Unreleased`

liamls:
Added one test to render the component without rules & modified changelog file. (Sorry I forgot to do a fixup commit).

kernicPanel (commented):
I think we should update the types as well.

PR #2696, opened 2025-10-06
💄(frontend) handle voucher error in sale tunnel

liamls:
## Purpose

Since we have an API rate limit in Joanie API, we want to display the good error for the user in the sale tunnel. If the voucher is not valid, already claimed or expired then we display the default one. But if it's about the API rate with 429 error code then we display the new one.


CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 - Add `discounted_price` to the course glimpse template
 - Add autoplay for carousel slides
 - Add voucher code handle in sale tunnel
+- Add custom error message display for voucher code in sale tunnel
```
kernicPanel: This line should go to the Unreleased section.

kernicPanel (commented):
Can we add a test for this message ?

PR #2741, opened 2026-02-18
Enrollement dashboard

liamls:
## Purpose

Display seats data related to a batch order in the buyer dashboard.

- [ ] Implement /batch-orders/{id}/seats API endpoint and update BatchOrderRead object
- [ ] Add subitem in dashboard with owned seats and to be owned seats
- [ ] Add subitem in organization dashboard 

Screenshots : 

image

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 - Handle French and English language in Storybook
 - Upgraded js dependencies, fixed tests and linting errors
+- Display batch order enrollment status in dashboard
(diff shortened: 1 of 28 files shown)


kernicPanel (commented):
src/frontend/js/widgets/Dashboard/components/DashboardItem/BatchOrder/BatchOrderSeatInfo.tsx and src/frontend/js/pages/TeacherDashboardOrganizationQuotes/BatchOrderSeatInfoQuote.tsx are almost the same. 
We should at least extract the messages to avoid their duplication.

jonathanreveille (approved):

PR #2817, opened 2026-05-22
🐛(frontend) fix select dropdown clipped in sale tunnel right column

liamls:
SaleTunnel was hiding overflowing elements in the right part, so we need to update the overflow style attribute to handle this.

**How to reproduce this case :** remove this fix and add at least 3 organizations to offering organizations to see that they are not displayed (you can mock the API answer with Tanstack Data Explorer). With the fix, it will be displayed.

image


jonathanreveille (approved):
works with even 5 organizations, nicely done !

PR #2841, opened 2026-08-14
Hotfix : string duplication in batch order seats

liamls:
Rename batchOrder.seatsCount used in TeacherDashboardOrganizationQuotes since it collided with the seatsCount id already used in the learner BatchOrder dashboard item, causing the French translation to silently fall back to English.


jonathanreveille (approved):
hopefully it's the last seat 🤞 
LGTM ! 🚀

Issue #2532, opened 2024-11-05
Incorrect condition on Weekly pace course search filter having an AND instead of OR


igobranco:
## Bug Report

**Problematic Behavior**
When we select multiple `Weekly pace` options on the course search filter page we don't see any results.

When we select:
- [x] One to two hours (4)
- [x] More than two hours (21)

On the page https://demo.richie.education/en/courses/?licences=3&licences=2&limit=21&offset=0&pace=1h-2h&pace=gt-2h

The results is always empty.

**Expected behavior/code**
The search result should merge (OR) the filter conditions, because the component is a select multiple on the Weekly pace.

**Steps to Reproduce**

Both Weekly pace's has courses:
https://demo.richie.education/en/courses/?licences=3&licences=2&limit=21&offset=0&pace=1h-2h
https://demo.richie.education/en/courses/?licences=3&licences=2&limit=21&offset=0&pace=gt-2h

When both are selected, if didn't show any couse: 
https://demo.richie.education/en/courses/?licences=3&licences=2&limit=21&offset=0&pace=1h-2h&pace=gt-2h

**Environment**
- Richie version: demo site and our NAU instance

**Possible Solution**

Fix the implementation to use and `OR` instead of `AND`.

**Additional context/Screenshots**

![image](https://github.com/user-attachments/assets/15f382a9-ce6a-49f4-93f7-4401951f715a)

cabralpedrofccn:
See if the issue was already solved.

PR #2856, opened 2026-09-13
🐛(search) combine several values of a choices filter with OR

Tiago-Salles:
## Purpose

When several values of a choices search filter are selected, the search returns no results. For example, on the demo site, selecting the weekly paces "One to two hours" and "More than two hours" returns nothing, while each of them alone returns courses:

| Query | Results |
|---|---|
| `?pace=1h-2h` | 1 |
| `?pace=gt-2h` | 1 |
| `?pace=1h-2h&pace=gt-2h` | 0 |

The cause is in `ChoicesQueryMixin.get_query_fragment`: it returns one query fragment per selected value, and `CourseSearchForm.build_es_query` flattens every fragment into a single bool `must` clause. Each selected value therefore becomes an extra AND condition, and since a course run has a single pace, no document can match two of them.

Terms based filters (subjects, levels, organizations, languages) are not affected because `TermsQueryMixin` emits a single `terms` query, which is an OR by nature. Only the filters built on `ChoicesQueryMixin` (pace, availability) have the problem.

## Proposal

When more than one value of a choices filter is selected, wrap the fragments of the selected values in a single bool `should` clause with `minimum_should_match: 1`, so the values are ORed together while the filter as a whole is still ANDed with the other filters. Each value's fragment is a list of clauses that must all match (availability values have two range clauses), hence the inner bool `must`.

- Combine the fragments of several selected values with `should` in `ChoicesQueryMixin.get_query_fragment`; a single selected value keeps the exact previous output and the fragment `key` is unchanged, so the facet aggregations that exclude a filter's own fragments keep working
- Add unit tests for `StaticChoicesFilterDefinition.get_query_fragment` with one and several values
- Add an Elasticsearch test selecting two paces and expecting the union of both course sets
- Add a changelog entry

Related to: openfun/richie#2532


Tiago-Salles:
Hey, @jonathanreveille  and @liamls, could you please review this? Thanks!

